PR #46, opened 2014-09-12
[FEATURE] Split 'sign' method into another call, remove offline sign

mDuo13:
After discussing with Norman, we agreed that removing the sign button would make it more obviously a "websocket" tool


wltsmrz:
> more obviously a "websocket" tool

Doesn't make sense. You should actually remove the sign / submit commands altogether

PR #58, opened 2014-09-19
[DOC] fixes in submit payment documentation

n0rmz:


n0rmz:
@mDuo13 I added some corrections in submit payment documentation in ripple-rest markdown. Can you  adjust the HTML version after the merge thanks!

mDuo13:
No further adjustments to the HTML version should be necessary -- right now only the rippled APIs page is compiled (because it's so big & slow otherwise).

We can talk about compiling the other pages the same way for SEO purposes later though

PR #70, opened 2014-09-26
Some issues

sublimator:
This is not a request to merge per se, just a way of highlighting some issues found upon perusal.


mDuo13:
I'm out of the office today, but I'll look into this in detail when I get back. Thanks for the feedback.

sublimator:
It might be worth pointing out that the `Expiration` timestamp uses the ripple epoch, which is seconds since `1 Jan 2000 00:00:00 GMT`

PR #72, opened 2014-09-30
Fix markup. s/DestinationAmount/DeliveredAmount/

sublimator:
DestinationAmount is actually `DeliveredAmount` and (for better or worse) is set only when the full amount wasn't paid.

``` c++
if (rc.result () == tesSUCCESS && rc.actualAmountOut != saDstAmount)
    mEngine->view ().setDeliveredAmount (rc.actualAmountOut);
```

References: 
[Payment Transactor](https://github.com/ripple/rippled/blob/2936bbfae86de2c94d5a1ee6ac129dd4f92de870/src/ripple/app/transactors/Payment.cpp#L292)
[Fields enumeration](https://github.com/ripple/rippled/blob/2936bbfae86de2c94d5a1ee6ac129dd4f92de870/src/ripple/data/protocol/SField.h#L358)

### DIFF
--- a/tx_format.md
+++ b/tx_format.md
@@ -226,7 +226,7 @@ In order to use AccountTxnID, you must first set the [asfAccountTxnID](#accounts
 The Memos field allows for arbitrary messaging data that can accompany the transaction. It is presented as an array of objects, where each object has the following fields:
 
 | Field | Type | [Internal Type](https://wiki.ripple.com/Binary_Format) | Description |
-|-------|------|-------------|
+|-------|------|--------------------------------------------------------|-------------|
 | MemoType | Hex String | VariableLength | A unique relation (according to [RFC 5988](http://tools.ietf.org/html/rfc5988#section-4)) that defines the format of this memo. |
 | MemoData | Hex String | VariableLength | TODO |
 | MemoFormat | Hex String | VariableLength | TODO |
@@ -331,7 +331,7 @@ By default, the `Amount` field of a Payment transaction specifies the amount of 
 
 The [*tfPartialPayment* flag](#payment-flags) allows you to make a "partial payment" instead. When this flag is enabled for a transaction, the costs of transaction fees and exchanging currencies are deducted from the amount sent instead of from the sender's balance. Consequently, the `Amount` field __*is not guaranteed to be the amount received*__. In fact, __*there is no minimum guaranteed amount*__ that a partial payment actually delivers.
 
-Validated partial payment transactions have a `meta.DestinationAmount` field, which indicates the amount of currency actually received by the destination account. *Note:* Early partial payments in historical ledgers do not have this field.
+Validated partial payment transactions have a `meta.DeliveredAmount` field, which indicates the amount of currency actually received by the destination account. *Note:* Early partial payments in historical ledgers do not have this field.
 
 A partial payment cannot provide the initial XRP to fund an account; this case returns the error code telNO_DST_PARTIAL. Direct XRP-to-XRP payments also cannot be partial payments temBAD_SEND_XRP_PARTIAL.
 


sublimator:
See also:  https://github.com/ripple/ripple-dev-portal/blame/master/tx_format.md#L524

The limit values of trust lines (See [TrustSet](#trustset)) do not affect >>> offer. <<<

sublimator:
See also: https://github.com/ripple/ripple-dev-portal/blame/master/tx_format.md#L172

`SigningPubKey   String  PubKey  (Omitted until signed)`

Not entirely clear what `Omitted until signed` means, but the SigningPubKey _does_ contribute to the blob that is hashed for signing.

sublimator:
> > > However, this flag is not enforced in rippled because it could potentially cause accounts to become unusable. (If an account did not have enough XRP to meet the reserve and make a transaction that disabled the flag, the account would never be able to send another transaction.) Instead, client applications should disallow or discourage XRP payments to accounts with the asfDisallowXRP flag enabled.

```
(If an account did not have enough XRP to meet the reserve and make a
 transaction that disabled the flag, the account would never be able to send
 another transaction.)
```

The implications of this are potentially misleading. I think transactions can always be sent, regardless of xrp balance in relation to reserve. I believe you can always use your reserve balance to send transactions. The possibility that's likely behind the non-enforcement policy is the case of burning up all the Balance with transaction Fee's.

Example: 

``` bash
#!/bin/bash

################################### CONSTANTS ##################################

export BOB='rPMh7Pi9ct699iZUTWaytJUoHcJ7cgyziK'
export ROOT='rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh'
export RIPPLED='./build/gcc.debug/rippled'

#################################### HELPERS ###################################

function submit_and_close() 
{
    ./build/gcc.debug/rippled submit $1 "$2"
    ./build/gcc.debug/rippled ledger_accept
}

#################################### SCRIPT ####################################

echo "1. Create the bob account"
submit_and_close masterpassphrase "$(cat <<TXN
    {
        "TransactionType" : "Payment",
        "Account" : "$ROOT",
        "Destination" : "$BOB",
        "Amount" : "1000000000"
    }
TXN
)"

echo "2. Bob trusts the root account"
submit_and_close bob "$(cat <<TXN
    {
        "TransactionType" : "TrustSet",
        "Account" : "$BOB",
        "LimitAmount" : {"value" : "500", "currency" : "USD", "issuer" : "$ROOT"}
    }
TXN
)"

echo "3. Try and send 1 drop past the balance (fails)"
submit_and_close bob "$(cat <<TXN
    {
        "TransactionType" : "Payment",
        "Account" : "$BOB",
        "Destination" : "$ROOT",
        "Amount" : "749999991"
    }
TXN
)"

echo "4. This should work "
echo "the Amount is 11 drops less"
echo "accounting for 10 drops Fee for the previous tec transaction"
submit_and_close bob "$(cat <<TXN
    {
        "TransactionType" : "Payment",
        "Account" : "$BOB",
        "Destination" : "$ROOT",
        "Amount" : "749999980"
    }
TXN
)"

echo "This won't work "
echo "Can not SEND from the reserve"
submit_and_close bob "$(cat <<TXN
    {
        "TransactionType" : "Payment",
        "Account" : "$BOB",
        "Destination" : "$ROOT",
        "Amount" : "1"
    }
TXN
)"

echo "But we can still make transactions and burn Fee"
submit_and_close bob "$(cat <<TXN
    {
        "TransactionType" : "AccountSet",
        "Account" : "$BOB",
        "Fee" : "249999980"
    }
TXN
)"

echo "Show how much balance is left"
$RIPPLED account_info bob
```

mDuo13:
I have made the relevant changes. They will get deployed in the next push.

In the future, can you open pull requests against the gh-pages branch? Thanks.

sublimator:
No problem! Sorry about that

PR #83, opened 2014-10-22
[DOC] update ripple-rest api changes

geertweening:


geertweening:
@mDuo13 changes to the ripple-rest API

one of the changes that might affect terminology through the some other parts of the document is the `account/new` to `wallet/new` change. I'll leave it up to you if we want to use the "wallet" through the explanation for the results of that endpoint as well.

mDuo13:
Not sure why we went back to "wallet" (a sort of confusing and overloaded term) when until now REST was blissfully lacking in that terminology -- also, you appear to have some accidental " marks in your edits?

geertweening:
Yeah, I'll take " marks out, my bad.

We went back to `wallet` since you're not really creating an account. You're generating an address and a key and it's not in the ledger yet. Looking up anything using the just generated "account" wouldn't do anything. To separate what an account meant and what this 'new' endpoint meant I was looking for a different term. We settled on "wallet". 

Other suggestions would be great as well, but I would like to keep it separate from "account"

mDuo13:
Maybe "address", then? That's actually closer to what you're generating.

geertweening:
It's an address and a secret. I was looking for something that captures both.

PR #95, opened 2014-11-18
[DOC] add technical start guide

geertweening:
proposal to add this to the dev portal


mDuo13:
I have some changes I want to make to the text, but I'll do that after merging.

PR #100, opened 2014-12-08
[DOC] add documentation for placing, canceling, and retrieving account orders

boxbag:


mDuo13:
I have some concerns with the section on specifying currency amounts in the request body
using the slash syntax
although not with the docs, per se, but the API they reflect
since we already use the JSON {amount, currency, issuer} syntax for currency amounts, even in requests
e.g. source_amount in Submit Payment
So introducing the slash-based syntax seems unnecessary, and the doc instructing people to use it appears to be giving people two separate sets of instructions for the same thing
It is my strong opinion that taker_gets and taker_pays should follow the same formatting conventions as source_amount and destination_amount

boxbag:
I agree that we should be consistent. After I update the API, I'll come back, rebase, and update the docs

PR #102, opened 2014-12-15
[DOC] add fixed fee docs to payments API

boxbag:


mDuo13:
What does this phrase mean? "If set below server's fee, the transaction will keep resubmitting with same fixed fee."

boxbag:
From what I understand, when the fee is set below the server's expected fee, we have logic to resubmit the transaction. Additionally, when a server's fee is updated, all pending transactions' fees are updated accordingly. But when you specify a fixed fee, the fee will not update and the same original fixed fee will be used each time the transaction is resubmitted.

mDuo13:
OK. I'll tweak the wording based on that explanation.

PR #103, opened 2014-12-15
[DOC]: Ripple-REST Update balances and trustlines limit query param (RLJS-116)

alandotcom:
- Allow ?limit=all

### DIFF
--- a/content/ripplerest_api.md
+++ b/content/ripplerest_api.md
@@ -569,7 +569,7 @@ Optionally, you can also include any of the following query parameters:
 | currency | String ([ISO 4217 Currency Code](http://www.xe.com/iso4217.php)) | If provided, only include balances in the given currency. |
 | counterparty | String (Address) | If provided, only include balances issued by the provided address (usually a gateway). |
 | marker | String | Server-provided value that marks where to resume pagination. |
-| limit | Integer | (Defaults to 200) Max results per response. Cannot be less than 10. Cannot be greater than 400. |
+| limit | String (Integer or 'all') | (Defaults to 200) Max results per response. Cannot be less than 10. Cannot be greater than 400. Use 'all' to return all results |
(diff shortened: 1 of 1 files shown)


mDuo13:
The "limit" system is in place to gracefully handle situations where the response would be too big. What error does the API return when the contents of "all" are just too big for it to reasonably return?

geertweening:
We don't have anything for that in place. "Too big to reasonably return" would probably be rippled taking too long to return all results and the connection timing out

alandotcom:
So here's an example with a very large result. I'm using [jq](http://stedolan.github.io/jq/) to get the count of balances in the result (I'm using snapswaps ripple address). The size of the result was 1549k and it took roughly 1-2 seconds.

```
» curl http://localhost:5990/v1/accounts/rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q/balances\?limit\=all 2>/dev/null | jq '.balances | length'
12706
```

This is an example of paginating through responses (using streams): https://gist.github.com/lumberj/88285637bd2b6a938161

@mDuo13 do you have any advice on where we can include these types of examples?

One caveat is that this example would probably crash the browser, so adding `limit=all` to the [rest tool](https://ripple.com/build/rest-tool/) may not be a great idea.

mDuo13:
We don't currently have a good place for this kind of large but highly-specific example in the Dev Portal. For now, I'd recommend putting it in the wiki. Meanwhile, I'll mull over the best way to incorporate this into the Dev Portal structure.

geertweening:
@lumberj @mDuo13 we have a `docs` folder in the ripple-rest repo where examples like this would fit as well. I would put it there over the wiki

mDuo13:
Sounds like a good plan.

alandotcom:
Not sure what you mean by 'wiki'. Is it this? https://wiki.ripple.com/Main_Page I don't see any documentation there for ripple-rest except for a link to the dev portal. Also, in my personal experience, I found the wiki to be generally out of date.

PR #106, opened 2015-01-08
[DOC] change issuer to counterparty for place order

boxbag:


mDuo13:
Why the change from "issuer" to "counterparty" for orders only, if I may ask?

boxbag:
It's not just for orders. It's planned to be a REST-wide change. Starting with orders since it's completed but planning to have all instances of issuer changed to counterparty. Reasoning is that issuer is not a domain term, whereas counterparty is more faithful to industry language

PR #109, opened 2015-01-25
Document can_delete method to rippled.

mtrippled:
I proof-read this in a mark-down editor. It didn't pick up all of the formatting, but the content is correct. I'm assuming that this file gets generated into https://ripple.com/build/rippled-apis/
Is that correct?


mDuo13:
This is the correct format. (I'd still like a little more time to look over the content more thoroughly before merging.) One question about the content: is can_delete an admin-only method?

mtrippled:
Yes, it is admin-only.

PR #112, opened 2015-02-05
[DOC] Modify to HTTPS

n0rmz:


mDuo13:
When I tested it, the API did not serve https. Has that changed?

n0rmz:
That's right, I'm preemptively changing it as there will be 443 support by the time we release the gateway bulletin.

mDuo13:
added separately

PR #113, opened 2015-02-09
[REST] Update API docs to add `Get Order Transaction`

alandotcom:
- Updates Payments API
- Adds a new response object `Order Change` to describe changes to
  orders in the order book
- Update all source code URLs to permalinks

### DIFF
--- a/content/ripplerest_api.md
+++ b/content/ripplerest_api.md
@@ -22,10 +22,12 @@ We recommend Ripple-REST for users just getting started with Ripple, since it pr
 * [Get Payment History - `GET /v1/accounts/{:address}/payments`](#get-payment-history)
 
 #### Orders ####
+
 * [Place Order - `POST /v1/accounts/{:address}/orders`](#place-order)
 * [Cancel Order - `DELETE /v1/accounts/{:address}/orders/{:sequence}`](#cancel-order)
 * [Get Account Orders - `GET /v1/accounts/{:address}/orders`](#get-account-orders)
 * [Get Order Book - `GET /v1/accounts/{:address}/order_book/{:base}/{:counter}`](#get-order-book)
+* [Get Order Transaction - `GET /v1/accounts{:address}/orders/{:hash}`](#get-order)
 
 #### Trustlines ####
 
@@ -239,22 +241,6 @@ openssl x509 -req -days 730 -in /etc/ssl/server.csr -signkey /etc/ssl/private/se
 
 ### Keeping the service running ###
 
-To make sure that the Ripple-REST process remains active even if it crashes for some reason, use the  [`forever`](https://www.npmjs.org/package/forever) Node module. Install `forever` using `sudo npm install -g forever`.
-
-Here is an example of running `ripple-rest` using `forever`:
-
-```
-forever start \
-    --pidFile /var/run/ripple-rest/ripple-rest.pid \
-    --sourceDir /opt/ripple-rest \
-    -a -o /var/log/ripple-rest/ripple-rest.log \
-    -e /var/log/ripple-rest/ripple-rest.err \
-    -l /var/log/ripple-rest/ripple-rest.for \
-    server.js
-```
-
-### Monitoring the service ###
-
 Monitor `ripple-rest` using [`monit`](http://mmonit.com/monit/). On Ubuntu you can install `monit` using `sudo apt-get install monit`.
 
 Here is an example of a monit script that will restart the server if:
@@ -273,7 +259,6 @@ check process ripple-rest with pidfile /var/run/ripple-rest/ripple-rest.pid
         and request "/v1/server"
     then restart
 ```
-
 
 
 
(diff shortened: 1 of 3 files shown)


alandotcom:
Updates for https://github.com/ripple/ripple-rest/pull/278 and https://github.com/ripple/ripple-rest/pull/280

content/ripplerest_api.md
```diff
@@ -1530,13 +1224,20 @@ Optionally, you can also include the following query parameters:
 
 If the length of the `payments` array is equal to `results_per_page`, then there may be more results. To get them, increment the `page` query paramter and run the request again.
 
-*Note:* It is not more efficient to specify more filter values, because Ripple-REST has to retrieve the full list of payments from the `rippled` before it can filter them.
```
mDuo13, line 1533: Is this no longer true?

content/ripplerest_api.md
```diff
@@ -239,22 +241,6 @@ openssl x509 -req -days 730 -in /etc/ssl/server.csr -signkey /etc/ssl/private/se
...
-```
-forever start \
-    --pidFile /var/run/ripple-rest/ripple-rest.pid \
-    --sourceDir /opt/ripple-rest \
-    -a -o /var/log/ripple-rest/ripple-rest.log \
-    -e /var/log/ripple-rest/ripple-rest.err \
-    -l /var/log/ripple-rest/ripple-rest.for \
-    server.js
-```
-
```
mDuo13, line 255: Why was the section on "forever" removed?

alandotcom, line 255: Running forever with monit causes issues. It's better to have monit monitor the process and restart it if necessary.

content/ripplerest_api.md
```diff
@@ -1721,159 +1419,252 @@ The response is an object with a `orders` array, where each member is a [order o
...
+    }
+  ]
+}
+
+```
+
+## Get Order Transaction ##
+[[Source]<br>](https://github.com/ripple/ripple-rest/blob/59ea02d634ac4a308db2ba21781efbc02f5ccf53/api/orders.js#L505 "Source")
+
+Get the details of an order transaction.
```
mDuo13: I'd like this blurb to be more informative. Why's there a separate method to get an order transaction? When would you use it?

content/ripplerest_api.md
alandotcom, line 1533: Didn't realize I removed this. It is still true

mDuo13, line 1533: Please add a commit fixing this to the PR

content/ripplerest_api.md
mDuo13, line 255: OK, that's fine. Next time, I'd appreciate a heads-up in the initial PR comment.

content/ripplerest_api.md
alandotcom: An order transaction is an `OfferCreate` or `OfferCancel`. The notifications API is now pointing to this endpoint (previously it would point to /v1/transactions/{:hash} which just returned the raw data from rippled.

I used order "transaction" because it's not an order (`OfferEntry`) on the ledger / orderbook. It's a transaction that could create or consume these.

I can add language to make this distinction. Perhaps calling it "Get Order Submission" would be clearer?

mDuo13: I am fine with either "Get Order Transaction" or "Get Order Submission". My priority is making sure that the blurb captures why you would use this method and not any other. I think I understand now, so I can write it myself if you don't want to.

alandotcom: I'll add something and update the PR

content/ripplerest_api.md
```diff
@@ -930,6 +926,11 @@ Submit a payment object to be processed and executed.
 
 ```
 POST /v1/accounts/{address}/payments?validated=true
+```
+
+</div>
```
alandotcom: Moved this tag by accident

content/ripplerest_api.md
alandotcom, line 1533: It wasn't removed. It was just moved. https://github.com/ripple/ripple-dev-portal/pull/113/files#diff-f7143e9cdcfd43266bc9e20b8dec1c36R1235

content/ripplerest_api.md
alandotcom: Updated

content/ripplerest_api.md
```diff
@@ -1721,159 +1416,298 @@ The response is an object with a `orders` array, where each member is a [order o
...
+
+| Field | Value | Description |
+|-------|-------|-------------|
+| address | String | The Ripple account address whose orders to look up. |
+| hash | String | The transaction hash for the order |
+
+
+#### Response ####
+
+An example response for an order transaction to place an order:
```
alandotcom, line 1603: This is also new. I have two example responses

content/ripplerest_api.md
```diff
@@ -1721,159 +1416,298 @@ The response is an object with a `orders` array, where each member is a [order o
...
-    "sequence": 105966,
-    "passive": false,
-    "sell": false
+    "sequence": 20,
+    "status": "canceled"
   }]
 }
 ```
 
+The response includes the original [`order`](#order-objects), if the `action` is `order_create`.
```
alandotcom, line 1700: New language

PR #143, opened 2015-10-02
[TASK] Remove yum install instructions

wilsonianb:
To be re-added when legit rippled rpm package is available


stevenzeiler:
Okay that is reasonable. Although it seems to work the delivery (unsigned and not encrypted) is sub-par to say the least.

stevenzeiler:
:+1:

stevenzeiler:
Alternatively we could simply replace `http` with `https`

PR #145, opened 2015-10-14
[TASK] Re-add rippled yum install instructions

wilsonianb:


mDuo13:
Is there any update regarding the signing keys for this thing?

wilsonianb:
Previously the yum repository didn't enable `gpgcheck`. That is fixed and the rpms are now signed.

I don't think the instructions need to be updated. Installation will now fail if the package signatures don't match.

PR #156, opened 2016-01-16
Peer review comments for KC-124

ryangyoung:
https://ripplelabs.atlassian.net/browse/KC-124

Some more comments:
## Freeze Article Review
- "There are three settings related to freezes" — Should you specify "account settings" here to distinguish from trust line settings?
- _(Nitpick)_ "Permanently give up the ability to freeze individual counterparties. Also gives up the ability to end a global freeze." — Breaking this up into two sentences seems a little weird to me, like the second part is an afterthought. But the meaning is clear.
- "the effect of a currency holder freezing an issuer is minimal" — Interesting. I didn't know this. What _is_ the effect? Is it worth describing here? Oh, I see that it's explained later. Perhaps add a reference or indicate that? The two are pretty close, though, so I can see the argument for leaving it as-is.
  - "The counterparty of that trust line can no longer decrease its balance on the frozen trust line, except in direct payments to the issuer." — Since this is twisty logic, perhaps add a second sentence along the lines of "This means that a counterparty can only send the frozen issuances back to the issuer."  
- "The gateway should also freeze the counterparty in any connector systems the gateway operates." — Define "connector systems" maybe? I don't know what that means exactly. But if a known term, that's fine. 
- "An individual can freeze the trust line to a gateway. This has no effect on transactions between the gateway and other users. It does, however, prevent other accounts, including hot wallets from sending that gateway's issued currency to the individual. It has no effect on offers." — Maybe specify "individual account" to make it absolutely clear? And should "gateway" here be "issuing account" to cover non-gateway cases (if any)? And in the second sentence, I think there should be a comma after "hot wallets". That last "it" is slightly ambiguous, but I can't think of a way to improve on it other than repeating most of the first sentence, as in "An individual account freezing the trust line to a gateway has no effect on offers created by the individual account."

### DIFF
--- a/content/freeze.md
+++ b/content/freeze.md
@@ -34,9 +34,10 @@ An account cannot enable the Individual Freeze setting if it has previously enab
 Global Freeze
 -------------
 
-The **Global Freeze** feature is a setting on an account. When an issuing account enables the Global Freeze feature, the following rules apply:
+The **Global Freeze** feature is a setting on an account. [//]: # (Is this not the case for the other Freeze settings?) When an issuing account enables the Global Freeze feature, the following rules apply:
 
 * All counterparties of the frozen issuing account can no longer decrease the balances in their trust lines to the frozen account, except in direct payments to the issuer. (This also affects any [hot wallet](gateway_guide.html#hot-and-cold-wallets) accounts.)
+    [//]: # (Again, maybe add a clarifying sentence here, "This means that..." or in the next point. )
 * Counterparties of the frozen issuing account can still send and receive payments directly to and from the issuing account.
 * All offers to sell currencies issued by the frozen account are [considered unfunded](transactions.html#lifecycle-of-an-offer).
 
@@ -55,7 +56,9 @@ No Freeze
 The **No Freeze** feature is a setting on an account that permanently gives up the ability to freeze counterparties. A business can use this feature to treat its issued funds as "more like physical money" in the sense that the business cannot interfere with customers trading it among themselves. The NoFreeze setting has two effects:
 
 * The issuing account can no longer use enable Individual Freeze on any counterparty.
+    [//]: # ("can no longer enable")
 * The issuing account can still enable Global Freeze to enact a global freeze, but the account cannot _disable_ Global Freeze.
+    [//]: # (To clarify, perhaps: "the account cannot disable Global Freeze once it's enabled")
 
 The Ripple Consensus Ledger cannot force a gateway to honor the obligations that its issued funds represent, so giving up the ability to enable a Global Freeze cannot protect customers. However, giving up the ability to _disable_ a Global Freeze ensures that the Global Freeze feature is not used unfairly against some customers.
 
@@ -409,3 +412,5 @@ Example JavaScript (ECMAScript 6) code to check whether an account has Global Fr
 # See Also #
 
 [Gateway Bulletin GB-2014-02 New Feature: Balance Freeze](https://ripple.com/files/GB-2014-02.pdf)
+
+[//] # (Ugh, I hate that this PDF is all we have for "See Also." No other alternative that I can think of, though.)


content/freeze.md
```diff
@@ -34,9 +34,10 @@ An account cannot enable the Individual Freeze setting if it has previously enab
 Global Freeze
 -------------
 
-The **Global Freeze** feature is a setting on an account. When an issuing account enables the Global Freeze feature, the following rules apply:
+The **Global Freeze** feature is a setting on an account. [//]: # (Is this not the case for the other Freeze settings?) When an issuing account enables the Global Freeze feature, the following rules apply:
```
mDuo13, line 37: Global Freeze and No Freeze are settings on an account.
Individual Freeze is a setting on a trust line.

So, in short, yes.

content/freeze.md
```diff
@@ -34,9 +34,10 @@ An account cannot enable the Individual Freeze setting if it has previously enab
...
 
 * All counterparties of the frozen issuing account can no longer decrease the balances in their trust lines to the frozen account, except in direct payments to the issuer. (This also affects any [hot wallet](gateway_guide.html#hot-and-cold-wallets) accounts.)
+    [//]: # (Again, maybe add a clarifying sentence here, "This means that..." or in the next point. )
```
mDuo13, line 40: I'd rather not add extraneous language to this point.

content/freeze.md
```diff
@@ -55,7 +56,9 @@ No Freeze
 The **No Freeze** feature is a setting on an account that permanently gives up the ability to freeze counterparties. A business can use this feature to treat its issued funds as "more like physical money" in the sense that the business cannot interfere with customers trading it among themselves. The NoFreeze setting has two effects:
 
 * The issuing account can no longer use enable Individual Freeze on any counterparty.
+    [//]: # ("can no longer enable")
```
mDuo13, line 59: Changed to "Can no longer enable Individual Freeze on trust lines to any counterparty."

content/freeze.md
```diff
@@ -55,7 +56,9 @@ No Freeze
...
 * The issuing account can still enable Global Freeze to enact a global freeze, but the account cannot _disable_ Global Freeze.
+    [//]: # (To clarify, perhaps: "the account cannot disable Global Freeze once it's enabled")
```
mDuo13, line 61: I think this sentence is clearer as it is.

content/freeze.md
```diff
@@ -409,3 +412,5 @@ Example JavaScript (ECMAScript 6) code to check whether an account has Global Fr
 # See Also #
 
 [Gateway Bulletin GB-2014-02 New Feature: Balance Freeze](https://ripple.com/files/GB-2014-02.pdf)
+
+[//] # (Ugh, I hate that this PDF is all we have for "See Also." No other alternative that I can think of, though.)
```
mDuo13, line 416: How about I add [Freeze Code Samples](https://github.com/ripple/ripple-dev-portal/tree/gh-pages/content/code_samples/freeze)?

content/freeze.md
ryangyoung, line 37: OK.

content/freeze.md
ryangyoung, line 40: OK.

content/freeze.md
ryangyoung, line 59: OK.

content/freeze.md
ryangyoung, line 61: OK. I see that the logic here is "Once you enable this setting, the following things are true," which covers my concern. Comment retracted.

content/freeze.md
ryangyoung, line 416: Sure, that's a good idea. My concern was more that we don't have a better place for the information in that PDF. But, like I said, I don't think there's really an alternative.

mDuo13:
1. "three settings" - One is a trust line setting. Two are account settings. The non-distinction here is by design.
2. Changed the sentence: "Permanently give up the ability to freeze individual counterparties, as well as the ability to end a global freeze."
3. I'd rather not clarify the minimal effect of user freezing a gateway in two places.
4. Added a sentence to the "twisty-logic" passage: "The counterparty of that trust line can no longer decrease its balance on the frozen trust line, except in direct payments to the issuer. The counterparty can only send the frozen issuances directly to the issuer."
5. The definition of "connector systems" would be too long, so I rephrased the passage: "The gateway should also freeze the counterparty in any other systems the gateway operates that are connected to the Ripple Consensus Ledger. (Otherwise, an account might still be able to engage in undesired activity by sending payments through the gateway.)"
6. Rephrased the passage: "An individual account can freeze its trust line to a gateway. This has no effect on transactions between the gateway and other users. It does, however, prevent other accounts, including [hot wallets](gateway_guide.html#hot-and-cold-wallets), from sending that gateway's issued currency to the individual account. This type of individual freeze has no effect on offers."

For the edits, see PR #157

PR #158, opened 2016-01-23
Remove ripplerest (and related)

mDuo13:
This massive overhaul of dev portal content removes references to Ripple-REST (which has been discontinued). In many cases, such references are replaced with references to its successor, RippleAPI. In some cases, there was too much and the references to REST were simply removed with no substitute. (I may later try to revisit those sections to add RippleAPI versions, but that's a low priority.)

This entirely removes the Ripple-REST article and the corresponding Ripple-REST API Tool.

Along the way, I also removed some other deprecated things, including the demurrage article (demurrage code has been removed) and a number of HTML files that were hanging around despite being removed (e.g. gateway services).

PR #159, opened 2016-01-29
RippleAPI Quick Start Guide

mDuo13:
(Not to be merged until after PR #158)

This creates a new RippleAPI Quick Start document, which shows basic usage of RippleAPI, including environment setup, in much greater detail than the API Reference does. It assumes very little familiarity with Node.js and modern JavaScript programming.

You can see the display version of the document here:
http://mduo13.github.io/ripple-dev-portal/rippleapi_quickstart.html


content/code_samples/rippleapi_quickstart/get-account-info.js
```diff
@@ -0,0 +1,24 @@
...
+  server: 'wss://s1.ripple.com' // Public rippled server
+});
+api.connect().then(() => {
+  /* begin custom code ------------------------------------ */
+  const my_address = 'rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn';
+
+  console.log('getting account info for', my_address);
+  return api.getAccountInfo(my_address);
+
+// info => {...} is just a shorter syntax for function(info) {...}
```
mDuo13: Clark says to drop this comment. Too pedantic.

content/code_samples/rippleapi_quickstart/get-account-info.js
```diff
@@ -0,0 +1,24 @@
...
+}).then(info => {
+  console.log(info);
+  console.log('getAccountInfo done');
+
+  /* end custom code -------------------------------------- */
+}).then(() => {
+  return api.disconnect();
+}).then(()=> {
```
clark800: Spacing is inconsistent; was eslint run on this?

content/rippleapi_quickstart.md
```diff
@@ -0,0 +1,280 @@
...
+### Script opening ###
+
+```
+'use strict';
+const {RippleAPI} = require('ripple-lib');
+```
+
+The opening line enables [strict mode](https://www.nczonline.net/blog/2012/03/13/its-time-to-start-using-javascript-strict-mode/). This is purely optional, but it helps you avoid some common pitfalls of JavaScript. See also: [Restrictions on Code in Strict Mode](https://msdn.microsoft.com/library/br230269%28v=vs.94%29.aspx#Anchor_1).
+
+The second line imports RippleAPI into the current scope using Node.js's require function. The [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) assigns it to the variable name `RippleAPI` instead of `ripple-lib` (which is the name of the package, for historical reasons).
```
clark800: I think you could drop "instead of ..." (if not this whole paragraph)

content/rippleapi_quickstart.md
```diff
@@ -0,0 +1,280 @@
...
+
+When a Promise finishes with its asynchronous operations, the Promise runs the callback function you passed it. The return value from the `then` method is another Promise object, so you can "chain" that into another `then` method, or the Promise's `catch` method, which also takes a callback. The callback you provide to `catch` gets called if something goes wrong.
+
+Finally, we have more new ECMAScript 6 syntax - an [arrow function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions). Arrow functions are just a shorter way of defining anonymous functions, which is pretty convenient when you're defining lots of one-off functions as callbacks, like we are here. The syntax `()=> {...}` is mostly equivalent to `function() {...}`. If you want an anonymous function with one parameter, you can use a syntax like `info => {...}` instead, which is basically just `function(info) {...}` as well.
```
clark800, line 163: For someone like me that likes to be precise, it's hard to read things like "mostly equivalent". They are not semantically equivalent, and addressing the difference is outside the scope of this document.

rippleapi_quickstart.html
```diff
@@ -0,0 +1,520 @@
...
+<h3 id="instantiating-the-api">Instantiating the API</h3>
+<pre><code>const api = new RippleAPI({
+  server: 'wss://s1.ripple.com' // Public rippled server
+});
+</code></pre>
+<p>This section creates a new instance of the RippleAPI class, assigning it to the variable <code>api</code>. (The <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const"><code>const</code> keyword</a> means you can't reassign the value <code>api</code> to some other value. The internal state of the object can still change, though.)</p>
+<p>The one argument to the constructor is an options object, which has <a href="rippleapi.html#parameters">a variety of options</a>. The <code>server</code> parameter tells it where it should connect to a <code>rippled</code> server.</p>
+<ul>
+<li>The example <code>server</code> setting uses a secure WebSocket connection to connect to one of the public servers that Ripple (the company) operates.</li>
+<li>If you don't include the <code>server</code> option, RippleAPI runs in <a href="rippleapi.html#offline-functionality">offline mode</a> instead, which severely limits what you can do with it.</li>
```
clark800: "severely" sounds a bit out of place, maybe just say something like "permits offline operations only"

content/code_samples/rippleapi_quickstart/get-account-info.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+const {RippleAPI} = require('ripple-lib');
+
+const api = new RippleAPI({
+  server: 'wss://s1.ripple.com' // Public rippled server
+});
+api.connect().then(() => {
+  /* begin custom code ------------------------------------ */
+  const my_address = 'rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn';
```
clark800: RippleAPI and javascript both use camelCase, I think it would be a bit more aesthetic to use it in the examples too.

content/rippleapi_quickstart.md
```diff
@@ -0,0 +1,280 @@
...
+
+Unfortunately, there are occasional tricky race conditions where very similar code might miss a confirmed transaction. This has to do with the fact that `rippled` returns a "not found" error if you look for a transaction that isn't in a ledger it has on hand. It's extremely unlikely to happen in the code as written, but it's possible that similar code with a larger interval between checks, or delayed due a power outage, may indicate that the transaction was not validated even when it was. This only occurs if the `rippled` server has an incomplete ledger history, either because it deletes or has not yet written the ledger where the submitted transaction was validated. The best solution is to check what ledger versions are available before taking any action to re-submit a transaction. See [Reliable Transaction Submission](reliable_tx.html) for a more thorough explanation.
```
clark800: RippleAPI uses a `MissingLedgerHistoryError` to address this concern.

content/rippleapi_quickstart.md
```diff
@@ -0,0 +1,280 @@
...
+
+# RippleAPI in Web Browsers #
+
+The process of using RippleAPI in a web browser is slightly different.
+
+## Build Instructions ##
+
+Before you can use RippleAPI in a browser, you need to compile a browser-compatible version. The following process creates a single JavaScript file you can include in a webpage.
+
+#### 1. Download a copy of the RippleAPI git repository.
```
clark800, line 231: Maybe we should re-enable the bower repo so users don't have to build from source?

content/rippleapi_quickstart.md
```diff
@@ -0,0 +1,280 @@
...
+npm WARN optional Skipping failed optional dependency /chokidar/fsevents:
+npm WARN notsup Not compatible with your operating system or architecture: fsevents@1.0.6
+```
+
+#### 3. Use Gulp to build a single JavaScript output
+
+RippleAPI comes with code to use the [gulp](http://gulpjs.com/) package to compile all its source code into browser-compatible JavaScript files. Gulp is automatically installed as one of the dependencies, so all you have to do is run it:
+
+```
+./node_modules/.bin/gulp
```
clark800: Can use `npm run build` here

content/code_samples/rippleapi_quickstart/get-account-info.js
mDuo13: It was... but maybe not on the most recent version?

clark800: Yeah probably, it looks like I added the rule for this just 12 days ago

rippleapi_quickstart.html
```diff
@@ -0,0 +1,520 @@
...
+&lt;html&gt;
+&lt;head&gt;
+  &lt;script src="ripple-0.16.5.js"&gt;&lt;/script&gt;
+  &lt;script&gt;
+    console.log(ripple);
+    var api = new ripple.RippleAPI({server:'wss://s1.ripple.com/'});
+    api.connect().then(()=&gt;{
+        return api.getServerInfo();
+    }).then(server_info=&gt;{
+    document.body.innerHTML += `&lt;p&gt;Connected to rippled server!&lt;/p&gt;
```
clark800: Might be good to note that this won't work in internet explorer

content/rippleapi_quickstart.md
mDuo13: Agree I could drop the whole paragraph, especially if I switch the samples to use the `require('ripple-lib').RippleAPI` syntax we talked about. In its place, I'll link to [a summary of other things ripple-lib exports](https://github.com/ripple/ripple-lib/blob/develop/src/index.js).

content/rippleapi_quickstart.md
mDuo13, line 231: I don't know what you're talking about, but it sounds like it's directed at the ripple-lib developers, not at the document.

clark800:
Ok, I'm done reviewing: looks good aside from comments! Just to capture the main points from our discussion on slack:
1. I think we should remove reference to Babel now that node officially supports much of ES6 (and Babel became more complicated!)
2. I think the embedded Javascript tutorial is a distraction and out of scope for this document

content/rippleapi_quickstart.md
clark800, line 231: Yes, I mean that we could remove this section if we re-enable bower releases. Might not happen before this document is released, but we should try to remember to update this if we do.

rippleapi_quickstart.html
mDuo13: Can we modify the script somehow so that it will? I'm pretty out of touch with the limitations of recent versions of IE.

clark800: I think you would have to get rid of the template string, which is an ES6 feature

content/code_samples/rippleapi_quickstart/get-account-info.js
mDuo13: I got my eslintrc from here -- which hasn't been updated since September:
https://github.com/ripple/javascript-style-guide/blob/master/eslintrc

clark800: Ah, the master branch is still the ES5 version, the ES6 version is on the es6 branch.

mDuo13: Updated the eslintrc, code fixed and now checks out against the new version.

content/rippleapi_quickstart.md
mDuo13, line 163: I'm going to leave this here, because I think it's useful for people who are still getting used to modern JavaScript. The MDN link for "arrow function" actually _does_ go into the details of how an arrow function is different from an ordinary anonymous function, so I think that's sufficient.

content/code_samples/rippleapi_quickstart/get-account-info.js
mDuo13: Changed.

content/rippleapi_quickstart.md
mDuo13: Updated this section.

mDuo13:
I've made various changes to the document and pushed changes. Regarding the comments from our Slack discussion:
1. I removed the destructuring assignments as per our discussion and confirmed that the `get-account-info.js` code actually runs in the current 5.x release of Node.js. However, I'm hesitant to remove mention of Babel-Node entirely, because I'm not sure which versions of Node support what and I'd rather not be the QA engineer who tests all those out. In order to reach audiences on OSes like Ubuntu LTS and CentOS, it's probably best to support a wide range of Node versions. That said, I've made [DEV-183](https://ripplelabs.atlassian.net/browse/DEV-183) to readdress this in the future.
2. Based on comments I've gotten from others, the JavaScript information is useful. I think it's easy enough to skim over if you already know it, and it's already written, so keeping it is a net positive.

Let me know if the other changes look good to you.

clark800:
Changes in the latest commit look good

mDuo13:
closing in favor of #164.

PR #161, opened 2016-02-04
Do not recommend m3.medium instance for rippled

wilsonianb:
@mtrippled


mDuo13:
Can you cd to the `tool` folder and re-gen the docs with `./parse_pages.py`?

Also, I think we may want to collect more data before we rule out m3.medium. Tim was having trouble with NuDB, but that's not the default.

wilsonianb:
I'm fine waiting to see what we decide.

PR #174, opened 2016-03-29
[Do not merge yet] Simple Multi-Sign Documentation

mDuo13:
Multi-signing
Amendments
Testing amendments in stand-alone mode
Fee Escalation
(and related tweaks)


content/reference-ledger-format.md
```diff
@@ -398,3 +398,72 @@ The `index` of a RippleState node is the SHA-512Half of the following values put
 * The AccountID of the low account
 * The AccountID of the high account
 * The 160-bit currency code of the trust line(s)
+
+
+## SignerList ##
+[[Source]<br>](https://github.com/ripple/rippled/blob/6d2e3da30696bd10e3bb11a5ff6d45d2c4dae90f/src/ripple/protocol/impl/LedgerFormats.cpp#L127 "Source")
+
+The `SignerList` node type represents a list of parties that, as a group, are authorized to sign a transaction in place of an individual account.
```
ryangyoung: _Nitpick:_ I think it might be slightly clearer to omit the commas, i.e., "The `SignerList` node type represents a list of parties that are authorized to sign a transaction as a group in place of an individual account." Will defer to you, though, if there's a reason not to do that.

content/reference-ledger-format.md
```diff
@@ -398,3 +398,72 @@ The `index` of a RippleState node is the SHA-512Half of the following values put
...
+```
+
+A SignerList node has the following fields:
+
+| Name            | JSON Type | Internal Type | Description |
+|-----------------|-----------|---------------|-------------|
+| OwnerNode       | String    | UInt64        | A hint indicating which page of the owner directory links to this node, in case the directory consists of multiple nodes. |
+| SignerQuorum    | Number    | UInt32        | A target number for signer weights. To produce a valid signature for the owner of this SignerList, the signers must provide valid signatures whose weights sum to this value or more. |
+| SignerEntries   | Array     | Array         | An array of SignerEntry objects representing the parties who are part of this signer list. |
+| SignerListID    | Number    | UInt32        | An ID for this signer list. Currently always set to `0`. If a future update allows multiple signer lists for an account, this may change. |
```
ryangyoung: Does "a future update" mean that another amendment is needed to allow multiple signer lists?

content/reference-ledger-format.md
```diff
@@ -398,3 +398,72 @@ The `index` of a RippleState node is the SHA-512Half of the following values put
...
+| PreviousTxnID   | String    | Hash256       | The identifying hash of the transaction that most recently modified this node. |
+| PreviousTxnLgrSeq | Number  | UInt32        | The sequence number (`ledger_index`) of the ledger that contains the transaction that most recently modified this node. |
+
+The `SignerEntries` may be any combination of funded and unfunded addresses that use either secp256k1 or ed25519 keys.
+
+### SignerEntry object ###
```
ryangyoung: Another nitpick: shouldn't all headers be in Title Case? Currently, there seems to be an inconsistent mix in this article. (I'm probably guilty of this, too).

content/reference-ledger-format.md
```diff
@@ -398,3 +398,72 @@ The `index` of a RippleState node is the SHA-512Half of the following values put
...
+
+| Name            | JSON Type | Internal Type | Description |
+|-----------------|-----------|---------------|-------------|
+| OwnerNode       | String    | UInt64        | A hint indicating which page of the owner directory links to this node, in case the directory consists of multiple nodes. |
+| SignerQuorum    | Number    | UInt32        | A target number for signer weights. To produce a valid signature for the owner of this SignerList, the signers must provide valid signatures whose weights sum to this value or more. |
+| SignerEntries   | Array     | Array         | An array of SignerEntry objects representing the parties who are part of this signer list. |
+| SignerListID    | Number    | UInt32        | An ID for this signer list. Currently always set to `0`. If a future update allows multiple signer lists for an account, this may change. |
+| PreviousTxnID   | String    | Hash256       | The identifying hash of the transaction that most recently modified this node. |
+| PreviousTxnLgrSeq | Number  | UInt32        | The sequence number (`ledger_index`) of the ledger that contains the transaction that most recently modified this node. |
```
ryangyoung: I think I understand why it must be done, but saying "The sequence number (`ledger_index`)" is a little confusing. Perhaps add a few explanatory words or a link in the parentheses.

content/reference-ledger-format.md
```diff
@@ -398,3 +398,72 @@ The `index` of a RippleState node is the SHA-512Half of the following values put
...
+
+The `SignerEntries` may be any combination of funded and unfunded addresses that use either secp256k1 or ed25519 keys.
+
+### SignerEntry object ###
+
+Each member of the `SignerEntries` field is an object that describes that signer in the list. A SignerEntry has the following fields:
+
+| Name            | JSON Type | Internal Type | Description |
+|-----------------|-----------|---------------|-------------|
+| Account         | String    | AccountID     | An address whose signature contributes to the multi-signature. This does not need to be a funded Ripple account. |
```
ryangyoung: But it has to be an Ripple address?

content/reference-ledger-format.md
```diff
@@ -398,3 +398,72 @@ The `index` of a RippleState node is the SHA-512Half of the following values put
...
+| SignerWeight    | Number    | UInt16        | The weight of signatures from this signer. A multi-signature is only valid of the sum weight of the signatures provided meets or exceeds the SignerList's `SignerQuorum` value. |
```
ryangyoung: Typo: "A multi-signature is only valid of the sum weight **_if**_ the signatures provided..."

Should the first line be "The weight of this signer's signature"? As is, it sounds like the signer might have to provide the number of signatures equal to the weight, or something like that.

content/reference-ledger-format.md
```diff
@@ -398,3 +398,72 @@ The `index` of a RippleState node is the SHA-512Half of the following values put
...
+
+When processing a multi-signed transaction, the server dereferences the `Account` values with respect to the ledger at the time of transaction execution. If the address _does not_ correspond to a funded [AccountRoot node](#accountroot), then only the master secret associated with that address can be used to produce a valid signature. If the account _does_ exist in the ledger, then it depends on the state of that account. If the account has a Regular Key configured, the Regular Key can be used. The account's master key can only be used if it is not disabled. Even if the account has a SignerList configured, a multi-signature cannot be used as a valid component to another multi-signature. (In other words, "multi-level" multi-signing is disallowed.)
```
ryangyoung: Maybe add a hyphen to "dereferences". I read it as "deferences" the first time through.

The last two sentences before the final parenthetical sentence are a little confusing (probably somewhat necessarily so). I'm missing the connection between the master key sentence (makes sense: it has to be enabled to use it) and the next one. If an account has configured a SignerList _and_ disabled the master key, does that mean they can't be a signer?

content/reference-ledger-format.md
```diff
@@ -398,3 +398,72 @@ The `index` of a RippleState node is the SHA-512Half of the following values put
...
+
+### SignerLists and Reserves ###
+
+A SignerList contributes to the [Account Reserve](https://wiki.ripple.com/Reserves). The SignerList itself counts as two objects, and each member of the list counts as one, so that the total owner reserve associated with a SignerList is anywhere from 3 times to 10 times the reserve required by a single trust line ([RippleState](#ripplestate)) or [Offer](#offer) node in the ledger.
```
ryangyoung: Suggest breaking this up into at least two (maybe even three) sentences.

content/reference-transaction-format.md
```diff
@@ -12,27 +12,48 @@ There are several different types of transactions that perform different actions
...
+
+* Only the master key can [disable the master key](#accountset-flags).
+* Only the master key can [permanently give up the ability to freeze](concept-freeze.html#no-freeze).
+* You can never remove the last method of signing transactions from an address.
+
+## Signing and Submitting Transactions ##
+
+Sending a transaction to the Ripple Consensus Ledger involves several steps:
+
+1. Create an unsigned transaction in JSON format.
```
ryangyoung: Maybe a link to give the user an idea of what an unsigned transaction looks like? e.g. https://ripple.com/build/transactions/#common-fields

Ah, you provide an example later on. Still think some kind of link would be good, though.

content/reference-transaction-format.md
```diff
@@ -157,6 +178,16 @@ After a transaction has been submitted, if it gets accepted into a validated led
 }
 ```
 
+
+### Multi-Signing ###
+
+Multi-signing in Ripple is the act of [authorizing transactions](#authorizing-transactions) for the Ripple Consensus Ledger by using a combination of multiple secret keys. Multi-signing is due to be enabled by an [Amendment](concept-amendments.html) to the Ripple Consensus Protocol. You can use multi-signing in addition to, or instead of, a master key, a [regular key](#setregularkey), or both.
```
ryangyoung: The last part of the last sentence is slightly unclear to me. You can use multi-signing and one or both of the other methods, or you can use multi-signing in addition to one or both of the other methods?

content/reference-transaction-format.md
```diff
@@ -935,13 +1062,20 @@ These codes indicate that the transaction was malformed, and cannot succeed acco
...
 | temINVALID | The transaction is otherwise invalid. For example, the transaction ID may not be the right format, the signature may not be formed properly, or something else went wrong in understanding the transaction. |
 | temINVALID\_FLAG | The transaction includes a [Flag](#flags) that does not exist, or includes a contradictory combination of flags. |
 | temREDUNDANT | The transaction would accomplish nothing; for example, it is sending a payment directly to the sending account, or creating an offer to buy and sell the same currency from the same issuer. |
 | temREDUNDANT\_SEND\_MAX | _(Removed in [rippled 0.28.0][])_ |
 | temRIPPLE\_EMPTY | The [Payment](#payment) transaction includes an empty `Paths` field, but paths are necessary to complete this payment. |
+| temBAD_WEIGHT | The [SignerListSet](#signerlistset) transaction includes a `SignerWeight` that is invalid, for example a zero or negative value. |
+| temBAD_SIGNER | The [SignerListSet](#signerlistset) transaction includes a signer who is invalid: for example, it might be a duplicate, or it might be the account to which the SignerList belongs. |
```
ryangyoung: Does "the account to which the SignerList belongs" mean that it's the owner of the SignerList?

content/reference-transaction-format.md
```diff
@@ -1009,18 +1147,17 @@ These codes indicate that the transaction failed, but it was applied to a ledger
...
 | tecUNFUNDED | 129 | **DEPRECATED.** Replaced by tecUNFUNDED\_OFFER and tecUNFUNDED\_PAYMENT. |
-| tecMASTER\_DISABLED | 130 | The [SetRegularKey transaction](#setregularkey) tried to unset the Regular Key, but the account has the `lsfDisableMaster` flag enabled. (Unsetting the Regular Key while also leaving the Master Key disabled would make the account unusable.) |
-| tecNO\_REGULAR\_KEY | 131 | The [AccountSet transaction](#accountset) tried to disable the Master Key, but the account does not have a Regular Key set. (Disabling the Master Key without having a Regular Key configured would make the account unusable.) |
+| tecNO\_ALTERNATIVE\_KEY | 130 | The transaction tried to remove the only available method of [authorizing transactions](#authorizing-transactions). This could be a [SetRegularKey transaction](#setregularkey) to remove the regular key, a [SignerListSet transaction](#signerlistset) to delete a SignerList, or an [AccountSet transaction](#accountset) to disable the master key. (Prior to [rippled 0.30.0](https://github.com/ripple/rippled/releases/tag/0.30.0), this was called `tecMASTER_DISABLED` instead.) |
```
ryangyoung: _Nitpick:_ Don't think "instead" is needed in the last parenthetical.

content/reference-ledger-format.md
scottschurr: Yup.  It would take writing a bunch of code and then introducing an amendment.

content/reference-ledger-format.md
scottschurr: I stumbled a bit on the first sentence.  Perhaps, "The weight of a signature from this signer."?  And I completely missed the typo in the second sentence.  Good eyes, @ryangyoung.

content/reference-ledger-format.md
ryangyoung: Ah! Then maybe "future update" should be "future amendment" with a link to the amendments page to make it totally clear.

content/reference-ledger-format.md
scottschurr: I would be inclined to omit the last two sentences.  Yeah, I implemented 2-level multi-sign and then threw it out.  But it might be best if we don't even suggest that multi-level multi-sign is a consideration at this point.

content/reference-transaction-format.md
```diff
@@ -157,6 +178,16 @@ After a transaction has been submitted, if it gets accepted into a validated led
 }
 ```
 
+
+### Multi-Signing ###
+
+Multi-signing in Ripple is the act of [authorizing transactions](#authorizing-transactions) for the Ripple Consensus Ledger by using a combination of multiple secret keys. Multi-signing is due to be enabled by an [Amendment](concept-amendments.html) to the Ripple Consensus Protocol. You can use multi-signing in addition to, or instead of, a master key, a [regular key](#setregularkey), or both.
+
+The [SignerListSet transaction](#signerlistset) defines which addresses can authorize transactions from your address. You can include up to 8 addresses in a SignerList. You can achieve several different configurations using the weight and quorum values of the SignerList.
```
scottschurr: Saying "You can achieve several different configurations" is a bit blurry to me.  Consider something along the lines of "You can give your account one of several different permission models by choosing different the weight and quorum values in the SignerList."  Does "permission models" come close to what the SignerList controls?  Hmmmm...

content/reference-transaction-format.md
```diff
@@ -279,10 +314,25 @@ Example of a transaction with a Memos field:
...
+| TxnSignature | String | Blob | A signature for this transaction, verifiable using the `SigningPubKey`. |
+| SigningPubKey | String | PubKey | The public key used to create this signature. |
+
+The `SigningPubKey` must be a key that is associated with the `Account` address. If the referenced `Account` is a funded account in the ledger, then the SigningPubKey can be that account's current Regular Key if one is set. It could also be that account's Master Key, unless the [lsfDisableMaster](ripple-ledger.html#accountroot-flags) flag is enabled. If the referenced `Account` address is not a funded account in the ledger, then the `SigningPubKey` must be the master key associated with that address.
+
+Because signature verification is a compute-intensive task, multi-signed transactions cost additional XRP to relay to the network. Each signature included in the multi-siganture increases the `Fee` required by the amount of the minimum transaction fee. For example, if the current minimum `Fee` value to relay a transaction to the network is `10000` drops, then a multi-signed transaction with 3 entries in the `Signers` array would cost `40000` drops to relay.
```
scottschurr: siganture -> signature.

I find "increases the `Fee` required by the amount of the minimum transaction fee" hard to comprehend.  Perhaps less detail here?  Consider: "Each signature included in the multi-signature increases the required `Fee`."  I think the "For example..." sentence gives a good description of how the fee increases.

content/reference-transaction-format.md
```diff
@@ -520,13 +570,15 @@ A SetRegularKey transaction changes the regular key used by the account to sign
...
+A regular key pair is generated in the same way as any other Ripple keys (for example, with [wallet_propose](reference-rippled.html#wallet-propose)), but it can be changed. A master key pair is an intrinsic part of an address's identity (the address is derived from the master public key). You can [disable](#accountset-flags) a master key but you cannot change it.
 
-If your regular key is compromised, but the master key is not, you can use this method to regain control of your account. In some cases, you can even send a [key reset transaction](concept-transaction-cost.html#key-reset-transaction) without paying the [transaction cost](#transaction-cost).
+You can protect your master secret by using a regular key instead of the master key to sign transactions where possible. If your regular key is compromised, but the master key is not, you can use a SetRegularKey transaction to regain control of your address. In some cases, you can even send a [key reset transaction](concept-transaction-cost.html#key-reset-transaction) without paying the [transaction cost](#transaction-cost).
```
scottschurr: Consider "where" -> "whenever".  Just a thought.

content/reference-transaction-format.md
```diff
@@ -730,6 +782,58 @@ Transactions of the TrustSet type support additional values in the [`Flags` fiel
...
+
+You cannot create a SignerList such that the SignerQuorum could never be met. The SignerQuorum must be greater than 0 but less than or equal to the sum of the `SignerWeight` values in the list. Otherwise, the transaction fails with the error [temMALFORMED](#tem-codes).
+
+You can create, update, or remove a SignerList using the master key, regular key, or the current SignerList, if those methods of signing transactions are available.
+
+You cannot remove the last method of signing transactions from an account. If an account's master key is disabled (it has the [`lsfDisableMaster` flag](ripple-ledger.html#accountroot-flags) enabled) and the account does not have a [Regular Key](#setregularkey) configured, then you cannot delete the SignerList from the account. Instead, the transaction fails with the error [tecNO\_ALTERNATIVE\_KEY](#tec-codes).
+
+
```
scottschurr, line 853: The SignerListSet section looks great!

content/reference-transaction-format.md
```diff
@@ -730,6 +782,58 @@ Transactions of the TrustSet type support additional values in the [`Flags` fiel
...
+                "SignerWeight": 1
+            }
+        }
+    ]
+}
+```
+
+| Field | JSON Type | [Internal Type](https://wiki.ripple.com/Binary_Format) | Description |
+|-------|-----------|--------------------------------------------------------|-------------|
+| SignerQuorum | Number | UInt32 | A target number for the signer weights. A multi-signature from this list is valid only if the sum weights of the signatures provided is equal or greater than this value. To delete a SignerList, use the value `0`. |
```
scottschurr: Did you perhaps want to write "equal to" instead of just "equal"?  I can make the same sense out of both forms, but for some reason "equal to" feels like more standard English.  Your call.

content/reference-transaction-format.md
```diff
@@ -773,6 +877,29 @@ A change in [transaction cost](concept-transaction-cost.html) or [account reserv
...
+### EnableAmendment Flags ###
+
+The `Flags` value of the EnableAmendment pseudo-transaction indicates the status of the amendment at the time of the ledger including the pseudo-transaction.
+
+A `Flags` value of `0` (no flags) indicates that the amendment has been enabled, and applies to all ledgers afterward. Other `Flags` values are as follows:
+
+| Flag Name | Hex Value | Decimal Value | Description |
+|-----------|-----------|---------------|-------------|
+| tfGotMajority | 0x00010000 | 65536 | Support for this amendment increased to at least 80% of trusted validators starting with this ledger version. |
+| tfLostMajority | 0x00020000 | 131072 | Support for this amendment decreased to less than 80% of trusted validators starting with this ledger version. |
```
scottschurr, line 919: I'm not really qualified as a technical reviewer of the EnableAmendment section.  But I read it anyway and it made sense to me.  You might want to ask Nik Bougalis to look it over in case I missed something.

content/reference-transaction-format.md
scottschurr: Hmmm.  I'm working on "it might be the account to which the SignerList belongs."  "Belonging" is not quite the right idea here.  Since the transaction is rejected this SignerList is never owned by anything.  Maybe "it might be the account to which the SignerList is being assigned."?  I don't know whether that helps...

content/tutorial-multisign.md
```diff
@@ -0,0 +1,603 @@
...
+At this point, your address is ready to [send a multi-signed transaction](#sending-a-multi-signed-transaction). You may also want to:
+
+* Disable the address's master key by sending an [AccountSet transaction](reference-transaction-format.html#accountset) using the `asfDisableMaster` flag.
+* Remove the address's regular key (if you previously set one) by sending a [SetRegularKey transaction](reference-transaction-format.html#setregularkey).
+
+
+Sending a Multi-Signed Transaction
+----------------------------------
+
+Before you can multi-sign a transaction, first check that [multi-sign is available](#availability-of-multi-signing) and [set up multi-signing](#set-up-multi-sign) for your address. The following procedure demonstrates how to create, sign, and submit a multi-signed transaction.
```
scottschurr: FWIW, if you successfully [set up multi-signing] then [multi-sign is available].  So the caution may be more appropriate in that section.  Just for your consideration.

content/reference-transaction-format.md
mDuo13: I have added a header to the unsigned transaction format and linked it.

content/reference-transaction-format.md
mDuo13: I've revised this sentence.

content/reference-transaction-format.md
mDuo13: I don't want to use the word "permission" because I think that implies more fine-grained control over sending different types of transactions than multi-signing currently provides.

I went with this instead:

> You can control how many signatures are needed, in which combinations, by using the quorum and weight values of the SignerList.

content/reference-transaction-format.md
mDuo13: Fixed this.

content/reference-transaction-format.md
mDuo13: Good call.

content/reference-transaction-format.md
mDuo13: Changed to "greater than or equal to".

content/reference-transaction-format.md
mDuo13: I rephrased it thus:

> For example, there may be duplicate entries, or the owner of the SignerList may also be a member.

In this case, I chose to disregard the fact that the SignerList technically doesn't have an owner in the ledger yet. As I see it, the owner of the list-as-it-appears-in-the-transaction is the sender. That's less confusing than saying something like, "the would-be owner of this SignerList."

content/reference-transaction-format.md
mDuo13: Fixed.

content/tutorial-multisign.md
mDuo13: Good point. I moved the part about checking whether multi-signing is available to the Setting Up Multi-Signing section.

content/reference-ledger-format.md
mDuo13: I'm worried rephrasing it as you suggested makes "as a group" into a misplaced modifier which could seem to refer to the transaction, not the parties.

I'd prefer to leave this one as-is.

content/reference-ledger-format.md
mDuo13: Good call. When I first wrote this, we didn't have amendment documentation, but now we do, so I've linked it.

content/reference-ledger-format.md
mDuo13: Good point.

content/reference-ledger-format.md
ryangyoung: Agreed.

content/reference-ledger-format.md
mDuo13: I added an include defining ledger index and linked it from here and many other similar places in the article.

content/reference-ledger-format.md
mDuo13: That question gets into the finer points of what constitutes a "Ripple address." Does it have to be funded to be an address? What if it's not funded, but potentially could be? Is a base-58-encoded validator pubkey a Ripple address?

That said, it's probably clearer to say, "A Ripple address" here, so I changed it.

content/reference-transaction-format.md
scottschurr: That looks great.  Thanks.

content/reference-ledger-format.md
mDuo13: Fixed both.

content/reference-transaction-format.md
scottschurr: That looks good.

content/reference-ledger-format.md
mDuo13: Regarding "dereferences" - it's tricky, but that's the proper spelling of the word, so I'm not going to add a hyphen.

As for the last two sentences, I trimmed them down to this:

> A multi-signature cannot be used as a component of another multi-signature.

content/reference-ledger-format.md
mDuo13: Done.

content/concept-amendments.md
```diff
@@ -0,0 +1,164 @@
+# Amendments #
+
+The Amendments system provides a means of introducing new features to the decentralized Ripple consensus network without causing disruptions. The amendments system works by utilizing the core consensus process of the network to approve any changes by showing continuous support before those changes go into effect. An amendment normally requires **80% support for two weeks** before it can apply.
+
+When an Amendment has been enabled, it applies permanently to all ledger versions after the one that included it. You cannot disable an Amendment, unless you introduce a new Amendment to do so.
+
+
+## Background ##
+
+Any changes to transaction processing could cause servers to build a different ledger with the same set of transactions. If only a portion of validators have upgraded to a new version of the software, this could cause anything from minor inconveniences (a minority of servers spend more time and bandwidth fetching the actual consensus ledger because they cannot rely on their own) to serious problems: consensus might be unable to validate new ledger versions because servers are unable to reach a consensus who agree to the exact same result.
```
ryangyoung: This is a long sentence that embeds two relatively complicated examples in it via two different methods (parentheses and a colon). I suggest breaking it up to give more dedicated space to each example and severity level.

content/reference-ledger-format.md
ryangyoung: Cool. That sounds good.

content/reference-ledger-format.md
ryangyoung: Interesting. Since you said it "does not need to be a funded Ripple ~~account~~ address," it made me wonder if anything other than a _Ripple_ address (or account or whatever) could sign. (I'm pretty sure the answer is no.) But, yeah, also probably a little clearer to use address than account, too.

content/reference-ledger-format.md
ryangyoung: Sounds good to me, on both counts.

content/concept-amendments.md
mDuo13: Good point. Even after revision, it's a bit of a challenge, but I think it's better:

> If only a portion of validators have upgraded to a new version of the software, this could cause anything from minor inconveniences to full outages. In the minor case, a minority of servers spend more time and bandwidth fetching the actual consensus ledger because they cannot build it using the transaction processing rules they already know. In the worst case, consensus might be unable to validate new ledger versions because servers with different rules could not reach a consensus on the exact ledger to build.

I wanted to say "cannot" instead of "could not" in the last sentence, but that makes it into a general statement about consensus that's not true. (Servers with different rules _can_ reach a consensus in some situations, namely, the cases where they omit anything they disagree on.)

ryangyoung: The revision looks a lot better. I thought "could" worked well even before you pointed out the technical aspects. 

In the last sentence, should it be "the consensus process" in the first instance (i.e., "In the worst case, the consenus process might be unable....")?

content/concept-amendments.md
```diff
@@ -0,0 +1,164 @@
...
+
+Amendments provide solution to this problem, so that new features can be enabled only when enough validators support those features.
```
ryangyoung: Typo: "Amendments provide **a** solution..."

content/concept-amendments.md
```diff
@@ -0,0 +1,164 @@
...
+
+Users and businesses who rely on the Ripple Consensus Ledger can also use Amendments to provide advance notice of changes in transaction processing that might affect their business. However, API changes that do not impact transaction processing or [the consensus process](https://ripple.com/knowledge_center/the-ripple-ledger-consensus-process/) do not require Amendments.
+
+
+## Amendment Process ##
+
+An amendment is a fully-functional feature or change, ready to apply when a consensus of servers can handle it. A `rippled` server that wants to use an amendment has business logic for two modes: without the amendment (previous behavior) and with the amendment (new behavior). Every amendment has a unique identifying hex value and a short name. The short name is for human use, and does not affect the status of the amendment.
```
ryangyoung: "...ready to apply when a consensus of servers can handle it":
- What is a "consensus" of servers? 
- Is it strictly when the servers can handle it, or when they agree to the amendment? I assume that being able to handle it would be part of their decision whether to support it. 

The part about a business using an amendment for business logic seems to belong in some other section, perhaps about uses of amendments. As is, it seems like a non-sequitur in this paragraph. 

I'm also confused by the last phrase--why would any kind of name affect the status of the amendment? 

Generally, this paragraph seems to be answering the question "What is an amendment and how can it be used?" than "What is the amendment process?" The answer to the latter seems to start with the next paragraph.

content/concept-amendments.md
```diff
@@ -0,0 +1,164 @@
...
+
+Every 256th ledger is called a "flag" ledger. The process of approving an amendment starts in the ledger version immediately before the flag ledger: at this time, `rippled` validator servers submit votes in favor of specific amendments alongside their validations for that ledger. ([Fee Voting](concept-fee-voting.html) also occurs around flag ledgers.)
```
ryangyoung: I stumbled at "...at this time" because I thought it might mean "at the time of this writing." Maybe make it more specific: "at this ledger," or something like that.

content/concept-amendments.md
mDuo13: Good catch. Fixed.

content/concept-amendments.md
```diff
@@ -0,0 +1,164 @@
...
+
+In the flag ledger itself, there is nothing unusual. However, during that time, the servers look at the votes of the validators they trust, and decide whether to insert an [`EnableAmendment` pseudo-transaction](reference-transaction-format.html#enableamendment) into the following ledger. The flags of an EnableAmendment pseudo-transaction indicate what the server thinks happened:
```
ryangyoung: Consider replacing the first sentence with something like "Flag ledgers are no different than other ledgers". Saying "there is nothing unusual" makes it sound slightly suspicious.

ryangyoung: And instead of "what the server thinks happened" perhaps "what happened according to the server"?

content/concept-amendments.md
```diff
@@ -0,0 +1,164 @@
...
+
+* The `tfGotMajority` flag means that support for the amendment has increased to at least 80% of trusted validators.
+* The `tfLostMajority` flag means that support for the amendment has decreased to less than 80% of trusted validators.
+* An EnableAmendment pseudo-transaction with no flags means that support for the amendment has been enabled. (The change in transaction processing applies to every ledger after this one.)
+
+A server only inserts the pseudo-transaction to enable an amendment if all of the following conditions are met:
+
+* The amendment has not already been enabled.
+* A previous ledger includes an EnableAmendment pseudo-transaction for this amendment with the `tfGotMajority` flag enabled.
+* The previous ledger in question is an ancestor of the current ledger.
```
ryangyoung, line 41: Is there a case where any previous ledger _wouldn't_ be an ancestor of the current ledger?

content/concept-amendments.md
```diff
@@ -0,0 +1,164 @@
...
+* The previous ledger in question has a close time that is at least **two weeks** before the close time of the latest flag ledger.
```
ryangyoung, line 42: Is there a reason to say "has a close time" instead of just "closed," as in "The previous ledger in question closed at least two weeks before the latest flag ledger closed"?  That phrasing makes more intuitive sense to me (which still doesn't mean it's better or more accurate).

content/concept-amendments.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Amendment Voting ##
+
+Operators of `rippled` validators can choose which amendments to support or reject using the [`feature` command](reference-rippled.html#feature). This decides which amendments the validator votes for in the [amendment process](#amendment-process). By default, `rippled` votes in favor of every amendment it knows about.
+
+The operator of a `rippled` validator can "veto" an amendment. In this case, that validator never sends a vote in favor of the amendment. If enough servers veto an amendment, that prevents it from reaching consistent 80% support, so the amendment does not apply.
+
+As with all aspects of the consensus process, amendment votes are only taken into account by servers that trust the validators sending those votes. Currently, Ripple recommends only trusting the 5 default validators that Ripple (the company) operates. For now, trusting only those validators is sufficient to coordinate with Ripple on releasing new features.
```
ryangyoung: Would it be worth linking to the validator registry here? 

Also, does the last instance of "Ripple" in the last sentence refer to the company or the network? (Sorry about this one.)

content/concept-amendments.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Configuring Amendment Voting ###
+
+You can temporarily configure an amendment using the [`feature` command](reference-rippled.html#feature). To make a persistent change to your server's support for an amendment, modify your server's `rippled.cfg` file.
+
+Use the `[veto_amendments]` stanza to list amendments you do not want the server to vote for. Each line should contains one amendment's unique ID, optionally followed by the short name for the amendment. For example:
```
ryangyoung: Typo: "Each line should **contain** one amendment's..."

content/concept-amendments.md
```diff
@@ -0,0 +1,164 @@
...
+
+```
+[veto_amendments]
+C1B8D934087225F509BEB5A8EC24447854713EE447D277F69545ABFA0E0FD490 Tickets
+DA1BD556B42D85EA9C84066D028D355B52416734D3283F85E216EA5DA6DB7E13 SusPay
+```
+
+Use the `[amendments]` stanza to list amendments you want to vote for. (Even if you do not list them here, by default a server votes for all the amendments it knows how to apply.) Each line should contains one amendment's unique ID, optionally followed by the short name for the amendment. For example:
```
ryangyoung: Typo: "Each line should **contain** one amendment's..."

content/concept-amendments.md
mDuo13: You're right that this is answering the question, "What is an amendment?" I picked this place to elaborate in more detail because you only really need the detail in order to understand the process.

I think you got confused by the term of art, "business logic" -- I really just mean computer code, though, so I changed it.

As for the phrase about why a name would affect the status of the amendment -- different servers can have different names for the same amendment ID. But that doesn't stop them from both supporting the amendment.

ryangyoung: Right, I was thinking of how an operator might implement some logic to veto the amendment or not, but I see now that that's explained later on. I agree that using "code" would prevent this confusion. 

I think I get what you're describing by "a consensus of servers" but I still think that  could be clarified a little more (if it hasn't been already). 

That's a good explanation for that phrase about the name and ID. It should go into the doc!

content/concept-amendments.md
```diff
@@ -0,0 +1,164 @@
...
+| 4C97EBA926031A7CF7D7B36FDE3ED66DDA5421192D63DE53FFB46E43B9DC8373 |
+
+Introduces simple [multi-signing](reference-transaction-format.html#multi-signing) as a way to authorize transactions. Creates the [`SignerList` ledger node type](reference-ledger-format.html#signerlist) and the [`SignerListSet` transaction type](reference-transaction-format.html#signerlistset). Adds the optional `Signers` field to all transaction types. Modifies some transaction result codes.
+
+This amendment allows addresses to have a list of signers who can authorize transactions from that address in a multi-signature. The list has a quorum and 1 to 8 weighted signers. This allows various configurations, such as "any 3-of-5" or "signature from A plus any other two signatures."
+
+Signers can be funded or unfunded addresses. Funded addresses in a signer list can sign using a regular key (if defined) or master key (unless disabled). Unfunded addresses can sign with a master key. This amendment does not allow second-level multi-signing (signers using multi-signatures to contribute to multi-signatures). Multi-signed transactions have the same permissions as transactions signed with a regular key.
```
ryangyoung: Should this be updated to go along with this change?: https://github.com/ripple/ripple-dev-portal/pull/174/files/482b453602c6e83ac4de2dd2a84621d3f2b7724b#r58143834

content/concept-amendments.md
mDuo13: Yep. Changed it and made it a (redundant) link to the Consensus Process article.

content/concept-amendments.md
mDuo13: I thought I had mentioned the name/ID thing in this doc already, but it seems I hadn't! It must've appeared in another doc. You're right I should bring that up here.

I made the first part an "About Amendments" section (wish I had a better name, but my other attempt, "Properties of Amendments", was worse). That section's been rewritten as follows:

> An amendment is a fully-functional feature or change, waiting to be enabled by the peer-to-peer network as a part of the consensus process. A `rippled` server that wants to use an amendment has code for two modes: without the amendment (previous behavior) and with the amendment (new behavior).
> 
> Every amendment has a unique identifying hex value and a short name. The short name is for human use, and is not used in the amendment process. Two servers can support the same amendment ID while using different names to describe it. An amendment's name is not guaranteed to be unique.

content/concept-amendments.md
mDuo13: Here's my rephrase:

> The process of approving an amendment starts in the ledger version immediately before the flag ledger. When `rippled` validator servers send validation messages for that ledger, those servers also submit votes in favor of specific amendments.

content/concept-amendments.md
mDuo13: Changed the first line to:

> The flag ledger itself has no special contents.

I prefer, "what the server thinks happened," because it's concise and conveys a reasonable sense of the uncertainty involved. Also, if we are to stop personifying servers, I'll have to edit quite a few more passages. (Just earlier in this paragraph there's a hat trick of server personification! "the servers _look_ at the votes of the validators _they trust_, and _decide_...")

content/concept-amendments.md
ryangyoung: That looks great. 

While it could be overused, I like "About <Topic>" as a title because that generally answers my question of "what is this thing exactly?" Actually, it might be good to start using that across all the docs so that we (and users) can find easily find conceptual information.

content/concept-amendments.md
ryangyoung: That rephrase looks great.

content/concept-amendments.md
mDuo13, line 41: Lots of cases. For example, the previous ledger might have been a dead offshoot that never got validated by consensus. Or maybe it's "earlier" in the sense of when it closed, but it's from an unrelated parallel continuity (e.g. the Test Net).

content/concept-amendments.md
mDuo13, line 42: Sadly, [yes, there is a difference](https://ripple.com/build/ledger-format/#close-flags) between "what close time it has" and "when it closed."

content/concept-amendments.md
mDuo13: You should be sorry. =p

Which instances of "(the company)" would be best to remove? Take your pick:

>  Currently, Ripple (the company) recommends only trusting the 5 default validators that Ripple (the company) operates. For now, trusting only those validators is sufficient to coordinate with Ripple (the company) on releasing new features.

content/concept-amendments.md
mDuo13: Fixed.

content/concept-amendments.md
mDuo13: Yep. Fixed.

content/concept-amendments.md
mDuo13: It would probably be OK as is, but as per Scott's suggestion earlier, I decided to nix mention of multi-level multi-signing here.

content/concept-amendments.md
ryangyoung: I wasn't objecting to the personification (that's quite useful!), rather to the phrase "thinks happened," which sounds very awkward, though there's nothing per se wrong with it. I think any rephrase would by necessity be less concise (the best I could do was "what the server understands to have happened" or my previous suggestion) but could still convey a sense of uncertainty.

content/concept-amendments.md
ryangyoung: I think keeping the first and third instances would probably be clearest, if far from ideal. It's the last one where I see the most potential confusion--coordinate with the network? the consensus ledger? --but using "(the company)" only there would make me wonder if the other two instances were also referring to the company.

mDuo13: The funny thing is, I see the most potential confusion in the second instance, which is why that was the one I clarified in the original doc. I'm not happy with any of the formulations we've come up with for this sentence. I'll see if #marketing has any useful suggestions there.

I also noticed that I failed to address your comment about validators. I've now added a relevant link to one of the earlier mentions of validators.

content/reference-rippled.md
```diff
@@ -6523,6 +6761,259 @@ __*Caution:*__ If this command results in an error messages, the message can con
...
+}'
+```
+
+<!-- </div> -->
+
+The request includes the following parameters:
+
+| Field    | Type   | Description |
+|----------|--------|-------------|
+| tx\_json | Object | [Transaction in JSON format](reference-transaction-format.html) with an array of `Signers`. To be successful, the weights of the signatures must be equal or higher than the quorum of the [SignerList](reference-ledger-format.html#signerlist). |
```
scottschurr, line 6896: "equal or higher than" -> "greater than or equal to"?  Your call.

content/reference-transaction-format.md
```diff
@@ -157,6 +184,16 @@ After a transaction has been submitted, if it gets accepted into a validated led
 }
 ```
 
+
+### Multi-Signing ###
+
+Multi-signing in Ripple is the act of [authorizing transactions](#authorizing-transactions) for the Ripple Consensus Ledger by using a combination of multiple secret keys. Multi-signing is due to be enabled by an [Amendment](concept-amendments.html) to the Ripple Consensus Protocol. You can have any combination of authorization methods enabled for your address, including multi-signing, a master key, and a [regular key](#setregularkey). (The only requirement is that _at least one_ method must be enabled.)
+
+The [SignerListSet transaction](#signerlistset) defines which addresses can authorize transactions from your address. You can include up to 8 addresses in a SignerList. You can control how many signatures are needed, in which combinations, by using the quorum and weight values of the SignerList.
```
scottschurr, line 192: The last sentence looks great.  Nicely done.

content/reference-transaction-format.md
```diff
@@ -730,6 +788,58 @@ Transactions of the TrustSet type support additional values in the [`Flags` fiel
 | tfClearFreeze | 0x00200000 | 2097152 | [Unfreeze](concept-freeze.html) the trustline. |
 
 
+
+## SignerListSet ##
+[[Source]<br>](https://github.com/ripple/rippled/blob/ef511282709a6a0721b504c6b7703f9de3eecf38/src/ripple/app/tx/impl/SetSignerList.cpp "Source")
+
+The SignerListSet transaction creates, modifies, or removes a list of signers that can be used to [multi-sign](#multi-signing) a transaction.
```
scottschurr: Here's just a thought for your consideration.  A SignerListSet can't really _modify_ a list of signers, but it can _replace_ a list.  I don't know whether that is worth bringing up, but you could consider replacing "modifies" with "replaces" in this sentence.  Your call.

content/concept-transaction-cost.md
```diff
@@ -17,14 +17,52 @@ You can also [query `rippled` for the current transaction cost](#querying-the-tr
 The transaction cost is not paid to any party: the XRP is irrevocably destroyed. Since no new XRP can ever be created, this makes XRP more scarce, and consequently benefits all holders of XRP by making XRP more valuable.
 
 
-## Load Scaling ##
+## Load Cost and Open Ledger Cost ##
+
+When the [FeeEscalation amendment](concept-amendments.html#feeescalation) is enabled, there are two thresholds for the transaction cost:
+
+* If the transaction cost does not meet a `rippled` server's [load-based transaction cost threshold](#local-load-cost), the server ignores the transaction completely. (This logic is essentially unchanged with or without the amendment.)
+* If the transaction cost does not meet a `rippled` server's [open ledger cost threshold](#open-ledger-cost), the queues the transaction for the next ledger.
```
ximinez: `the server queues the transaction for a later ledger`

"server" is just a missing word. I suggest "a later" instead of "the next" because there's no telling when a queued transaction will be reattempted. It certainly isn't guaranteed that it will be the next ledger, especially if there are a lot of items in the queue, and a particular transaction is cheap.

content/concept-transaction-cost.md
```diff
@@ -17,14 +17,52 @@ You can also [query `rippled` for the current transaction cost](#querying-the-tr
...
+
+The open ledger cost requirement is proportional to the normal cost of the transaction, not the absolute transaction cost. Transaction types that have a higher-than-normal requirement, such as [multi-signed transactions](reference-transaction-format.html#multi-signing) must pay more to meet the open ledger cost than transactions which have minimum transaction cost requirements.
+
+See also: [Fee Escalation explanation in `rippled` repository](https://github.com/ripple/rippled/blob/release/src/ripple/app/misc/FeeEscalation.md).
+
+### Queued Transactions ###
+
+(Requires the [FeeEscalation amendment](concept-amendments.html#feeescalation))
+
+When `rippled` receives a transaction that meet the server's local load cost but not the open ledger cost, the server checks the transaction to see if it is "likely to succeed" in the following ledger. If so, the transaction goes into the transaction queue. Transactions that are "likely to succeed" are properly-formed and are [authorized](reference-transaction-format.html#authorizing-transactions) with valid signatures.
```
ximinez: To be pedantic, "likely to claim a fee" may be a safer description. "Likely to succeed" may cause the user to expect `tesSUCCESS`, but the queue is happy with anything that gets a `tec`, too.

ximinez: The description of "likely to succeed / claim a fee" is true, but not complete. However, the complete description is partially dependent on the transaction type, and probably far more complicated than appropriate for this document. For example, most transactors check that the account has enough balance to pay the fee, that the sequence number is valid, etc. An OfferCancel won't queue a transaction that tries to cancel an offer with a sequence number greater than the account's current sequence number. How far down this rabbit hole do we want to go?

mDuo13: That's a tough call. I think the ideal solution would be to make and link a new section towards the end of the document that defines precisely what "likely to claim a fee" means. That would have to be updated with related changes in `rippled`, but I think it would be best to communicate any such changes anyway.

I'm not sure how long and complex a very precise definition is going to be. If I can get it down to half a page of bullet points or something and then link the source code, I'll be happy with that.

content/concept-transaction-cost.md
mDuo13: Fixed both.

content/concept-transaction-cost.md
mDuo13: Revised this section to allude to the idea that there are lots of heuristics, without actually listing them all. I can go back and add the proper list post-publication when it's ready.

content/concept-transaction-cost.md
```diff
@@ -7,24 +7,74 @@ Every transaction must [specify how much XRP it will destroy](#specifying-the-tr
...
+When the [FeeEscalation amendment](concept-amendments.html#feeescalation) is enabled, there are two thresholds for the transaction cost:
+
+* If the transaction cost does not meet a `rippled` server's [load-based transaction cost threshold](#local-load-cost), the server ignores the transaction completely. (This logic is essentially unchanged with or without the amendment.)
+* If the transaction cost does not meet a `rippled` server's [open ledger cost threshold](#open-ledger-cost), the server queues the transaction for a later ledger.
+
+This divides transactions into roughly three categories:
+
+* Transactions that specify a transaction cost so low that they get rejected by the load-based transaction cost.
+* Transactions that specify a transaction cost high enough to be included in the current open ledger.
+* Transactions in between, which get [queued for the next open ledger](#queued-transactions).
```
ximinez: Sorry, I missed this one. "the next open ledger" -> "a later open ledger".

mDuo13:
With 3 peer reviews and no technical reviews outstanding, I'm merging and deploying this documentation.

content/concept-stand-alone-mode.md
```diff
@@ -0,0 +1,103 @@
...
+
+
+
+Load Saved Ledger
+-------------------------------------------------------------------------------
+You can start with a ledger version that was saved to disk if your `rippled` server was previously synced with the Ripple peer-to-peer network (either the production network or the [Test Net](tutorial-rippled-setup.html#parallel-networks)).
+
+### 1. Start `rippled` normally. ###
+
+To load an existing ledger, you must first retrieve that ledger from the network. Start `rippled` in networked mode as normal:
```
NATTSiM, line 39: 'networked mode' doesn't sound correct.

content/concept-stand-alone-mode.md
```diff
@@ -0,0 +1,103 @@
...
+
+```
+rippled --conf=/path/to/rippled.cfg
+```
+
+### 2. Wait until `rippled` is synced. ###
+
+Use the [`server_info` command](reference-rippled.html#server-info) to check the state of your server relative to the network. Your server is synced when the `server_state` value shows any of the following values:
+
```
NATTSiM, line 48: Suggest adding a link for more information on server states.

content/concept-stand-alone-mode.md
```diff
@@ -0,0 +1,103 @@
...
+* `proposing`
+* `validating`.
+
+### 3. (Optional) Retrieve specific ledger versions. ###
+
+If you only want the most recent ledger, you can skip this step.
+
+If you want to load a specific historical ledger version, use the [`ledger_request` command](reference-rippled.html#ledger-request) to make `rippled` fetch it. If `rippled` does not already have the ledger version, you may have to run the `ledger_request` command multiple times until it has finished retrieving the ledger.
+
+If you want to replay a specific historical ledger version, you must fetch both the ledger version to replay and the ledger version immediately before it.
```
NATTSiM, line 59: Why is the "ledger version immediately before it" required?

content/concept-stand-alone-mode.md
```diff
@@ -0,0 +1,103 @@
...
+
+* `full`
+* `proposing`
+* `validating`.
+
+### 3. (Optional) Retrieve specific ledger versions. ###
+
+If you only want the most recent ledger, you can skip this step.
+
+If you want to load a specific historical ledger version, use the [`ledger_request` command](reference-rippled.html#ledger-request) to make `rippled` fetch it. If `rippled` does not already have the ledger version, you may have to run the `ledger_request` command multiple times until it has finished retrieving the ledger.
```
NATTSiM, line 57: Explain why running the same command multiple times is required.

content/concept-stand-alone-mode.md
```diff
@@ -0,0 +1,103 @@
...
+* Create a [new genesis ledger](#new-genesis-ledger) from scratch.
+* [Load an existing ledger version](#load-saved-ledger) from disk.
+
+**Caution:** In stand-alone mode, you must [manually advance the ledger](#advancing-ledgers-in-stand-alone-mode).
+
+New Genesis Ledger
+-------------------------------------------------------------------------------
+In stand-alone mode, you can have `rippled` create a new genesis ledger. This provides a known state, with none of the ledger history from the production Ripple Consensus Ledger. (This is very useful for unit tests, among other things.)
+
+* To start `rippled` in stand-alone mode with a new genesis ledger, use the `-a` and `--start` options:
```
NATTSiM, line 17: It would be helpful to explicitly define the meaning of these flags.

content/concept-stand-alone-mode.md
```diff
@@ -0,0 +1,103 @@
...
+
+Use the [`stop` command](reference-rippled.html#stop):
+
+```
+rippled stop --conf=/path/to/rippled.cfg
+```
+
+### 5. Start `rippled` in stand-alone mode. ###
+
+To load the most recent ledger version, you can use the `-a` and `--load` options when starting the server:
```
NATTSiM, line 71: Explicitly define the --load flag.

content/concept-stand-alone-mode.md
```diff
@@ -0,0 +1,103 @@
...
+### 6. Manually advance the ledger. ###
+
+When you load a ledger with `--ledger` in stand-alone mode, it goes to the current open ledger, so you must [manually advance the ledger](#advancing-ledgers-in-stand-alone-mode):
+
+```
+rippled ledger_accept --conf=/path/to/rippled.cfg
+```
+
+
+Advancing Ledgers in Stand-Alone Mode
```
NATTSiM, line 92: This seems very similar to step 6. Maybe you can merge the two.

content/concept-stand-alone-mode.md
```diff
@@ -0,0 +1,103 @@
...
+
+```
+rippled ledger_accept --conf=/path/to/rippled.cfg
+```
+
+In stand-alone mode, `rippled` makes no distinction between a "closed" ledger version and a "validated" ledger version. (For more information about the difference, see [The Ripple Ledger Consensus Process](https://ripple.com/knowledge_center/the-ripple-ledger-consensus-process/).)
+
+Whenever `rippled` closes a ledger, it reorders the transactions according to a deterministic but hard-to-game algorithm. (This is an important part of consensus, since transactions may arrive at different parts of the network in different order.) When using `rippled` in stand-alone mode, you should manually advance the ledger before submitting a transaction that depends on the result of a previous transaction. Otherwise, the second transaction might be executed before the first transaction when you manually advance the ledger. **Note:** You can safely submit multiple transactions from a single address to a single ledger, because `rippled` sorts transactions from the same address in ascending order by `Sequence` number.
```
NATTSiM, line 103: "depends on the result of a previous transaction" 
--> Wouldn't sequence number resolve this anyway? I think the example should be more specific. You cannot broadcast a transaction for which the previous transaction removed the possibility of success.

NATTSiM, line 103: Also, why is sequence highlighted? If you highlight it, why not link to the sequence number section of the documentation?

content/concept-stand-alone-mode.md
mDuo13, line 39: Changed to "online mode."

content/concept-stand-alone-mode.md
mDuo13, line 48: The linked server_info command already has a link to a summary of possible server states, but I also added a link directly to the list of server states.

content/concept-stand-alone-mode.md
mDuo13, line 59: Added this clarification:

> (The previous ledger version sets up the initial state upon which you apply the changes described by the ledger version you replay.)

content/concept-stand-alone-mode.md
mDuo13, line 57: I think it's more confusing than helpful to explain why. (Basically, requesting the ledger tends to time out before rippled finishes, and gets distracted with "more important" work related to keeping up with the current ledger.) I think this rippled behavior is obnoxious, but I'm going to leave the documentation as-is for simplicity's sake.

content/concept-stand-alone-mode.md
mDuo13, line 17: This tutorial is not the right place to go into an explicit definition of the flags. I've linked to the rippled wiki page for now. Anything not there will have to wait for https://ripplelabs.atlassian.net/browse/DOC-368

content/concept-stand-alone-mode.md
mDuo13, line 71: Linked to https://wiki.ripple.com/Rippled#--load

content/concept-stand-alone-mode.md
mDuo13, line 92: It's a little weird to see them side-by-side like that, I agree, but they're separate for good reason. Step 6 is a step of the process with an example. The "Advancing Ledgers in Stand-Alone Mode" header is meant to be linked from other docs (e.g. the multi-signing tutorial)

content/concept-stand-alone-mode.md
mDuo13, line 103: Linked "Sequence number" to https://ripple.com/build/transactions/#common-fields -- and it's "highlighted" because it's a field name in JSON, which we always typeset in monospace font.

Rephrased the part about advancing the ledger as follows:

> When using `rippled` in stand-alone mode, you should manually advance the ledger before submitting a transaction that depends on the result of a transaction from a different address. Otherwise, the two transactions might be executed in reverse order when the ledger is closed.

content/concept-transaction-cost.md
```diff
@@ -7,24 +7,74 @@ Every transaction must [specify how much XRP it will destroy](#specifying-the-tr
...
 
 You can also [query `rippled` for the current transaction cost](#querying-the-transaction-cost).
 
+### Special Transaction Costs ###
+
+Some transactions have different transaction costs:
+
+| Transaction           | Cost Before Load Scaling |
+|-----------------------|--------------------------|
+| Reference Transaction (Standard cost of most transactions) | 10 drops |
```
ryangyoung, line 20: Probably my ignorance, but this seems to be different than the amount mentioned above, i.e. "The current transaction cost required by the network for a standard transaction is typically **0.01 XRP** (10,000 drops)."

content/concept-transaction-cost.md
```diff
@@ -7,24 +7,74 @@ Every transaction must [specify how much XRP it will destroy](#specifying-the-tr
...
 
 ## Beneficiaries of the Transaction Cost ##
 
 The transaction cost is not paid to any party: the XRP is irrevocably destroyed. Since no new XRP can ever be created, this makes XRP more scarce, and consequently benefits all holders of XRP by making XRP more valuable.
 
 
-## Load Scaling ##
+## Load Cost and Open Ledger Cost ##
+
+When the [FeeEscalation amendment](concept-amendments.html#feeescalation) is enabled, there are two thresholds for the transaction cost:
```
ryangyoung, line 32: There's a tense conflict here, which is tricky. Perhaps separate this into two sentences to get around it. It would be clunky, but I'm thinking of something like: "This section describes the thresholds that will apply once the FeeEscalation amendment is enabled. With FeeEscalation enabled, there are two thresholds..."

content/concept-transaction-cost.md
```diff
@@ -7,24 +7,74 @@ Every transaction must [specify how much XRP it will destroy](#specifying-the-tr
...
+This divides transactions into roughly three categories:
+
+* Transactions that specify a transaction cost so low that they get rejected by the load-based transaction cost.
+* Transactions that specify a transaction cost high enough to be included in the current open ledger.
+* Transactions in between, which get [queued for a later ledger version](#queued-transactions).
+
+
+## Local Load Cost ##
+
+Each `rippled` server maintains a cost threshold based on its current load. If you submit a transaction with a `Fee` value that is lower than current load-based transaction cost of the `rippled` server, that server neither applies nor relays the transaction. (**Note:** If you submit a transaction through an [admin connection](reference-rippled.html#connecting-to-rippled), the server applies and relays the transaction as long as the transaction cost meets the overall minimum.) A transaction is very unlikely to survive [the consensus process](https://ripple.com/knowledge_center/the-ripple-ledger-consensus-process/) unless its `Fee` value meets the requirements of a majority of servers.
```
ryangyoung, line 46: What does "overall minimum" refer to? The open ledger cost?

content/concept-transaction-cost.md
```diff
@@ -7,24 +7,74 @@ Every transaction must [specify how much XRP it will destroy](#specifying-the-tr
...
+
+## Open Ledger Cost ##
+
+A `rippled` server with the [FeeEscalation amendment](concept-amendments.html#feeescalation) enabled has a second mechanism for enforcing the transaction cost, called the _open ledger cost_. The open ledger cost starts out equal to the minimum transaction cost, but increases exponentially when an in-progress ledger has more transactions than the previous one. Only transactions which pay more than the open ledger cost can be included in the current open ledger.
```
ryangyoung, line 50: "when an in-progress ledger has more transactions than the previous one" --> "when an in-progress ledger has more transactions than the previous ledger"

ryangyoung, line 50: "Only transactions which pay more than the open ledger cost can be included in the current open ledger."

That's always true? Based on the previous sentence, it seems like it would only be true when the current ledger has more transactions than the previous ledger?

content/concept-transaction-cost.md
mDuo13, line 20: I should clarify that these transaction costs are "before load scaling" whereas the ones in the Current Transaction Cost section are "after scaling for typical load."

content/concept-transaction-cost.md
mDuo13, line 50: > Only transactions which pay more than the open ledger cost can be included in the current open ledger.

Yes, this is always true, with the clarification that "pay more" is in "fee levels" (so more relative to the minimum cost for that particular transaction) not absolute cost.

As stated here, the open ledger cost starts out very low, but increases when the ledger becomes larger than the previous one. (It also increases if consensus takes longer than 5 seconds, I think, which is a detail that I should probably add.)

content/concept-transaction-cost.md
mDuo13, line 46: Probably better phrased as "un-scaled" minimum.

content/concept-transaction-cost.md
mDuo13, line 20: Rather than revise this section (which already has "before load scaling" in the table header) I revised the section about the Current Transaction Cost to clarify that the 0.01 XRP number is after load scaling.

content/concept-transaction-cost.md
mDuo13, line 32: I'd rather leave the sentence as is. There's no tense conflict if you think of the sentence as "If the FeeEscalation amendment is enabled..." but I don't want to write that since it suggests FeeEscalation won't get enabled (a pretty unlikely scenario I'd rather people not worry about).

content/concept-transaction-cost.md
mDuo13, line 50: Rephrased as follows:

> The open ledger cost starts out equal to the un-scaled minimum transaction cost, but increases exponentially when an in-progress ledger has more transactions than the previous ledger. The open ledger cost also increases if consensus took longer than 5 seconds to approve the previous ledger. A transaction can only be included in the open ledger if it meets the open ledger cost requirement in XRP. Transactions that do not meet the open ledger cost are [queued for a following ledger](#queued-transactions) instead.

(the following paragraph, unchanged, explains how the open ledger cost is proportional)

content/concept-transaction-cost.md
ryangyoung, line 32: But...I'm reading it as "When the..." because those are the words that are written. I'd still suggest a prefatory sentence to explain the situation, which is complex, and which not everyone may totally get right away.

content/concept-transaction-cost.md
ryangyoung, line 46: Still, "the un-scaled minimum" what?

content/concept-transaction-cost.md
ryangyoung, line 50: Much better. That was the missing piece for me.

content/concept-transaction-cost.md
mDuo13, line 32: I played around with a number of rephrasings and I think all of them are less clear than the current one.

Part of the reason I'm leaving it as-is has to do with parallel ledger chains, such as the ones created in stand-alone mode from a fresh genesis ledger. Those ledgers don't have amendments enabled (unless you force them on with the `[feature]` config stanza) so it's not just a matter of "when" the amendment gets enabled, but "in situations where the amendment is enabled". I think "When" with present tense is the most efficient way to explain the situation without introducing a large amount of additional information that may also need to be updated. It's like saying, "When on vacation, there are..."

content/concept-transaction-cost.md
mDuo13, line 46: rephrased:

> ...the server applies and relays the transaction as long as the transaction meets the un-scaled minimum transaction cost.

content/concept-transaction-cost.md
mDuo13, line 50: I had to rephrase this section again because the line about the open ledger cost increasing if consensus took longer than 5 seconds was not _quite_ technically accurate. New version:

> A `rippled` server with the [FeeEscalation amendment](concept-amendments.html#feeescalation) enabled has a second mechanism for enforcing the transaction cost, called the _open ledger cost_. A transaction can only be included in the open ledger if it meets the open ledger cost requirement in XRP. Transactions that do not meet the open ledger cost are [queued for a following ledger](#queued-transactions) instead.
> 
> For each new ledger version, the server picks a soft limit on the number of transactions to be included in the open ledger, based on the number of transactions in the previous ledger. The open ledger cost is equal to the minimum un-scaled transaction cost until the number of transactions in the open ledger is equal to the soft limit. After that, the open ledger cost increases exponentially for each additional transaction included in the open ledger. For the next ledger, the server increases the soft limit if the current ledger contained more transactions than the soft limit, and decreases the soft limit if the consensus process takes more than 5 seconds.

content/concept-transaction-cost.md
ryangyoung, line 32: Makes sense. I yield my case.

PR #175, opened 2016-04-16
data v2.1.0 changes

mDuo13:
Four new methods (documentation mostly done by the Data team; I've just polished it up slightly):
- Get Top Currencies
- Get Top Markets
- Get Account Transaction Stats
- Get Account Value Stats

Various whitespace changes (this happened automatically because I switched text editors)

`convertedAmount` -> `converted_amount`

Found a few cases of "Ripple, Inc" and replaced them accordingly

Fixed one or two cases of invalid JSON

### DIFF
--- a/content/reference-data-api.md
+++ b/content/reference-data-api.md
@@ -1422,7 +1422,7 @@ Each object in the `components` array of the Volume Objects represent the volume
 | amount | Number | The amount of volume in the market, in units of the base currency. |
 | base   | Object | The `currency` and `issuer` that identify the base currency of this market. There is no `issuer` for XRP. |
 | counter | Object | The `currency` and `issuer` that identify the counter currency of this market. There is no `issuer` for XRP. |
-| converted\_amount | Number | The total amount of volume in the market, converted to the display currency. |
+| converted\_amount | Number | The total amount of volume in the market, converted to the display currency. _(Before [v2.1.0][], this was `convertedAmount`.)_ |
 
 #### Example ####
 
@@ -1570,7 +1570,7 @@ Each object in the `components` array of the Volume Objects represent the volume
 | amount | Number | Total payment volume for this currency during the interval, in units of the currency itself. |
 | count  | Number | The total number of payments in this currency |
 | rate   | Number | The exchange rate between this currency and the display currency. |
-| converted\_amount | Number | Total payment volume for this currency, converted to the display currency. |
+| converted\_amount | Number | Total payment volume for this currency, converted to the display currency. _(Before [v2.1.0][], this was `convertedAmount`.)_ |
 
 #### Example ####
 
--- a/content/reference-data-api.md
+++ b/content/reference-data-api.md
@@ -9,7 +9,7 @@ Ripple provides a live instance of the Data API with as complete a transaction r
 
 
 ## More Information ##
-The Ripple Data API v2 replaces the Historical Database v1 and the [Charts API](https://github.com/ripple/ripple-data-api/). 
+The Ripple Data API v2 replaces the Historical Database v1 and the [Charts API](https://github.com/ripple/ripple-data-api/).
 
 * [API Methods](#api-method-reference)
 * [API Conventions](#api-conventions)
@@ -22,6 +22,7 @@ The Ripple Data API v2 replaces the Historical Database v1 and the [Charts API](
 [v2.0.6]: https://github.com/ripple/rippled-historical-database/releases/tag/v2.0.6
 [v2.0.7]: https://github.com/ripple/rippled-historical-database/releases/tag/v2.0.7
 [v2.0.8]: https://github.com/ripple/rippled-historical-database/releases/tag/v2.0.8
+[v2.1.0]: https://github.com/ripple/rippled-historical-database/releases/tag/v2.1.0
 
 
 # API Method Reference #
@@ -44,6 +45,8 @@ General Methods:
 * [Get Exchange Volume - `GET /v2/network/exchange_volume`](#get-exchange-volume)
 * [Get Payment Volume - `GET /v2/network/payment_volume`](#get-payment-volume)
 * [Get Issued Value - `GET /v2/network/issued_value`](#get-issued-value)
+* [Get Top Currencies - `GET /v2/network/top_currencies`](#get-top-currencies)
+* [Get Top Markets - `GET /v2/network/top_markets`](#get-top-markets)
 * [Get All Gateways - `GET /v2/gateways`](#get-all-gateways)
 * [Get Gateway - `GET /v2/gateways/{:gateway}`](#get-gateway)
 * [Get Currency Image - `GET /v2/currencies/{:currencyimage}`](#get-currency-image)
@@ -60,6 +63,8 @@ Account Methods:
 * [Get Account Exchanges - `GET /v2/accounts/{:address}/exchanges`](#get-account-exchanges)
 * [Get Account Balance Changes - `GET /v2/accounts/{:address}/balance_changes`](#get-account-balance-changes)
 * [Get Account Reports - `GET /v2/accounts/{:address}/reports`](#get-account-reports)
+* [Get Account Transaction Stats - `GET /v2/accounts/{:address}/stats/transactions`](#get-account-transaction-stats)
+* [Get Account Value Stats - `GET /v2/accounts/{:address}/stats/value`](#get-account-value-stats)
 
 Health Checks:
 
@@ -234,7 +239,7 @@ Response (trimmed for size):
                         }
                     }
                 },
-                
+
                 ...
             ],
             "TransactionResult": "tesSUCCESS"
@@ -476,7 +481,7 @@ If the request specifies a `currency` and an `interval`, the result includes obj
 
 #### Example ####
 
-Request: 
+Request:
 
 ```
 GET /v2/payments/BTC+rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q?limit=2
@@ -736,7 +741,7 @@ The rate is derived from the volume weighted average over the calendar day speci
 
 #### Example ####
 
-Request: 
+Request:
 
 ```
 GET /v2/exchange_rates/USD+rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q/XRP?date=2015-11-13T00:00:00Z
@@ -800,7 +805,7 @@ All exchange rates are calculating by converting both currencies to XRP.
 
 #### Example ####
 
-Request: 
+Request:
 
 ```
 GET /v2/normalize?amount=100&currency=XRP&exchange_currency=USD&exchange_issuer=rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q
@@ -902,9 +907,9 @@ Response (trimmed for size):
                     "amount": "40",
                     "type": "received"
                 },
-                
+
                 ...(additional results trimmed)...
-                
+
                 {
                     "tx_hash": "76041BD6546389B5EC2CDBAA543200CF7B8D300F34F908BA5CA8523B0CA158C8",
                     "amount": "1400",
@@ -916,18 +921,18 @@ Response (trimmed for size):
             "receiving_counterparties": [
                 "rDMFJrKg2jyoNG6WDWJknXDEKZ6ywNFGwD",
                 "r4XXHxraHLuCiLmLMw96FTPXXywZSnWSyR",
-                
+
                 ...(additional results trimmed)...
-                
-                
+
+
                 "rp1C4Ld6uGjurFpempUJ8q5hPSWhak5EQf"
             ],
             "sending_counterparties": [
                 "rwxcJVWZSEgN2DmLZYYjyagHjMx5jQ7BAa",
-                
+
                 ...(additional results trimmed)...
-                
-                
+
+
                 "rBK1rLjbWsSU9EuST1cAz9RsiYdJPVGXXA"
             ],
             "total_value": "210940",
@@ -945,10 +950,10 @@ Response (trimmed for size):
                     "amount": "900",
                     "type": "sent"
                 },
-                
+
                 ...(additional results trimmed)...
-                
-                
+
+
                 {
                     "tx_hash": "EC25427964419394BB5D06343BC74235C33655C1F70523C688F9A201957D65BA",
                     "amount": "100",
@@ -959,19 +964,19 @@ Response (trimmed for size):
             "payments_sent": 62,
             "receiving_counterparties": [
                 "rB4cyZxrBrTmJcWZSBc8YoW2t3bafiKRp",
-                
+
                 ...(additional results trimmed)...
-                
-                
+
+
                 "rKybkw3Pu74VfJfrWr7QJbVPJNarnKP2EJ"
             ],
             "sending_counterparties": [
                 "rNRCXw8PQRjvTwMDDLZVvuLHSKqqXUXQHv",
                 "r7CLMVEuNvK2yXTPLPnkWMqzkkXuopWeL",
-                
+
                 ...(additional results trimmed)...
-                
-                
+
+
                 "ranyeoYRhvwiFABzDvxSVyqQKp1bMkFsaX"
             ],
             "total_value": "117600",
@@ -1023,7 +1028,7 @@ The `family` and `metrics` query parameters provide a way to filter results to a
 
 | Family | Included Metrics | Meaning |
 |--------|------------------|---------|
-| type | All Ripple [transaction types](reference-transaction-format.html), including `Payment`, `AccountSet`, `SetRegularKey`, `OfferCreate`, `OfferCancel`, `TrustSet`. | Number of transactions of the given type that occurred during the interval. |
+| type | All Ripple [transaction types](reference-transaction-format.html), including `Payment`, `AccountSet`, `OfferCreate`, and others. | Number of transactions of the given type that occurred during the interval. |
 | result | All [transaction result codes](reference-transaction-format.html#transaction-results) (string codes, not the numeric codes), including `tesSUCCESS`, `tecPATH_DRY`, and many others. | Number of transactions that resulted in the given code during the interval. |
 | metric | Data-API defined Special Transaction Metrics. | (Varies) |
 
@@ -1054,7 +1059,7 @@ A successful response uses the HTTP code **200 OK** and has a JSON body with the
 
 #### Example ####
 
-Request: 
+Request:
 
 ```
 GET /v2/stats/?start=2015-08-30&end=2015-08-31&interval=day&family=metric&metrics=accounts_created,exchanges_count,ledger_count,payments_count
@@ -1151,7 +1156,7 @@ Each **issuer capitalization object** has the following fields:
 
 #### Example ####
 
-Request: 
+Request:
 
 ```
 GET /v2/capitalization/USD+rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q?start=2015-01-01T00:00:00Z&end=2015-10-31&interval=month
@@ -1324,9 +1329,9 @@ Response:
             "counter_volume": 52.4909286454546,
             "count": 1
         },
-        
+
         ... (additional results trimmed)...
-        
+
         {
             "buy": {
                 "base_volume": 1.996007,
@@ -1417,7 +1422,7 @@ Each object in the `components` array of the Volume Objects represent the volume
 | amount | Number | The amount of volume in the market, in units of the base currency. |
 | base   | Object | The `currency` and `issuer` that identify the base currency of this market. There is no `issuer` for XRP. |
 | counter | Object | The `currency` and `issuer` that identify the counter currency of this market. There is no `issuer` for XRP. |
-| convertedAmount | Number | The total amount of volume in the market, converted to the display currency. |
+| converted\_amount | Number | The total amount of volume in the market, converted to the display currency. |
 
 #### Example ####
 
@@ -1448,7 +1453,7 @@ Response:
                     "counter": {
                         "currency": "XRP"
                     },
-                    "convertedAmount": 117720.99268355068
+                    "converted_amount": 117720.99268355068
                 },
                 {
                     "count": 1977,
@@ -1461,11 +1466,11 @@ Response:
                     "counter": {
                         "currency": "XRP"
                     },
-                    "convertedAmount": 74003.51871932109
+                    "converted_amount": 74003.51871932109
                 },
-                
+
                 ... (additional results trimmed) ...
-                
+
                 {
                     "count": 3,
                     "rate": 0.022999083584408355,
@@ -1478,7 +1483,7 @@ Response:
                         "currency": "USD",
                         "issuer": "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B"
                     },
-                    "convertedAmount": 12.72863756671683
+                    "converted_amount": 12.72863756671683
                 },
                 {
                     "count": 3,
@@ -1492,7 +1497,7 @@ Response:
                         "currency": "BTC",
                         "issuer": "rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q"
                     },
-                    "convertedAmount": 4.4137945368632545
+                    "converted_amount": 4.4137945368632545
                 }
             ],
             "count": 11105,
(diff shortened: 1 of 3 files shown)


content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
 
 
 
+## Get Top Currencies ##
+[[Source]<br>](https://github.com/ripple/rippled-historical-database/blob/develop/api/routesV2/network/topCurrencies.js "Source")
+
+Returns the top currencies on the Ripple Consensus Ledger over a rolling 30 day window. The currencies are ordered from highest ranked to lowest, where the rank is determined by the volume and count of transactions and the number of unique counterparties. By default, returns the top currencies for the most recent date available. You can query historical data by date. _(New in [v2.1.0][])_
```
ryangyoung: A "rolling 30 day window" is a little ambiguous. What exactly does it mean here? That it will always use the last 30 days of data? Should clarify that. 

Also, perhaps rephrase the second sentence so that the definition of the ranking comes first. Depending on how you rephrase the first sentence, you might be able to wedge the sort order description there (i.e., "Returns the top currencies on the RCL for the last 30 days, sorted from highest to lowest ranking. The rank is determined by...").

Finally, maybe rephrase the last sentence to something like "You can also query the top currencies for a specified date range using historical data" to make it's connection to the rest of the summary a little clearer.

content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
...
+[Try it! >](data-api-v2-tool.html#get-top-currencies)
+
+
+#### Response Format ####
+
+A successful response uses the HTTP code **200 OK** and has a JSON body with the following:
+
+| Field  | Value | Description |
+|--------|-------|-------------|
+| result | `success` | Indicates that the body represents a successful response. |
```
ryangyoung: Shouldn't the "Value" be "String" here? Seems odd to specify an actual value here.

content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
...
+| date   | String - [Timestamp][] | The time at which this data was measured. |
```
ryangyoung, line 1795: "Measured" doesn't seem like quite the right word here. Instead, maybe: "Collected," "obtained," or "calculated"?

content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
...
+| count  | Integer | Number of results returned. |
```
ryangyoung: What does a "result" mean here? The number of "Top Currency Objects"?

content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
...
+| currencies | Array of Top Currency Objects | A currency + issuer, and the metrics measured for inclusion and ranking |
```
ryangyoung: Shouldn't the "Value" (Type?) field define just the type? And any further description would be in the "Description" field (i.e., "Value: Array"; "Description: Array of Top Currency objects")? I could see that could be a little redundant, but that's what makes sense to me. Maybe there's a good reason not to do that.

ryangyoung: The description could probably be more expansive, e.g.: "Array of Top Currency objects. Each Top currency object includes `currency` and `issuer` fields, and additional metrics used for inclusion and ranking.  See the Top Currency Object table for a detailed description of each field."

content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
...
+
+Each Top Currency Object has the following fields:
+
+| Field  | Value | Description |
+|--------|-------|-------------|
+| currency | String - [Currency Code][] | The currency this object describes |
```
ryangyoung, line 1803: Would be great to have a standard include for common fields like this. As is, maybe mention the standard "in ISO 4217 format." And maybe link to the following definition, though that might not be necessary: https://ripple.com/build/rippled-apis/#specifying-currency-amounts (or https://ripple.com/build/rippled-apis/#currency-codes)

Never mind: I see the "Value" already links there.

content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
...
+| issuer | String - [Address][] | The Ripple address that issues this currency |
```
ryangyoung, line 1804: Maybe reuse the description from the link above, i.e.: "Unique account address of the entity issuing the currency. In other words, the person or business where the currency can be redeemed."

content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
...
+*Most Recent*
+
+```
+GET /v2/network/top_currencies
+```
+
+*By Date*
+
+```
+GET /v2/network/top_currencies/2016-01-01
```
ryangyoung, line 1779: Maybe, per the [API Template](https://ripplelabs.atlassian.net/wiki/display/DOCS/REST+API+Method+Template), add a note about this method not taking any URL parameters.

content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
...
+| count  | Integer | Number of results returned. |
+| currencies | Array of Top Currency Objects | A currency + issuer, and the metrics measured for inclusion and ranking |
+
+Each Top Currency Object has the following fields:
+
+| Field  | Value | Description |
+|--------|-------|-------------|
+| currency | String - [Currency Code][] | The currency this object describes |
+| issuer | String - [Address][] | The Ripple address that issues this currency |
+| avg_exchange_count | [String - Number][] | Daily average number of exchanges |
```
ryangyoung: Link to Exchange Objects section: http://mduo13.github.io/ripple-dev-portal/reference-data-api.html#exchange-objects

content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
...
+| avg_exchange_volume | [String - Number][] | Daily average volume of exchanges, normalized to XRP |
+| avg_payment_count | [String - Number][] | Daily average number of payments |
```
ryangyoung: Link to Payment Objects section: http://mduo13.github.io/ripple-dev-portal/reference-data-api.html#payment-objects

content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
...
+  ]
+}
+```
+
+
+
+## Get Top Markets ##
+[[Source]<br>](https://github.com/ripple/rippled-historical-database/blob/develop/api/routesV2/network/topMarkets.js "Source")
+
+Returns the top exchange markets on the Ripple Consensus Ledger over a rolling 30 day window.  The markets are ordered from highest ranked to lowest, where the rank is determined by the number and volume of exchanges and the number of counterparties participating.  By default, returns top markets for the latest date available.  You can query historical data by date. _(New in [v2.1.0][])_
```
ryangyoung: Rephrase according to suggestions for "Get Top Currencies."

content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
...
+*Most Recent*
+
+```
+GET /v2/network/top_markets
+```
+
+*By Date*
+
+```
+GET /v2/network/top_markets/2016-01-01
```
ryangyoung, line 1879: Add a URL Parameters section, similar to Get Top Currencies.

content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
...
+[Try it! >](data-api-v2-tool.html#get-top-markets)
+
+
+#### Response Format ####
+
+A successful response uses the HTTP code **200 OK** and has a JSON body with the following:
+
+| Field  | Value | Description |
+|--------|-------|-------------|
+| result | `success` | Indicates that the body represents a successful response. |
```
ryangyoung: Same comments as for the Get Top Currencies response fields:
- Actual field name shouldn't be in Value field.
- "Measured" doesn't seem quite right for the `date` field. 
- `Value` field for `markets` should probably only define type, with description going into `Description` field. 
- `Description` field for `markets` should be rephrased, similar to suggestion for Top Currencies object.

content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
...
+
+| Field  | Value | Description |
+|--------|-------|-------------|
+| base_currency | String - [Currency Code][] | The base currency for this market |
+| base_issuer | String - [Address][] | (Omitted if `base_currency` is XRP) The Ripple address that issues the base currency |
+| counter_currency | String - [Currency Code][] | The counter currency for this market |
+| counter_issuer | String - [Address][] | (Omitted if `counter_currency` is XRP) The Ripple address that issues the counter currency |
+| avg_base_volume | String | Daily average volume in terms of the base currency |
+| avg_counter_volume | String | Daily average volume in terms of the counter currency |
+| avg_exchange_count | String | Daily average number of exchanges |
```
ryangyoung: Link to Exchanges Objects section.

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
 
 
 
+## Get Account Transaction Stats ##
+[[Source]<br>](https://github.com/ripple/rippled-historical-database/blob/develop/api/routesV2/accountStats.js "Source")
+
+Retrieve daily summaries of transaction activity for an account. _(New in [v2.1.0][].)_
```
ryangyoung, line 3220: Sounds like the default is a daily summary for today's date, but you can optionally get summaries for other time ranges. Perhaps revise the description to reflect that.

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+
+<!--</div>-->
+
+[Try it! >](data-api-v2-tool.html#get-account-transaction-stats)
+
+This method requires the following URL parameters:
+
+| Field    | Value  | Description |
+|----------|--------|-------------|
+| :address | String | Ripple address to query |
```
ryangyoung: _Minor and superficial:_ It looks like all the existing open-source reference docs have "Fields" in regular font. For internal docs, I used backticks around the field names because I thought the monospace font looked better in the PDF. Probably better to maintain consistency at this point than bother changing what's there already, but I'd be interested in your take for future docs, style guide, etc.

ryangyoung: Another minor point: some descriptions end with periods, some don't. I think being consistent by adding periods even to sentence fragments is probably the way to go.

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+
+
+Optionally, you can also include the following query parameters:
+
+| Field      | Value   | Description |
+|------------|---------|-------------|
+| start      | String  | UTC start time of query range. Defaults to start of current date. |
```
ryangyoung: Maybe: "Start time of query range, in UTC". 

Also: "Defaults to current date", unless the description for `end` is changed to "Defaults to end of current date." Whichever is more technically accurate.

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+| end        | String  | UTC end time of query range. Defaults to current date. |
+| limit      | Integer | Max results per page (defaults to 200). Cannot be more than 1000. |
```
ryangyoung: "Maximum results per page." Default should probably be freed from parentheses.

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+| descending | Boolean | If true, sort results with most recent first. By default, sort results with oldest first. |
```
ryangyoung, line 3248: Suggest: "If true, .... If false, sort results with oldest first. Defaults to `false.`"

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+| marker     | String  | [Pagination](#pagination) key from previously returned response. |
+| format     | String  | Format of returned results: `csv`,`json` defaults to `json` |
+
+
+#### Response Format ####
+A successful response uses the HTTP code **200 OK** and has a JSON body with the following:
+
+| Field  | Value | Description |
+|--------|-------|-------------|
+| result | `success` | Indicates that the body represents a successful response. |
```
ryangyoung: See comments on this for Get Top Currencies and Get Top Markets.

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+| count | Integer | Number of reports returned. |
```
ryangyoung: "Number of Report Objects returned." And link to Report Objects section.

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+| rows | Array of Transaction Stats Objects | Daily summaries of account transaction activity for the given account. |
```
ryangyoung, line 3260: See comments on this for Get Top Currencies and Get Top Markets.

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+
+Each Transaction Stats Object has the following fields:
+
+| Field  | Value | Description |
+|--------|-------|-------------|
+| date   | String - [Timestamp][] | The date this transaction stats object describes. |
```
ryangyoung: The object doesn't describe the date exactly, but rather the transactions that occurred on that date, right? Suggest rephrasing to reflect that (if it is correct).

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+  ]
+}
+```
+
+
+
+## Get Account Value Stats ##
+[[Source]<br>](https://github.com/ripple/rippled-historical-database/blob/develop/api/routesV2/accountStats.js "Source")
+
+Retrieve daily summaries of transaction activity for an account. _(New in [v2.1.0][].)_
```
ryangyoung, line 3318: As with Get Account Transaction Stats, suggest revising this to reflect the fact that it is by default a daily summary, but can also be a summary of other time ranges.

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+| Field    | Value  | Description |
+|----------|--------|-------------|
+| :address | String | Ripple address to query |
+
+
+Optionally, you can also include the following query parameters:
+
+| Field      | Value   | Description |
+|------------|---------|-------------|
+| start      | String - [Timestamp][] | Start time of query range. Defaults to the start of the most recent interval. |
```
ryangyoung, line 3343: In UTC? Suggest revising this and `end` in-line with description for Get Account TX Stats.

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+| end        | String - [Timestamp][] | End time of query range. Defaults to the end of the most recent interval. |
+| limit      | Integer | Max results per page (defaults to 200). Cannot be more than 1000. |
```
ryangyoung: Default is probably too important to be closed off in parens.

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+| marker     | String  | [Pagination](#pagination) key from previously returned response. |
+| descending | Boolean | If true, sort results with most recent first. By default, sort results with oldest first. |
```
ryangyoung, line 3347: Suggest: "If true, .... If false, sort results with oldest first. Defaults to false."

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+| format     | String  | Format of returned results: `csv` or `json`. Defaults to `json`. |
+
+
+#### Response Format ####
+A successful response uses the HTTP code **200 OK** and has a JSON body with the following:
+
+| Field  | Value | Description |
+|--------|-------|-------------|
+| result | `success` | Indicates that the body represents a successful response. |
```
ryangyoung: See comments about this field for other methods.

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+| count | Integer | Number of reports returned. |
```
ryangyoung: "Report Objects" and link to that section.

content/reference-data-api.md
```diff
@@ -3006,6 +3212,190 @@ Response:
...
+| rows | Array of Value Stats Objects | Daily summaries of account value for the given account. |
+
+Each Value Stats Object has the following fields:
+
+| Field  | Value | Description |
+|--------|-------|-------------|
+| date   | String - [Timestamp][] | This date this object describes. |
```
ryangyoung: Same as Get Account TX Stats: The object doesn't describe the date exactly, but rather the value of that account on the specified date, right? Suggest rephrasing to reflect that (if it is correct).

content/reference-data-api.md
mDuo13: I don't want to lose the term "rolling window" because that is a well-defined statistics term, which makes it clear that multiple sub-samples overlap. That said, I rephrased as follows:

> Returns the top currencies on the Ripple Consensus Ledger, ordered from highest rank to lowest. The ranking is determined by the volume and count of transactions and the number of unique counterparties. By default, returns results for the 30-day rolling window ending on the current date. You can specify a date to get results for the 30-day window ending on that date.

content/reference-data-api.md
mDuo13, line 1779: Done.

content/reference-data-api.md
mDuo13, line 1795: Rephrased to be more descriptive:

> The end of the rolling window over which this data was calculated.

content/reference-data-api.md
mDuo13: Well, the number of such objects in _this_ response. (This description is recycled from elsewhere in the API doc, where the `count` in one response may be less than the total number of results over a paginated set of responses.)

Changed to:

> Number of results in the `markets` field.

(and the appropriate variation for `currencies`)

content/reference-data-api.md
ryangyoung: Much better!

content/reference-data-api.md
mDuo13, line 1803: For the record, the standard include already exists. See lines 3605-3608 of the full file.

content/reference-data-api.md
mDuo13: For statically-typed programming languages, it's standard for the type of object contained in an array to be part of the data-type of that array, because arrays (typically) are limited to holding only one type of variable. Although JSON documents are not required to follow those rules, by putting the type of objects as part of the array definition, we can indicate clearly that users can make assume that the contents are all the same object type if they parse the API response using a statically-typed language.

I don't like saying, "See the Top Currency Object table" when the table is not explicitly named, but I did revise the description:

> The currency code, issuer, and metrics used for ranking this currency.

content/reference-data-api.md
mDuo13, line 1804: I'd rather keep the description shorter in the table. They can click the "Address" link to get the detailed version.

content/reference-data-api.md
mDuo13: Done.

content/reference-data-api.md
mDuo13: Done.

content/reference-data-api.md
mDuo13: On second thought, the description I wrote is misleading because it describes one member, not the field itself. Here's a better Description for the `currencies` field:

> The top currencies for this data sample. Each member represents one currency, by currency code and issuer.

content/reference-data-api.md
```diff
@@ -1752,6 +1757,207 @@ Response:
...
+#### Response Format ####
+
+A successful response uses the HTTP code **200 OK** and has a JSON body with the following:
+
+| Field  | Value | Description |
+|--------|-------|-------------|
+| result | `success` | Indicates that the body represents a successful response. |
+| date   | String - [Timestamp][] | The time at which this data was measured. |
+| count  | Integer | Number of results returned. |
+| markets | Array of Top Market Objects | A currency pair and the metrics measured for inclusion and ranking |
```
mDuo13: Changed the "Description" as per Top Currencies above:

> The top markets for this data sample. Each member represents a currency pair.

content/reference-data-api.md
mDuo13, line 3220: Not quite. The default is an array of daily summaries covering the entire active history of the account. The query parameters can filter it down to fewer total days, results, or reverse order, but it pretty much always returns "daily summaries of transaction activity for an account."

content/reference-data-api.md
mDuo13: Yes, I had noticed that about the "Fields" column. I chose non-monospace because I think it looks better and it's redundant to put everything in monospace when you know from context that they're field names. That said, I'm not wedded to one convention.

I'm with you on adding periods to the fragments for consistency. It's hard to catch them all though.

content/reference-data-api.md
mDuo13: Rephrased this to match the better wording elsewhere in the doc.

content/reference-data-api.md
mDuo13: Changed this and ~20 other cases to free the default from parentheses and replace "Max results" with "Maximum results"

content/reference-data-api.md
mDuo13, line 3248: I _think_ that should be accurate, but I'm hesitant to make this change across many API calls because there may be some weird cases where the API checks for the presence of the parameter rather than its value. (I don't _think_ the Data API has any of that, but there's at least one case in the `rippled` APIs, so I'm wary...)

I'm going to leave this as is for consistency with the rest of the document, rather than investigating every API method just to be sure.

content/reference-data-api.md
mDuo13: Yeah, these aren't Report Objects. Changed the description to:

> Number of transaction stats objects in the `rows` field.

content/reference-data-api.md
mDuo13, line 3260: Leaving as-is, per my response there.

content/reference-data-api.md
mDuo13: I think it's debatable. Yes, in the broad sense it doesn't describe the date. I think it does describe the date "in terms of transaction stats." And it's clearly plainer language to phrase it this way than something like, "The date to which this Transaction Stats Object pertains."

I tried rephrasing this, but the best I could come up with is this:

> This object describes activity on this date.

content/reference-data-api.md
mDuo13, line 3318: As with the other, it's an array of daily summary objects, so no change here.

content/reference-data-api.md
mDuo13, line 3343: This description matches the convention used throughout the doc. The **Timestamp** link mentions the requirements (e.g. whether it has to be in UTC or not, which I think has changed over the course of the API's evolution)

content/reference-data-api.md
mDuo13, line 3347: As before, leaving this as-is.

content/reference-data-api.md
mDuo13: Changed to:

> This object describes activity on this date.

Not totally happy with that, but I guess it's better than the original wording.

content/reference-data-api.md
ryangyoung: Sounds good on both counts.

content/reference-data-api.md
ryangyoung, line 1803: Awesome! Didn't see the include until after I wrote my comment.

content/reference-data-api.md
ryangyoung, line 1804: Yep, that sounds good.

content/reference-data-api.md
ryangyoung, line 3220: Interesting, not what I would've expected. But if that's the case, then disregard my comment.

content/reference-data-api.md
ryangyoung: Cool. It might just be my eyes, but using monospace, bold, etc, seems more important for PDFs than HTML. Something to think about for the future.

content/reference-data-api.md
ryangyoung, line 3248: Oh man, I didn't think about a case where it just checks whether the field is there. In that case, makes sense to leave as is.

content/reference-data-api.md
ryangyoung: The rephrase is more what I was thinking of. I see your point, though, that, for the purposes of this call, the date is described by the transaction stats it returns. I'd vote for the rephrase, but it's your call.

PR #187, opened 2016-05-23
rippled - peer protocol crawl command (DOC-74)

mDuo13:
The task DOC-74 is for documenting the crawl method. However, it turns out that the crawl method is a special feature of the (mostly undocumented in the ripple-dev-portal) peer protocol instead of a standard JSON-RPC/WebSocket command. I stubbed out a new Peer Protocol section to house this, which we can flesh out in future doc efforts as time permits.

I intentionally deviated from the standard REST method template because this HTTP request for the crawl command is unique (even the peer protocol messages mostly don't work this way), so I have to clarify information that's usually put in another section that applies to all the methods in an API.


content/reference-rippled.md
```diff
@@ -10955,6 +10957,182 @@ The response follows the [standard format](#response-formatting), with a success
 
 * Any of the [universal error types](#universal-errors).
 
+
+
+
+# Peer Protocol #
+
+Servers in the Ripple Consensus Ledger communicate to each other using the Ripple peer protocol, also known as RTXP. Peer servers connect via HTTPS and then _upgrade_ to RTXP. (For more information, see the [Overlay Network article in the `rippled` repository](https://github.com/ripple/rippled/blob/906ef761bab95f80b0a7e0cab3b4c594b226cf57/src/ripple/overlay/README.md#handshake).)
```
vinniefalco: Suggested: "and then perform a HTTP Upgrade to RTXP"
Link:
https://tools.ietf.org/html/rfc7230#section-6.7

content/reference-rippled.md
```diff
@@ -10955,6 +10957,182 @@ The response follows the [standard format](#response-formatting), with a success
...
+
+## Configuring the Peer Protocol ##
+
+In order to participate in the Ripple Consensus Ledger, `rippled` servers connects to arbitrary peers using the peer protocol. (All such peers are treated as untrusted.)
```
vinniefalco: Partially correct. rippled supports "cluster peers" which are designated node public keys that have special privileges. For example, we trust that they verify signatures correctly. We operate such a cluster ourselves.

content/reference-rippled.md
```diff
@@ -10955,6 +10957,182 @@ The response follows the [standard format](#response-formatting), with a success
...
+			"uptime": 328776,
+			"version": "rippled-0.31.0-rc1"
+		}]
+	}
+}
+```
+
+### Concealing Peer Information ###
+
+You can use the `[peer_private]` stanza of the `rippled` config file to request that peer servers do not report your IP address in the Peer Crawler response. You cannot force peers you do not control to follow this request, if they run custom software. However, you can use this to hide the IP addresses of `rippled` servers you control that _only_ connect to peers you control (using `[ips_fixed]` and a firewall). This can help to protect important [validating servers](tutorial-rippled-setup.html#types-of-rippled-servers).
```
vinniefalco, line 11121: Good!!

reference-rippled.html
```diff
@@ -228,6 +228,10 @@
 <li class="level-2"><a href="#json">json</a></li>
 <li class="level-2"><a href="#connect">connect</a></li>
 <li class="level-2"><a href="#stop">stop</a></li>
+<li class="level-1"><a href="#peer-protocol">Peer Protocol</a></li>
```
vinniefalco, line 231: Is generated HTML supposed to be committed to the repository as well?

content/reference-rippled.md
mDuo13: Good catch. Mentioned that and linked to [the existing explanation of clustering](https://ripple.com/build/rippled-setup/#clustering).

reference-rippled.html
mDuo13, line 231: Yes. That way we can stage draft docs with GitHub Pages.

content/reference-rippled.md
```diff
@@ -10955,6 +10957,182 @@ The response follows the [standard format](#response-formatting), with a success
 
 * Any of the [universal error types](#universal-errors).
 
+
+
+
+# Peer Protocol #
+
+Servers in the Ripple Consensus Ledger communicate to each other using the Ripple peer protocol, also known as RTXP. Peer servers connect via HTTPS and then perform an [HTTP upgrade](https://tools.ietf.org/html/rfc7230#section-6.7) to RTXP. (For more information, see the [Overlay Network article in the `rippled` repository](https://github.com/ripple/rippled/blob/906ef761bab95f80b0a7e0cab3b4c594b226cf57/src/ripple/overlay/README.md#handshake).)
```
ryangyoung: Quibble: break this link into two parts: one for the title of the article (i.e., [Overlay Network](https://github.com/ripple/rippled/blob/906ef761bab95f80b0a7e0cab3b4c594b226cf57/src/ripple/overlay/README.md#handshake)) and one for the `rippled` repo (i.e., [`rippled` repository](https://github.com/ripple/rippled)). The second part might be redundant, but I feel like it's a "nice" thing to do for the user.

content/reference-rippled.md
```diff
@@ -10955,6 +10957,182 @@ The response follows the [standard format](#response-formatting), with a success
...
+```
+[port_peer]
+port = 51235
+ip = 0.0.0.0
+protocol = peer
+```
+
+## Peer Crawler ##
+
+The Peer Crawler reports information about other servers a `rippled` server is connected to as peers. The [`peers` command](#peers) in the [WebSocket and JSON-RPC APIs](#websocket-and-json-rpc-apis) also returns a similar, more comprehensive set of information, but requires admin access to the server. The Peer Crawler response is available to other servers on a non-privileged basis through the Peer Protocol (RTXP) port.
```
ryangyoung: "The Peer Crawler reports information about other servers a `rippled` server is connected to as peers."

There might not be a fix for this, but I had to read that sentence a couple of times to understand it. The only possible improvement I could think of is to add "the", as in "The Peer Crawler reports information about _the_ other servers a `rippled` server is connected to as peers."

content/reference-rippled.md
```diff
@@ -10955,6 +10957,182 @@ The response follows the [standard format](#response-formatting), with a success
...
+
+#### Request Format ####
+
+To request the Peer Crawler information, make the following HTTP request:
+
+* **Protocol:** https
```
ryangyoung, line 10990: Why is the request in this format? I assume there's a good reason, but it seems to deviate from the rest of the doc.

mDuo13, line 10990: As stated in the intro to this pull request (no surprise you missed it, the link kind of skips past it):


I intentionally deviated from the standard REST method template because this HTTP request for the crawl command is unique (even the peer protocol messages mostly don't work this way), so I have to clarify information that's usually put in another section to describe all the methods in an API.

content/reference-rippled.md
mDuo13: The lack of specificity ("about other servers...") is intentional. Since the Crawl command can be instructed (as described in the doc) to leave out some servers from its response, I felt that saying, "the other servers..." was too strong because it implies "all the other servers..."

content/reference-rippled.md
ryangyoung, line 10990: Ah! The good reason I knew existed.

content/reference-rippled.md
ryangyoung: The sentence is like LOTR: it seems to end, but then keeps on going. The especially confusing part for me is the "peers" at the end, which refers back to "servers." But, like I said, there might not really be a way to improve the sentence (to my level of understanding, anyway). That said, I'm always in favor of explaining these nuances more verbosely upfront (for example, maybe with the explanation in your comment and a link to the "Concealing Peer Information" section where that exception is fully explained. My concern is that if I had to read the sentence more than once, other users probably will, too, and that could make a reader's experience just a little less than pleasant.

PR #197, opened 2016-06-24
rippled 0.32.0

mDuo13:
Changes for `rippled` 0.32.0 -
- new `telCAN_NOT_QUEUE` transaction result code
- `server_info` and `server_state` command now include load factors
- `fee` command is no longer admin-only
- transaction queue can now hold up to 10 transactions from the same sending address
- `validations` subscription stream has new fields
- `wallet_propose` command now returns a warning when key may be low-entropy


mDuo13:
Merging this so we can deploy the changes. Still up for comments; we can revise and deploy again later after incorporating review feedback.

content/concept-transaction-cost.md
```diff
@@ -57,17 +57,20 @@ See also: [Fee Escalation explanation in `rippled` repository](https://github.co
...
+#### Queuing Restrictions ####
+
+The `rippled` server uses a variety of heuristics to estimate which transactions are "likely to be included in a ledger." The current implementation uses the following rules to decide which transactions to queue:
+
+* Transactions must be properly-formed and [authorized](reference-transaction-format.html#authorizing-transactions) with valid signatures.
+* Transactions with an `AccountTxnID` field cannot be queued.
+* A single sending address can have at most 10 transactions queued at the same time. In order for a transaction to be queued, the sender must have enough XRP to pay all the XRP costs of all the sender's queued transactions including both the `Fee` fields and the sum of the XRP that each transaction could send. If a transaction affects how the address authorizes accounts, no other transactions from the same address can be queued behind it. _(New in [`rippled` 0.32.0](https://github.com/ripple/rippled/releases/tag/0.32.0))_
+* If the transaction includes a `LastLedgerSequence` field, the value of that field must not be more than **the current ledger index + 2**.
```
ximinez, line 71: ...must not be less than the current ledger index + 2.

content/reference-rippled.md
```diff
@@ -8624,8 +8654,14 @@ The `state` object may have some arrangement of the following fields:
...
+| load\_base | Integer | This amount of server load is the baseline that is used to decide how much to charge in transaction fees; if the `load_factor` is equal to the `load_base` then only the base fee is enforced; if the `load_factor` is double the `load_base` then transaction costs are doubled. |
+| load\_factor | Number | The load factor the server is currently enforcing. The ratio between this value and the `load_base` determines the multiplier for transaction fees. The load factor is determined by the highest of the individual server's load factor, cluster's load factor, and the overall network's load factor. |
+| load\_factor\_local | Integer | (May be omitted) The load factor based on load to this server. |
+| load\_factor\_net | Integer | (May be omitted) The load factor being used by the rest of the network (estimated). |
+| load\_factor\_cluster | Integer | (May be omitted) The load factor based on load to servers in [this cluster](tutorial-rippled-setup.html#clustering). |
```
ximinez, line 8661: (I think this is the block for `server_state`, but it's hard to tell in the diff.)
The load_factor_local, load_factor_net, and load_factor_cluster fields are never included in the `server_state` output. https://github.com/ximinez/rippled/blob/5dd8f282905c70d92790fde3f1d384e43c925ff3/src/ripple/app/misc/NetworkOPs.cpp#L2099-L2118

content/reference-rippled.md
```diff
@@ -8624,8 +8654,14 @@ The `state` object may have some arrangement of the following fields:
...
+| load\_factor\_fee\_escalation | Integer | (May be omitted) The current multiplier to the [transaction cost][] that a transaction must pay to get into the open ledger, in fee levels. _(New in [version 0.32.0][])_ |
+| load\_factor\_fee\_queue | Integer | (May be omitted) The current multiplier to the [transaction cost][] that a transaction must pay to get into the queue, if the queue is full, in fee levels. _(New in [version 0.32.0][])_ |
+| load\_factor\_fee\_reference | Integer | (May be omitted) The [transaction cost][] with no load scaling, in fee levels. _(New in [version 0.32.0][])_ |
```
ximinez, line 8664: Similarly, the load_factor_fee_\* fields will always be included if the fee escalation amendment is enabled.

content/reference-transaction-format.md
```diff
@@ -1049,6 +1049,7 @@ These codes indicate an error in the local server processing the transaction; it
 | telBAD\_DOMAIN | The transaction specified a domain value (for example, the `Domain` field of an [AccountSet transaction](#accountset)) that cannot be used, probably because it is too long to store in the ledger. |
 | telBAD\_PATH_COUNT | The transaction contains too many paths for the local server to process. |
 | telBAD\_PUBLIC\_KEY | The transaction specified a public key value (for example, as the `MessageKey` field of an [AccountSet transaction](#accountset)) that cannot be used, probably because it is too long. |
+| telCAN\_NOT\_QUEUE | The transaction did not meet the [open ledger cost](concept-transaction-cost.html), but this server did not queue it because it failed one of the server's heuristics for being "likely to succeed". You can try again later or sign and submit a replacement transaction with a higher transaction cost in the `Fee` field. |
```
ximinez, line 1052: I don't think there are any cases where a higher fee will overcome `telCAN_NOT_QUEUE`. (Any where it would should still return `telINSUF_FEE_P`.)

Here's a complete list of what will return a `telCAN_NOT_QUEUE` as of 0.32.0:
- Attempting to replace a tx sequence already in the queue (but not the first or last) with a tx that affects how the address authorizes accounts
- Attempting to queue a new (later sequence) tx when a tx already in the queue affects how the address authorizes accounts.
- Attempting to queue a new tx or replace a queued tx when all the other queued txs already potentially leave the account with too low of a balance to cover the fee or account reserve.
- The tx can't go into the open ledger and would otherwise pass the heuristics to be queued, but the tx has too low a `LastLedgerSequence`, or the tx has a `AccountTxnID` or `PreviousTxnID`. (Because of efforts to save work, if the account has 10 txs in the queue, and this would be the 11th,  the tx will get a `terPRE_SEQ`.)
- The tx can't go into the open ledger and would otherwise pass heuristics, but the queue is full, and lowest fee tx in the queue is for the same account.

In short, there really isn't much a user can do to the current candidate transaction to overcome `telCAN_NOT_QUEUE`, except to increase the `LastLedgerSequence` if that's the problem. That's pretty much why it needed a new result code. The only options I can think of are to wait or to try to replace / reprocess some of the earlier transactions in the queue (for example, replace the first queued tx with one that pays the open ledger fee. This will, of course, require a lot more work on the client to track such things.)

content/concept-transaction-cost.md
```diff
@@ -47,7 +47,7 @@ Each `rippled` server maintains a cost threshold based on its current load. If y
 
 ## Open Ledger Cost ##
 
-A `rippled` server with the [FeeEscalation amendment](concept-amendments.html#feeescalation) enabled has a second mechanism for enforcing the transaction cost, called the _open ledger cost_. A transaction can only be included in the open ledger if it meets the open ledger cost requirement in XRP. Transactions that do not meet the open ledger cost are [queued for a following ledger](#queued-transactions) instead.
+The `rippled` server has a second mechanism for enforcing the transaction cost, called the _open ledger cost_. A transaction can only be included in the open ledger if it meets the open ledger cost requirement in XRP. Transactions that do not meet the open ledger cost are [queued for a following ledger](#queued-transactions) instead.
```
ximinez, line 50: Suggest:
`...do not meet the open ledger cost may be queued for a following ledger instead.`

content/reference-rippled.md
mDuo13, line 8664: I think I'll leave this one as-is because it's better to assume the fields might be omitted, for example if starting up in stand-alone mode such that the amendment isn't applied.

content/reference-transaction-format.md
mDuo13, line 1052: Why can't you just increase the `Fee` value enough to make it into the open ledger?

ximinez, line 1052: Most of those cases are for transactions which are not for the next sequence for the account. If they didn't get `telCAN_NOT_QUEUE`, they'd get `terPRE_SEQ` or some other error. Any exceptions to those rules (ie. if I wasn't detailed enough), where the transaction is the next one for the account should return `telINSUFF_FEE`.

Example 1:
Account's next sequence number is 100. Fee is escalated. Transaction 100 pays 10 drops, gets `terQUEUED`. Transaction 101 would affect the ability to claim a fee, gets `terQUEUED`. Transaction 102 gets `telCAN_NOT_QUEUE` pretty much no matter what, not because it doesn't have a high enough fee, but because transaction 101 is preventing it from getting queued and relayed. Once 101 gets into the open ledger, 102 may or may not also be able to get into the open ledger or be queued itself.

Example 2:
Account's next sequence number is 100. Current ledger sequence is 500. Fee is escalated. Transaction 100 pays 10 drops, has LastLedgerSequence == 501. That gets `telINSUFF_FEE`. User resubmits transaction 100 with the open ledger fee, it succeeds.

Example 3:
Same scenario as example 3. Transaction 100 pays 10 drops, has LastLedgerSequence == 503, gets `terQUEUED`. Transaction 101 pays the open ledger fee, has LastLedgerSequence == 501. That'll get `telCAN_NOT_QUEUE`. Again, no matter how high the fee on 101, it can't get into the open ledger, because 100 is in the queue, and it can't get into the queue because the LastLedgerSequence is too low. Once 100 gets into the open ledger, then this situation becomes Example 2.

content/reference-rippled.md
```diff
@@ -3012,7 +3012,7 @@ The response follows the [standard format](#response-formatting), with a success
 | account\_id | String | The [Address][] of the account. |
 | public\_key | String | The public key of the account, in encoded string format. |
 | public\_key\_hex | String | The public key of the account, in hex format. |
-{#| warning | String | (May be omitted) If the request specified a seed value, this field provides a warning that it may be insecure. TODO: uncomment when this gets added in 0.32.0 |#}
+| warning | String | (May be omitted) If the request specified a seed value, this field provides a warning that it may be insecure. _(New in [version 0.32.0][])_ |
```
ryangyoung, line 3015: This looks fine, but noticed a few other, minor things in the rest of this section: 
- "The wallet_propose request is an admin command that cannot be run by unpriviledged users!"
  1) "unpriviledged" --> "unprivileged" 2) is this the way the other admin commands are identified? Seems fine, just thinking the format should be uniform 3) "admin command" links to "Connecting to rippled," which has a link to all the Admin commands. Should there also be a reciprocal link from the "List of Admin Commands" to "Connecting to rippled"?
- "The commandline version of this command cannot generate Ed25519 keys."
  Shouldn't this be in the description of the `key_type` parameter?
- "Specifying a Seed"
  Should this be a header? Also, it seems a little weird to start a section with a warning. Maybe the text in the "Caution" callout could just be text under this header? (Oh, it looks like it _is_ level 5 header, it just renders as paragraph text. I guess our styles don't handle 5th level headers, but maybe it should.)

content/reference-rippled.md
```diff
@@ -8624,8 +8654,14 @@ The `state` object may have some arrangement of the following fields:
...
 | load.job\_types | Array | *Admin only* Information about the rate of different types of jobs the server is doing and how much time it spends on each. |
 | load.threads | Number | *Admin only* The number of threads in the server's main job pool. |
-| load\_base | Number | This amount of server load is the baseline that is used to decide how much to charge in transaction fees; if the `load_factor` is equal to the `load_base` then only the base fee is enforced; if the `load_factor` is double the `load_base` then transaction fees are doubled. |
-| load\_factor | Number | The load factor the server is currently enforcing. The ratio between this value and the load\_base determines the multiplier for transaction fees. The load factor is determined by the highest of the individual server's load factor, cluster's load factor, and the overall network's load factor. |
+| load\_base | Integer | This amount of server load is the baseline that is used to decide how much to charge in transaction fees; if the `load_factor` is equal to the `load_base` then only the base fee is enforced; if the `load_factor` is double the `load_base` then transaction costs are doubled. |
```
ryangyoung, line 8657: "This amount of server load is the baseline that is used to decide how much to charge in transaction fees; if the load_factor is equal to the load_base then only the base fee is enforced; if the load_factor is double the load_base then transaction costs are doubled."

Suggested rephrase: "The baseline used [by `rippled`?] to decide how much to charge in [transaction fees](content/concept-fees.html). If the `load_factor`is equal to the `load_base`, then only the base fee is enforced. If the `load_factor` is double the `load_base`, then [transaction costs](content/concept-transaction-cost.html) are doubled."

Also, is it worth explaining, here or somewhere else, what happens if the `load_factor` is 1.5 or 3 times the `load_base`? I think the answer is implied, but would probably be making that explicit since this is a tricky issue.

content/reference-rippled.md
```diff
@@ -8345,6 +8368,11 @@ The `info` object may have some arrangement of the following fields:
 | load.job\_types | Array | *Admin only* Information about the rate of different types of jobs the server is doing and how much time it spends on each. |
 | load.threads | Number | *Admin only* The number of threads in the server's main job pool. |
 | load\_factor | Number | The load factor the server is currently enforcing, as a multiplier on the base transaction cost. The load factor is determined by the highest of the individual server's load factor, cluster's load factor, and the overall network's load factor. **Note:** This `load_factor` is calculated as the ratio of the `load_factor` and the `load_base` that are reported by the [`server_state` command](#server-state) |
```
ryangyoung, line 8370: Maybe add an example of this value?

content/reference-rippled.md
```diff
@@ -8345,6 +8368,11 @@ The `info` object may have some arrangement of the following fields:
...
+| load\_factor\_local | Number | (May be omitted) Current multiplier to the [transaction cost][] based on load to this server. |
```
ryangyoung, line 8371: Maybe a couple of words about why this would be omitted (ie, if there is no multiplier, I think?).

content/reference-rippled.md
```diff
@@ -8345,6 +8368,11 @@ The `info` object may have some arrangement of the following fields:
...
+| load\_factor\_net | Number | (May be omitted) Current multiplier to the [transaction cost][] being used by the rest of the network (estimated). |
```
ryangyoung, line 8372: I'm curious how this is estimated. Worth explaining that here?

content/reference-rippled.md
```diff
@@ -8345,6 +8368,11 @@ The `info` object may have some arrangement of the following fields:
...
+| load\_factor\_cluster | Number | (May be omitted) Current multiplier to the [transaction cost][] based on load to servers in [this cluster](tutorial-rippled-setup.html#clustering). |
+| load\_factor\_fee\_escalation | Number | (May be omitted) The current multiplier to the [transaction cost][] that a transaction must pay to get into the open ledger. _(New in [version 0.32.0][])_ |
+| load\_factor\_fee\_queue | Number | (May be omitted) The current multiplier to the [transaction cost][] that a transaction must pay to get into the queue, if the queue is full. _(New in [version 0.32.0][])_ |
```
ryangyoung, line 8375: I find the phrase "The current multiplier to the [transaction cost][] that a transaction must pay" a little confusing. Is the value in this field what I need to pay? Or do I need to multiply my transaction cost by this? (Again, I think this might be implied, but would probably be better to make it totally clear.) Also, what are "fee levels" (don't see this in the diff, but I see it here: http://mduo13.github.io/ripple-dev-portal/reference-rippled.html#server-state).

content/reference-rippled.md
```diff
@@ -8624,8 +8654,14 @@ The `state` object may have some arrangement of the following fields:
...
-| load\_factor | Number | The load factor the server is currently enforcing. The ratio between this value and the load\_base determines the multiplier for transaction fees. The load factor is determined by the highest of the individual server's load factor, cluster's load factor, and the overall network's load factor. |
+| load\_base | Integer | This amount of server load is the baseline that is used to decide how much to charge in transaction fees; if the `load_factor` is equal to the `load_base` then only the base fee is enforced; if the `load_factor` is double the `load_base` then transaction costs are doubled. |
+| load\_factor | Number | The load factor the server is currently enforcing. The ratio between this value and the `load_base` determines the multiplier for transaction fees. The load factor is determined by the highest of the individual server's load factor, cluster's load factor, and the overall network's load factor. |
```
ryangyoung, line 8658: Maybe add a link to [transaction fees](content/concept-transfer-fee.html)?

content/reference-rippled.md
```diff
@@ -8624,8 +8654,14 @@ The `state` object may have some arrangement of the following fields:
...
+| load\_factor\_local | Integer | (May be omitted) The load factor based on load to this server. |
+| load\_factor\_net | Integer | (May be omitted) The load factor being used by the rest of the network (estimated). |
+| load\_factor\_cluster | Integer | (May be omitted) The load factor based on load to servers in [this cluster](tutorial-rippled-setup.html#clustering). |
+| load\_factor\_fee\_escalation | Integer | (May be omitted) The current multiplier to the [transaction cost][] that a transaction must pay to get into the open ledger, in fee levels. _(New in [version 0.32.0][])_ |
```
ryangyoung, line 8662: Add a link to "fee levels" description (here and other instances): http://mduo13.github.io/ripple-dev-portal/concept-transaction-cost.html#fee-levels

content/reference-rippled.md
```diff
@@ -7648,23 +7648,46 @@ The validations stream sends messages whenever it receives validation messages,
...
 The fields from a validations stream message are as follows:
 
 | Field | Type | Description |
 |-------|------|-------------|
-| type | String | `validationReceived` indicates this is from the validations stream |
-| ledger\_hash | String | The identifying hash of the proposed ledger is being validated. |
-| ledger\_index | String - Integer | The [Ledger Index][] of the proposed ledger. _(New in [version 0.31.0][])_ |
-| signature | String | The signature that the validator used to sign its vote for this ledger. |
-| validation\_public\_key | String | The base-58 encoded public key from the key-pair that the validator used to sign the message. This identifies the validator sending the message and can also be used to verify the `signature`. |
+| `type` | String | The value `validationReceived` indicates this is from the validations stream. |
```
ryangyoung, line 7678: Noticed that `backticks` are used for fields in some pages but not others. Should try to standardize that across pages, products, doc sets, etc.

content/reference-rippled.md
```diff
@@ -7648,23 +7648,46 @@ The validations stream sends messages whenever it receives validation messages,
...
+| `amendments` | Array of Strings | (May be omitted) The [amendments](concept-amendments.html) this server wants to be added to the protocol. _(New in [version 0.32.0][])_ |
+| `base_fee` | Integer | (May be omitted) The unscaled transaction cost (`reference_fee` value) this server wants to set by [Fee Voting](concept-fee-voting.html). _(New in [version 0.32.0][])_ |
+| `flags` | Number | Bit-mask of flags added to this validation message. The flag 0x80000000 indicates that the validation signature is fully-canonical. The flag 0x00000001 indicates that this is a full validation; otherwise it's a partial validation. <!--{# TODO: what's a partial validation? #}--> _(New in [version 0.32.0][])_ |
+| `full` | Boolean | If `true`, this is a full validation. Otherwise, this is a partial validation. _(New in [version 0.32.0][])_ |
```
ryangyoung, line 7682: Maybe link to explanation of what the difference is (if one exists)?

content/concept-transaction-cost.md
```diff
@@ -57,17 +57,20 @@ See also: [Fee Escalation explanation in `rippled` repository](https://github.co
 
 ### Queued Transactions ###
 
-(Requires the [FeeEscalation amendment](concept-amendments.html#feeescalation))
-
 When `rippled` receives a transaction that meet the server's local load cost but not the open ledger cost, the server estimates whether the transaction is "likely to be included" in a later ledger. If so, the server adds the transaction to the transaction queue and relays the transaction to other members of the network. Otherwise, the server discards the transaction. The server tries minimize the amount of network load caused by transactions that would not pay a transaction cost, since [the transaction cost only applies when a transaction is included in a validated ledger](#transaction-costs-and-failed-transactions).
```
ryangyoung, line 60: Link back to "open ledger cost" section.

content/reference-rippled.md
mDuo13, line 3015: - wallet_propose:
  1. Fixed 17 cases by find/replace
  2. Yes, the format is uniform for all 17 admin commands
  3. Good idea. Added.
- Actually, I realized this whole section changed again with 0.32.0 so I needed to reorganize it. The commandline syntax doesn't support the "seed" or "seed_hex" fields, either, but that's not a limitation in functionality since the passphrase field can cover those values. I joined the Ed25519 experimental note and the comment that the commandline version can't generate Ed25519 keys together.
- "Specifying a Seed" -
  1. The problem is that our style actually does "handle" level 5 headers, insomuch as it explicitly styles them to look a lot like paragraph text. I can raise with Ben.
  2. Agreed, it makes sense to just make this text instead of a callout. Done.

content/concept-transaction-cost.md
ryangyoung, line 60: "The server tries minimize the amount of network load caused by transactions that would not pay a transaction cost," --> missing "to"

content/reference-rippled.md
mDuo13, line 8658: I think you're confusing transfer fees with transaction fees. Transaction fees (old terminology) = transaction costs (new terminology). Transfer fees are different, which is one of the reasons we changed it.

This was one a few cases of old terminology persisting in this section. Fixed.

content/concept-transaction-cost.md
```diff
@@ -57,17 +57,20 @@ See also: [Fee Escalation explanation in `rippled` repository](https://github.co
...
 
-The `rippled` server uses a variety of heuristics to estimate which transactions are "likely to be included in a ledger."  Most importantly, those transactions must be properly-formed and [authorized](reference-transaction-format.html#authorizing-transactions) with valid signatures.
-
 When the current open ledger closes and the server starts a new open ledger, the server starts taking transactions from the queue to include in the new open ledger. The transaction queue is sorted with the transactions that would pay the highest transaction cost first, proportional to the un-scaled cost of those transactions. Transactions that pay the same transaction cost are queued in the order the server receives them.
```
ryangyoung, line 62: Maybe add a link to `server_info` or `server_state` to explain scaled vs. unscaled. Or somewhere else if there's a better description elsewhere.

content/reference-rippled.md
mDuo13, line 8657: Changed to:

> This is the baseline amount of server load used in [transaction cost](concept-transaction-cost.html) calculations. If the `load_factor` is equal to the `load_base` then only the base transaction cost is enforced. If the `load_factor` is higher than the `load_base`, then transaction costs are multiplied by the ratio between them. For example, if the `load_factor` is double the `load_base`, then transaction costs are doubled.

content/concept-transaction-cost.md
```diff
@@ -57,17 +57,20 @@ See also: [Fee Escalation explanation in `rippled` repository](https://github.co
...
 
-**Note:** As of `rippled` **version 0.31.0**, the transaction queue supports at most 1 transaction per sending address. This is expected to change in later versions.
+#### Queuing Restrictions ####
+
+The `rippled` server uses a variety of heuristics to estimate which transactions are "likely to be included in a ledger." The current implementation uses the following rules to decide which transactions to queue:
```
ryangyoung, line 66: This is great! Should really help users understand this new feature, which I know has caused some confusion.

content/reference-rippled.md
mDuo13, line 8371: I'm not 100% sure on the logic of when it can be omitted, but I don't think it's as simple as "when there's no multiplier" so I'd rather just leave it vague.

content/reference-rippled.md
mDuo13, line 8372: The explanation would be too long for this section. It involves servers reporting their current load as part of their validation votes. I changed "estimated" to "estimated from other servers' reported load values" as a compromise.

content/reference-rippled.md
mDuo13, line 7678: Of course. This edit is working incrementally towards using backticks in all cases.

content/reference-rippled.md
mDuo13, line 7682: Yeah, I just found out after writing this what a partial validation is. It's complicated, of course.

content/reference-rippled.md
mDuo13, line 8375: Is this better?

> The current multiplier to the [transaction cost][] to get into the open ledger, in [fee levels][].

content/reference-rippled.md
ryangyoung, line 8372: Sounds like that's enough for here.

content/reference-rippled.md
ryangyoung, line 8375: Yep, that clears it up for me.

content/reference-rippled.md
ryangyoung, line 8658: I know it changed, just didn't catch that this was a "legacy" usage, where the old meant the new. Looks good now!

PR #199, opened 2016-07-11
Correct 'Going Below Reserve Requirement'

brettmollin:
Brett Mollin [1:17 PM]  
In contrast to this document, I seem to be able to keep burning XRP in transaction fees well after I am below the reserve. https://ripple.com/build/reserves/#going-below-the-reserve-requirement

"When an address holds less XRP than its current reserve requirement, it cannot send new transactions. "

Vinnie Falco [1:18 PM]  
It can't send XRP. But it can use that XRP to pay for transactions.

[1:18]  
"it cannot send new transactions" should read "it cannot send XRP to other accounts"


mDuo13:
Edits need to go to the markdown files in the `content/` folder. Edits to the HTML files will be lost when the docs are re-built.

Also, I'm not sure what you mean by this:

> [An address below its reserve requirement] cannot... process transactions that increase the reserve

Since _addresses_ never process transactions. Maybe what you mean is that the ledger won't execute transactions that increase an address's owner reserve if the address is below the requirement already?

brettmollin:
Got ya

you are correct in what I was attempting to say

mDuo13:
This got resolved by https://github.com/ripple/ripple-dev-portal/commit/55a2ea08f2037c9627aeebc444b8314bac902da8

PR #200, opened 2016-07-12
Update concept-reserves.md

brettmollin:
make clear that transactions that do not bring up the reserve can still be made once you are under the reserve.


mDuo13:
Looks a lot better. I think there's still some minor rewording necessary, but I can take care of that.

PR #202, opened 2016-08-01
Update FlowV2 amendment description

mDuo13:
Also updates the TrustSetAuth sections to properly reflect the amendment being live. (It's a little late, but I was on vacation when it happened.)


ryangyoung:
>  (It's a little late, but I was on vacation when it happened.)

Nevertheless, I should've remembered to update it. To that end, I updated the [rippled Release Communications](https://ripplelabs.atlassian.net/wiki/display/DOCS/rippled+Release+Communications) page to include updating that section as part of the Amendment announcement process.

PR #205, opened 2016-08-19
amendments - clarify flowv2 status, add flow

mDuo13:
Addresses [DOC-525](https://ripplelabs.atlassian.net/browse/DOC-525) and [DOC-526](https://ripplelabs.atlassian.net/browse/DOC-526):
- Clarifies how vetoing amendments works (voting goes on forever, but removing the amendment makes it unlikely to get a majority)
- Clarifies that FlowV2 was vetoed and Flow is coming to replace it.

### DIFF
--- a/concept-amendments.html
+++ b/concept-amendments.html
@@ -178,7 +178,7 @@
 <p>Every amendment has a unique identifying hex value and a short name. The short name is for human use, and is not used in the amendment process. Two servers can support the same amendment ID while using different names to describe it. An amendment's name is not guaranteed to be unique.</p>
 <p>See also: <a href="#known-amendments">Known Amendments</a></p>
 <h2 id="amendment-process">Amendment Process</h2>
-<p>Every 256th ledger is called a "flag" ledger. The process of approving an amendment starts in the ledger version immediately before the flag ledger. When <code>rippled</code> validator servers send validation messages for that ledger, those servers also submit votes in favor of specific amendments. (<a href="concept-fee-voting.html">Fee Voting</a> also occurs around flag ledgers.)</p>
+<p>Every 256th ledger is called a "flag" ledger. The process of approving an amendment starts in the ledger version immediately before the flag ledger. When <code>rippled</code> validator servers send validation messages for that ledger, those servers also submit votes in favor of specific amendments. If a validator does not vote in favor of an amendment, that is the same as voting against the amendment. (<a href="concept-fee-voting.html">Fee Voting</a> also occurs around flag ledgers.)</p>
 <p>The flag ledger itself has no special contents. However, during that time, the servers look at the votes of the validators they trust, and decide whether to insert an <a href="reference-transaction-format.html#enableamendment"><code>EnableAmendment</code> pseudo-transaction</a> into the following ledger. The flags of an EnableAmendment pseudo-transaction show what the server thinks happened:</p>
 <ul>
 <li>The <code>tfGotMajority</code> flag means that support for the amendment has increased to at least 80% of trusted validators.</li>
@@ -195,10 +195,9 @@
 </ul>
 <p>Theoretically, a <code>tfLostMajority</code> EnableAmendment pseudo-transaction could be included in the same ledger as the pseudo-transaction to enable an amendment. In this case, the pseudo-transaction with the <code>tfLostMajority</code> pseudo-transaction has no effect.</p>
 <h2 id="amendment-voting">Amendment Voting</h2>
-<p>Operators of <code>rippled</code> validators can choose which amendments to support or reject using the <a href="reference-rippled.html#feature"><code>feature</code> command</a>. This decides which amendments the validator votes for in the <a href="#amendment-process">amendment process</a>. By default, <code>rippled</code> votes in favor of every amendment it knows about.</p>
-<p>The operator of a <code>rippled</code> validator can "veto" an amendment. In this case, that validator never sends a vote in favor of the amendment. If enough servers veto an amendment, that prevents it from reaching consistent 80% support, so the amendment does not apply. The amendment can still become enabled later if an 80% majority of trusted validators support the amendment, either because validator operators changed their minds or new validators that support the amendment became trusted.</p>
-<p>An amendment can gain and lose a majority an unlimited number of times before it becomes permanently enabled. As long as there are servers voting in favor of an amendment, the amendment is not permanently rejected. New versions of <code>rippled</code> can remove an amendment to ensure that they do not vote for it.</p>
-<p>As with all aspects of the consensus process, amendment votes are only taken into account by servers that trust the validators sending those votes. Ripple (the company) recommends only trusting the 5 default validators that Ripple (the company) operates. For now, trusting only those validators is enough to coordinate with Ripple (the company) on releasing new features.</p>
+<p>Each version of <code>rippled</code> is compiled with a list of known amendments and the code to implement those amendments. By default, <code>rippled</code> supports known amendments and opposes unknown amendments. Operators of <code>rippled</code> validators can <a href="#configuring-amendment-voting">configure their servers</a> to explicitly support or oppose certain amendments, even if those amendments are not known to their <code>rippled</code> versions.</p>
+<p>To become enabled, an amendment must be supported by at least 80% of trusted validators continuously for two weeks. If support for an amendment goes below 80% of trusted validators, the amendment is temporarily rejected. The two week period starts over if the amendment regains support of at least 80% of trusted validators. (This can occur if validators vote differently, or if there is a change in which validators are trusted.) An amendment can gain and lose a majority an unlimited number of times before it becomes permanently enabled. An amendment cannot be permanently rejected, but it becomes very unlikely for an amendment to become enabled if new versions of <code>rippled</code> do not have the amendment in their known amendments list.</p>
+<p>As with all aspects of the consensus process, amendment votes are only taken into account by servers that trust the validators sending those votes. At this time, Ripple (the company) recommends only trusting the 5 default validators that Ripple (the company) operates. For now, trusting only those validators is enough to coordinate with Ripple (the company) on releasing new features.</p>
 <h3 id="configuring-amendment-voting">Configuring Amendment Voting</h3>
 <p>You can temporarily configure an amendment using the <a href="reference-rippled.html#feature"><code>feature</code> command</a>. To make a persistent change to your server's support for an amendment, change your server's <code>rippled.cfg</code> file.</p>
 <p>Use the <code>[veto_amendments]</code> stanza to list amendments you do not want the server to vote for. Each line should contain one amendment's unique ID, optionally followed by the short name for the amendment. For example:</p>
@@ -238,7 +237,7 @@ TrustSetAuth
 <tr>
 <th>Name</th>
 <th>Introduced</th>
-<th>Enabled</th>
+<th>Status</th>
 </tr>
 </thead>
 <tbody>
@@ -250,7 +249,7 @@ TrustSetAuth
 <tr>
 <td><a href="#flowv2">FlowV2</a></td>
 <td>v0.32.1</td>
-<td>Vetoed</td>
+<td>To be removed</td>
 </tr>
 <tr>
 <td><a href="#tickets">Tickets</a></td>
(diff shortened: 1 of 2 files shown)


content/concept-amendments.md
```diff
@@ -138,16 +141,23 @@ A transaction remains in the queue until one of the following happens:
...
-The FlowV2 Engine also makes it easier to improve and expand the payment engine with further Amendments.
+The Flow Engine also makes it easier to improve and expand the payment engine with further Amendments.
+
+## FlowV2 ##
+
+| Amendment ID | Status |
+|--------------|--------|
+| 5CC22CFF2864B020BD79E0E1F048F63EF3594F95E650E43B3F837EF1DF5F4B26 | Failed to hold a majority. To be removed. |
 
+This amendment was intended to replace the payment processing engine with a more robust and efficient rewrite called the FlowV2 engine. However, a critical bug was found during the voting period, so key validators vetoed the Amendment and it lost its majority. Ripple plans to remove the FlowV2 amendment in future versions of `rippled` and replace it with the [Flow](#flow) amendment.
```
nbougalis: I don't know if we want to say "critical bug" as opposed to using the wording we used on the announcement: "flaw". Additionally, should we link to the commit that addresses this? If so, it's here: https://github.com/ripple/rippled/commit/b92a7d415eecb07ace2b72b6792d9dfa489c5a04

concept-amendments.html
```diff
@@ -297,6 +304,23 @@ <h2 id="feeescalation">FeeEscalation</h2>
...
+</tr>
+</thead>
+<tbody>
+<tr>
+<td>740352F2412A9909880C23A559FCECEDA3BE2126FED62FC7660D628A06927F11</td>
+<td>TBD</td>
+</tr>
+</tbody>
+</table>
+<p>Replaces the payment processing engine with a more robust and efficient rewrite called the Flow engine. The new version of the payment processing engine is intended to follow the same rules as the old one, but occasionally produces different results due to floating point rounding. This Amendment was created to replace the <a href="#flowv2">FlowV2</a> amendment when a critical bug was found in the FlowV2 amendment.</p>
```
nbougalis: Propose rewording last sentence: 

> This amendment supersedes the <a href="#flowv2">FlowV2</a> amendment.

concept-amendments.html
```diff
@@ -308,12 +332,11 @@ <h2 id="flowv2">FlowV2</h2>
...
 <tr>
 <td>5CC22CFF2864B020BD79E0E1F048F63EF3594F95E650E43B3F837EF1DF5F4B26</td>
-<td>Planned for voting, expected to be enabled 2016-08-24</td>
+<td>Failed to hold a majority. To be removed.</td>
 </tr>
 </tbody>
 </table>
-<p>Replaces the payment processing engine with a more robust and efficient rewrite called the FlowV2 engine. The new version of the payment processing engine is intended to follow the same rules as the old one, but occasionally produces different results due to floating point rounding.</p>
-<p>The FlowV2 Engine also makes it easier to improve and expand the payment engine with further Amendments.</p>
+<p>This amendment was intended to replace the payment processing engine with a more robust and efficient rewrite called the FlowV2 engine. However, a critical bug was found during the voting period, so key validators vetoed the Amendment and it lost its majority. Ripple plans to remove the FlowV2 amendment in future versions of <code>rippled</code> and replace it with the <a href="#flow">Flow</a> amendment.</p>
```
nbougalis: Again, "critical bug"

content/concept-amendments.md
```diff
@@ -1,6 +1,6 @@
 # Amendments #
 
-_(New in [version 0.31.0](https://wiki.ripple.com/Rippled-0.31.0))_
+_(Introduced in [version 0.31.0](https://wiki.ripple.com/Rippled-0.31.0))_
```
nbougalis, line 3: Just a suggestion for some visual spice, should we consider using badges for this? For example:

# Amendments

Introduced in version 0.31.0

This lets us do other fancy things. For example:

# FlowV2

Vetoed

# MultiSign

Ratified

mDuo13, line 3: That's an interesting idea. I'll look into it.

content/concept-amendments.md
```diff
@@ -48,7 +48,9 @@ Theoretically, a `tfLostMajority` EnableAmendment pseudo-transaction could be in
 
 Operators of `rippled` validators can choose which amendments to support or reject using the [`feature` command](reference-rippled.html#feature). This decides which amendments the validator votes for in the [amendment process](#amendment-process). By default, `rippled` votes in favor of every amendment it knows about.
 
-The operator of a `rippled` validator can "veto" an amendment. In this case, that validator never sends a vote in favor of the amendment. If enough servers veto an amendment, that prevents it from reaching consistent 80% support, so the amendment does not apply.
+The operator of a `rippled` validator can "veto" an amendment. In this case, that validator never sends a vote in favor of the amendment. If enough servers veto an amendment, that prevents it from reaching consistent 80% support, so the amendment does not apply. The amendment can still become enabled later if an 80% majority of trusted validators support the amendment, either because validator operators changed their minds or new validators that support the amendment became trusted.
```
nbougalis: "consistent 80% support" isn't defined. We should explain that once the amendment reaches an 80% majority it must hold it for two weeks, at which point it becomes enabled. If that majority is lost, and later regained, a new two week timer is started.

"The amendment can still become enabled later if an 80% majority of trusted validators support the amendment" is confusing. Doesn't say the amendment must _get_ and _hold_ that majority for two weeks.

mDuo13: The section above (not changed in this diff) explains the details in more specifics, but I agree this could be clearer.

content/concept-amendments.md
```diff
@@ -48,7 +48,9 @@ Theoretically, a `tfLostMajority` EnableAmendment pseudo-transaction could be in
...
+
+An amendment can gain and lose a majority an unlimited number of times before it becomes permanently enabled. As long as there are servers voting in favor of an amendment, the amendment is not permanently rejected. New versions of `rippled` can remove an amendment to ensure that they do not vote for it.
 
 As with all aspects of the consensus process, amendment votes are only taken into account by servers that trust the validators sending those votes. Ripple (the company) recommends only trusting the 5 default validators that Ripple (the company) operates. For now, trusting only those validators is enough to coordinate with Ripple (the company) on releasing new features.
```
nbougalis: This "recommendation" seems overly draconian. Should we attempt to soften it a bit? Something like:

> At this time, Ripple (the company) recommends only trusting the 5 default validators that it operates.

nbougalis:
Left some comments. None are _strictly_ needed, but do consider them. :+1: from me.

content/concept-amendments.md
ryangyoung: The previous paragraph mentions "operators of `rippled`" and this one mentions "The operator of a `rippled`". I think being consistent with either singular or plural would make these points just a bit clearer.

content/concept-amendments.md
ryangyoung: I like the re-write, but how are the validators Ripple (the company) operates "default"? I think we could omit that term.

mDuo13: They're default because the default config file includes their IDs, so anyone running rippled will use those validators unless they change the config file. And since we don't recommend people change the defaults, that's how that works.

content/concept-amendments.md
```diff
@@ -48,7 +48,9 @@ Theoretically, a `tfLostMajority` EnableAmendment pseudo-transaction could be in
...
 
-The operator of a `rippled` validator can "veto" an amendment. In this case, that validator never sends a vote in favor of the amendment. If enough servers veto an amendment, that prevents it from reaching consistent 80% support, so the amendment does not apply.
+The operator of a `rippled` validator can "veto" an amendment. In this case, that validator never sends a vote in favor of the amendment. If enough servers veto an amendment, that prevents it from reaching consistent 80% support, so the amendment does not apply. The amendment can still become enabled later if an 80% majority of trusted validators support the amendment, either because validator operators changed their minds or new validators that support the amendment became trusted.
+
+An amendment can gain and lose a majority an unlimited number of times before it becomes permanently enabled. As long as there are servers voting in favor of an amendment, the amendment is not permanently rejected. New versions of `rippled` can remove an amendment to ensure that they do not vote for it.
```
ryangyoung: The first sentence is a little confusing. Perhaps something like: "There is no limit to the number of times an amendment can gain and lose a majority before the amendment becomes permanently enabled."

ryangyoung: > New versions of `rippled` can remove an amendment to ensure that they do not vote for it.

Suggest: "New versions of `rippled` can remove an amendment to ensure that no servers vote for the amendment"

content/concept-amendments.md
ryangyoung: Ah, then Nik's suggestion is great without reservations in that case.

content/concept-amendments.md
```diff
@@ -111,7 +113,8 @@ The following is a comprehensive list of all known amendments and their status o
 
 | Name                            | Introduced | Enabled |
```
ryangyoung: Should the column name be changed to "Status"  since not all amendments are going to be enabled?

content/concept-amendments.md
```diff
@@ -111,7 +113,8 @@ The following is a comprehensive list of all known amendments and their status o
...
 |---------------------------------|------------|---------|
-| [FlowV2](#flowv2)               | v0.32.1    | Expected 2016-08-24 |
+| [Flow](#flow)                   | TBD        | TBD |
+| [FlowV2](#flowv2)               | v0.32.1    | Vetoed |
```
ryangyoung: If my understanding is correct, "Vetoed" is always potentially just temporary. I know that's explained just above, but listing the status as "Vetoed" makes it seem slightly misleading. If I were just glancing at this, I'd assume (incorrectly) that the status was more permanent. Maybe link "Vetoed" to the section that explains how vetoing works (even though, yes, they're proximate)?

content/concept-amendments.md
```diff
@@ -138,16 +141,23 @@ A transaction remains in the queue until one of the following happens:
...
-## FlowV2 ##
+## Flow ##
 
 | Amendment ID | Status |
 |--------------|--------|
-| 5CC22CFF2864B020BD79E0E1F048F63EF3594F95E650E43B3F837EF1DF5F4B26 | Planned for voting, expected to be enabled 2016-08-24 |
+| 740352F2412A9909880C23A559FCECEDA3BE2126FED62FC7660D628A06927F11 | TBD |
 
-Replaces the payment processing engine with a more robust and efficient rewrite called the FlowV2 engine. The new version of the payment processing engine is intended to follow the same rules as the old one, but occasionally produces different results due to floating point rounding.
+Replaces the payment processing engine with a more robust and efficient rewrite called the Flow engine. The new version of the payment processing engine is intended to follow the same rules as the old one, but occasionally produces different results due to floating point rounding. This Amendment was created to replace the [FlowV2](#flowv2) amendment when a critical bug was found in the FlowV2 amendment.
```
ryangyoung: I'd vote for breaking the second sentence into two sentences. Something like: "The new version of the payment processing engine is intended to follow the same rules as the old engine. However, due to floating point rounding, the new engine occasionally produces different results"

As is, it sounds like the exception was somehow part of the design.

content/concept-amendments.md
ryangyoung: @mDuo13 This goes to your point in [DOC-525](https://ripplelabs.atlassian.net/browse/DOC-525). I wonder if there's a more accurate term we could use that doesn't connote finality in the same way, i.e. suspended or deferred.

content/concept-amendments.md
mDuo13: The problem here is that servers can still vote for the amendment, especially servers haven't been updated to the new version. (But also, you can explicitly configure a server to vote in favor of an amendment it doesn't know.) A minority of updated validators can be overruled if an 80% majority still supports the amendment, so it's definitely misleading to say that it "ensures that no servers vote for the amendment."

I'm still trying to figure out a better wording.

content/concept-amendments.md
```diff
@@ -46,13 +46,11 @@ Theoretically, a `tfLostMajority` EnableAmendment pseudo-transaction could be in
...
 
-An amendment can gain and lose a majority an unlimited number of times before it becomes permanently enabled. As long as there are servers voting in favor of an amendment, the amendment is not permanently rejected. New versions of `rippled` can remove an amendment to ensure that they do not vote for it.
-
-As with all aspects of the consensus process, amendment votes are only taken into account by servers that trust the validators sending those votes. Ripple (the company) recommends only trusting the 5 default validators that Ripple (the company) operates. For now, trusting only those validators is enough to coordinate with Ripple (the company) on releasing new features.
+As with all aspects of the consensus process, amendment votes are only taken into account by servers that trust the validators sending those votes. At this time, Ripple (the company) recommends only trusting the 5 default validators that Ripple (the company) operates. For now, trusting only those validators is enough to coordinate with Ripple (the company) on releasing new features.
```
ryangyoung, line 53: "5" --> "five"

content/concept-amendments.md
```diff
@@ -147,7 +145,7 @@ A transaction remains in the queue until one of the following happens:
 |--------------|--------|
 | 740352F2412A9909880C23A559FCECEDA3BE2126FED62FC7660D628A06927F11 | TBD |
 
-Replaces the payment processing engine with a more robust and efficient rewrite called the Flow engine. The new version of the payment processing engine is intended to follow the same rules as the old one, but occasionally produces different results due to floating point rounding. This Amendment was created to replace the [FlowV2](#flowv2) amendment when a critical bug was found in the FlowV2 amendment.
+Replaces the payment processing engine with a more robust and efficient rewrite called the Flow engine. The new version of the payment processing engine is intended to follow the same rules as the old one, but occasionally produces different results due to floating point rounding. This Amendment supersedes the [FlowV2](#flowv2) amendment.
```
ryangyoung, line 148: I'd still vote for breaking up the second sentence. But I realize now that that might require more discussion of floating point rounding than is feasible at the moment. We should probably document that somewhere since it has caused some confusion, but I think this is OK for now.

ryangyoung:
Aside from my very minor comments, LGTM.

PR #213, opened 2016-09-29
amendments - 0.33 release updates

mDuo13:
updates the Amendments page now that that SusPay, PayChan, etc. have graduated from being "in development" and adds the expected enable-date for Flow.


ryangyoung:
This looks fine to me. The only thing that struck me is that there are a lot of different statuses. I know that they're inherently complicated, but I wonder if there's a way to standardize them so we can clearly define each one beyond the note at the end of the known amendments table. 

But that shouldn't block merging this PR. 👍  +2

PR #219, opened 2016-10-18
rippled - table format cleanup

mDuo13:
matches the rippled table syntax to more recent conventions (`codefont` left column, aligned column widths via table-formatter atom plugin)


mDuo13:
build failure is a benign deadlink (w3schools currently 503ing)

PR #220, opened 2016-10-19
rippled 0.33 - load_factor changes

mDuo13:
Changes also include the `closed_ledger` field which is returned in some circumstances. That field has been around since before 0.33.0 but was previously undocumented.


ryangyoung:
LGTM 👍

content/reference-rippled.md
```diff
@@ -8488,36 +8488,40 @@ The `info` object may have some arrangement of the following fields:
...
+| `load_factor`                       | Number                    | The load factor the server is currently enforcing, as a multiplier on the base transaction cost. For example, at `1000` load factor and a reference transaction cost of 10 drops of XRP, the load-scaled transaction cost is 10,000 drops (0.01 XRP). The load factor is determined by the highest of the [individual server's load factor](concept-transaction-cost.html#local-load-cost), the cluster's load factor, the [open ledger cost](concept-transaction-cost.html#open-ledger-cost) and the overall network's load factor. |
```
ximinez: Should this include `[Updated in: rippled 0.33.0][]`?

content/reference-rippled.md
```diff
@@ -8778,15 +8782,17 @@ The `state` object may have some arrangement of the following fields:
...
-| `load_factor`                    | Number           | The load factor the server is currently enforcing. The ratio between this value and the `load_base` determines the multiplier for transaction costs. The load factor is determined by the highest of the individual server's load factor, cluster's load factor, and the overall network's load factor. |
+| `load_factor`                    | Number           | The load factor the server is currently enforcing. The ratio between this value and the `load_base` determines the multiplier for transaction costs. The load factor is determined by the highest of the individual server's load factor, cluster's load factor, the [open ledger cost](concept-transaction-cost.html#open-ledger-cost), and the overall network's load factor. |
```
ximinez: Should this include `[Updated in: rippled 0.33.0][]`?

ximinez (changes requested):
Generally looks good. I'm unsure about the need for change notation...

content/reference-rippled.md
ximinez: On further thought, would it be worth including that this is the "load-scaled open ledger transaction cost" or something like that?  `load_factor_server` could then note that factor is sufficient to get into the transaction queue.  The link to "open ledger cost" is probably sufficient, but I'm not sure.

PR #221, opened 2016-10-19
rippled 0.33.0 - fee averaging explanation

mDuo13:

PR #225, opened 2016-10-26
fee averaging - edits per peer review

mDuo13:
incorporates some of the feedback from #221 from after it got merged


ryangyoung (approved):
LGTM 👍

PR #229, opened 2016-12-03
PayChan - transaction types 1st draft

mDuo13:
This is the first in a batch of documentation updates describing new functionality to be introduced by the PayChan amendment. This one covers the three new transaction types. 

Also on the radar but not included yet are a conceptual summary of payment channels, a tutorial for how to use them, an update to the ledger format page with the new payment channel ledger node type, and an update to the rippled API reference for the new payment channel commands. (Cross-references between the sections will be added as they become available.)


content/reference-transaction-format.md
```diff
@@ -724,6 +725,125 @@ In the above example with a ¥95/$15 offer and a ¥5/$2 offer, the situation is
 
 
 
+## PaymentChannelClaim
+[[Source]<br>](https://github.com/ripple/rippled/blob/develop/src/ripple/app/tx/impl/PayChan.cpp "Source")
+
+_Requires the [PayChan Amendment](concept-amendments.html#paychan)._
+
+Claim XRP from a payment channel, adjust its expiration, or both. This transaction can be used differently depending on the transaction sender's role in the specified channel:
```
ryangyoung: The antecedent of "Its" is slightly ambiguous here (XRP or the payment channel?). Based on context, most readers will probably, correctly, assume that it refers to "payment channel" (esp. if they know that XRP doesn't have an expiration). Still, might be worth rewording to prevent confusion, i.e. 

> Claim XRP from a payment channel, adjust the payment channel's expiration, or both.

content/reference-transaction-format.md
```diff
@@ -724,6 +725,125 @@ In the above example with a ¥95/$15 offer and a ¥5/$2 offer, the situation is
...
+
+The **source address** of a channel can:
+
+- Send XRP from the channel to the destination with _or without_ a signed Claim.
+- Set the channel to expire as soon as the channel's `SettleDelay` has passed.
+- Clear a pending `Expiration` time.
+- Close a channel immediately if it has no XRP remaining in it.
```
ryangyoung: What does "immediately" mean here? Right after the XRP balance is depleted?

content/reference-transaction-format.md
```diff
@@ -724,6 +725,125 @@ In the above example with a ¥95/$15 offer and a ¥5/$2 offer, the situation is
...
+  "PublicKey": "32D2471DB72B27E3310F355BB33E339BF26F8392D5A93D3BC0FC3B566612DA0F0A"
+}
+```
+
+<!--{# TODO: replace the above example with one where the channel, pubkey, signature, and balance match #}-->
+
+| Field       | JSON Type | [Internal Type][] | Description                    |
+|:------------|:----------|:------------------|:-------------------------------|
+| `Channel`   | String    | Hash256           | The unique ID of the channel, as a 64-character hexadecimal string. |
+| `Balance`   | String    | Amount            | _(Optional)_ Total amount of XRP, in drops, delivered by this channel after processing this claim. Required in order to deliver XRP. Must be more than the total amount delivered by the channel so far, but not greater than the `Amount` of the signed claim. Must be provided except when closing the channel. |
```
ryangyoung: super nitpick: "in order to" --> "to"

content/reference-transaction-format.md
```diff
@@ -724,6 +725,125 @@ In the above example with a ¥95/$15 offer and a ¥5/$2 offer, the situation is
...
+
+### PaymentChannelClaim Flags
+
+Transactions of the PaymentChannelClaim type support additional values in the [`Flags` field](#flags), as follows:
+
+| Flag Name | Hex Value  | Decimal Value | Description                         |
+|:----------|:-----------|:--------------|:------------------------------------|
+| `tfRenew` | 0x00010000 | 65536         | Clear the channel's `Expiration` time. (`Expiration` is different from the channel's immutable `CancelAfter` time.) Only the source address of the payment channel can use this flag. |
+| `tfClose` | 0x00020000 | 131072        | Request to close the channel. Only the channel source and destination addresses can use this flag. This flag closes the channel immediately if it has no more XRP allocated to it after processing the current claim, or if the destination address uses it. If the source address uses this flag when the channel still holds XRP, this schedules the channel to close after `SettleDelay` seconds have passed. (Specifically, this sets the sets the `Expiration` of the channel to the close time of the previous ledger plus the channel's `SettleDelay` time, unless the channel already has a sooner `Expiration` time.) |
```
ryangyoung: typo: "sets the sets the"

ryangyoung: This might be covered elsewhere, but what happens to the XRP if the channel still holds XRP when it closes?

Ah, I see:

> After the `Expiration` time, the channel closes automatically and returns any unspent XRP to the sender. 

Might be worth adding that explanation here as well.

ryangyoung: > unless the channel already has a sooner `Expiration` time.

Maybe "earlier `Expiration` time"?

content/reference-transaction-format.md
```diff
@@ -724,6 +725,125 @@ In the above example with a ¥95/$15 offer and a ¥5/$2 offer, the situation is
...
+| `Destination`    | String    | AccountID         | Address to receive XRP claims against this channel. This is also known as the "destination address" for the channel. |
+| `SettleDelay`    | Number    | UInt32            | Amount of time the source address must wait before closing the channel if it has unclaimed XRP. |
+| `PublicKey`      | String    | PubKey            | The public key of the key pair the source will use to sign claims against this channel, in hexadecimal. This can be any secp256k1 or Ed25519 public key. |
+| `CancelAfter`    | Number    | UInt32            | The time, in [seconds since the Ripple Epoch](reference-rippled.html#specifying-time), when this channel expires. Any transaction that would modify the channel after this time closes the channel without otherwise affecting it. This value is immutable; the channel can be closed sooner than this time but not later. |
```
ryangyoung: > the channel can be closed sooner than this time but not later. 

Maybe "before this time but not after"?

content/reference-transaction-format.md
```diff
@@ -724,6 +725,125 @@ In the above example with a ¥95/$15 offer and a ¥5/$2 offer, the situation is
...
+    "Amount": "200000",
+    "Expiration": 543171558
+}
+```
+
+| Field        | JSON Type | [Internal Type][] | Description                   |
+|:-------------|:----------|:------------------|:------------------------------|
+| `Channel`    | String    | Hash256           | The unique ID of the channel to fund, as a 64-character hexadecimal string. |
+| `Amount`     | String    | Amount            | How many [drops of XRP][Currency Amount] to add to the channel. To set the expiration for a channel without adding more XRP, set this to `"0"`. |
+| `Expiration` | Number    | UInt32            | _(Optional)_ New `Expiration` time to set for the channel, in seconds since the Ripple Epoch. This cannot be sooner than the close time of the previous ledger plus the `SettleDelay` of the channel. After the `Expiration` time, the channel closes automatically and returns any unspent XRP to the sender. (`Expiration` is separate from the channel's immutable `CancelAfter` time.) |
```
ryangyoung: Again, I think "earlier" or "before" would work better than "sooner."

ryangyoung (approved):
Looks good! A little hard to grok without the other set of docs, but even so, the text is clear enough to give me a fairly good idea of how the payment channels work. A few comments about minor style things, but that's all.

content/reference-transaction-format.md
mDuo13: It means at the time the transaction is processed. (Normally you have to wait out a delay.)

How about this?

> Close a channel immediately. You can only do this if it has no XRP remaining in it.

content/reference-transaction-format.md
```diff
@@ -724,6 +725,125 @@ In the above example with a ¥95/$15 offer and a ¥5/$2 offer, the situation is
...
+    "Amount": "200000",
+    "Expiration": 543171558
+}
+```
+
+| Field        | JSON Type | [Internal Type][] | Description                   |
+|:-------------|:----------|:------------------|:------------------------------|
+| `Channel`    | String    | Hash256           | The unique ID of the channel to fund, as a 64-character hexadecimal string. |
+| `Amount`     | String    | Amount            | How many [drops of XRP][Currency Amount] to add to the channel. To set the expiration for a channel without adding more XRP, set this to `"0"`. |
+| `Expiration` | Number    | UInt32            | _(Optional)_ New `Expiration` time to set for the channel, in seconds since the Ripple Epoch. This cannot be earlier than the close time of the previous ledger plus the `SettleDelay` of the channel. After the `Expiration` time, any transaction that would access the channel closes the channel without taking its normal action. Any unspent XRP is returned to the source address when the channel closes. (`Expiration` is separate from the channel's immutable `CancelAfter` time.) |
```
seelabs: `Expiration` must be later than _either_ the current expiration time (if any), _or_ the previous ledger close time plus the settle delay. So if the current expiration is an hour from now, and the settle delay is five hours, it's OK to set a expiration of two hours from now, even though that is less than the close time + settle delay.

content/reference-transaction-format.md
```diff
@@ -724,6 +725,125 @@ In the above example with a ¥95/$15 offer and a ¥5/$2 offer, the situation is
...
+}
+```
+
+<!--{# TODO: replace the above example with one where the channel, pubkey, signature, and balance match #}-->
+
+| Field       | JSON Type | [Internal Type][] | Description                    |
+|:------------|:----------|:------------------|:-------------------------------|
+| `Channel`   | String    | Hash256           | The unique ID of the channel, as a 64-character hexadecimal string. |
+| `Balance`   | String    | Amount            | _(Optional)_ Total amount of XRP, in drops, delivered by this channel after processing this claim. Required to deliver XRP. Must be more than the total amount delivered by the channel so far, but not greater than the `Amount` of the signed claim. Must be provided except when closing the channel. |
+| `Amount`    | String    | Amount            | _(Optional)_ The amount of XRP, in drops, authorized by the `Signature`. This must match the amount in the signed message.  |
```
seelabs: I suspect `Amount` will be a source of confusion. `Amount` is the value that a channel's balance will be allowed to be adjusted to. For example, to authorize an initial 10 drops, the amount will be 10. To authorize an additional 10 drops, the amount will be 20.

mDuo13:
The build failures are from a broken link not touched by this build. (Fixed in master.)

PR #247, opened 2017-04-27
Accounts concept

mDuo13:
Adds a conceptual page explaining "Accounts" in the RCL in general, plus some technical details (from [Encodings](https://wiki.ripple.com/Encodings)) to explain how addresses are derived. Replaces the [Accounts wiki page](https://wiki.ripple.com/Accounts).

The link checker is expected to fail this PR because the link to the JavaScript source file for address derivation won't exist until the PR is merged.

This PR also needs to be coordinated with some prep work on the Ripple.com side to prepare the new page.


reference-rippled.html
```diff
@@ -627,53 +628,10 @@ <h3 id="addresses">Addresses</h3>
 <li>Between 25 and 35 characters in length</li>
 <li>Starts with the character <code>r</code></li>
 <li>Case-sensitive</li>
-<li><a href="https://wiki.ripple.com/Encodings">Base-58</a> encoded using only the following characters: <code>rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz</code> That's alphanumeric characters, excluding zero (<code>0</code>), capital O (<code>O</code>), capital I (<code>I</code>), and lowercase L (<code>l</code>).</li>
 <li>Contains error-checking that makes it unlikely that a randomly-generated string is a valid address.</li>
+<li><strong>Base58-encoded</strong>, using alphanumeric characters, excluding zero (<code>0</code>), capital O (<code>O</code>), capital I (<code>I</code>), and lowercase L (<code>l</code>).</li>
```
jbheron: Could the sentence be more readable with quotes around the letters?

> ... using alphanumeric characters, excluding zero, "`0`", capital "`O`", capital "`I`", and lowercase "`l`."

jbheron (approved):
LGTM  -- nice! (I made a single comment with suggestion, separately).

content/data_types/address.md
```diff
@@ -1,9 +1,9 @@
-Ripple Accounts are identified by a base-58 Ripple Address, which is derived from the account's master public key. An address is represented as a String in JSON, with the following characteristics:
+Ripple Accounts are identified by a base58 Ripple Address, which is derived from the account's master public key. An address is represented as a string in JSON, with the following characteristics:
```
jbheron: Nice catch on base58 naming.

content/concept-accounts.md
```diff
@@ -0,0 +1,123 @@
+# RCL Accounts
+
+An "Account" in the Ripple Consensus Ledger represents a holder of XRP and a sender of transactions. The core elements of an account are:
+
+- An identifying **address**, such as `rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn`
+- An account can send [transactions](reference-transaction-format.html) to modify the Ripple Consensus Ledger, for a variety of purposes including:
```
ryangyoung: This phrasing isn't quite parallel with the introductory sentence: "The core elements of an account are: ... An account can send transactions..." All of the sub-bullets seem to be about what an account can do, so maybe it would make sense to move this to its own section?

content/concept-accounts.md
```diff
@@ -0,0 +1,123 @@
...
+    - Trading in the decentralized exchange.
+    - Changing settings such as authorization settings, balance limits, or the behavior of assets issued by the account.
+- An **XRP balance**. Some of this XRP is set aside for the [Reserve](concept-reserves.html).
+- A **sequence number**, starting at 1 and increasing with each transaction sent from this account. No transaction can be included in a ledger unless the transaction's sequence number matches its sender's next sequence number.
+- A **history of transactions** that affected this account and its balances.
+
+In the ledger's data tree, an account's core data is stored in the [AccountRoot](reference-ledger-format.html#accountroot) ledger node type. An account can also be the owner (or partial owner) of several other types of data.
+
+**Tip:** An "Account" in the Ripple Consensus Ledger is somewhere between the financial usage (like "bank account") and the computing usage (like "UNIX account"). Non-XRP currencies and assets aren't stored in an RCL Account itself; each such asset is stored in an accounting relationship called a "Trust Line" that connects two parties.
```
ryangyoung: > an accounting relationship called a "Trust Line" that connects two parties.

I was going to say add a link to the trust line section, but I don't think that exists yet. Something to remember for when we do create that!

content/concept-accounts.md
```diff
@@ -0,0 +1,123 @@
...
+
+
+## Addresses
+
+{% include 'data_types/address.md' %}
+
+The conversion from a public key to an address involves a one-way hash function, so it is possible to confirm that a public key matches an address but it is impossible to derive the public key from the address alone. (This is part of the reason why signed transactions include the public key _and_ the address of the sender.) For more technical details, see [Address Encoding](#address-encoding).
+
+The process of "creating" an address is the purely mathematical task of generating a key pair and calculating the address from it, so an address can exist separately from its presence as an account object in the ledger. To actually make an account for an address, someone must send the address enough XRP to satisfy the reserve. This is called _funding_ the account. Whether or not the address is a funded account, you can use the address (and the key pair it's derived from) as a [regular key](reference-transaction-format.html#setregularkey) or as part of a [signer list](reference-transaction-format.html#multi-signing).
```
ryangyoung: > The process of "creating" an address is the purely mathematical task of generating a key pair and calculating the address from it, so an address can exist separately from its presence as an account object in the ledger.

Two small nitpicks:

1. It might just be me, but "the purely mathematical task..." sounds ever so slightly condescending, similar to saying "Simply do this." (I guess it's not the same thing since it is just stating a fact, but adverbs always make me a little nervous.)
2. That's a longish sentence. The second clause, after the comma, should probably stand on its own.

ryangyoung: Right now, the whole paragraph is bit confusing. The first sentence talks about creating an address. The second sentence talks about "actually" making an account via funding, which I understand is closely related, but it still seems to call for its own section, or at least paragraph. Maybe the concepts might be too intertwined to do that, though.

content/concept-accounts.md
```diff
@@ -0,0 +1,123 @@
...
+| rrrrrrrrrrrrrrrrrrrn5RM1rHd | NaN Address | Previous versions of [ripple-lib](https://github.com/ripple/ripple-lib) generated this address when base58 encoding the value [NaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN). | Yes |
+
+
+## Permanence of Accounts
+
+Once created, an account exists in the Ripple Consensus Ledger's data tree forever. This is because the current sequence number for a transaction must be tracked forever, so that old transactions cannot be processed a second time.
+
+Unlike Bitcoin and many other crypto-currencies, each new version of the Ripple Consensus Ledger's public ledger chain contains the full state of the ledger, which increases in size with each new account. For that reason, Ripple discourages creating new accounts unless entirely necessary. Institutions who send and receive value on behalf of many users can use **Source Tags** and **Destination Tags** to distinguish payments from and to their customers while only using one (or a handful) of accounts in the Ripple Consensus Ledger.
```
ryangyoung: For the source and destination tags, add a link to: https://ripple.com/build/gateway-guide/#source-and-destination-tags

content/concept-accounts.md
```diff
@@ -0,0 +1,123 @@
...
+
+
+## Transaction History
+
+In the Ripple Consensus Ledger, transaction history is tracked by a "thread" of transactions linked by a transaction's identifying hash and the ledger index. The `AccountRoot` ledger node has hash and ledger of the transaction that most recently modified it; the metadata of that transaction includes the previous state of the `AccountRoot` node, so it is possible to iterate through the history of a single account this way. This transaction history includes any transactions that modify the `AccountRoot` node directly, including:
```
ryangyoung: > The `AccountRoot` ledger node has hash and ledger

typo: "has **a** hash..."

content/concept-accounts.md
```diff
@@ -0,0 +1,123 @@
...
+
+- Transactions sent by the account, because the modify the account's `Sequence` number. These transactions also modify the account's XRP balance because of the [transaction cost](concept-transaction-cost.html).
```
ryangyoung: > because the modify

typo: "they"

ryangyoung (changes requested):
This does an excellent job of migrating and updating the old wiki content. I was even able to understand most of it!

I'm requesting changes because there are just a couple of places where some of the language could be tweaked a bit to be even clearer.

content/concept-accounts.md
mDuo13: You're right that it didn't fit in. Ultimately I think a better bit of information to capture in this section is the unit(s) of authentication the account contains, rather than the types of transactions it can send, so I've replaced this bullet point with the following:

> - One or more ways to [authorize transactions](reference-transaction-format.html#authorizing-transactions), possibly including:
>    - A master key pair intrinsic to the account. (This can be [disabled](reference-transaction-format.html#accountset-flags) but not changed.)
>    - A "regular" key pair that [can be rotated](reference-transaction-format.html#setregularkey).
>    - A signer list for [multi-signing](reference-transaction-format.html#multi-signing). (Stored separately from the account's core data.)

content/concept-accounts.md
mDuo13: I restructured this section a little to clear up some of the confusion. After agonizing for quite a while trying to come up with a wording that made more sense to me than "purely mathematical task," I threw in the towel and settled for "Creating a valid address is a strictly mathematical task starting with a key pair." My thought process is:

- The condescending feel may have been coming in part from some other pieces of the passage, such as the scare quotes around "creating" and the the word "actually" splitting the infinitive in the following sentence. By removing those, I hope the adverb is less likely to offend.
- I changed to "strictly" because the strictness sounds more academic than the ideological connotation of "purely."
- I felt it was still important to retain an adverb to rule out any other bits of non-math that people could unknowingly assume are involved in creating an address.

I also moved the more important bits about funding accounts and using unfunded accounts first. I tried to include a paragraph explaining that anyone can fund an account but only the holder of the secret key can control the account, but ultimately cut it in the name of simplicity. The resulting passage looks like this:

> Any valid address can become an account in the Ripple Consensus Ledger by receiving a [Payment][] of XRP, as long as the amount of XRP delivered is greater than or equal to the [account reserve](concept-reserves.html). This is called _funding_ the account. You can also use an address that has not been funded to represent a [regular key](reference-transaction-format.html#setregularkey) or a member of a [signer list](reference-transaction-format.html#multi-signing). Only a funded account can be the sender of a transaction.
>
> Creating a valid address is a strictly mathematical task starting with a key pair. You can generate a key pair and calculate its address entirely offline without communicating to the Ripple Consensus Ledger or any other party. The conversion from a public key to an address involves a one-way hash function, so it is possible to confirm that a public key matches an address but it is impossible to derive the public key from the address alone. (This is part of the reason why signed transactions include the public key _and_ the address of the sender.)
> 
> For more technical details of how to calculate a Ripple address, see [Address Encoding](#address-encoding).

ryangyoung: That's a terrific revision. Much clearer and it breaks up the address/account relationship nicely. I didn't even break a sweat when I read "strictly."

ryangyoung (approved):

PR #248, opened 2017-05-02
EscrowFinish transaction cost

mDuo13:
The EscrowFinish transaction type has a higher than usual transaction cost if has a fulfillment. This update clarifies that.

Technically the formula is:

    base fee + 32 + (crypto-condition cost÷16) drops

However, since the base fee is fairly stable at 10 and the only currently-supported crypto-condition is PREIMAGE-SHA-256, whose cost is equal to the size of the preimage in bytes, I simplified the formula to:

    42 drops + (preimage size ÷ 16)

### DIFF
--- a/concept-transaction-cost.html
+++ b/concept-transaction-cost.html
@@ -195,7 +195,7 @@
 </tr>
 <tr>
 <td><a href="reference-transaction-format.html#escrowfinish">EscrowFinish Transaction with Fulfillment</a></td>
-<td>42 drops + (Fulfillment size in bytes divided by 16)</td>
+<td>10 drops + (32 drops × (Fulfillment size in bytes ÷ 16))</td>
 </tr>
 </tbody>
 </table>
@@ -281,9 +281,9 @@
 </tr>
 <tr>
 <td><a href="reference-transaction-format.html#escrowfinish">EscrowFinish transaction</a> with 32-byte preimage.</td>
-<td>44</td>
+<td>74</td>
 <td>245</td>
-<td>88</td>
+<td>148</td>
 <td>512</td>
 </tr>
 </tbody>
--- a/concept-transaction-cost.html
+++ b/concept-transaction-cost.html
@@ -195,7 +195,7 @@
 </tr>
 <tr>
 <td><a href="reference-transaction-format.html#escrowfinish">EscrowFinish Transaction with Fulfillment</a></td>
-<td>10 drops + (32 drops × (Fulfillment size in bytes ÷ 16))</td>
+<td>10 drops × (32 + (Fulfillment size in bytes ÷ 16))</td>
 </tr>
 </tbody>
 </table>
@@ -281,9 +281,9 @@
 </tr>
 <tr>
 <td><a href="reference-transaction-format.html#escrowfinish">EscrowFinish transaction</a> with 32-byte preimage.</td>
-<td>74</td>
+<td>340</td>
 <td>245</td>
-<td>148</td>
+<td>680</td>
 <td>512</td>
 </tr>
 </tbody>
--- a/concept-transaction-cost.html
+++ b/concept-transaction-cost.html
@@ -190,8 +190,12 @@
 <td>0</td>
 </tr>
 <tr>
-<td><a href="reference-transaction-format.html#multi-signing">Multi-signed transaction</a></td>
+<td><a href="reference-transaction-format.html#multi-signing">Multi-signed Transaction</a></td>
 <td>10 drops × (1 + Number of Signatures Provided)</td>
+</tr>
+<tr>
+<td><a href="reference-transaction-format.html#escrowfinish">EscrowFinish Transaction with Fulfillment</a></td>
+<td>42 drops + (Fulfillment size in bytes divided by 16)</td>
 </tr>
 </tbody>
 </table>
@@ -274,6 +278,13 @@
 <td>(Effectively infinite)</td>
 <td>N/A</td>
 <td>(Effectively infinite)</td>
+</tr>
+<tr>
+<td><a href="reference-transaction-format.html#escrowfinish">EscrowFinish transaction</a> with 32-byte preimage.</td>
+<td>44</td>
+<td>245</td>
+<td>88</td>
+<td>512</td>
 </tr>
 </tbody>
 </table>
--- a/concept-transaction-cost.html
+++ b/concept-transaction-cost.html
@@ -195,7 +195,7 @@
 </tr>
 <tr>
 <td><a href="reference-transaction-format.html#escrowfinish">EscrowFinish Transaction with Fulfillment</a></td>
-<td>10 drops × (32 + (Fulfillment size in bytes ÷ 16))</td>
+<td>10 drops × (33 + (Fulfillment size in bytes ÷ 16))</td>
 </tr>
 </tbody>
 </table>
@@ -281,9 +281,9 @@
 </tr>
 <tr>
 <td><a href="reference-transaction-format.html#escrowfinish">EscrowFinish transaction</a> with 32-byte preimage.</td>
-<td>340</td>
-<td>245</td>
-<td>680</td>
+<td>350</td>
+<td>256</td>
+<td>700</td>
 <td>512</td>
 </tr>
 </tbody>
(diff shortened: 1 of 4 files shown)


content/concept-transaction-cost.md
```diff
@@ -19,7 +19,8 @@ Some transactions have different transaction costs:
 |-----------------------|--------------------------|
 | [Reference Transaction](#reference-transaction-cost) (Most transactions) | 10 drops |
 | [Key Reset Transaction](#key-reset-transaction) | 0 |
-| [Multi-signed transaction](reference-transaction-format.html#multi-signing) | 10 drops × (1 + Number of Signatures Provided) |
+| [Multi-signed Transaction](reference-transaction-format.html#multi-signing) | 10 drops × (1 + Number of Signatures Provided) |
+| [EscrowFinish Transaction with Fulfillment](reference-transaction-format.html#escrowfinish) | 42 drops + (Fulfillment size in bytes divided by 16) |
```
jbheron: Mixing '+' and 'divided by' makes this formula a bit hard for me to read. Perhaps:

> | [EscrowFinish Transaction with Fulfillment](reference-transaction-format.html#escrowfinish) | 42 drops + (Fulfillment size (bytes) ÷ 16) |

wilsonianb:
The formula is 
```
base fee * (32 + (crypto-condition cost÷16)) drops
```

content/reference-transaction-format.md
```diff
@@ -581,6 +581,8 @@ Any account may submit an EscrowFinish transaction.
 * If the corresponding [EscrowCreate transaction][] specified a `FinishAfter` time that is after the close time of the most recently-closed ledger, the EscrowFinish transaction fails.
 * If the corresponding [EscrowCreate transaction][] specified a `CancelAfter` time that is before the close time of the most recently-closed ledger, the EscrowFinish transaction fails.
 
+**Note:** The [transaction cost](concept-transaction-cost.html) to submit an EscrowFinish transaction is higher if it contains a fulfillment. Instead of the usual base cost of 10 [drops of XRP](reference-rippled.html#specifying-currency-amounts), an EscrowFinish transaction costs 42 drops plus another 1 drop for every 16 bytes in size of the preimage. (For example, a fulfillment with a 32-byte preimage would cost 44 drops of XRP.)
```
jbheron: The superlative here suggests you will include a comparator, such as 'higher than...'. For clarity, how about:

> **Note:** The [transaction cost](concept-transaction-cost.html) to submit an EscrowFinish transaction increases if it contains a fulfillment.

reference-transaction-format.html
```diff
@@ -1028,6 +1028,7 @@ <h2 id="escrowfinish">EscrowFinish</h2>
 <li>If the corresponding <a href="#escrowcreate">EscrowCreate transaction</a> specified a <code>FinishAfter</code> time that is after the close time of the most recently-closed ledger, the EscrowFinish transaction fails.</li>
 <li>If the corresponding <a href="#escrowcreate">EscrowCreate transaction</a> specified a <code>CancelAfter</code> time that is before the close time of the most recently-closed ledger, the EscrowFinish transaction fails.</li>
 </ul>
+<p class="devportal-callout note"><strong>Note:</strong> The <a href="concept-transaction-cost.html">transaction cost</a> to submit an EscrowFinish transaction is higher if it contains a fulfillment. Instead of the usual base cost of 10 <a href="reference-rippled.html#specifying-currency-amounts">drops of XRP</a>, an EscrowFinish transaction costs 42 drops plus another 1 drop for every 16 bytes in size of the preimage. (For example, a fulfillment with a 32-byte preimage would cost 44 drops of XRP.)</p>
```
jbheron: Again, I'd lobby for 'increases' over 'is higher'.

jbheron (commented):
LGTM - I made three suggestions, which you may find useful.

reference-transaction-format.html
mDuo13: For reference, this is just generated from the same text as the other comment. (The HTML files are all sourced from the markdown, so no need to comment on the HTML versions.)

jbheron: Ah, derp, that makes sense. Thanks, Rome.

content/reference-transaction-format.md
```diff
@@ -581,7 +581,7 @@ Any account may submit an EscrowFinish transaction.
...
 * If the corresponding [EscrowCreate transaction][] specified a `CancelAfter` time that is before the close time of the most recently-closed ledger, the EscrowFinish transaction fails.
 
-**Note:** The [transaction cost](concept-transaction-cost.html) to submit an EscrowFinish transaction is higher if it contains a fulfillment. Instead of the usual base cost of 10 [drops of XRP](reference-rippled.html#specifying-currency-amounts), an EscrowFinish transaction costs 42 drops plus another 1 drop for every 16 bytes in size of the preimage. (For example, a fulfillment with a 32-byte preimage would cost 44 drops of XRP.)
+**Note:** The [transaction cost](concept-transaction-cost.html) to submit an EscrowFinish transaction increases if it contains a fulfillment. An EscrowFinish transaction has a base cost of 10 [drops of XRP](reference-rippled.html#specifying-currency-amounts), plus an additional 32 drops for every 16 bytes in size of the preimage. For example, a fulfillment with a 32-byte preimage would cost 10 + (32 drops × (32 bytes / 16)) = 74 drops of XRP.
```
jbheron: We're using "÷" and "/" to indicate division on this page. We should probably choose one that is most readable, standardized or familiar to our readers.

concept-transaction-cost.html
```diff
@@ -275,6 +279,13 @@ <h3 id="fee-levels">Fee Levels</h3>
 <td>N/A</td>
 <td>(Effectively infinite)</td>
 </tr>
+<tr>
+<td><a href="reference-transaction-format.html#escrowfinish">EscrowFinish transaction</a> with 32-byte preimage.</td>
+<td>340</td>
+<td>245</td>
```
wilsonianb: Should this be 256?

content/concept-transaction-cost.md
```diff
@@ -19,7 +19,8 @@ Some transactions have different transaction costs:
 |-----------------------|--------------------------|
 | [Reference Transaction](#reference-transaction-cost) (Most transactions) | 10 drops |
 | [Key Reset Transaction](#key-reset-transaction) | 0 |
-| [Multi-signed transaction](reference-transaction-format.html#multi-signing) | 10 drops × (1 + Number of Signatures Provided) |
+| [Multi-signed Transaction](reference-transaction-format.html#multi-signing) | 10 drops × (1 + Number of Signatures Provided) |
+| [EscrowFinish Transaction with Fulfillment](reference-transaction-format.html#escrowfinish) | 10 drops × (32 + (Fulfillment size in bytes ÷ 16)) |
```
wilsonianb: Sorry, now I was wrong.
What you have is the calculation for how much extra is added to the base fee.
So the total is actually:
```
10 drops + (10 drops × (32 + (Fulfillment size in bytes ÷ 16)))
```
the second 10 drops is the `fee_ref`/`TRANSACTION_FEE_BASE`
https://github.com/ripple/rippled/blob/develop/src/ripple/core/Config.h#L135

content/concept-transaction-cost.md
```diff
@@ -101,6 +102,7 @@ _Fee levels_ represent the proportional difference between the minimum cost and
 | Reference transaction (most transactions) | 10 | 256 | 20 | 512 |
 | [Multi-signed transaction](reference-transaction-format.html#multi-signing) with 4 signatures | 50 | 256 | 100 | 512 |
 | [Key reset transaction](concept-transaction-cost.html#key-reset-transaction) | 0 | (Effectively infinite) | N/A | (Effectively infinite) |
+| [EscrowFinish transaction](reference-transaction-format.html#escrowfinish) with 32-byte preimage. | 340 | 245 | 680 | 512 |
```
wilsonianb: 340 -> 350
680 -> 360

concept-transaction-cost.html
mDuo13: Yes. Good catch.

content/concept-transaction-cost.md
mDuo13: Wait, at 512 fee units the tx cost goes to 360 drops? Not 700 drops? How does _that_ formula work?

wilsonianb (approved):
LGTM :+1:

PR #249, opened 2017-05-03
Payment Channels Tutorial

mDuo13:
A thorough walkthrough of payment channels with examples of API methods and transactions.

(The important text is in [tutorial-paychan.md](https://github.com/ripple/ripple-dev-portal/pull/249/files#diff-075115445d6de595743a0e599bca55e1)—the .html files are all auto-generated.)

EDIT: Removed the in-progress diagram. I may add a diagram later, but not for this version.


content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
+# Payment Channels Tutorial
+
+Payment Channels are an advanced feature for sending "asynchronous" XRP payments that can be divided into very small increments and settled later. This tutorial walks through the entire process of using a payment channel, with examples using the [JSON-RPC API](reference-rippled.html) of a local `rippled` server.
```
ryangyoung, line 3: Add a link to https://ripple.com/build/amendments/#paychan

(And we should update the link if/when we come up with another conceptual article that replaces the amendment description.)

ryangyoung: Maybe add some info about prerequisites? What do I need to start this tutorial?

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+
+1. **The payer creates a payment channel to a particular recipient.** In doing so, the payer pre-funds the channel with XRP.
```
ryangyoung: Maybe inline here or in the prerequisite section, explain who will be involved (i.e., payer and payee) and their roles.

ryangyoung: Who is the audience for this tutorial? Right now it's a little weird because it seems to be describing the process more than actually telling you (the payer? payee? someone else?) how to do these things.

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+
+    This is a [PaymentChannelCreate transaction][]. As part of this process, the payer sets certain specifics of the channel like an expiration time and a settlement delay, which affect the guarantees around the claims in the channel. The payer also sets the public key that will be used to verify claims against the channel.
```
ryangyoung: This paragraph is conceptual information, which gets back to the question of who the audience is. If it's a strict "do this, then that" sort of tutorial, this would be better-suited to some other place in the docs (like that yet-to-be written conceptual article maybe). Since there isn't a better place for this currently, it's probably ok to leave it here for now. But it is something to think about.

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+
+    **Tip:** The "settlement delay" does not delay the settlement, which can happen as fast as a ledger version closes (3-5 seconds). The "settlement delay" is a forced delay on closing the channel so that the payee has a chance to finish with settlement.
```
ryangyoung: Should this maybe be part of the description below? Or moved to after that description? It feels a little odd to clarify something that hasn't been introduced yet.

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+
+    The following example shows creation of a payment channel by submitting to a local `rippled` server with the JSON-RPC API. The payment channel allocates 100 XRP from rN7n7... to rf1Bi... with a settlement delay of 1 day. The public key happens to be rN7n7...fzRH's master public key.
```
ryangyoung: > from rN7n7... to rf1Bi...

Maybe "from account rN7n7... to account rf1Bi..." with a link to that new accounts page?

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+                ...
+                "tx_json": {
+                    ...
+                    "TransactionType": "PaymentChannelCreate",
+                    "hash": "3F93C482C0BC2A1387D9E67DF60BECBB76CC2160AE98522C77AF0074D548F67D"
+                }
+            }
+        }
+
+    The payer should check the transaction's final result in a validated ledger and get the Channel ID from the metadata. This can be done with the `tx` command:
```
ryangyoung: It seems like this should be its own step or sub-step? Also, is the payer the intended reader here?

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+                    "TransactionResult": "tesSUCCESS"
+                },
+                "status": "success",
+                "validated": true
+            }
+        }
+
+    In the response from the JSON-RPC, the payer should look for the following:
+
+    - Confirm that the response has `"validated":true` to indicate the data comes from a validated ledger, so the transaction's result is final.
```
ryangyoung: Maybe break this up into two sentences? The second clause seems well worth explaining a little more, and perhaps linking out to https://ripple.com/build/transactions/#finality-of-results

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+    - In the `AffectedNodes` of the transaction's `meta` field, look for a `CreatedNode` with the `LedgerEntryType` of `PayChannel`. The `LedgerIndex` field indicates the Channel ID. (In the above example, this is a hex string starting with "5DB0...") The Channel ID is necessary later to sign claims.
```
ryangyoung: > In the `AffectedNodes` 

"In the `AffectedNode` array"?

>  look for a `CreatedNode` with 

"look for a `CreatedNode` object"?

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+
+2. **The payee confirms that the specifics of the payment channel are satisfactory for the situation.**
+
+    In particular, the payee should confirm that the settlement delay is long enough that the channel cannot close before the payee has a chance to redeem any outstanding claims against the channel.
```
ryangyoung: Maybe the solution to audience issue is to preface each step (or task?) with a "Payer" or "Payee" to make it totally clear who the intended audience is (who needs to do the thing that's described).

ryangyoung: This is another paragraph that feels more like a conceptual description than the kind of imperative I'd expect from a tutorial. Not sure how best to address that.

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+            "result": {
+                "signature": "304402204EF0AFB78AC23ED1C472E74F4299C0C21F1B21D07EFC0A3838A420F76D783A400220154FB11B6F54320666E4C36CA7F686C16A3A0456800BBC43746F34AF50290064",
+                "status": "success"
+            }
+        }
+
+
+4. **The payer transmits a claim to the payee as payment for goods or services.**
+
+    This communication happens "off-ledger" in any communication system the payer and payee can agree to. It's recommended that the two parties use secure communications for this, but it's not strictly necessary. Only the payer or payee of a channel can redeem claims against that channel.
```
ryangyoung: > It's recommended that the two parties use secure communications for this, but it's not strictly necessary.

To make this active, maybe: "The two parties should use secure communications, but that is not technically necessary."

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+    |:------------------------|:---------------------------------------------------|
+    | Channel ID              | `5DB01B7FFED6B67E6B0414DED11E051D2EE2B7619CE0EAA6286D67A3A4D5BDB3` |
+    | Amount of XRP, in drops | `1000000`                                          |
+    | Signature               | `304402204EF0AFB78AC23ED1C472E74F4299C0C21F1B21D07EFC0A3838A420F76D783A400220154FB11B6F54320666E4C36CA7F686C16A3A0456800BBC43746F34AF50290064` |
+
+    The payee also needs to know the Public Key associated with the channel, which is the same throughout the channel's life.
+
+5. **The payee verifies the claims and provides the goods or services.**
+
+    You can verify claims using the `channel_verify` API method. The payee should confirm that the amount of the claim is equal or greater than the total price of goods and services provided. (Since the amount is cumulative, this is the total price of all goods and services purchased so far.)
```
ryangyoung: > the claim is equal or greater than the total price of goods and services provided.

"equal to or greater..."

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+                    "Account": "rN7n7otQDd6FczFgLdSqtcsAUxDkw6fzRH",
+                    "TransactionType": "PaymentChannelClaim",
+                    "Channel": "5DB01B7FFED6B67E6B0414DED11E051D2EE2B7619CE0EAA6286D67A3A4D5BDB3",
+                    "Flags": 2147614720
+                },
+                "fee_mult_max": 1000
+            }]
+        }
+
+    After the transaction is included in a validated ledger, either party can look up the currently-scheduled expiration of the channel using the `account_channels` method. Be sure to specify `"ledger_index": "validated"` to get data from the latest validated ledger version.
```
ryangyoung: This also seems like it should be its own step or sub-step.

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+                        "settle_delay": 86400
+                    }
+                ],
+                "status": "success"
+            }
+        }
+
+    In this example, the `expiration` value 547073182 in [seconds since the Ripple Epoch](reference-rippled.html#specifying-time) maps to 2017-05-02T20:46:22Z, so any claims not redeemed by that time are no longer valid.
+
+9. **After the settlement delay has passed or the channel has reached its planned expiration time, the channel is expired.** Any further transaction that would affect the channel can only close it, returning unclaimed XRP to the payer.
```
ryangyoung: To be consistent with the other steps, the non-bold text should start on its own line.

ryangyoung: Also, that description doesn't sound like a step per-se. If we're thinking in tutorial terms, there's nothing here for the reader to do, strictly speaking.

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+
+    The channel can remain on the ledger in an expired state indefinitely. This is because the ledger cannot change except as the results of a transaction, so _someone_ must send a transaction to cause the expired channel to close.
```
ryangyoung: Are there any consequences to letting an expired channel linger around?

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+
+    We recommend the payer sends a second [PaymentChannelClaim transaction][] with the `tfClose` flag for this purpose. However, other accounts, even those not involved in the payment channel, can cause an expired channel to close.
```
ryangyoung: Who is "we"?

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+                    "Account": "rN7n7otQDd6FczFgLdSqtcsAUxDkw6fzRH",
+                    "TransactionType": "PaymentChannelClaim",
+                    "Channel": "5DB01B7FFED6B67E6B0414DED11E051D2EE2B7619CE0EAA6286D67A3A4D5BDB3",
+                    "Flags": 2147614720
+                },
+                "fee_mult_max": 1000
+            }]
+        }
+
+    When the transaction has been included in a validated ledger, you can look at the metadata of the transaction to confirm that it deleted the channel and returned the XRP to the sender.
```
ryangyoung: Again, this seems like a separate (sub-)step.

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+        }
+
+    In the transaction's metadata, look for the following:
+
+    - A `DeletedNode` entry with `"LedgerEntryType": "PayChannel"`. The `LedgerIndex` field should match the Channel ID. This indicates that the channel was deleted.
+    - A `ModifiedNode` entry with `"LedgerEntryType": "AccountRoot"`. The change in the `Balance` field in `PreviousFields` and `FinalFields` reflects the unspent XRP being returned to the payer.
+
+    Those fields indicate that the payment channel is closed.
+
+This concludes the tutorial of Payment Channel usage. Ripple encourages users to find unique and interesting use cases to take full advantage of the speed and convenience of payment channels.
```
ryangyoung: This might be asking too much, but it would probably help to frame this tutorial with one example use case. Alice needs to pay Bob for some custom matryoshka dolls? I dunno. Some concrete grounding would help, especially with the last couple of steps.

ryangyoung (changes requested):
This is a good first draft that explains a somewhat abstruse feature clearly. My main critique is that it feels more like a description of the process rather than a strict tutorial. I suspect this is partly due to the nature of the feature, which inherently requires two different people to participate. I've made some suggestions about how this might be worked around (not sure if we want to recommend people to create a "dummy" account or channel). Probably the most helpful thing would be to create a standalone conceptual article that explains the process; then this would be a nice supplement.

content/tutorial-paychan.md
```diff
@@ -0,0 +1,489 @@
...
+        200 OK
+
+        {
+            "result": {
+                "signature_verified":true,
+                "status":"success"
+            }
+        }
+
+    If the response shows `"signature_verified": true` then the claim is genuine. At this point, the payee has not yet received the XRP, but can be certain that he or she _can_ redeem the XRP as long as the transaction to do so is processed before the channel expires. ***Rome's question: Is there anything else the payee should check, or any tricks the payer could use to make a "valid" signature for a balance that couldn't actually be redeemed?***
```
seelabs: `channel_verify` shows the claim is genuine, as you state. However note it does _not_ verify the channel has enough funds available to honor the claim. The payee may want to check that the channel has enough funds to honor the claim.

seelabs (approved):

content/tutorial-paychan.md
seelabs: @ryangyoung Payment channels are part of the payer's owner count, which effects the payer's reserve: https://ripple.com/build/reserves/

content/tutorial-paychan.md
mDuo13: Respectfully, I disagree. This is _contextual_ information that reminds users what they are doing in this step.

Technically, you are correct that some clauses, like "which affect the guarantees around the claims in the channel," are conceptual information. However, short phrases connect the conceptual information to the steps of the tutorial, which I think is the point of an educational tutorial like this.

content/tutorial-paychan.md
mDuo13: But "settlement delay" is first introduced in the paragraph right before this!

content/tutorial-paychan.md
mDuo13: I'd rather not make every "submit transaction; now check the results" into multiple sub-steps. That's better covered in detail by the Reliable Transaction Submission document.

content/tutorial-paychan.md
mDuo13: I made a new "Example Values" section with this and linked that from here.

content/tutorial-paychan.md
mDuo13: I like what you're getting at here, but I'm going to avoid "technically" here because taking security precautions is almost never "technically" necessary. (Just hope you'll get lucky and not pwned!) That's not quite the meaning I'm going for here: even if someone does eavesdrop your communication, they almost certainly can't use the claim for much.

content/tutorial-paychan.md
```diff
@@ -0,0 +1,611 @@
...
+Payment Channels are an advanced feature for sending "asynchronous" XRP payments that can be divided into very small increments and settled later. This tutorial walks through the entire process of using a payment channel, with examples using the [JSON-RPC API](reference-rippled.html) of a local `rippled` server.
+
+## Background
+
+The process of using a payment channel always involves two parties, a payer and a payee. The payer is an individual person or institution using the Ripple Consensus Ledger who is a customer of the payee. The payee is a person or business who is doing business on the Ripple Consensus Ledger (RCL), receiving XRP as payment for goods or services.
+
+The types of goods and services are not defined by the software or in this tutorial. However, the types of goods and services that are a good fit for payment channels are:
+
+- Things that can be transmitted near-instantly, like digital items
+- Inexpensive things, where the cost of processing a transaction is a non-trivial portion of the price
```
ryangyoung, line 12: Just a question for my own understanding: what's an example of this type of use case?

content/tutorial-paychan.md
```diff
@@ -0,0 +1,611 @@
...
+The example addresses used in this tutorial are:
+
+| | |
+|--|--|
+| Payer's address | rN7n7otQDd6FczFgLdSqtcsAUxDkw6fzRH |
+| Public key used for channel (in base58) | aB44YfzW24VDEJQ2UuLPV2PvqcPCSoLnL7y5M1EzhdW4LnK5xMS3
+| Public key used for channel (in hex) | 023693F15967AE357D0327974AD46FE3C127113B1110D6044FD41E723689F81CC6 |
+| Payee's address | rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn |
+
+**Tip:** In this example, we use the public key from the payer's master key pair as the key pair for the channel. This is perfectly safe and valid. It is also perfectly safe and valid to use a different key pair, as long only the payer knows the public and secret keys for that key pair. <!-- Editor's note: We don't have a good page to link to explain key pairs as of time of this writing. -->
```
ryangyoung: > as long only the payer knows 

typo: "as long *as* only the payer knows..."

content/tutorial-paychan.md
```diff
@@ -0,0 +1,611 @@
...
+
+To test without transferring real XRP, you can use [Ripple Test Net](https://ripple.com/build/ripple-test-net/) addresses with Test Net XRP. If you do use the Ripple Test Net, you can use the Test Net servers' JSON-RPC API by connecting to `https://api.altnet.rippletest.net:51234` instead of `http://localhost:5005/`.
+
+You can use any amount of XRP for the payment channels. The example values in this tutorial set aside 100 XRP (`100000000` drops) in a payment channel for at least 1 day.
+
+## Flow Diagram
+
+The following diagram summarizes the lifecycle of a payment channel:
+
+[![Payment Channel Flow Diagram](img/paychan-flow.png)](img/paychan-flow.png)
```
ryangyoung: That's a great diagram. Very helpful in clarifying the process!

content/tutorial-paychan.md
```diff
@@ -0,0 +1,611 @@
...
+        }
+    }
+
+The payee should check that the parameters of the payment channel are suitable for their specific use case, including all of the following:
+
+- Confirm the `destination_account` field has the payee's correct address.
+- Confirm the `settle_delay` field has a settlement delay in seconds that provides enough time for the payee to redeem outstanding claims.
+- Confirm the fields `cancel_after` (immutable expiration) and `expiration` (mutable expiration), if they are present, are not too soon. The payee should take note of these times so they can be sure to redeem claims before then.
+- Take note of the `public_key` and `channel_id` fields. These are necessary later to verify and redeem claims.
+- _(Optional)_ Confirm the `destination_tag` field, if present, has the correct destination tag (if desired).
```
ryangyoung: nitpick: Using "Optional" and "(if desired)" seems a little redundant. One or the other seems sufficient.

content/tutorial-paychan.md
```diff
@@ -0,0 +1,611 @@
...
+    - The `rippled` server the payee queried has lost sync with the rest of the network or is experiencing an unknown bug. Use the [`server_info` command](reference-rippled.html#server-info) to check the state of the server. (If you can reproduce this situation, please [report an issue](https://github.com/ripple/rippled/issues/).)
+
+After confirming both the signature and the current state of the payment channel, the payee has not yet received the XRP, but is certain that he or she _can_ redeem the XRP as long as the transaction to do so is processed before the channel expires.
+
+
+## 6. Payee provides goods or services.
+
+At this point, the payee can provide goods and services to the payer, knowing that payment is already assured.
+
+For purposes of this tutorial, the payee can give the payer a high-five or equivalent online message as the "goods and services".
```
ryangyoung, line 355: Or fist bump?  :fist_right:  :fist_left:

content/tutorial-paychan.md
```diff
@@ -0,0 +1,611 @@
...
+
+    {
+        "method": "submit",
+        "params": [{
+                "secret": "s████████████████████████████",
+                "tx_json": {
+                    "Account": "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn",
+                    "TransactionType": "PaymentChannelClaim",
+                    "Amount": "1000000",
+                    "Balance": "1000000",
```
ryangyoung, line 390: Probably a naive question, but do the `Balance` and `Amount` need to be the same? Would they only differ if the payee doesn't want to redeem the full amount?

content/tutorial-paychan.md
```diff
@@ -0,0 +1,611 @@
...
+                "TransactionResult": "tesSUCCESS"
+            },
+            "status": "success",
+            "validated": true
+        }
+    }
+
+In the transaction's metadata, look for the following:
+
+- A `DeletedNode` entry with `"LedgerEntryType": "PayChannel"`. The `LedgerIndex` field should match the Channel ID. This indicates that the channel was deleted.
```
ryangyoung, line 599: This is a nitpick, but is there another way to specify which `LedgerIndex` field? Some lazy readers (ahem, not me of course) might skip over the first sentence and be confused by the multiple `LedgerIndex` fields in the response. (Though if the reader made it this far, they're probably invested enough to pay sufficient attention to those details, so feel free to ignore this if it doesn't make sense.)

content/tutorial-paychan.md
ryangyoung: But it's only defined in the Tip, which still seems a little backward to me. Shouldn't it be defined before it's clarified? Currently, it's only introduced in a kind of vague way.

content/tutorial-paychan.md
ryangyoung: To me, it seems like separating these into distinct actions might be helpful for letting the reader know that there are N number of steps involved in the overall task.

ryangyoung (approved):

ryangyoung:
I left a few more comments, most of which are small nitpicks or questions that can be safely ignored. I still think adding sub-steps would be helpful, but I'm ok with publishing this as-is.

content/tutorial-paychan.md
mDuo13: That's exactly correct

content/tutorial-paychan.md
mDuo13: I dunno. Maybe a maker space could charge half a cent for every washer and nail you use from a communal supply?

mDuo13:
Addressed Ryan's re-review comments and rebased onto master to fix the missing accounts concept link.

PR #251, opened 2017-05-19
rippled: paychan API methods

mDuo13:
Adds documentation for three new rippled RPC API methods related to payment channels.

As usual, the HTML files are generated from the MD files so you can ignore those.

### DIFF
--- a/content/reference-rippled.md
+++ b/content/reference-rippled.md
@@ -1277,18 +1277,18 @@ An example of a successful response:
 
 <!-- MULTICODE_BLOCK_END -->
 
-The response follows the [standard format](#response-formatting), with a successful result containing the address of the account and an array of trust-line objects. Specifically, the result object contains the following fields:
+The response follows the [standard format](#response-formatting), with a successful result containing the address of the account and an array of trust line objects. Specifically, the result object contains the following fields:
 
 | `Field`                | Type                                       | Description |
 |:-----------------------|:-------------------------------------------|:-------|
 | `account`              | String                                     | Unique [Address][] of the account this request corresponds to. This is the "perspective account" for purpose of the trust lines. |
-| `lines`                | Array                                      | Array of trust-line objects, as described below. If the number of trust-lines is large, only returns up to the `limit` at a time. |
+| `lines`                | Array                                      | Array of trust line objects, as described below. If the number of trust lines is large, only returns up to the `limit` at a time. |
 | `ledger_current_index` | Integer                                    | (Omitted if `ledger_hash` or `ledger_index` provided) Sequence number of the ledger version used when retrieving this data. [New in: rippled 0.26.4-sp1][] |
 | `ledger_index`         | Integer                                    | (Omitted if `ledger_current_index` provided instead) Sequence number, provided in the request, of the ledger version that was used when retrieving this data. [New in: rippled 0.26.4-sp1][] |
 | `ledger_hash`          | String                                     | (May be omitted) Hex hash, provided in the request, of the ledger version that was used when retrieving this data. [New in: rippled 0.26.4-sp1][] |
 | `marker`               | [(Not Specified)](#markers-and-pagination) | Server-defined value indicating the response is paginated. Pass this to the next call to resume where this call left off. Omitted when there are no additional pages after this one. [New in: rippled 0.26.4][] |
 
-Each trust-line object has some combination of the following fields:
+Each trust line object has some combination of the following fields:
 
 | `Field`          | Type             | Description                            |
 |:-----------------|:-----------------|:---------------------------------------|
--- a/content/reference-rippled.md
+++ b/content/reference-rippled.md
@@ -418,6 +418,7 @@ Some API methods require you to specify an amount of currency. Depending on whet
 
 #### XRP ####
 [drops of XRP]: #xrp
+[XRP, in drops]: #xrp
 
 Amounts of XRP are represented as strings. (XRP has precision equivalent to a 64-bit integer, but JSON integers are limited to 32 bits, so XRP can overflow if represented in a JSON integer.) XRP is formally specified in "drops", which are equivalent to 0.000001 (one 1-millionth) of an XRP each. Thus, to represent 1.0 XRP in a JSON document, you would write:
 
@@ -513,12 +514,15 @@ API methods for the Websocket and JSON-RPC APIs are defined by command names, an
 ## List of Public Commands ##
 
 * [`account_currencies` - Get a list of currencies an account can send or receive](#account-currencies)
+* [`account_channels` - Get a list of payment channels where the account is the source](#account-channels)
 * [`account_info` - Get basic data about an account](#account-info)
 * [`account_lines` - Get info about an account's trust lines](#account-lines)
 * [`account_objects` - Get all ledger objects owned by an account](#account-objects)
 * [`account_offers` - Get info about an account's currency exchange offers](#account-offers)
 * [`account_tx` - Get info about an account's transactions](#account-tx)
 * [`book_offers` - Get info about offers to exchange two currencies](#book-offers)
+* [`channel_authorize` - Sign a claim for money from a payment channel](#channel-authorize)
+* [`channel_verify` - Check a payment channel claim's signature](#channel-verify)
 * [`fee` - Get information about transaction cost](#fee)
 * [`gateway_balances` - Calculate total amounts issued by an account](#gateway-balances)
 * [`ledger` - Get info about a ledger version](#ledger)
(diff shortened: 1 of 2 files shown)


content/reference-rippled.md
```diff
@@ -513,12 +514,15 @@ API methods for the Websocket and JSON-RPC APIs are defined by command names, an
 ## List of Public Commands ##
 
 * [`account_currencies` - Get a list of currencies an account can send or receive](#account-currencies)
+* [`account_channels` - Get a list of payment channels where the account is the source](#account-channels)
```
jbheron: Perhaps this is obvious to your intended reader, but '...the source' of what?

content/reference-rippled.md
```diff
@@ -513,12 +514,15 @@ API methods for the Websocket and JSON-RPC APIs are defined by command names, an
...
 * [`account_info` - Get basic data about an account](#account-info)
 * [`account_lines` - Get info about an account's trust lines](#account-lines)
 * [`account_objects` - Get all ledger objects owned by an account](#account-objects)
 * [`account_offers` - Get info about an account's currency exchange offers](#account-offers)
 * [`account_tx` - Get info about an account's transactions](#account-tx)
 * [`book_offers` - Get info about offers to exchange two currencies](#book-offers)
+* [`channel_authorize` - Sign a claim for money from a payment channel](#channel-authorize)
+* [`channel_verify` - Check a payment channel claim's signature](#channel-verify)
```
jbheron, line 525: I dislike inanimate objects having possessive qualities (anthro), so, how about:

> Check the signature of a payment channel claim.

content/reference-rippled.md
```diff
@@ -737,6 +741,181 @@ The response follows the [standard format](#response-formatting), with a success
...
+rippled account_channels rN7n7otQDd6FczFgLdSqtcsAUxDkw6fzRH rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn validated
+```
+
+<!-- MULTICODE_BLOCK_END -->
+
+The request includes the following parameters:
+
+| Field                 | Type                                       | Description |
+|:----------------------|:-------------------------------------------|:--------|
+| `account`             | String                                     | A unique identifier for the account whose channels to retrieve. Typically this is the account's [Address][]. |
```
jbheron: I can't parse this sentence, particularly the end,  'whose channels to retrieve.' Does the following express what you mean? 

> A unique identifier to retrieve the channels for the account, typically the account address.

content/reference-rippled.md
```diff
@@ -737,6 +741,181 @@ The response follows the [standard format](#response-formatting), with a success
...
+| `destination_account` | String                                     | _(Optional)_ If provided, filter results to payment channels where the destination of the channel. Typically this is an [Address][]. |
```
jbheron: I can't parse this sentence. I see that 'where' has been used throughout the document, despite its spatial and proximate meaning. (very sad) So, I'll just encourage less use of it, in general. Beyond that (small detail?) though, does this sentence truncate before revealing some missing meaning? What about the destination of the channel? Or is there a typo?

content/reference-rippled.md
```diff
@@ -737,6 +741,181 @@ The response follows the [standard format](#response-formatting), with a success
...
+| `ledger_hash`         | String                                     | _(Optional)_ A 20-byte hex string for the ledger version to use. (See [Specifying a Ledger](#specifying-ledgers)) |
+| `ledger_index`        | String or Unsigned Integer                 | _(Optional)_ The sequence number of the ledger to use, or a shortcut string to choose a ledger automatically. (See [Specifying a Ledger](#specifying-ledgers)) |
```
jbheron: Tiny picky thing, I prefer '...to automatically choose a ledger'. Is that what you mean?

content/reference-rippled.md
```diff
@@ -737,6 +741,181 @@ The response follows the [standard format](#response-formatting), with a success
...
+| `limit`               | Integer                                    | _(Optional)_ Limit the number of transactions to retrieve. The server is not required to honor this value. The default is 200. Cannot be smaller than 10 or larger than 400. |
```
jbheron: I'd combine the last two sentences:

> The default is 200 and the acceptable range is 10 to 400.

content/reference-rippled.md
```diff
@@ -737,6 +741,181 @@ The response follows the [standard format](#response-formatting), with a success
...
+| `marker`              | [(Not Specified)](#markers-and-pagination) | _(Optional)_ Value from a previous response to specify where to resume retrieving data from. |
```
jbheron: I find it difficult to follow this idea. I understand it better in present-active tense, as:

> Specifies previous response value from which to resume retrieving data. 

OR (even more clear to me but may not be as accurate)

> Specifies previous response value from which to resume data retrieval.

content/reference-rippled.md
```diff
@@ -737,6 +741,181 @@ The response follows the [standard format](#response-formatting), with a success
...
+}
+```
+
+<!-- MULTICODE_BLOCK_END -->
+
+The response follows the [standard format](#response-formatting), with a successful result containing the following fields:
+
+| Field      | Type                                       | Description        |
+|:-----------|:-------------------------------------------|:-------------------|
+| `account`  | String                                     | The address of the account from the request; the owner of the payment channels. |
```
jbheron: I had thought to join these using 'and' without a semi-colon, as:

> The address of the account from the request and the owner of the payment channels.

But, as I dig in and re-read, I'm confused by 'address of the account from the request'. Does it mean the account address that initiated the request? If so, perhaps:

> The account address of the payment channels.

Hmmm... 

Or:

>  The account address and owner of the payment channels.

content/reference-rippled.md
```diff
@@ -737,6 +741,181 @@ The response follows the [standard format](#response-formatting), with a success
...
+| `channels` | Array of Channel Objects                   | Payment channels owned by this `account`. |
+| `limit`    | Number                                     | _(May be omitted)_ The limit to how many channel objects were actually returned by this request. |
```
jbheron: Using 'were actually' seems either colloquial or intentional to indicate that not all objects may be returned by any given request. If the latter, and if I even understand it, might I suggest:

> Limits the number of channel objects the request returns.

content/reference-rippled.md
```diff
@@ -737,6 +741,181 @@ The response follows the [standard format](#response-formatting), with a success
...
+| `marker`   | [(Not Specified)](#markers-and-pagination) | _(May be omitted)_ Server-defined value for pagination. Pass this to the next call to resume getting results where this call left off. Omitted when there are no additional pages after this one. |
```
jbheron: I think the use of 'where this call left off' is confusing, how about something like:

> Server-defined value for pagination. Pass this value to the next call to return a continuation of the truncated results. Omitted when there are no additional pages after this one.

content/reference-rippled.md
```diff
@@ -737,6 +741,181 @@ The response follows the [standard format](#response-formatting), with a success
...
+
+Each Channel Object has the following fields:
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `account` | String | The owner of the channel, as an [Address][]. |
+| `amount` | String | The amount of [XRP, in drops](#specifying-currency-amounts) allocated to this channel in total. |
```
jbheron: > The total amount of [XRP, in drops](#specifying-currency-amounts), allocated to this channel.

content/reference-rippled.md
```diff
@@ -737,6 +741,181 @@ The response follows the [standard format](#response-formatting), with a success
...
+| `balance` | String | The total amount of XRP, in drops, paid out from this channel so far. (You can calculate the amount of XRP remaining in the channel by subtracting `balance` from `amount`.) |
```
jbheron: 'So far' creates some confusion, how about 'to date':

The total amount of XRP, in drops, paid out from this channel to date.

content/reference-rippled.md
```diff
@@ -737,6 +741,181 @@ The response follows the [standard format](#response-formatting), with a success
...
+| `channel_id` | String | A unique ID for this channel, as a 64-character hexadecimal string. This is also the [index of the channel](reference-ledger-format.html#paychannel-index-format) in the ledger's state data. |
+| `destination_account` | String | The destination of the channel, as an [Address][]. Only this account can receive the XRP in the channel while it remains open. |
```
jbheron: Nit picking ensues: 

> The destination of the channel, as an [Address][]. The only account that can receive the XRP in the channel while it is open.

content/reference-rippled.md
```diff
@@ -737,6 +741,181 @@ The response follows the [standard format](#response-formatting), with a success
...
+| `public_key` | String | _(May be omitted)_ The public key for the payment channel in base58 format, if one was specified at channel creation. Signed claims against this channel must be redeemed with the matching key pair. |
+| `public_key_hex` | String | _(May be omitted)_ The public key for the payment channel in hexadecimal format, if one was specified at channel creation. Signed claims against this channel must be redeemed with the matching key pair. |
+| `settle_delay` | Unsigned Integer | The number of seconds the payment channel must remain open after the owner of the channel requests to close it. |
+| `expiration` | Unsigned Integer | _(May be omitted)_ Time, in seconds since the [Ripple Epoch](#specifying-time), when this channel is set to expire. This expiration date is mutable. If this is before the close time of the most recent validated ledger, the channel is expired. |
```
jbheron: A small adjustment for clarity and active voice: 

> If the expiration occurs before the close time of the most recent validated ledger, the channel expires.

content/reference-rippled.md
```diff
@@ -737,6 +741,181 @@ The response follows the [standard format](#response-formatting), with a success
...
+| `cancel_after` | Unsigned Integer | _(May be omitted)_ Time, in seconds since the [Ripple Epoch](#specifying-time), of this channel's immutable expiration, if one was specified at channel creation. If this is before the close time of the most recent validated ledger, the channel is expired. |
```
jbheron: Again, 'the channel expires' instead of 'is expired'.

content/reference-rippled.md
```diff
@@ -951,7 +1130,7 @@ Each object in the `transactions` array, if present, may contain any or all of t
 ## account_lines ##
 [[Source]<br>](https://github.com/ripple/rippled/blob/master/src/ripple/rpc/handlers/AccountLines.cpp "Source")
 
-The `account_lines` method returns information about the account's lines of trust, including balances in all non-XRP currencies and assets. All information retrieved is relative to a particular version of the ledger.
+The `account_lines` method returns information about an account's trust lines, including balances in all non-XRP currencies and assets. All information retrieved is relative to a particular version of the ledger.
```
jbheron, line 1133: I see that we typically hyphenate trust-lines:

> The `account_lines` method returns information about an account's trust-lines, including balances in all non-XRP currencies and assets. All information retrieved is relative to a particular version of the ledger.

content/reference-rippled.md
```diff
@@ -7571,6 +7750,247 @@ In addition to the standard Offer fields, the following fields may be included i
 
 
 
+## channel_authorize
+[[Source]<br>](https://github.com/ripple/rippled/blob/d4a56f223a3b80f64ff70b4e90ab6792806929ca/src/ripple/rpc/handlers/PayChanClaim.cpp#L41 "Source")
+
+_(Requires the [PayChan amendment](concept-amendments.html#paychan) to be enabled. [New in: rippled 0.33.0][])_
+
+The `account_channels` method creates a signature that can be used to redeem a specific amount of XRP from a payment channel.
```
jbheron: Less is more?

> +The `account_channels` method creates a signature to redeem a specific amount of XRP from a payment channel.

content/reference-rippled.md
```diff
@@ -7571,6 +7750,247 @@ In addition to the standard Offer fields, the following fields may be included i
...
+```
+
+<!-- MULTICODE_BLOCK_END -->
+
+The request includes the following parameters:
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `channel_id` | String | The unique ID of the payment channel to use.
+| `secret` | String | The secret key to use to sign the claim. This must be the same key pair as the public key specified in the channel. |
```
jbheron, line 7807: More 'less is more':

> The secret key to sign the claim. It must be the same key pair as the public key specified in the channel.

content/reference-rippled.md
```diff
@@ -7571,6 +7750,247 @@ In addition to the standard Offer fields, the following fields may be included i
...
+| `amount` | String | Cumulative amount of XRP, in drops, to authorize. If the destination has already received a lesser amount of XRP from this channel, the signature created by this method can be redeemed for the difference. |
+
+**Note:** You cannot use Ed25519 keys to sign claims with this method. This is a known bug.
```
jbheron, line 7810: Do we reference known bugs with links?

content/reference-rippled.md
```diff
@@ -7571,6 +7750,247 @@ In addition to the standard Offer fields, the following fields may be included i
...
+}
+```
+
+<!-- MULTICODE_BLOCK_END -->
+
+The response follows the [standard format](#response-formatting), with a successful result containing the following fields:
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `signature` | String | The signature for this claim, as a hexadecimal value. To verify this signature or process the claim, the destination of the payment channel must send a [PaymentChannelClaim transaction][] with this signature, the exact Channel ID, XRP amount, and public key of the channel. |
```
jbheron: While a little wordy in the iteration of requisite payload for claim, I'm confused by the destination being the sender. Do you mean that the sender is the receiver of the payment channel or the destination account or something else? Or perhaps the destination being the sender is an actual thing. (very sad)

If the answer is that the destination account is the sender, then:

> The signature for this claim, as a hexadecimal value. To verify this signature or process the claim, the destination account of the payment channel must send a [PaymentChannelClaim transaction][] with this signature, the exact Channel ID, XRP amount, and public key of the channel.

content/reference-rippled.md
```diff
@@ -7571,6 +7750,247 @@ In addition to the standard Offer fields, the following fields may be included i
...
+
+#### Possible Errors
+
+* Any of the [universal error types](#universal-errors).
```
jbheron: Even more 'less is more':

> Any [universal error types](#universal-errors).

content/reference-rippled.md
```diff
@@ -7571,6 +7750,247 @@ In addition to the standard Offer fields, the following fields may be included i
...
+* `badSeed` - The value specified in the `secret` field was not a valid secret key.
```
jbheron, line 7865: Present tense FTW:

>* `badSeed` - The value specified in the `secret` field is not a valid secret key.

content/reference-rippled.md
```diff
@@ -7571,6 +7750,247 @@ In addition to the standard Offer fields, the following fields may be included i
...
+* `channelAmtMalformed` - The value specified in the `amount` field was not a valid XRP amount. See [Specifying Currency Amounts](#specifying-currency-amounts) for details.
```
jbheron, line 7866: ditto:

> * `channelAmtMalformed` - The value specified in the `amount` field is not a valid XRP amount. See [Specifying Currency Amounts](#specifying-currency-amounts) for details.

content/reference-rippled.md
```diff
@@ -7571,6 +7750,247 @@ In addition to the standard Offer fields, the following fields may be included i
...
+* `channelMalformed` - The value specified in the `channel_id` field of the reqeuest was not a valid Channel ID. The Channel ID should be a 256-bit (64-character) hexadecimal string.
```
jbheron, line 7867: again:

>* `channelMalformed` - The value specified in the `channel_id` field of the reqeuest is not a valid Channel ID. The Channel ID should be a 256-bit (64-character) hexadecimal string.

(Perhaps I missed a bunch of other past tense references -- if so, I'd suggest present tense where possible)

content/reference-rippled.md
```diff
@@ -7571,6 +7750,247 @@ In addition to the standard Offer fields, the following fields may be included i
...
+
+
+
+## channel_verify
+[[Source]<br>](https://github.com/ripple/rippled/blob/d4a56f223a3b80f64ff70b4e90ab6792806929ca/src/ripple/rpc/handlers/PayChanClaim.cpp#L89 "Source")
+
+_(Requires the [PayChan amendment](concept-amendments.html#paychan) to be enabled. [New in: rippled 0.33.0][])_
+
+The `channel_verify` method checks the validity of a signature that can be used to redeem a specific amount of XRP from a payment channel.
```
jbheron, line 7876: Terse makes clear:

> The `channel_verify` method checks the validity of a signature to redeem a specific amount of XRP from a payment channel.

content/reference-rippled.md
```diff
@@ -7571,6 +7750,247 @@ In addition to the standard Offer fields, the following fields may be included i
...
+
+<!-- MULTICODE_BLOCK_END -->
+
+The request includes the following parameters:
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `amount` | String | The amount of [XRP, in drops][], the provided `signature` authorizes. |
+| `channel_id` | String | The Channel ID of the channel that provides the XRP. This is a 64-character hexadecimal string. |
+| `public_key` | String | The public key of the channel and the key pair that was used to create the signature, in base58 format. (One way to get the public key in base58 format is from the [`wallet_propose` command](#wallet-propose).) |
```
jbheron: Strikes 'from':

> | `public_key` | String | The public key of the channel and the key pair that was used to create the signature, in base58 format. (One way to get the public key in base58 format is the [`wallet_propose` command](#wallet-propose).) |

content/reference-rippled.md
```diff
@@ -7571,6 +7750,247 @@ In addition to the standard Offer fields, the following fields may be included i
...
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `signature_verified` | Boolean | If `true`, the signature is valid for the stated amount, channel, and public key. |
+
+**Caution:** This does not indicate check that the channel has sufficient XRP allocated to it. Before considering a claim valid, you should look up the channel in the latest validated ledger and confirm that the channel is open and its `amount` value is equal or greater than the `amount` of the claim. To do so, use the [`account_channels` method](#account-channels).
+
+#### Possible Errors ####
+
+* Any of the [universal error types](#universal-errors).
```
jbheron, line 7986: > * Any [universal error types](#universal-errors).

content/reference-rippled.md
```diff
@@ -7571,6 +7750,247 @@ In addition to the standard Offer fields, the following fields may be included i
...
+* `invalidParams` - One or more fields are specified incorrectly, or one or more required fields are missing.
+* `publicMalformed` - The value specified in the `public_key` field of the request was not a valid public key in the correct format. Public keys should be 33 bytes, represented in base58. The base58 representation should start with letter `a`.
```
jbheron: Present tense preferred, and do you mean 'must'?

> +* `publicMalformed` - The value specified in the `public_key` field of the request is not a valid public key in the correct format. Public keys must be 33 bytes, represented in base58. The base58 representation must start with letter `a`.

content/reference-rippled.md
```diff
@@ -7571,6 +7750,247 @@ In addition to the standard Offer fields, the following fields may be included i
...
+* `channelMalformed` - The value specified in the `channel_id` field of the reqeuest was not a valid Channel ID. The Channel ID should be a 256-bit (64-character) hexadecimal string.
+* `channelAmtMalformed` - The value specified in the `amount` field was not a valid XRP amount. See [Specifying Currency Amounts](#specifying-currency-amounts) for details.
```
jbheron: Present tense and 'should' >> 'must':
> * `channelMalformed` - The value specified in the `channel_id` field of the reqeuest is not a valid Channel ID. The Channel ID must be a 256-bit (64-character) hexadecimal string.
 * `channelAmtMalformed` - The value specified in the `amount` field is not a valid XRP amount. See [Specifying Currency Amounts](#specifying-currency-amounts) for details.

jbheron (changes requested):
Looks really good, I learned a bunch. I've mostly pointed out some style and grammar items, and, where appropriate, highlighted where I think the language could use some clarity. 

Feel free to disregard comments that you feel migrate the work away from its typical tone (precedent), if you find their update onerous or they would change the docs too much.

content/reference-rippled.md
```diff
@@ -737,6 +741,181 @@ The response follows the [standard format](#response-formatting), with a success
...
+| `amount` | String | The amount of [XRP, in drops](#specifying-currency-amounts) allocated to this channel in total. |
+| `balance` | String | The total amount of XRP, in drops, paid out from this channel so far. (You can calculate the amount of XRP remaining in the channel by subtracting `balance` from `amount`.) |
+| `channel_id` | String | A unique ID for this channel, as a 64-character hexadecimal string. This is also the [index of the channel](reference-ledger-format.html#paychannel-index-format) in the ledger's state data. |
+| `destination_account` | String | The destination of the channel, as an [Address][]. Only this account can receive the XRP in the channel while it remains open. |
+| `public_key` | String | _(May be omitted)_ The public key for the payment channel in base58 format, if one was specified at channel creation. Signed claims against this channel must be redeemed with the matching key pair. |
+| `public_key_hex` | String | _(May be omitted)_ The public key for the payment channel in hexadecimal format, if one was specified at channel creation. Signed claims against this channel must be redeemed with the matching key pair. |
```
seelabs: For `public_key` and `public_key_hex`. I'd remove the "if one was specified at channel creation". Channels should always have a public key associated with them. I know there is code that checks if the key exists, but that's me being paranoid and to make sure rippled doesn't crash.

content/reference-rippled.md
seelabs: You mean `channel_authorize` here, not `account_channels`.

content/reference-rippled.md
mDuo13: This is probably a style discussion to have elsewhere, but I disagree on the possessive. As a student of Latin, whose genitive case is quite flexible (and ultimately quite similar English's possessive, partitive genitive aside), I'm inclined to use whichever form conveys the intended information most succinctly. In this case, "Check the payment channel claim's signature" reads more clearly than "Check the signature of the payment channel claim" to me so I say we keep it.

content/reference-rippled.md
mDuo13: Well, technically neither my example nor yours are sentences; they're merely noun phrases. That said, although I'm confident my original phrasing is linguistically valid, I can agree it's not the clearest. How's this for a better restructuring:

> The unique identifier of an account, typically the account's Address. The request returns channels where this account is the channel's source.

content/reference-rippled.md
mDuo13: It's a typo. It was intended to say, "Filter results to payment channels where this account is the destination of the channel." Per your stylistic suggestions, I revised this table cell a little more:

> _(Optional)_ The unique identifier of an account, typically the account's [Address](https://ripple.com/build/rippled-apis/#addresses). If provided, filter results to payment channels whose destination is this account.

content/reference-rippled.md
mDuo13: "To automatically choose X" is a split infinitive, which I habitually avoid. "To choose X automatically" is equivalent.

content/reference-rippled.md
mDuo13: I rewrote this and other `limit` parameter descriptions to be a little more consistent, but I kept them as separate sentences so that we can be consistent with "Defaults to" instead of "The default is"

content/reference-rippled.md
mDuo13: I rephrased all instances of the `marker` field (in responses) like this:

> Value from a previous paginated response. Resume retrieving data where that response left off.

content/reference-rippled.md
mDuo13: This is a little tricky, for a few reasons:

- Anyone can initiate a request for any account. The API request isn't "from" an account.
- It really means the `account` address from the request. However, the `account` value in the request may not actually be an address; it could be a public key or one or two other formats. Regardless of what format was used in the request, the response always formats `account` as an address. Thus, "the address of the account from the request".
- The "owner" and "source" refer to the same thing in payment channels. ("Source" emphasizes the direction the XRP is flowing and who created the channel and is a dualism to the term "destination" used in the API. "Owner" emphasizes who's responsible for the Owner Reserve, and who authorizes the flow of the XRP.)

content/reference-rippled.md
mDuo13: I get where you're coming from, but this is phrased the weird way it is for good reasons:

1. This is the description of a _response_ parameter. So it's not correct to say it limits what the request returns. That would imply it's an instruction, which it isn't.
2. This parameter exists because the server is not required to honor the `limit` value of the request. (The server may use a stricter limit if it's under heavy load, or the request specified a limit above the server's configured maximum limit, or something.) So this response tells you what limit the server enforced in this particular case.

content/reference-rippled.md
mDuo13: I'm going to leave this as-is for now since changing it would involve changing the description of pagination in a lot of places.

content/reference-rippled.md
mDuo13: If the request specified a historical ledger version, "to date" would be inaccurate. Changed to "as of the ledger version used" instead.

content/reference-rippled.md
mDuo13: I think the active voice is actively deceptive in this particular case (pun intended). To say "expiration occurs" suggests that an action has taken place. The RCL and the rippled servers don't actually do anything though. They may not even notice that the expiration time has passed.

"Expiration" is more a matter of perspective than an actual occurrence, if that makes sense.

It would be more accurate to say "the RCL rules define a channel in a particular ledger version to be expired if and only if the channel's `Expiration` time is prior to the close time of that particular ledger version's direct parent ledger version." But that's a mouthful.

content/reference-rippled.md
jbheron: Ah, I see, that makes sense. I would still remove 'actually', one way or another, perhaps as:
 
> The enforced limit of the number of channel objects returned in the response.

content/reference-rippled.md
mDuo13: Per the [style guide](https://ripplelabs.atlassian.net/wiki/display/DOCS/Ripple+Documentation+Style+Guide) we do not hyphenate trust lines.

content/reference-rippled.md
mDuo13: Not when the bugs are in JIRA, because our JIRA is not public (anymore).

content/reference-rippled.md
mDuo13: I'd argue your phrasing makes it less clear. It's easy to understand your phrasing as meaning that you use channel verify to redeem XRP. That's a key misunderstanding of the API.

One uses `channel_verify` to check the validity of a signature, so that one knows one is _able_ to redeem a claim if one decides to do so later. (One might decide not to do so, because one redeems a subsequent claim that supersedes this claim.)

content/reference-rippled.md
mDuo13: Disagree. The phrasing is already using a past participle ("specified") so past tense is required for tense consistency.

mDuo13: On second thought, I've come around to using present tense for these.

content/reference-rippled.md
jbheron: Hmmm, I see now that 25 times we do not, but four times we do hyphenate trust lines. Not sure if you want to get those instances in this PR, or pick them up as typos somewhere else.

jbheron (approved):

PR #259, opened 2017-06-15
rippled v0.70.0 changes: amendments, queue_data

mDuo13:
- Adds 3 new amendments (EnforceInvariants, fix1373, FlowCross) to the
Known Amendments table with descriptions for each. (DOC-963)
- Adds the new queue parameter to the ledger command (DOC-801) and
reformats some of the surrounding text.
- Updates the path specifications page to clarify some rules that relate
to the FlowCross amendment.

The link to the rippled 0.70.0 release tag is expected to fail until
that version is released.

As per usual, the .html files are generated from the .md files, so no need to review the .html versions.

PR #261, opened 2017-06-16
rippled: corrections to ledger/queue dump

mDuo13:
This incorporates @ximinez's feedback from after PR #259 was merged


ximinez (approved):
👍

PR #270, opened 2017-09-05
Reorg: remove generated HTML files, move dactyl-config to op level

mDuo13:
Reorganizes the ripple-dev-portal repo to work a little more like the docs-private repo:

- Build from the top folder of the repo instead of the `tool/` folder
- Removes all the generated `*.html` files from the top level of the repo. (So you no longer have to build the generated HTML to keep the repo internally consistent, and PRs will no longer have generated HTML content to distract from the source files.)
- The default build goes to an `out/` folder at the top level, which is now part of the `.gitignore` file

As discussed internally, this gives up on fixing GitHub Pages compatibility in favor of simpler repo management. To get GH Pages to work with Dactyl, we'd need to track static files in multiple places or do something else weird, just for the rare occasions we'd use it. It's easy to stage the files on another server or just check out the repo and build instead.

(Resolves DOC-955)


jbheron (approved):
This looks like a great improvement. I haven't worked much, if at all, with this repo, but this organization looks like a solid step forward!

PR #272, opened 2017-09-20
Escrow concept article (with diagrams!)

mDuo13:
Adds a concept article that describes the Escrow feature, including some diagrams of how it works, with copious links to the other Escrow documentation that already exists.

- Moves the "Availability of Escrow" blurb out of the Escrow tutorials into the new concept page.
- Discusses escrow cancellation. (Tutorial for canceling an escrow still forthcoming, see DOC-1064.)
- Contains the generalized formula for an EscrowFinish transaction with a fulfillment.


content/concept-escrow.md
```diff
@@ -0,0 +1,89 @@
...
+
+**Step 1:** To send an escrow, the sender uses an [EscrowCreate transaction][] to set aside some XRP. This transaction defines a finish time, an expiration time, or both. The transaction may also define a crypto-condition that must be fulfilled to finish the escrow. This transaction must define an intended recipient for the XRP; the recipient _may_ be the same as the sender.
+
+**Step 2:** After this transaction has been processed, the XRP Ledger has an [Escrow object](reference-ledger-format.html#escrow) that holds the escrowed XRP. This object tracks the properties of the escrow as defined by the transaction that created it. If this escrow has a finish time, no one can access the XRP before then.
+
+**Step 3:** The recipient, or any other XRP Ledger address, sends an [EscrowFinish transaction][] to deliver the XRP. If the correct conditions are met, this destroys the Escrow object in the ledger and credits the XRP to the intended recipient. If the escrow has a crypto-condition, this transaction must include a fulfillment for that condition. If the escrow has an expiration time, this cannot happen after that time.
```
jbheron: Wow, so, anyone could finish an escrow transaction, even if they're not the sender or recipient? That's wild.

In the last sentence, 'this cannot happen after that time', I'd suggest just stating again to what 'this' refers:

>  If the escrow has an expiration time, condition fulfillment cannot happen after that time.

content/concept-escrow.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Expiration Case
+
+[![Escrow Flow Diagram (Expired escrow)](img/escrow-cancel-flow.png)](img/escrow-cancel-flow.png)
+
+All escrows start the same way, so **Steps 1 and 2** are the same as in the successful case.
+
+**Step 3a:** If the escrow has an expiration time, and it has not been successfully finished before then, the escrow is considered expired. It continues to exist in the XRP Ledger, but can no longer successfully finish. (Expired objects remain in the ledger until a transaction modifies them. Time-based triggers cannot change the ledger contents.)
```
jbheron, line 25: Nit pick, some passive voice. How about:

> If the escrow has an expiration time, and it does not successfully finish before then, the escrow expires.

content/concept-escrow.md
```diff
@@ -0,0 +1,89 @@
...
+
+**Step 4a:** The sender, or any other XRP Ledger address, sends an [EscrowCancel transaction][] to cancel the expired escrow. This destroys the [Escrow object](reference-ledger-format.html#escrow) in the ledger and returns the XRP to the sender.
+
+## Limitations
+
+Escrow is designed as a feature to enable XRP Ledger to be used in the [Interledger Protocol][] and with other smart contracts. The current version has a modest scope to avoid complexity.
+
+- Escrow only works with XRP, not issued currencies.
+- Escrow requires sending at least two transactions: one to create the escrow, and one to finish or cancel it. Thus, it is not financially sensible to escrow payments for very small amounts, since you have to destroy the [transaction cost](concept-transaction-cost.html) of two transactions.
```
jbheron: Nit pick: I don't think we should opine whether escrow is or isn't financial sensible, though it may not be. Also, I tend to think of 'since' as relating to time (from its primary definition), and opt for 'as' in this case; or, perhaps 'because' is really the most appropriate, here. How about:

> Thus, it may not be financially sensible to escrow payments for very small amounts, as you have to destroy the [transaction cost](concept-transaction-cost.html) of two transactions.

content/concept-escrow.md
```diff
@@ -0,0 +1,89 @@
...
+
+The XRP set aside in an escrow is locked up. No one can use or destroy the XRP until the escrow has been successfully finished or canceled. Before the expiration time, only the intended receiver can get the XRP. After the expiration time, the XRP can only be returned to the sender.
+
+## Usage
+
+<!--{# Diagram sources: https://docs.google.com/presentation/d/1C-_TLkkoQEH7KJ6Gjwa1gO6EX17SLiJ8lxvFcAl6Rxo/ #}-->
+
+[![Escrow Flow Diagram (Successful finish)](img/escrow-success-flow.png)](img/escrow-success-flow.png)
+
+**Step 1:** To send an escrow, the sender uses an [EscrowCreate transaction][] to set aside some XRP. This transaction defines a finish time, an expiration time, or both. The transaction may also define a crypto-condition that must be fulfilled to finish the escrow. This transaction must define an intended recipient for the XRP; the recipient _may_ be the same as the sender.
```
jbheron: Perhaps it's worth the parallel use of 'lock up' here, instead of 'set aside'? If so:

> To send an escrow, the sender uses an [EscrowCreate transaction][] to lock up some XRP.

jbheron (commented):
Overall, this is a great concept document. It's well presented and the images are gorgeous. 

I made some comments in-line, most nit picking language, but thought you might want to readdress the images to make them a bit more readable. I had to look at them pretty closely, but they might be fine as is, totally up to you.

In `escrow-success-flow.png`, I'd like to suggest these changes:

- For Step 1, the image well portrays the idea, but I'd center the EscrowCreate dotted box, to highlight it, even despite it not yet being an object.
- For Step 2, I'm really not sure what the solid arrow from the creator to the escrow object means. Looking at it quite closely, I'm thinking it indicates that the 100 XRP are locked up. If so, could you put a label on the line? Maybe, 'locked XRP', or '100 XRP', or 'escrowed amount'?
- For Step 3, would you consider moving the dotted box to the center, above the 'destroyed/vanished object' icon?

For `escrow-cancel-flow.png`, if you do change the images, this would be quite similar in scope:

- Step 2, label the flow line.
- Step 3a, looks good.
- Step 4, center the dotted box

jbheron (approved):

content/concept-escrow.md
```diff
@@ -0,0 +1,89 @@
...
+
+**Step 2:** After this transaction has been processed, the XRP Ledger has an [Escrow object](reference-ledger-format.html#escrow) that holds the escrowed XRP. This object tracks the properties of the escrow as defined by the transaction that created it. If this escrow has a finish time, no one can access the XRP before then.
```
scottschurr: Regarding "This object tracks...", I usually think of "tracking" as being an active operation.  What may be more accurate would be to say, "This object contains the escrow's properties as defined...".  The escrow object really just sits in the ledger, taking up reserve in the escrow owner's account, until the object is acted on by a transaction.  I leave to your discretion whether that would be a useful change.

content/concept-escrow.md
scottschurr: As counterpoint, passive voice may be appropriate here.  The passive voice emphasizes that the escrow object is just a record in the ledger that must be acted upon in order to change.  Not saying passive is required here, but that it may be appropriate.

content/concept-escrow.md
```diff
@@ -0,0 +1,89 @@
...
+
+All escrows start the same way, so **Steps 1 and 2** are the same as in the successful case.
+
+**Step 3a:** If the escrow has an expiration time, and it has not been successfully finished before then, the escrow is considered expired. It continues to exist in the XRP Ledger, but can no longer successfully finish. (Expired objects remain in the ledger until a transaction modifies them. Time-based triggers cannot change the ledger contents.)
+
+**Step 4a:** The sender, or any other XRP Ledger address, sends an [EscrowCancel transaction][] to cancel the expired escrow. This destroys the [Escrow object](reference-ledger-format.html#escrow) in the ledger and returns the XRP to the sender.
+
+## Limitations
+
+Escrow is designed as a feature to enable XRP Ledger to be used in the [Interledger Protocol][] and with other smart contracts. The current version has a modest scope to avoid complexity.
```
scottschurr: Suggest "enable _the_ XRP Ledger..."?

content/concept-escrow.md
scottschurr: The transaction cost issue is a little more interesting.  The "you" in this paragraph is likely (not required) to be two different entities.  One of the entities is always the escrow creator.  But the other entity is likely to be the escrow recipient or whoever computes the completion.  So, while there certainly is the cost of two transactions, that cost is split between the participants.

That level of detail may not be appropriate here.  But I thought I'd mention it...

scottschurr (approved):
Nicely written and clear.

For me, at least, the document feels a bit abstract.  It talks about what escrow can do, but not why someone would want to use it.

Another take, that I might have found more approachable, would be to start with the two use cases that motivated adding escrow to the ledger: XRP lockup and cryptoconditions.  Then walk through those two use cases as motivating examples.

Oh, and I also find it's easier to follow a document like this if the participants have names (the now legendary Alice and Bob) rather than letters.  But that may not fit within Ripple's documentation stylistic guidelines.

Just my thoughts after reading.  No action required.

Despite these notes, it's a nice document.  Thanks for writing it.

content/concept-escrow.md
mDuo13: Changed to:

> Thus, it may not be financially sensible to escrow payments for very small amounts, because the participants must destroy the [transaction cost](concept-transaction-cost.html) of the two transactions.

content/concept-escrow.md
mDuo13: > Wow, so, anyone could finish an escrow transaction, even if they're not the sender or recipient? That's wild.

Yep, it's a little surprising. On the other hand, there's not really any downside other than it being unexpected: if someone else finishes an escrow for you, they're just saving you the effort and the XRP equivalent of a postage stamp. In the case of a conditional escrow, it's unlikely that anyone other than the sender or recipient can actually fulfill the condition, anyway.

mDuo13:
Re: @jbheron 

> For Step 1, the image well portrays the idea, but I'd center the EscrowCreate dotted box, to highlight it, even despite it not yet being an object.

I tried doing this, but I think the result ends up being more confusing. It becomes less obvious that the EscrowCreate is a transaction associated with the sender, not an object.

>  For Step 2, I'm really not sure what the solid arrow from the creator to the escrow object means. Looking at it quite closely, I'm thinking it indicates that the 100 XRP are locked up. If so, could you put a label on the line? Maybe, 'locked XRP', or '100 XRP', or 'escrowed amount'?

Yes, it represents the flow of XRP. I've added labels in the corresponding steps (in both diagrams) to indicate as such.

>    For Step 3, would you consider moving the dotted box to the center, above the 'destroyed/vanished object' icon?

Tried this, but as with the Step 1 case, I think it ends up more confusing.

jbheron:
@mDuo13, thanks for updating the images, I think the labels really do the trick to make them more readable. Nice!

content/concept-escrow.md
```diff
@@ -28,10 +28,11 @@ All escrows start the same way, so **Steps 1 and 2** are the same as in the succ
...
 
-Escrow is designed as a feature to enable XRP Ledger to be used in the [Interledger Protocol][] and with other smart contracts. The current version has a modest scope to avoid complexity.
+Escrow is designed as a feature to enable the XRP Ledger to be used in the [Interledger Protocol][] and with other smart contracts. The current version has a modest scope to avoid complexity.
 
 - Escrow only works with XRP, not issued currencies.
-- Escrow requires sending at least two transactions: one to create the escrow, and one to finish or cancel it. Thus, it is not financially sensible to escrow payments for very small amounts, since you have to destroy the [transaction cost](concept-transaction-cost.html) of two transactions.
+- Escrow requires sending at least two transactions: one to create the escrow, and one to finish or cancel it. Thus, it may not be financially sensible to escrow payments for very small amounts, because the participants must destroy the [transaction cost](concept-transaction-cost.html) of the two transactions.
+    - When using Crypto-Conditions, the [cost of the transaction to finish the escrow](#escrowfinish-transaction-cost) is higher than usual.
```
scottschurr: :+1:

content/concept-escrow.md
```diff
@@ -10,11 +10,11 @@ The XRP set aside in an escrow is locked up. No one can use or destroy the XRP u
...
 
-**Step 3:** The recipient, or any other XRP Ledger address, sends an [EscrowFinish transaction][] to deliver the XRP. If the correct conditions are met, this destroys the Escrow object in the ledger and credits the XRP to the intended recipient. If the escrow has a crypto-condition, this transaction must include a fulfillment for that condition. If the escrow has an expiration time, this cannot happen after that time.
+**Step 3:** The recipient, or any other XRP Ledger address, sends an [EscrowFinish transaction][] to deliver the XRP. If the correct conditions are met, this destroys the Escrow object in the ledger and credits the XRP to the intended recipient. If the escrow has a crypto-condition, this transaction must include a fulfillment for that condition. If the escrow has an expiration time, the EscrowFinish transaction fails instead, with the code [`tecNO_PERMISSION`](reference-transaction-format.html#tec-codes).
```
scottschurr: I think this is almost right.  I think this is closer:

"If the escrow has an expiration time, and that expiration is passed, the EscrowFinish transaction instead fails with the `tecNO_PERMISSION` code."

content/concept-escrow.md
```diff
@@ -28,10 +28,11 @@ All escrows start the same way, so **Steps 1 and 2** are the same as in the succ
 
 ## Limitations
 
-Escrow is designed as a feature to enable XRP Ledger to be used in the [Interledger Protocol][] and with other smart contracts. The current version has a modest scope to avoid complexity.
+Escrow is designed as a feature to enable the XRP Ledger to be used in the [Interledger Protocol][] and with other smart contracts. The current version has a modest scope to avoid complexity.
 
 - Escrow only works with XRP, not issued currencies.
-- Escrow requires sending at least two transactions: one to create the escrow, and one to finish or cancel it. Thus, it is not financially sensible to escrow payments for very small amounts, since you have to destroy the [transaction cost](concept-transaction-cost.html) of two transactions.
+- Escrow requires sending at least two transactions: one to create the escrow, and one to finish or cancel it. Thus, it may not be financially sensible to escrow payments for very small amounts, because the participants must destroy the [transaction cost](concept-transaction-cost.html) of the two transactions.
```
scottschurr: :+1:

content/concept-escrow.md
```diff
@@ -65,6 +66,37 @@ If [Fee Voting](concept-fee-voting.html) changes the `reference_fee` value, the
...
+
+Placing the money into escrow does not directly protect Ripple's holdings from malicious actors, but it sharply reduces the amount of XRP that can be quickly stolen or redirected if a malicious actor gains temporary control over Ripple's XRP accounts. This reduces the risk of catastrophic losses of XRP and increases the time for Ripple to detect, prevent, and track down unintended uses of Ripple's XRP assets.
+
+### Use Case: Interledger Payments
+
+**Background:** In the quickly-developing world of financial technology, one of the core challenges is coordinating activities that cross multiple digital money systems, or ledgers. Many proposed solutions to this problem (including early views of the XRP Ledger!) can be reduced to creating "one ledger to rule them all." Ripple believes no one system can meet the needs of everyone in the world: in fact, some desirable features are mutually exclusive. Instead, Ripple believes that an interconnected network of ledgers—an _interledger_—is the true future of financial technology. The [Interledger Protocol][] defines standards for making as many systems as possible able to connect securely and smoothly.
```
scottschurr: Suggest "no one" -> "no single".  Otherwise looks great.

scottschurr (approved):
Nice! I think "Why Escrow" is a great addition.  Well done saying that there might be other possible uses, but identifying the exact current use cases.

PR #278, opened 2017-10-13
Add validator RPC updates

bachase:
Adds documentation for upcoming `validator_lists` and `validator_sites` RPC calls. See ripple/rippled#2242. I *think* the RPC formats have settled, but will update this PR if anything changes.


content/reference-rippled.md
```diff
@@ -10693,6 +10701,311 @@ The response follows the [standard format](#response-formatting), with a success
 * Any of the [universal error types](#universal-errors).
 * `badSeed` - The request provided an invalid secret value. This usually means that the secret value appears to be a valid string of a different format, such as an account address or validation public key.
 
+## validator_lists ##
+[[Source]<br>](https://github.com/ripple/rippled/blob/master/src/ripple/rpc/handlers/ValidatorLists.cpp "Source")
+
+The `validator_lists` command returns human readable information about the current list of published and trusted validators used by the server. [New in: rippled 0.90.0]
```
mDuo13: > `[New in: rippled 0.90.0]`

Please add an extra `[]` to the end there. (Markdown style / compatibility thing)

content/reference-rippled.md
```diff
@@ -10693,6 +10701,311 @@ The response follows the [standard format](#response-formatting), with a success
...
+| `validator_list_expires` | String | Either the human readable time when the current validator list will expire or the string `unknown` if the server has yet to load a published validator list. |
+
+Each member of the `publisher_lists` array is an object with the following fields:
+
+| `Field`            | Type             | Description                              |
+|:-------------------|:-----------------|:-----------------------------------------|
+| `available`        | Boolean          | If `false`, the validator keys in `list` may no longer be supported by this publisher. |
+| `expiration`       | String           |  Either the human readable time when this published list will expire or the string `unknown` if the server has yet to load a list for this publisher.      |
+| `list`             | Array            | Array of published validator keys.   |
+| `pubkey_publisher` | String           | Hex encoded public key of publisher. |
```
mDuo13: What type of key must publishers use? ECDSA? Are there other options?

content/reference-rippled.md
```diff
@@ -10693,6 +10701,311 @@ The response follows the [standard format](#response-formatting), with a success
...
+
+#### Possible Errors ####
+
+* Any of the [universal error types](#universal-errors).
+
+
+## validator_sites ##
+[[Source]<br>](https://github.com/ripple/rippled/blob/master/src/ripple/rpc/handlers/ValidatorSites.cpp "Source")
+
+The `validator_sites` command returns status information of sites serving validator lists. [New in: rippled 0.90.0]
```
mDuo13: Please change `[New in: rippled 0.90.0]` to `[New in: rippled 0.90.0][]` for greater Markdown compatibility

mDuo13 (changes requested):
Left a couple really small nitpicky comments, but largely, this looks great. Thanks so much for the work here!

When we have an official validator publisher live at its final URL, it would be great to replace the examples with versions using the production site, list, and quorum. Until then, it's fine as-is.

content/reference-rippled.md
bachase: It supports whatever the `PublicKey` class supports, which is currently secp25k1 or ed25519, but I think the external tooling for generating keys does ed25519 by default. What is the best designation for that?

bachase:
I think we are close to having the official list ready, so I'll push the updates after that. Do you prefer I squash down to a single commit?

mDuo13:
No need to squash commits

content/reference-rippled.md
mDuo13: I'd just change the description to something like this:

> Ed25519 or ECDSA public key of the list publisher, as hexadecimal.

mDuo13 (approved):

mDuo13:
merged as https://github.com/ripple/ripple-dev-portal/commit/df621af60cb7c9cb54121e9a0718c8480ed8917d

PR #279, opened 2017-10-21
Amendments ledger object & ledger format cleanup

mDuo13:
- Style changes to the "Ledger Format" reference
- Terminology changes. Uses "ledger object ID" instead of "ledger node index" to avoid ambiguity with ledger indexes and server nodes.
- Adds the missing `Amendments` object type

### DIFF
--- a/content/reference-ledger-format.md
+++ b/content/reference-ledger-format.md
@@ -182,7 +182,7 @@ Example `Amendments` object:
 
 Amendments are added to the `Amendments` table by an [EnableAmendment][] pseudo-transaction with no flags as the result of the [amendment process](concept-amendments.html#amendment-process).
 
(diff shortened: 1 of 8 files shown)


content/reference-ledger-format.md
```diff
@@ -1,4 +1,4 @@
-# The Ledger #
+# XRP Ledger Data Format
```
ryangyoung, line 1: _Much_ better title 😺

ryangyoung (approved):
👍

mDuo13:
Two changes in the latest commit:

1. @JoelKatz helpfully pointed out that I had missed the `Majorities` field entirely
2. Changed several other files that were calling things "ledger nodes" to call them "ledger objects" (or "ledger state objects" in some cases)

PR #283, opened 2017-11-15
Amendments changes for 0.80.0

mDuo13:
Adds descriptions of the new amendments added in 0.80.0 and updates the relevant docs in other places.


content/concept-amendments.md
```diff
@@ -290,6 +339,18 @@ Creates three new transaction types:[PaymentChannelCreate][], [PaymentChannelCla
...
 
+## SortedDirectories
+
+| Amendment ID                                                     | Status    |
+|:-----------------------------------------------------------------|:----------|
+| CC5ABAE4F3EC92E94A59B1908C2BE82D2228B6485C00AFF8F22DF930D89C194E | Enabled   |
+
+Sorts the entries in [DirectoryNode ledger objects](reference-ledger-format.html#directorynode) and fixes a bug that occasionally caused pages of owner directories not to be deleted when they should have been.
+
+**Warning:** Older versions of `rippled` that do not know about this amendment may crash when they encounter a DirectoryNode sorted by the new rules. To avoid this problem,
```
bachase: Is the rest of the sentence missing?

bachase (approved):
:+1: on fix1528

content/concept-amendments.md
```diff
@@ -209,6 +225,37 @@ Fixes a minor bug in transaction processing that causes failures when trying to
...
+
+With this amendment, the transactions fail with a more appropriate result code, `temBAD_AMOUNT`, instead.
+
+## fix1523
+
+| Amendment ID                                                     | Status    |
+|:-----------------------------------------------------------------|:----------|
+| B9E739B8296B4A1BB29BE990B17D66E21B62A300A909F25AC55C22D6C72E1F9D | Enabled   |
+
+Adds tracking by destination account to [escrows](concept-escrow.html). Without this amendment, pending escrows are only tracked by sender. This amendment makes it possible to look up pending escrows by the destination address using the [`account_objects` command](reference-rippled.html#account-objects), excluding any pending escrows that were created before this amendment became enabled. This amendment also makes [EscrowCreate transactions][] appear in the destination's transaction history, as viewed with the [`account_tx` command](reference-rippled.html#account-tx).
```
ryangyoung, line 244: Sounds like this means we should add a "Look up escrows by receiver" section to the Escrow tutorial? (Not for this PR, though.)

ryangyoung (approved):
LGTM. Just left one comment about how the escrow amendment probably affects the escrow tutorial, but that's outside the scope of this PR.

content/concept-amendments.md
mDuo13: Yep, the fact that we couldn't write that section originally is what inspired this amendment.

But that's a good reminder because I don't think we have a JIRA task on the backlog for said update.

ryangyoung: Created [DOC-1165](https://ripplelabs.atlassian.net/browse/DOC-1165) to track it.

intelliot (approved):
Thank you for documenting these! Good and complete documentation is crucial for new folks like me :)

content/concept-amendments.md
```diff
@@ -191,6 +196,17 @@ A transaction remains in the queue until one of the following happens:
...
 
+## fix1201
+
+| Amendment ID                                                     | Status    |
+|:-----------------------------------------------------------------|:----------|
+| B4D44CC3111ADD964E846FC57760C8B50FFCD5A82C86A72756F6B058DDDF96AD | Enabled   |
+
+Correctly implements a limit on [transfer fees](concept-transfer-fees.html) to a 100% fee, represented by a maximum `TransferRate` value of `2000000000`. (A 100% fee in this case means you must send 2 units of the issued currency for every 1 unit you want to deliver.) Without the amendment, the effective limit is a `TransferRate` value of 2<sup>32</sup>-1, for approximately a 329% fee.
+
+With this amendment enabled, an [AccountSet][] transaction that attempts to set `TransferRate` higher than `2000000000` fails with the result code `temBAD_TRANSFER_RATE`. Any existing `TransferRate` already set to a value higher than the limit is treated as equal to the limit instead.
```
scottschurr: Doing my best to stand in for Niraj.  I'm not seeing any evidence in the code base for the second sentence.  To me it looks like any previously existing very high transfer rate will continue to work as it did.  This is the only place that I see `fix1201` being checked in the code base (other than tests): https://github.com/ripple/rippled/blob/develop/src/ripple/app/tx/impl/SetAccount.cpp#L126.  So a new very high `TransferRate` will be rejected.  An old very high TransferRate will continue to exhibit the old behavior.

content/concept-amendments.md
```diff
@@ -209,6 +225,37 @@ Fixes a minor bug in transaction processing that causes failures when trying to
...
+
+## fix1523
+
+| Amendment ID                                                     | Status    |
+|:-----------------------------------------------------------------|:----------|
+| B9E739B8296B4A1BB29BE990B17D66E21B62A300A909F25AC55C22D6C72E1F9D | Enabled   |
+
+Adds tracking by destination account to [escrows](concept-escrow.html). Without this amendment, pending escrows are only tracked by sender. This amendment makes it possible to look up pending escrows by the destination address using the [`account_objects` command](reference-rippled.html#account-objects), excluding any pending escrows that were created before this amendment became enabled. This amendment also makes [EscrowCreate transactions][] appear in the destination's transaction history, as viewed with the [`account_tx` command](reference-rippled.html#account-tx).
+
+With this amendment, new escrows are added to the [owner directories](reference-ledger-format.html#directorynode) of both the sender and receiver. This amendment also adds a new `DestinationNode` field to [Escrow ledger objects](reference-ledger-format.html#escrow), containing the address of the destination of the escrow.
```
scottschurr: As someone noted earlier (I think you, Rome), the content of the `DestinationNode` is not the address of the destination.  It  contains the destination directory page that was used to store the Escrow reference.

content/concept-amendments.md
```diff
@@ -270,12 +317,14 @@ An address with a SignerList can disable the master key even if a regular key is
...
-| 9178256A980A86CF3D70D0260A7DA6402AAFE43632FDBCB88037978404188871 | Released but not enabled |
+| 9178256A980A86CF3D70D0260A7DA6402AAFE43632FDBCB88037978404188871 | In development |
 
 Fixes an inconsistency in the way [transfer fees](concept-transfer-fees.html) are calculated between [OfferCreate](reference-transaction-format.html#offercreate) and [Payment](reference-transaction-format.html#payment) transaction types. Without this amendment, the holder of the issuances pays the transfer fee if an offer is executed in offer placement, but the initial sender of a transaction pays the transfer fees for offers that are executed as part of payment processing. With this amendment, the holder of the issuances always pays the transfer fee, regardless of whether the offer is executed as part of a Payment or an OfferCreate transaction. Offer processing outside of payments is unaffected.
 
 This Amendment requires the [Flow Amendment](#flow) to be enabled.
 
+**Note:** An incomplete version of this amendment was introduced in v0.33.0 and removed in v0.80.0. (It was never enabled.) Ripple plans to re-add the amendment when the code is stable enough.
```
scottschurr, line 326: FWIW, to the best of my knowledge there is no active development on the OwnerPaysFee amendment.  Or any identified bugs.  `FlowCross` (also not enabled) relies on the OwnerPaysFee code, and I've not seen any bugs in the OwnerPaysFee stuff (bugs have been found (and fixed) in FlowCross).

As to why OwnerPaysFee is not enabled on the network I'm unclear.  I think it's good idea to have offer crossing and payments exhibit the same behavior.  But my vote doesn't count.

content/concept-transfer-fees.md
```diff
@@ -27,21 +27,23 @@ In this scenario, Salazar (the sender) holds EUR issued by ACME, and wants to de
...
+The transfer fee is represented by a setting on the [issuing address][]. The transfer fee has a maximum precision of 9 digits, and cannot be less than 0% or more than 100%. The TransferRate setting applies to all currencies issued by the same account. If you want to have different transfer fee percentages for different currencies, use different [issuing addresses][issuing address] for each currency.
+
+**Note:** The [fix1201 amendment](concept-amendments.html), introduced in `rippled` v0.80.0 and enabled on 2017-11-14, lowered the maximum transfer fee to 100% from an effective limit of approximately 329% (based on the maximum size of a 32-bit integer). The ledger may still contain accounts with a transfer fee setting higher than 100% (a `TransferRate` of `2000000000`), but all higher values are treated as 100% during transaction processing with the fix1201 amendment enabled.
```
scottschurr: See earlier note.  I think you'll need to fix the last sentence.

scottschurr (approved):
Looks good.  I noted a few places you may want to tidy up.  Thanks for doing this!

content/concept-amendments.md
mDuo13: Good catch. I fixed that in the ledger format table but not here.

content/concept-amendments.md
mDuo13: aww man, old grandfathered high fees still work?

I'm starting to want to put together a scavenger hunt for ledger objects in the production ledger that couldn't be created under the current rules...

content/concept-amendments.md
mDuo13: It's baffling to me, then, that [OwnerPaysFee is commented out and FlowCross isn't](https://github.com/ripple/rippled/blob/0.80.0/src/ripple/app/main/Amendments.cpp#L48-L58) in rippled v0.80.0. Might want to look into that?

scottschurr: That's fair.  The amendments that are commented out are the ones that we don't run (by default) against our unit tests.  We want to run the unit tests against code that we think is likely to be on the MainNet soon.  I'm personally hoping that `FlowCross` will make the cut sometime soon.  But I'm not in control of that.  As I recall, `MultiSign` was sitting, ready, in the code base for about a year before we finally had the grit to enable it.  In the 0.90.0 release you'll see `Tickets` and `SHAMapV2` are also commented out since `SHAMapV2` is not moving fast and the implementation for `Tickets` needs work.

At any rate, `OwnerPaysFee` was specifically commented out because Mike made a great change so unit tests automatically pick up the latest amendments.  When he did that a bunch of unit tests failed because `OwnerPaysFee` was enabled.  Nothing terrible happened, just the expected balances didn't turn out the way they used to.  So the simplest thing to do was remove `OwnerPaysFee` from being enabled by default in the unit tests.  If the product folks decide that the time is right for `OwnerPaysFee`, then the first thing to do will be fixing up the unit tests.

content/concept-amendments.md
scottschurr: It would be really cool to make a list like that and let the data folks loose on it to see if any of those objects are actually hanging around in the MainNet.  But we'd have to be careful about removing support altogether since we still need to be able to replay old ledgers.

PR #285, opened 2017-11-29
Fixed example for account_tx

jhaaaa:
https://ripplelabs.atlassian.net/browse/DOC-1160

Updated request examples for account_tx as requested. 

Updated note to explicitly state that if a deprecated field is used in a request, pagination won't work. 

Question: The JSON-RPC example includes the offset field - should I remove it as well?

### DIFF
--- a/content/reference-rippled.md
+++ b/content/reference-rippled.md
@@ -14,9 +14,11 @@ If you want to communicate directly with a `rippled` server, you can use either 
 
 * The WebSocket API uses the [WebSocket protocol](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API), available in most browsers and Javascript implementations, to achieve persistent two-way communication. There is not a 1:1 correlation between requests and responses. Some requests prompt the server to send multiple messages back asynchronously; other times, responses may arrive in a different order than the requests that prompted them. The `rippled` server can be configured to accept secured (wss:), unsecured (ws:) WebSocket connections, or both.
 * The JSON-RPC API relies on request-response communication via HTTP or HTTPS. (The `rippled` server can be configured to accept HTTP, HTTPS, or both.) For commands that prompt multiple responses, you can provide a callback URL.
-* The `rippled` program can also be used as a quick commandline client to make JSON-RPC requests to a running `rippled` server. This is only intended for administrative purposes, and is not a supported API.
 
 In general, we recommend using WebSocket, because WebSocket's push paradigm has less latency and less network overhead. WebSocket is also more reliable; you can worry less about missing messages and establishing multiple connections. On the other hand, there is widespread support for JSON-RPC because you can use a standard HTTP library to connect to `rippled`'s JSON-RPC API.
+
+**Note:** The `rippled` program can also be used as a quick commandline client to make JSON-RPC requests to a running `rippled` server. The commandline interface is intended for administrative purposes only and is _not a supported API_.
+
 
 ## Changes to the APIs ##
 
@@ -98,6 +100,9 @@ The commandline interface connects to the same service as the JSON-RPC one, so t
 ```
 rippled --conf=/etc/rippled.cfg server_info
 ```
+
+**Note:** The commandline interface is intended for administrative purposes only and is _not a supported API_.
+
 
 #### Request Formatting ####
 
@@ -590,6 +595,7 @@ You can use the `rippled` application (as a separate instance) as a JSON-RPC cli
 
 * [`json` - Pass JSON through the commandline](#json)
 
+**Note:** The commandline interface is intended for administrative purposes only and is _not a supported API_.
 
 
 # Account Information #
@@ -2149,7 +2155,6 @@ An example of the request format:
   "ledger_index_min": -1,
   "ledger_index_max": -1,
   "binary": false,
-  "count": false,
   "limit": 2,
   "forward": false
 }
@@ -2164,13 +2169,10 @@ An example of the request format:
         {
             "account": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",
             "binary": false,
-            "count": false,
-            "descending": false,
             "forward": false,
             "ledger_index_max": -1,
             "ledger_index_min": -1,
-            "limit": 2,
-            "offset": 1
+            "limit": 2
         }
     ]
 }
@@ -2179,8 +2181,8 @@ An example of the request format:
 *Commandline*
 
 ```
-#Syntax account_tx account [ledger_index_min [ledger_index_max [limit]]] [binary] [count] [forward]
-rippled account_tx r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59 -1 -1 2 false false false
+#Syntax account_tx account ledger_index_min ledger_index_max [offset] [limit] [binary] [count] [forward]
+rippled -- account_tx r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59 -1 -1 2 5 1 0 1
 ```
 
 <!-- MULTICODE_BLOCK_END -->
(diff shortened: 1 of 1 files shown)


mDuo13:
> Question: The JSON-RPC example includes the offset field - should I remove it as well?

Yes, that's a good idea.

content/reference-rippled.md
```diff
@@ -2199,8 +2197,9 @@ The request includes the following parameters:
...
 | `marker`           | [(Not Specified)](#markers-and-pagination) | Value from a previous paginated response. Resume retrieving data where that response left off. This value is stable even if there is a change in the server's range of available ledgers. |
 
-[[Source]<br>](https://github.com/ripple/rippled/blob/master/src/ripple/rpc/handlers/AccountTxSwitch.cpp "Source")
-There is also a deprecated legacy variation of the `account_tx` method. For that reason, we recommend *not using any of the following fields*: `offset`, `count`, `descending`, `ledger_max`, `ledger_min`.
+[[Source]<br>](https://github.com/ripple/rippled/blob/master/src/ripple/rpc/handlers/AccountTxSwitch.cpp "Source")<br>
+
+**Note:** There is also a deprecated legacy variation of the `account_tx` method. For that reason, we recommend *not using any of the following fields*: `offset`, `count`, `descending`, `ledger_max`, and `ledger_min`. If you include any of these deprecated fields in the request, the method will not support pagination.
```
mDuo13: Nitpick: avoid future tense here. Suggest you change "will not support" to "does not support".

content/reference-rippled.md
```diff
@@ -2177,7 +2175,7 @@ An example of the request format:
 *Commandline*
 
 ```
-#Syntax account_tx account [ledger_index_min [ledger_index_max [limit]]] [binary] [count] [forward]
+#Syntax account_tx account [ledger_index_min [ledger_index_max [limit]]] [binary] [forward]
```
mDuo13: I think the commandline syntax still has the `count` parameter, and because the arguments are positional, if you leave it off, it'll interpret the next value as "count" even if you meant to to be "forward". So you may have to revert this line.

content/reference-rippled.md
```diff
@@ -2177,7 +2175,7 @@ An example of the request format:
...
 rippled account_tx r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59 -1 -1 2 false false false
```
mDuo13: If you don't revert the change from the previous line, then you need to change this to have only two "false"s.

mDuo13 (commented):
I'm not sure if the example responses change as a result of the changes to the example requests. It's probably not a bad chance to update the example responses in case other things have changed anyway.

I think we might need a second look from an engineer to confirm whether the docs on the commandline syntax should/shouldn't change.

content/reference-rippled.md
jhaaaa: Thank you! I see your point. All I did was edit a comment - not the actual command. The second "false" represents the count command -- if it still exists for command line. Added count back. 

@nbougalis - any thoughts on Rome's comment here? Does the command line syntax still include the count parameter?

content/reference-rippled.md
jhaaaa: Reverted the change. Checking with Nik on whether count still exists for command line.

content/reference-rippled.md
jhaaaa: I appreciate all nitpicks. Crucial to keeping things in order! Fixed. I saw it as I was committing and should have fixed it. Embarassed 💯 . Hehe. Thank you...

jhaaaa:
Nik says that the command line hasn't changed. Leaving the request and response examples as-is for now.

@mDuo13: Regarding updating the response examples:

- *For Websocket*, the response has changed EVER SO SLIGHTLY. I removed the `count` parameter from the existing request example and made the call. The only difference I saw is that the `offset` parameter is not included in the response. I'll go ahead and update the response code sample to reflect that. (The `offset` parameter is not in the existing request example.)

- *For JSON-RPC*, I removed the `count` and `offset` parameters from the existing request example and made the call. The `offset` parameter and all else continue to be included in the response. So, there is no change needed for the JSON-RPC response example, but it seems like the JSON-RPC response should exclude the `offset` parameter. I believe the `offset` value in the response, is just the `offset` value sent in the request. If `offset` is never included in the request, is seems that the `offset` value in the response will always be `0`. I can explain this in the doc for now, but I think this may be request for dev - what do you think?

If you'd like to take a look, here are the responses [on Google Drive](https://drive.google.com/open?id=1hy5Ltrp9fHQhhBIML0Xnf22OR6uXiRn0). To call out the difference in the Websocket responses, I inserted a # for your reference. But, of course, the # won't be in the published version.

content/reference-rippled.md
```diff
@@ -2698,7 +2695,7 @@ The response follows the [standard format](#response-formatting), with a success
 | `ledger_index_max` | Integer                                    | The sequence number of the most recent ledger actually searched for transactions. |
 | `limit`            | Integer                                    | The `limit` value used in the request. (This may differ from the actual limit value enforced by the server.) |
 | `marker`           | [(Not Specified)](#markers-and-pagination) | Server-defined value indicating the response is paginated. Pass this to the next call to resume where this call left off. |
-| `offset`           | Integer                                    | The `offset` value used in the request. |
+| `offset`           | Integer                                    | Included only in the JSON-RPC response. `offset` is a field from a deprecated legacy variation of the `account_tx` method. We recommend that you do not include the deprecated `offset` field in your request. When you don't include `offset` in your request, the `offset` value is returned in the response, but is always set to `0`. If you include an `offset` value in your request, it is returned as the `offset` value in the response. |
```
jhaaaa: @mDuo13 - This is absolutely awkward, but it is because I'm explaining some awkward functionality. Sure, blame the functionality... If `offset` is not sent in the request, it should not be returned in the response, but it is. 

How about this less verbose, but slightly more mysterious version?

> This legacy field is included in the JSON-RPC response only. It is always set to `0`.

content/reference-rippled.md
```diff
@@ -2199,8 +2195,9 @@ The request includes the following parameters:
...
 | `marker`           | [(Not Specified)](#markers-and-pagination) | Value from a previous paginated response. Resume retrieving data where that response left off. This value is stable even if there is a change in the server's range of available ledgers. |
 
-[[Source]<br>](https://github.com/ripple/rippled/blob/master/src/ripple/rpc/handlers/AccountTxSwitch.cpp "Source")
-There is also a deprecated legacy variation of the `account_tx` method. For that reason, we recommend *not using any of the following fields*: `offset`, `count`, `descending`, `ledger_max`, `ledger_min`.
+[[Source]<br>](https://github.com/ripple/rippled/blob/master/src/ripple/rpc/handlers/AccountTxSwitch.cpp "Source")<br>
+
+**Note:** There is also a deprecated legacy variation of the `account_tx` method. For that reason, we recommend *not using any of the following fields*: `offset`, `count`, `descending`, `ledger_max`, and `ledger_min`, with the exception of `count` via command line. If you include any of these deprecated fields, the method does not support pagination.
```
ryangyoung: > we recommend

To be clear about who "we" is, let's change that to "Ripple recommends..."

content/reference-rippled.md
```diff
@@ -14,10 +14,12 @@ If you want to communicate directly with a `rippled` server, you can use either
...
 * The JSON-RPC API relies on request-response communication via HTTP or HTTPS. (The `rippled` server can be configured to accept HTTP, HTTPS, or both.) For commands that prompt multiple responses, you can provide a callback URL.
-* The `rippled` program can also be used as a quick commandline client to make JSON-RPC requests to a running `rippled` server. This is only intended for administrative purposes, and is not a supported API.
 
 In general, we recommend using WebSocket, because WebSocket's push paradigm has less latency and less network overhead. WebSocket is also more reliable; you can worry less about missing messages and establishing multiple connections. On the other hand, there is widespread support for JSON-RPC because you can use a standard HTTP library to connect to `rippled`'s JSON-RPC API.
 
+**Note:** The `rippled` program can also be used as a quick commandline client to make JSON-RPC requests to a running `rippled` server. The commandline interface is intended for administrative purposes only and is _not a supported API_.
```
jhaaaa, line 20: Moved this text out of a discussion that is meant to focus on WebSocket and JSON-RPC and put into a callout.

content/reference-rippled.md
```diff
@@ -99,6 +101,9 @@ The commandline interface connects to the same service as the JSON-RPC one, so t
 rippled --conf=/etc/rippled.cfg server_info
 ```
 
+**Note:** The commandline interface is intended for administrative purposes only and is _not a supported API_.
+
+
```
jhaaaa, line 106: Added this important callout to this Commandline section.

content/reference-rippled.md
```diff
@@ -588,6 +593,7 @@ You can use the `rippled` application (as a separate instance) as a JSON-RPC cli
 
 * [`json` - Pass JSON through the commandline](#json)
 
+**Note:** The commandline interface is intended for administrative purposes only and is _not a supported API_.
```
jhaaaa, line 598: Added this important callout to this Commandline section.

content/reference-rippled.md
```diff
@@ -2147,7 +2153,6 @@ An example of the request format:
   "ledger_index_min": -1,
   "ledger_index_max": -1,
   "binary": false,
-  "count": false,
```
jhaaaa, line 2152: removed deprecated field

content/reference-rippled.md
```diff
@@ -2162,13 +2167,10 @@ An example of the request format:
...
             "binary": false,
-            "count": false,
-            "descending": false,
             "forward": false,
             "ledger_index_max": -1,
             "ledger_index_min": -1,
-            "limit": 2,
-            "offset": 1
+            "limit": 2
         }
```
jhaaaa, line 2176: removed three deprecated fields

content/reference-rippled.md
```diff
@@ -2177,8 +2179,8 @@ An example of the request format:
 *Commandline*
 
 ```
-#Syntax account_tx account [ledger_index_min [ledger_index_max [limit]]] [binary] [count] [forward]
-rippled account_tx r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59 -1 -1 2 false false false
+#Syntax account_tx account ledger_index_min ledger_index_max [offset] [limit] [binary] [count] [forward]
+rippled -- account_tx r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59 -1 -1 2 5 1 0 1
```
jhaaaa, line 2185: Commandline still requires `offset`. Added it back to syntax and commandline example. Added `--` to example to ensure that API treats `-_x_` as a field value, not a flag. Changed Boolean `true`/`false` values to `1`/`0` values as commandline is not accepting true/false. I do not explicitly tell the user that for commandline Boolean only you must use `1` or `0` instead of `true` or `false`. I feel like I should -- but the doc is already a bit complex -- should I further complicate it to explain a subtlety of the unsupported commandline interface? I'm open.

content/reference-rippled.md
```diff
@@ -2190,17 +2192,18 @@ The request includes the following parameters:
...
+| `ledger_index_max` | Integer                                    | Use to specify the most recent ledger to include transactions from. A value of `-1` instructs the server to use the most recent validated ledger version available. This field is optional if you use `ledger_index_min`, however you can use these fields in tandem to define a range of ledgers. This field is optional and unnecessary if you use `ledger_hash` or `ledger_index`. |
+| `ledger_hash`      | String                                     | Use to look for transactions from a single ledger only. This field is optional and unnecessary if you use `ledger_index_min`, `ledger_index_max`, or `ledger_index`. (See [Specifying a Ledger](#specifying-ledgers).) |
+| `ledger_index`     | String or Unsigned Integer                 | Use to look for transactions from a single ledger only. This field is optional and unnecessary if you use `ledger_index_min`, `ledger_index_max`, or `ledger_hash`. (See [Specifying a Ledger](#specifying-ledgers).) |
+| `binary`           | Boolean                                    | _(Optional)_ Defaults to `false`. If set to `true`, returns transactions as hex strings instead of JSON. |
```
jhaaaa: Changed lowercased `False` and `True`. WebSocket and JSON-RPC do not accept init capped Boolean values.

content/reference-rippled.md
```diff
@@ -2190,17 +2192,18 @@ The request includes the following parameters:
...
 | `marker`           | [(Not Specified)](#markers-and-pagination) | Value from a previous paginated response. Resume retrieving data where that response left off. This value is stable even if there is a change in the server's range of available ledgers. |
 
-[[Source]<br>](https://github.com/ripple/rippled/blob/master/src/ripple/rpc/handlers/AccountTxSwitch.cpp "Source")
-There is also a deprecated legacy variation of the `account_tx` method. For that reason, we recommend *not using any of the following fields*: `offset`, `count`, `descending`, `ledger_max`, `ledger_min`.
+[[Source]<br>](https://github.com/ripple/rippled/blob/master/src/ripple/rpc/handlers/AccountTxSwitch.cpp "Source")<br>
+
+**Note:** For WebSocket and JSON-RPC, there is also a deprecated legacy variation of the `account_tx` method. For that reason, Ripple recommends *not using any of the following fields*: `offset`, `count`, `descending`, `ledger_max`, and `ledger_min`. If you use any of these deprecated fields, the method does not support pagination.
```
jhaaaa: Updated note text to make it clear that if you use deprecated fields, pagination isn't supported. Updated note text to be WebSocket- and JSON-RPC-specific because the commandline still requires use of some deprecated fields.

content/reference-rippled.md
```diff
@@ -2190,17 +2192,18 @@ The request includes the following parameters:
...
+| `ledger_index_max` | Integer                                    | Use to specify the most recent ledger to include transactions from. A value of `-1` instructs the server to use the most recent validated ledger version available. This field is optional if you use `ledger_index_min`, however you can use these fields in tandem to define a range of ledgers. This field is optional and unnecessary if you use `ledger_hash` or `ledger_index`. |
+| `ledger_hash`      | String                                     | Use to look for transactions from a single ledger only. This field is optional and unnecessary if you use `ledger_index_min`, `ledger_index_max`, or `ledger_index`. (See [Specifying a Ledger](#specifying-ledgers).) |
+| `ledger_index`     | String or Unsigned Integer                 | Use to look for transactions from a single ledger only. This field is optional and unnecessary if you use `ledger_index_min`, `ledger_index_max`, or `ledger_hash`. (See [Specifying a Ledger](#specifying-ledgers).) |
```
jhaaaa: Totally open to feedback on clarity here. `ledger_index_min`, `ledger_index_max`, `ledger_hash`, and `ledger_index` are all ... for lack of a better phrase ... conditionally required/optional. Per feedback from Ryan, instead of making them all optional, I made them all required, so that users don't gloss over them. I then explain, for each field, the cases in which it becomes optional. And in some cases, when it becomes not only optional but unnecessary.

jhaaaa:
Updated the PR. @mDuo13 - I'd greatly appreciate your feedback. Again, not the most elegant language, but I believe it accurately describes the behavior of the API. See my comments inline. Thank you!

@ryangyoung @jbheron - totally optional, but if you are interested, please feel free to review and comment.

content/reference-rippled.md
mDuo13: I think it's fine not to add extra confusion for the sake of explaining the unsupported commandline interface.

content/reference-rippled.md
mDuo13: A simpler way to describe this case may be to simply mark all of the *ledger_\** fields as _(Optional)_ and then add a line after the parameter table to the effect of, "You must specify at least one of `ledger_index`, `ledger_hash`, `ledger_index_min`, or `ledger_index_max`."

mDuo13 (approved):
Looks good. I left one optional suggestion for one of the tougher segments to describe, but I leave that change to your discretion. The rest of the work looks like an excellent improvement over the existing docs. Thanks!

mDuo13:
For the record, if you rebase onto the latest master branch, that should fix the broken link problem (which doesn't originate in this PR). You could also take the opportunity to combine some of the in-progress commits here to just one with all your combined edits.

mDuo13:
!rebuild

mDuo13:
!rebuild

PR #287, opened 2017-12-01
updated escrow tut & samples - look up by receiver

jhaaaa:
https://ripplelabs.atlassian.net/browse/DOC-1165

- Updated Escrow Tutorial to include coverage of how to look up escrows by receiver. 
- Updated code samples. Single file illustrates request for sender and receiver. Single file illustrates response for sender and receiver.

Questions:
- Who would be a good engineering person to include in the review? Nik?
- In "Look up escrows by receiver," I may have gone overboard in describing when the receiver lookup functionality became available. :-) Do you have a preferred method for conveying this kind of info?
- The existing tutorial doc says, "You can look up escrow nodes **owned by your address** ..." -- how does this work? I may just not understand the depths of the tech yet, but it seems that the request to look up by destination does not include a value that identifies the owner of the escrow.
- What is the difference between an "escrow" and an "escrow node"?
- Are "account" and "address" synonymous?


content/tutorial-escrow.md
```diff
@@ -5,7 +5,7 @@ The XRP Ledger supports held payments, or _escrows_, that can be executed only a
 - [Send a time-held escrow](#send-a-time-held-escrow)
 - [Send a conditionally-held escrow](#send-a-conditionally-held-escrow)
 - [Cancel an expired escrow](#cancel-an-expired-escrow)
-- [Look up escrows by sender](#look-up-escrows-by-sender)
+- [Look up escrows](#look-up-escrows)
 <!-- {# Doesn't work yet:- Look up escrows by destination #}-->
```
mDuo13: You can remove this comment line now

mDuo13:
Taking them one by one...

> -  Who would be a good engineering person to include in the review? Nik?

@nbougalis has some fingerprints on the escrow stuff, so he's not a bad choice. But he's also a pretty busy guy, so maybe he wants to suggest another rippled team member who can do the tech review?

> - In "Look up escrows by receiver," I may have gone overboard in describing when the receiver lookup functionality became available. :-) Do you have a preferred method for conveying this kind of info?

The way you handled it is just fine. We don't have a set format for conveying this information, but [here](https://ripple.com/build/how-to-multi-sign/#availability-of-multi-signing) [are](https://ripple.com/build/ledger-format/#escrow) [some](https://ripple.com/build/transactions/#signerlistset) examples:

> -  The existing tutorial doc says, "You can look up escrow nodes owned by your address ..." -- how does this work? I may just not understand the depths of the tech yet, but it seems that the request to look up by destination does not include a value that identifies the owner of the escrow.

The "sender" is the "owner" of the escrow object in the ledger while the escrow is pending. A more elegant phrasing would be, "You can look up escrow objects by the sender's address."

> -  What is the difference between an "escrow" and an "escrow node"?

Not much of one. We should remove the term "escrow node"—the updated/preferred phrase is "escrow object in the ledger" (with the "in the ledger" part being optional based on contextual needs).

> -  Are "account" and "address" synonymous?

It's a long story. The short answer is no, but we use the terms inconsistently because "account" is a very overloaded term. The address is the unique identifier of an account object in the ledger, so in an act of [synecdoche](https://en.wikipedia.org/wiki/Synecdoche) we sometimes use "address" to refer to the whole thing. A tricky exception is that there can also be addresses which don't have accounts in the ledger (yet).

content/tutorial-escrow.md
```diff
@@ -539,6 +549,36 @@ _Websocket_
 
 <!-- MULTICODE_BLOCK_END -->
 
-**Tip:** If you don't know what `OfferSequence` to use in the [EscrowFinish transaction][] to execute an escrow, use the [`tx` method](reference-rippled.html) to look up the transaction that created the escrow, using the identifying hash of the transaction in the Escrow's `PreviousTxnID` field. Use the `Sequence` value of that transaction as the `OfferSequence` value when finishing the escrow.
+###Look up escrows by receiver
+
+You can use the [`account_objects`](reference-rippled.html#account-objects) method to look up escrows by receiver address.
+
+Look up by receiver is possible for only pending escrows created after [fix1523](concept-amendments.html#fix1523) was enabled for rippled 0.80.0 on 2017-11-14.
```
mDuo13: Change this from passive to active voice:

> You can only look up pending escrows by receiver if those escrows were created after the [fix1523 amendment](concept-amendments.html#fix1523) was enabled on 2017-11-14.

You may also note in the above suggested revision: it's _slightly_ inaccurate to say that the amendment was enabled for `rippled` 0.80.0. A more precise explanation is that it was enabled for the XRP Ledger. `rippled` is the server that manages the XRP ledger, and technically the amendment applies to the whole ledger regardless of which server implementation is running it and what software version they're on. (`rippled` is the only server implementation for the XRP Ledger, but it's theoretically possible for there to be others, and at one point we even aspired for there to be other implementations.)

content/tutorial-escrow.md
```diff
@@ -511,9 +511,19 @@ _Websocket_
...
+All pending escrows are stored in the ledger as [Escrow objects](reference-ledger-format.html#escrow).
+
+You can look up escrow nodes by [sender](#look-up-escrows-by-sender) or [receiver](#look-up-escrows-by-receiver) using the [`account_objects`](reference-rippled.html#account-objects) method.
 
-All pending escrows are stored in the ledger as [Escrow objects](reference-ledger-format.html#escrow). You can look up escrow nodes owned by your address using the [`account_objects` method](reference-rippled.html#account-objects).
+###Look up escrows by sender
+
+You can use the [`account_objects`](reference-rippled.html#account-objects) method to look up escrows by sender.
+
+The following examples illustrate an escrow lookup by sender. Note that the `account` in the request is the sender `Account` of the escrow `account_objects` returned in the response.
```
mDuo13: This may warrant a little more clarification. I think that the `account_objects` list (now) contains all escrows sent to or from an address, so you may get a bunch sent _by_ an address if you try to look up ones being sent _to_ that address.

The examples you have don't have a mix because you set them up that way, but in the wild, it's entirely possible for that to occur. Unfortunately, I don't think the API method has an easy way to filter, so you'll just have to tell people to check whether their address is the `Account` (sender) or `Destination` (recipient), or both, for each escrow in the response, when reading a response from `account_objects`.

mDuo13 (changes requested):
Looks very good... except unfortunately, I think that there's no simple API way to distinguish between incoming/outgoing escrows now, so you'll have to add a note telling the person to check the `Account` and `Destination` fields of the escrow objects to figure out which direction the escrow is going.

content/tutorial-escrow.md
jhaaaa: Done! Thank you for the clarification of rippled vs the XRP Ledger. Fantastic. 🥇

content/tutorial-escrow.md
jhaaaa: Done. 👍

content/tutorial-escrow.md
jhaaaa: Yeppppp... got it. Thank you for taking the time to call this out. 💯  I oversimplified the samples. Indeed, in the wild it would/could not be as neat. 

I updated the samples to show a mix and have provided updated text in both the sender and receiver sections that I hope provides the necessary clarifications. 

I'm using the same samples in both sections. Likely the reader will not notice, but I'm hoping that it reinforces the use case of the "mixture" being returned. 

I'll be sure to delete the new code samples I added. I guess it wouldn't hurt to rename the code samples to not be sender specific.

content/tutorial-escrow.md
```diff
@@ -537,6 +536,11 @@ _Websocket_
 
 <!-- MULTICODE_BLOCK_END -->
 
+
+The response will look something like the following example. Note that the response includes all pending escrow objects with `rfztBskAVszuS3s5Kq7zDS74QtHrw893fm` as the sender or receiver address, where the sender address is the `Account` value and the receiver address is the `Destination` value.
```
mDuo13: Avoid future tense. Suggested:

> The response resembles the following example.

content/tutorial-escrow.md
```diff
@@ -537,6 +536,11 @@ _Websocket_
...
+
+In this case, the second and fourth escrow objects have `rfztBskAVszuS3s5Kq7zDS74QtHrw893fm` as their sender address (`Account`) value and are what we are looking for.
```
mDuo13: Style nitpick: we avoid phrases like "what we are looking for". Suggested rephrase:

> In this example, `rfztBskAVszuS3s5Kq7zDS74QtHrw893fm` is the the `Account` (sender) address in the second and fourth escrow objects.

(To be honest, my phrasing here still feels awkward, so feel free to tweak some more.)

content/tutorial-escrow.md
```diff
@@ -564,19 +568,24 @@ Request:
...
-{% include 'code_samples/escrow/websocket/account_objects-request-receiver.json' %}
+{% include 'code_samples/escrow/websocket/account_objects-request-sender.json' %}
 ```
 
 <!-- MULTICODE_BLOCK_END -->
 
+
+The response will look something like the following example. Note that the response includes all pending escrow objects with `rfztBskAVszuS3s5Kq7zDS74QtHrw893fm` as the receiver or sender address, where the receiver address is the `Destination` value and the sender address is the `Account` value.
+
+In this case, the first and third escrow objects have `rfztBskAVszuS3s5Kq7zDS74QtHrw893fm` as their receiver address (`Destination`) value and are what we are looking for.
```
mDuo13: Another case where I'd prefer you replace "and are what we are looking for" (or make it a stand-alone sentence just for ease of reading).

mDuo13 (approved):
I left a few minor style edits that I'd like to see implemented, but substantially, this looks good now, so you have my approval. Thanks!

content/tutorial-escrow.md
jhaaaa: Done - thank you!

content/tutorial-escrow.md
jhaaaa: Yep - got it - thank you! 👍  Just as an FYI, I went with this general sentence structure for sender and receiver:

> In this example, the second and fourth escrow objects meet our lookup criteria because their `Account` (sender address) values are set to `rfztBskAVszuS3s5Kq7zDS74QtHrw893fm`.

content/tutorial-escrow.md
jhaaaa: Addressed, my friend... Thank you!

mDuo13:
The build failures are unrelated to this PR, so feel free to merge.

content/code_samples/escrow/websocket/account_objects-request.json
```diff
@@ -1,7 +1,7 @@
 {
   "id": 5,
   "command": "account_objects",
-  "account": "rEhw9vD98ZrkY4tZPvkZst5H18RysqFdaB",
+  "account": "rfztBskAVszuS3s5Kq7zDS74QtHrw893fm",
   "ledger_index": "validated",
   "type": "escrow"
 }
```
jhaaaa: This file replaces account_objects_request_sender.json and serves as the request code sample for both sender and receiver.

content/code_samples/escrow/websocket/account_objects-request.json
```diff
@@ -1,7 +1,7 @@
 {
```
jhaaaa, line 1: `account_objects-request.json` replaces `account_objects_request_sender.json` and serves as the request code sample for both sender and receiver.

content/code_samples/escrow/websocket/account_objects-response-sender.json
```diff
@@ -1,26 +0,0 @@
-{
```
jhaaaa, line 1: `account_objects-response-sender.json` has been deleted. It is replaced by `account_objects_response.json`.

content/code_samples/escrow/websocket/account_objects-response.json
```diff
@@ -0,0 +1,60 @@
+{
```
jhaaaa, line 1: `account_objects-response.json` has been added. It replaces `account_objects_response_sender.json` and serves as the response code sample for both sender and receiver.

content/tutorial-escrow.md
```diff
@@ -522,23 +531,63 @@ Request:
...
 ```
 
 <!-- MULTICODE_BLOCK_END -->
 
-**Tip:** If you don't know what `OfferSequence` to use in the [EscrowFinish transaction][] to execute an escrow, use the [`tx` method](reference-rippled.html) to look up the transaction that created the escrow, using the identifying hash of the transaction in the Escrow's `PreviousTxnID` field. Use the `Sequence` value of that transaction as the `OfferSequence` value when finishing the escrow.
+###Look up escrows by receiver address
+
+You can use the [`account_objects`](reference-rippled.html#account-objects) method to look up escrow objects by receiver address.
+
+You can only look up pending escrow objects by receiver address if those escrows were created after the [fix1523 amendment](concept-amendments.html#fix1523) was enabled on 2017-11-14.
```
ryangyoung: Consider making this a callout.

ryangyoung (approved):
LGTM! Just one spot where you might want to use a note rather than plain body text.

nbougalis (approved):

scottschurr:
Hi Jennifer, I'm one of the rippled team members.  Welcome.  Nice work with the tutorial.

I have a minor suggestion for your consideration.  Your writeup uses the term `receiver` of the escrow.  The current escrow tutorial (https://ripple.com/build/escrow-tutorials/) does not use `receiver`.  You could consider using `destination` or `recipient`, which are already used in the tutorial, in place of `receiver`.  To me it seems that `destination` is a particularly good choice, since it is also the tag used in the JSON, so the correlation is very clear.

I guess another choice might be to replace the few instances of `recipient` in the tutorial with `receiver`.  Again, this is just with an eye to manage the number of terms that mean the same thing.

Just for your consideration.  I don't feel strongly about this, and you're the professional writer, not me.

Everything else looked good to me.  (Although, to be honest, I did not look through the JSON in detail.)

Thanks.

content/tutorial-escrow.md
jhaaaa: Thank you for your review - done!

jhaaaa:
Thank you for your approval @nbougalis! 

Thank you for your thoughtful review, @scottschurr - excellent point about receiver vs recipient vs destination. I’ve updated all instances of "receiver" to “destination.” I agree that using destination is the clearest option due to the direct correlation with the `Destination` tag. Much gratitude!

PR #288, opened 2017-12-15
Fix typo in transaction format reference

martijnjanssen:
The second sentence has a typo: 
`She might fund that this uncommon currency market has few offers.`

The `i` is unknowingly mistyped as a `u`. The new sentence now reads:
`She might find that this uncommon currency market has few offers.`


mDuo13 (approved):
Thanks for catching this.

PR #289, opened 2017-12-16
Build and Run rippled on Ubuntu

jhaaaa:


content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+
+Use these instructions to build a `rippled` binary file and run it as a stock `rippled` server on Ubuntu 15.04 or later. These instructions were tested on Ubuntu 16.04 LTS.
+
+For information about building `rippled` for other platforms, see [Builds](https://github.com/ripple/rippled/tree/develop/Builds) in the `rippled` GitHub repository.
+
+
+## System Requirements
+
+Ripple recommends meeting the following **minimum** system requirements for building and running `rippled`.
+
```
jhaaaa: Are these system requirements current? Kept the build reqs separate from run reqs. Would a user build in one environment and then run in another?

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+sudo apt-get -y update
+sudo apt-get install -y libboost-all-dev
+```
+11. Get `rippled` source code.
+```
+git clone https://github.com/ripple/rippled.git
+cd rippled
+git checkout master
+```
+12. Build `rippled` binary executable from source code. This may take about 30 minutes.
```
jhaaaa: this is how long it took for me on an AWS m3.large instance.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+    `rippled.cfg` is the configuration file for `rippled`.
+
+    You can find an example configuration file (`rippled-example.cfg`) in `rippled/doc`. See the file for a description of all configuration options. You should be able to connect to the XRP Ledger using the default configuration defined in the example file.
+
+    Put `rippled.cfg` in `etc/opt/ripple/`.
+
+2. Configure `validators.txt`.
+
+    When you start `rippled` for the first time, it uses an initial list of validators defined in `validators.txt` to retrieve validation information and form its UNL (unique node list). `validators.txt` also enables `rippled` clients to indirectly locate IPs they can use to contact the Ripple network.
+
```
jhaaaa: Not sure about the reference to "Ripple network" here. What is the difference between the XRP Ledger and the Ripple network? Is the Ripple network different from RippleNet?

Also, I say that these instructions are for a stock server. If you are use the example validators.txt as-is, are you indeed running a stock server?

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+## 2. Configure `rippled`
+
+1. Configure `rippled.cfg`.
+
+    `rippled.cfg` is the configuration file for `rippled`.
+
+    You can find the `rippled-example.cfg` example file in `rippled/doc`. See the file for a description of all configuration options. You should be able to connect to the XRP Ledger using the default configuration defined in the example file.
+
+    Put your configured `rippled.cfg` file in `etc/opt/ripple/`.
```
jhaaaa: what needs to change about the configuration to run a standalone server? I ask because if it is a simple configuration change - we may want to give the user the instructions on how to configure and run either flavor --

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+
+2. Configure `validators.txt`.
+
+    When you start your `rippled` server for the first time, it uses an initial list of validators defined in `validators.txt` to retrieve validation information and to form its UNL (unique node list). The information in `validators.txt` also enables `rippled` clients to indirectly locate IPs they can use to contact the Ripple network.
+
+    You can find the `validators-example.txt` example file in `rippled/doc`. See the file for a description of all validator options. You should be able to connect to the XRP Ledger using the default validators defined in the example file.
+
+    Put your configured `validators.txt` file in `etc/opt/ripple/`.
+
```
jhaaaa: Using the validators.txt configurations as-is -- are you indeed running a stock server? Just making sure that it isn't putting the user somewhere between a stock server and a validating server.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+To run your stock `rippled` server from the executable you built, using the configurations you defined:
+```
+cd build
+sudo ./rippled
+```
+
+
+### What to Expect
+
```
jhaaaa, line 146: I added this to help prepare the user for what they see in the terminal once they are running rippled. I would have found this useful when I first started rippled. But, for it to be truly useful, I think we need to add a few explanations of what you are seeing.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+2017-Dec-15 18:27:37 Amendments:DBG Amendment 4C97EBA926031A7CF7D7B36FDE3ED66DDA5421192D63DE53FFB46E43B9DC8373 is supported.
+...
+2017-Dec-15 18:27:37 ValidatorList:DBG Loading configured trusted validator list publisher keys
+...
+2017-Dec-15 18:27:37 ValidatorList:DBG Loading configured validator keys
+...
+2017-Dec-15 18:27:37 ValidatorSite:DBG Loading configured validator list sites
+...
+2017-Dec-15 18:27:38 LedgerConsensus:NFO Entering consensus process, watching, synced=no
+...
```
jhaaaa: What are we seeing below?

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+	"ledger_hash" : "014816C71F4B9AD5924AFBD7B02E0DF1522C4BC4BC45A69F97F51D4F2234FA10",
+	"ledger_index" : "2",
+	"parent_close_time" : 0,
+	"parent_hash" : "AB868A6CFEEC779C2FF845C0AF00A642259986AF40C01976A7F842B6918936C7",
+	"seqNum" : "2",
+	"totalCoins" : "100000000000000000",
+	"total_coins" : "100000000000000000",
+	"transaction_hash" : "0000000000000000000000000000000000000000000000000000000000000000"
+}
+...
```
jhaaaa: It would be great to provide a description of each of the following: 

* NetworkOPs: WRN We are not running on the consensus ledger. Who is not running on the consensus ledger and what does this mean?
* Ledgerconsensus:WRN Need consensus ledger. Who need it and why?
* OpenLedger:WRN Offer involves frozen asset. What does this mean?
* LedgerConsensus:WRN Offer involves frozen asset. What does this mean?

jhaaaa:
@mDuo13 Here's the build doc to meet your acceptance criteria for https://ripplelabs.atlassian.net/browse/DOC-369. Is this what you had in mind? Thank you!

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+Use these instructions to build a `rippled` binary file and run it as a stock `rippled` server on Ubuntu 15.04 or later. These instructions were tested on Ubuntu 16.04 LTS.
+
+For information about building `rippled` for other platforms, see [Builds](https://github.com/ripple/rippled/tree/develop/Builds) in the `rippled` GitHub repository.
+
+
+## System Requirements
+
+Here are the **minimum** system requirements for building and running `rippled`.
+
+**_To build rippled:_**
```
mDuo13: need `codefont` for `rippled` here too

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+
+* RAM: 8GB
+
+**_To run rippled:_**
```
mDuo13: Let's not duplicate the requirements for running `rippled` on this page. They're already in the other page, so we can link them. (Or we could make a snippet and include it in both.)

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+* CPU: 64-bit x86_64, 2+ cores
+* Disk: 50GB SSD with 500 IOPS for the database partition
+* Network: A fast network connection is preferable
+
+If you are running `rippled` in an Amazon EC2 instance, Ripple recommends selecting a **Type** of **_m3.large_**. If you choose to run a validating `rippled` server, you may need more resources.
+
+
+## 1. Build `rippled`
+
+These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the software prerequisites you need to build and run `rippled`.
```
mDuo13, line 25: I think this is the first time I've ever seen an explanation of what the "apt"  in `apt-get` stands for. 👍

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+
+1. Update the list of packages that are available for `apt-get` to install or upgrade.
+```
```
mDuo13: "Code fences" like this don't work properly in lists. (The best case scenario you'll get is for each list item to be a one-item list starting with the correct number. Many Markdown parsers will restart the numbering at 1 for each list item, though.)

You can create code blocks with extra indentation, though. There's an example page ([rendered](https://github.com/ripple/dactyl/blob/master/examples/content/lists-and-codeblocks.md) and [raw](https://raw.githubusercontent.com/ripple/dactyl/master/examples/content/lists-and-codeblocks.md)) that exemplifies some of the corner cases

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+11. Get `rippled` source code.
+```
+git clone https://github.com/ripple/rippled.git
+cd rippled
+git checkout master
+```
+12. Build `rippled` binary executable from source code. This may take about 30 minutes.
+
+    SCons saves the built executable in `rippled/build`.
```
mDuo13: I would put this after the example code (which will work if you adjust the example code to use indentation instead of code fences, as explained in my comment above)

content/tutorial-build-run-rippled-ubuntu.md
mDuo13: Not sure what you mean by "a standalone server". Each server stands alone unless you explicitly configure it to be part of a cluster. If you're asking what configuration changes you need to make to the example config file before you can run `rippled` at all, see my comment on line 94 above.

If you're asking about [stand-alone mode](https://ripple.com/build/stand-alone-mode/) that's a little different, and mostly for testing things. In stand-alone mode, the server ignores the rest of the network and does its own thing.

content/tutorial-build-run-rippled-ubuntu.md
mDuo13: BTW, I opened a task ( DOC-1222 ) to improve the nuance and detail of these hardware specs and how they relate to config options and usage patterns.

content/tutorial-build-run-rippled-ubuntu.md
mDuo13: I would add "depending on your hardware specs" just for clarity.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+
+```
+scons
+```
+
+
+## 2. Configure `rippled`
+
+1. Configure `rippled.cfg`.
```
mDuo13: I would break this up into multiple steps:

> 1. Create a copy of the example config file (assumes you're in the `rippled` folder already):
>    - To run `rippled` as a non-root user (recommended):
>
>               cp doc/rippled-example.cfg ~/.config/ripple/rippled.cfg
> 2. Edit the config file to set necessary file paths.  The user you plan to run `rippled` as must have write permissions to all of the paths you specify here.
>    1. Set the `[node_db]`'s `path` to the location where you want to store the ledger database.
>    2. Set the `[database_path]` to the location where you want to store other database data (this includes a sqlite database with configuration data, and is typically one level above the `[node_db]` `path` field)
>    3. Set the `[debug_logfile]` to a path where rippled can write logging information.

All other configuration is optional and can be tweaked after you have a working server, but that information must be set before `rippled` can start up successfully.

mDuo13: For what it's worth, I believe this section on where `rippled` looks for a config file (if you don't explicitly pass a config file on the commandline) is still accurate: https://wiki.ripple.com/Rippled#--conf.3Dpath

Also, `/etc/opt/ripple/` might be the default folder where Alien installs `rippled` from the RPM on Ubuntu, but I think there's nothing else special about that path and it might actually be a good idea not to put it there if you're compiling it yourself.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+
+    `rippled.cfg` is the configuration file for `rippled`.
+
+    You can find the `rippled-example.cfg` example file in `rippled/doc`. See the file for a description of all configuration options. You should be able to connect to the XRP Ledger using the default configuration defined in the example file.
+
+    Put your configured `rippled.cfg` file in `etc/opt/ripple/`.
+
+2. Configure `validators.txt`.
```
mDuo13: <s>I believe that all the info about the `validators.txt` file is two or three times outdated and you should strike this entire section.

A `validators.txt` file is not necessary (it's just a file that can contain options that are, by default, configured in `rippled.cfg`) and specifying anything other than the default is not wise at this point in time.</s>

(Sorry, this is wrong. Writing a correction now.)

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+
+    When you start your `rippled` server for the first time, it uses an initial list of validators defined in `validators.txt` to retrieve validation information and to form its UNL (unique node list). The information in `validators.txt` also enables `rippled` clients to indirectly locate IPs they can use to contact the Ripple network.
+
+    You can find the `validators-example.txt` example file in `rippled/doc`. See the file for a description of all validator options. You should be able to connect to the XRP Ledger using the default validators defined in the example file.
+
+    Put your configured `validators.txt` file in `etc/opt/ripple/`.
```
mDuo13: Nitpick (even though this section should be stricken entirely anyway): There's a very important difference between `/etc/opt/ripple/` (a very specific absolute path) and `etc/opt/ripple/` (a relative path, which could be anywhere). On UNIX-like operating systems, the leading slash indicates an absolute path—"start from the top". If you don't have a leading slash, it's a relative path—"start from wherever you are" instead.

Incidentally, the trailing slash usually doesn't matter, so `/etc/opt/ripple` and `/etc/opt/ripple/` are functionally equivalent (the only difference being that the latter is _definitely_ a directory, whereas the former _could_ be a directory or a different type of file.)

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+
+
+## 3. Run `rippled`
+
+To run your stock `rippled` server from the executable you built, using the configurations you defined:
+```
+cd build
+sudo ./rippled
```
mDuo13: If you set your paths in the `rippled.cfg` file to folders you can write to as a regular user, then you don't need `sudo`.

That's a good thing, because by running `rippled` as a non-root user, it means `rippled` cannot overwrite stuff it shouldn't be able to, even if it bugs out and goes berserk or is taken over by an exploit. (Not that it should be able to do either of those under normal circumstances, but we're talking defense in depth here.)

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+2. Configure `validators.txt`.
+
+    When you start your `rippled` server for the first time, it uses an initial list of validators defined in `validators.txt` to retrieve validation information and to form its UNL (unique node list). The information in `validators.txt` also enables `rippled` clients to indirectly locate IPs they can use to contact the Ripple network.
+
+    You can find the `validators-example.txt` example file in `rippled/doc`. See the file for a description of all validator options. You should be able to connect to the XRP Ledger using the default validators defined in the example file.
+
+    Put your configured `validators.txt` file in `etc/opt/ripple/`.
+
+
+## 3. Run `rippled`
```
mDuo13: Another item that _mayyyy_ be necessary is to raise the limit on the number of open file descriptors per process. That's described in [a forum thread](https://www.xrpchat.com/topic/2342-how-to-fix-the-max-file-descriptors-error-when-running-rippled/) but not documented properly in any official place yet.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+
+### What to Expect
+
+Once you've run `rippled`, here are excerpts of what you can expect to see in your terminal:
+
+```
+Loading: "/etc/opt/ripple/rippled.cfg"
+Watchdog: Launching child 1
+2017-Dec-15 18:27:37 JobQueue:NFO Auto-tuning to 4 validation/transaction/proposal threads for non-validator.
+2017-Dec-15 18:27:37 Amendments:DBG Amendment C6970A8B603D8778783B61C0D445C23D1633CCFAEF0D43E7DBCD1521D34BD7C3 is supported.
```
mDuo13: That's weird. These are debug-level log messages, which probably should not be output if you use [the example config](https://github.com/ripple/rippled/blob/develop/doc/rippled-example.cfg#L1082), which limits log messages to warning and up.

content/tutorial-build-run-rippled-ubuntu.md
mDuo13: > `NetworkOPs:WRN We are not running on the consensus ledger`

This indicates that this server does not have the latest complete copy of the ledger data from the network. It's normal to see this for the first few minutes after starting up the server since it needs time to download the ledger state and catch up with the rest of the network. If it continues for more than 5-10 minutes, or recurs after the server has been running and stable for a while, that's a problem.

> `LedgerConsensus:WRN Need consensus ledger `

This is a similar warning from a different part of the code which indicates the same thing (not synced to the network state), and is also normal when it occurs within the first 5-10 minutes of starting up.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+	"seqNum" : "2",
+	"totalCoins" : "100000000000000000",
+	"total_coins" : "100000000000000000",
+	"transaction_hash" : "0000000000000000000000000000000000000000000000000000000000000000"
+}
+...
+2017-Dec-15 18:27:50 NetworkOPs:WRN We are not running on the consensus ledger
+2017-Dec-15 18:27:50 LedgerConsensus:WRN Need consensus ledger 8459443546CCBE4E89A41E1D0C17F24FBEA7B344C1ED6A26292B4C5CD19B4321
+...
+2017-Dec-15 18:38:28 OpenLedger:WRN Offer involves frozen asset
```
mDuo13: > `OpenLedger:WRN Offer involves frozen asset`

This is probably a log message that should not be warning-level. I think it's benign and indicates someone is trying to trade an issued currency that's [frozen](https://ripple.com/build/freeze/).

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+2017-Dec-15 18:38:31 LedgerConsensus:WRN Offer involves frozen asset
+```
+
+
+## Explore Next Steps
+
+Now that you have a stock `rippled` server running, you may want to consider running it as a validating server. For information about validating servers and why you might want to run one, see the [rippled Setup Tutorial](tutorial-rippled-setup.html).
+
+For information about updating and maintaining your `rippled` server, see [Updating rippled](tutorial-rippled-setup.html#updating-rippled).
```
mDuo13: Those instructions for updating `rippled` don't apply if you built it yourself. They're only for if you installed it from the RPM.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,173 @@
...
+
+For information about communicating with your `rippled` server using the `rippled` API, see [rippled](reference-rippled.html).
```
mDuo13: A better name for that page is probably "the `rippled` API reference"

dactyl-config.yml
```diff
@@ -247,6 +249,15 @@ pages:
             - local
             - ripple.com
 
+# Jha added entry below. Is there where we want this tutorial to be accessbile?
```
mDuo13: This looks good. (You can remove the comment)

dactyl-config.yml
```diff
@@ -59,6 +59,8 @@ targets:
             "tutorial-paychan.html": https://ripple.com/build/payment-channels-tutorial/
             "tutorial-escrow.html": https://ripple.com/build/escrow-tutorials/
             "tutorial-listing-xrp.html": https://ripple.com/build/listing-xrp-exchange/
+            # Jha added placeholder link_sub below - need to get the permalink from Wordpress
```
mDuo13: Good idea. Let's get the real permalink, set it here, and remove this comment before merging.

mDuo13 (changes requested):
An excellent handling of tough source material. As usual, you dug deep into the task and uncovered some further work that should be done.

I left some advice on a couple formatting things and recommend that you remove the `validators.txt` section from this guide.

I also suggest caution about expanding the scope of this task again. Let's keep it simple for now so we can publish what we know and dig into the more complex stuff—like the meaning of the log messages, for example—later.

mDuo13:
!rebuild

content/tutorial-build-run-rippled-ubuntu.md
mDuo13: Per [DEC-745](https://ripplelabs.atlassian.net/browse/DEC-745), the default config file does not list the validators explicitly (I was basing that on a mistaken previous assumption) but instead uses `[validators_file]` to specify a file with the default validator config, with the idea being that `validators.txt` should be automatically replaced when there are updates while `rippled.cfg` should not. That's because `rippled.cfg` may contain settings that are specific and/or sensitive (like a validating server's secret or token) whereas `validators.txt` is general and everyone should generally use the same one so there's no chance of forking away from the majority of the network.

People (like me) who have a config file derived from an older (pre-0.32.0) example may have their validators explicitly configured in `rippled.cfg` instead. They should be sure to update their config when the 0.81.0 release happens. That's outside the scope of this task.

What's in-scope for this task is that the install should include copying the default `validators.txt` file to an appropriate location. If you put it in the same folder as `rippled.cfg` (which you should) then you don't even need to edit the `[validators_file]` stanza of `rippled.cfg`. So under "configure `rippled`" you'd add another step like this:

> 3. Copy the example `validators.txt` file to the same folder as `rippled.cfg`:
> 
>         cp doc/validators-example.txt ~/.config/ripple/validators.txt

mDuo13: Also, unlike the language in the 1st draft which mentions "a description of all validator options", the final version should have a strongly-worded warning that you _SHOULD NOT_ modify the example validators.txt file except as directed by Ripple, or else you may fork away from the production network and could lose money.

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: This is GREAT. Thank you - yes, it would be great to describe the hierarchy of where `rippled` looks for the config files. I'll open a ticket (witness the self-control... lol) 

I'll put this in the ticket, but just because I think you might be interested - for this task, when I run `rippled` with no config files present, this is the error message: 

image

This is why I put the config files in that location. 😂  Perhaps this is the wrong location for it to first look for the config files if we recommend that users run `rippled` as a non-root users?

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: Yes, sorry for the incorrect language. I am asking about stand-alone mode. 😄  How do you put your `rippled` server in stand-alone mode? Sorry if this is a dumb question - but does stand-alone mode mean that my `rippled` server still connects to the XRP Ledger, but not to any other `rippled` servers in the network? And this way, I can still query the XRP Ledger using the `rippled` API.

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: 👍  Done - I used the language from Nik in cpp-informal and included a link to the decentralization strategy update provided by Arthur.

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: I learn a LOT from your "nitpicks" -- so I appreciate your taking the time to provide them. Got it.

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: Opening a ticket to document .

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: Yep. I totally get it now. I removed the `sudo` from the command. The text you provided at line 94 states the requirement for write permissions so the user will not need `sudo`.

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: Ah, weird. Yes, I am using the example config as-is.

image

content/tutorial-build-run-rippled-ubuntu.md
mDuo13: The short answer is, you just add `-a` to the commandline to start your `rippled` server in stand-alone mode.

And I should probably schedule some time to explain some higher-level concepts of the XRP Ledger. The short version is, the servers in the network _are_ the XRP Ledger (sort of), so it doesn't make sense to ask if it connects to one but not the other. I mean, really, the XRP Ledger is just a concept, the shared state created out of the imagination of a bunch of servers dreaming in unison, and you can't truly connect to a concept.

Basically stand-alone mode is a lab setting where your server does all the imagining of the XRP Ledger all by its lonesome, so that you can do stuff like:

- submit transactions no one else will ever see
- freeze-frame, manually advance, and replay the ledger
- test out experimental code, force amendments to be on and see what they do
- explore alternate realities from a fresh starting point

content/tutorial-build-run-rippled-ubuntu.md
mDuo13: When you use `sudo` it runs it as root, so it looks in the directory that's appropriate for a root account.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,165 @@
...
+9. Install `python-software-properties`.
+
+        sudo apt-get install -y python-software-properties
+
+10. Install Boost.
+
+        sudo add-apt-repository -y ppa:boost-latest/ppa
+        sudo apt-get -y update
+        sudo apt-get install -y libboost-all-dev
+
```
jhaaaa: Based this entire doc on https://gist.github.com/stevenzeiler/1c6d421dfec45a26b09a. 

FYI - I changed `sudo apt-get install -y libboost1.55-all-dev` from the gist to `sudo apt-get install -y libboost-all-dev` in this doc because `libboost1.55-all-dev` it didn't work for me.

MarkusTeufelberger: Upstream boost builds just fine on Ubuntu 16.04 fyi - https://github.com/MarkusTeufelberger/rippled-distrotest/blob/master/ubuntu-16.04/Dockerfile

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,165 @@
...
+./rippled
+```
+
+
+### What to Expect
+
+Once you've run `rippled`, here are excerpts of what you can expect to see in your terminal.
+
+```
+Loading: "/etc/opt/ripple/rippled.cfg"
```
MarkusTeufelberger: This path shouldn't exist if following this guide.

Also /etc/opt/ripple/rippled.cfg is not the correct location for a config file on Debian based systems.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,165 @@
...
+...
+2017-Dec-15 18:27:50 NetworkOPs:WRN We are not running on the consensus ledger
+2017-Dec-15 18:27:50 LedgerConsensus:WRN Need consensus ledger 8459443546CCBE4E89A41E1D0C17F24FBEA7B344C1ED6A26292B4C5CD19B4321
+...
+2017-Dec-15 18:38:28 OpenLedger:WRN Offer involves frozen asset
+2017-Dec-15 18:38:31 LedgerConsensus:WRN Offer involves frozen asset
+```
+
+
+## Explore Next Steps
```
MarkusTeufelberger: ...such as moving the binary to /usr/bin/rippled and installing a systemd unit for it.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,165 @@
...
+
+        sudo apt-get -y install git
+
+4. Install SCons.
+
+        sudo apt-get -y install scons
+
+5. Install Ctags.
+
+        sudo apt-get -y install ctags
```
MarkusTeufelberger: Why ctags?

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,165 @@
...
+        sudo apt-get -y install protobuf-compiler
+        sudo apt-get -y install libprotobuf-dev
+
+8. Install Secure Socket Layer (SSL) toolkit development files.
+
+        sudo apt-get -y install libssl-dev
+
+9. Install `python-software-properties`.
+
+        sudo apt-get install -y python-software-properties
```
MarkusTeufelberger: Why this package? Where in the build process is it used?

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,165 @@
...
+
+10. Install Boost.
+
+        sudo add-apt-repository -y ppa:boost-latest/ppa
```
MarkusTeufelberger: This ppa is horribly out of date: https://launchpad.net/~boost-latest/+archive/ubuntu/ppa, Xenial already has 1.58 upstream: https://packages.ubuntu.com/xenial/libboost-all-dev

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,165 @@
...
+        git checkout master
+
+12. Build `rippled` binary executable from source code. This may take about 30 minutes, depending on your hardware specs.
+
+        scons
+
+    SCons saves the built executable in `rippled/build`.
+
+
+## 2. Configure `rippled`
```
MarkusTeufelberger: The proper next step would be to build a deb file at this point by the way: https://wiki.debian.org/HowToPackageForDebian

content/tutorial-build-run-rippled-ubuntu.md
mDuo13: Yeah, I'm not sure what's up with that. I think it's supposed to use an XDG_* environment variable. But also if you run `rippled` _not_ as root, it should look in `~/.config/ripple/rippled.cfg` first.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,165 @@
...
+
+    2. Set the `[database_path]` to the location where you want to store other database data. (This includes an SQLite database with configuration data, and is typically one level above the `[node_db]` path field.)
+
+    3. Set the `[debug_logfile]` to a path where `rippled` can write logging information.
+
+3. Copy the example `validators.txt` file to the same folder as `rippled.cfg`:
+
+        cp doc/validators-example.txt ~/.config/ripple/validators.txt
+
+    **Warning:** Ripple has designed and [documented](https://ripple.com/dev-blog/decentralization-strategy-update/) a decentralization plan with maximum safety in mind. During the transition, you *should not* modify the `validators.txt` file except as recommended by Ripple. Even minor modifications to your validator settings could cause your server to diverge from the rest of the network and report out of date, incomplete, or inaccurate data. Acting on such data can cause you to lose money.
```
mDuo13: Nitpick: rather than "documented", I think the link text should be "a decentralization plan"

content/tutorial-build-run-rippled-ubuntu.md
wilsonianb: The `rpc_startup` doesn't get processed right away.
https://github.com/ripple/rippled/blob/d2fc4e3569d79d3cade78533f673f642a8d26845/src/ripple/app/main/Application.cpp#L1266-L1269
Everything gets logged before that point

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,165 @@
...
+    SCons saves the built executable in `rippled/build`.
+
+
+## 2. Configure `rippled`
+
+Complete the following configurations that are required for `rippled` to start up successfully. All other configuration is optional and can be tweaked after you have a working server.
+
+1. Create a copy of the example config file (assumes you're in the `rippled` folder already). Saving the config file to this location enables you to run `rippled` as a non-root user (recommended).
+
+         cp doc/rippled-example.cfg ~/.config/ripple/rippled.cfg
```
wilsonianb: This could be preceded with 
`mkdir -p ~/.config/ripple`

mDuo13:
Looks like the broken link on this is inherited from an unrelated file. If you rebase onto master, it should go away. Let's be sure to get the build passing before we merge.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,165 @@
...
+
+        git clone https://github.com/ripple/rippled.git
+        cd rippled
+        git checkout master
+
+12. Build `rippled` binary executable from source code. This may take about 30 minutes, depending on your hardware specs.
+
+        scons
+
+    SCons saves the built executable in `rippled/build`.
```
wilsonianb: Scons will likely be deprecated in favor of cmake in the near future

This would mean installing `cmake` instead of `scons` (Step 4)
The build step could be either 
```
BUILD=cmake CC=gcc TARGET=release.unity bin/ci/ubuntu/build-and-test.sh
```
or
```
mkdir -p build/gcc.release
cd build/gcc.release
cmake ../.. -Dtarget=gcc.release
cmake --build .
```
For the former, `rippled` will be in `build/gcc.release.unity/` and unittests will run automatically
For the latter, `rippled` will be in `build/gcc.release`

MarkusTeufelberger: https://github.com/ripple/rippled/blob/develop/bin/ci/ubuntu/build-and-test.sh is a bit of a mess that expects all kinds of environment variables to be maybe existing, I wouldn't recommend it as the documented build step. Even `Builds/Test.py [and some options]` already gets messy with cmake.

The latter version is still better, even though it still contains lots of "magic" (like the "target" having to match a certain subfolder or the target variable containing a definition which compiler should be used in which configuration...).

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: Thank you for calling this out. Indeed, installation of ctags is optional. It is a tool that a developer, in general, may want to install, but ctags is not specifically required by rippled.  Confirmed that build and run works fine without it. Removing this step.

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: From what I can tell, installing `python-software-properties` makes it possible to run `sudo add-apt-repository...` Removing this step per comments below that make `sudo add-apt-repository -y ppa:boost-latest/ppa` unnecessary. Confirmed that build and run works fine without it.

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: Thank you for pointing this out. Removing `sudo add-apt-repository -y ppa:boost-latest/ppa` and `sudo apt-get -y update`. Install Boost using `sudo apt-get install -y libboost-all-dev` only. Confirmed that build and run works fine without removed commands.

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: Thank you for this. Indeed, we’ll just use `libboost-all-dev only`.

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: Thank you for the heads up, @wilsonianb. Thank you for the detailed feedback, @MarkusTeufelberger. I've opened a ticket to capture these details -- we'll update the doc when scons is deprecated in favor of cmake.

nbougalis:
Love this work @jhaaaa! Thanks @wilsonianb & @MarkusTeufelberger for the thorough review and feedback.

Once we migrate to `CMake` the documentation will, hopefully, be a lot cleaner.

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: Will cover this behavior along with more comprehensive coverage of rippled log/warning/error messages. Added to ticket.

MarkusTeufelberger:
Looking forward to having someone to help mduo13 with reducing the documentation backlog. :-)

Sorry for dropping in unannounced and as an external contributor, but I had a suspicion that I might have a bit of experience building and running rippled that could be useful.

mDuo13:
No need to apologize. External contributions are more than welcomed.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -0,0 +1,184 @@
...
+
+        sudo apt-get -y install libssl-dev
+
+8. Install `wget` to be able to download Boost in the next step.
+
+        sudo apt-get install wget
+
+8. Compile Boost.
+
+    Starting in `rippled` 0.90, the compatible Boost version is 1.65.0. Because Boost version 1.65.0 isn't available in the Ubuntu 16.04 repos, you'll need to compile it yourself. ***TODO: Per Scott D, the recommendation will be either 1.65.0 or 1.66.0+ for 0.90.0. Per Scott D, test with 1.65.0 first; this doc reflects that test. If this PR is accepted, the recommendation will be 1.66.0+***
```
jhaaaa: Clarification - by "this PR," I mean Scott's PR: https://github.com/ripple/rippled/pull/2348

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: Thank you, Markus! Added this as a possible next step for the user to consider in "Explore Next Steps."

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: Added installing a systemd unit to "Explore Next Steps." Thank you for the suggestion!

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -4,7 +4,7 @@
 
 For an overview of `rippled`, see [Operating rippled Servers](tutorial-rippled-setup.html).
 
-Use these instructions to build a `rippled` binary file and run it as a stock `rippled` server on Ubuntu 15.04 or later. These instructions were tested on Ubuntu 16.04 LTS.
+Use these instructions to build a `rippled` 0.90.0+ binary file and run it as a stock `rippled` server on Ubuntu 15.04 or later. ***TODO: Question: I added the specific `rippled` release series to this text - is that okay? Should I remove the reference to Ubuntu 15.04 and just say that this is for Ubuntu 16.04 LTS or later?*** These instructions were tested on Ubuntu 16.04 LTS.
```
mDuo13: 1. It's good to specify to which `rippled` version(s) these instructions apply. I would change the wording slightly, from:
    > build a `rippled` 0.90.0 binary file and run it as a stock `rippled` server on Ubuntu 15.04 or later.
    
    To:
    
    > build a `rippled` executable from source for version 0.90.0 or higher on supported versions of Ubuntu Linux.
2. Let's just say this is for supported versions of Ubuntu and then, in the next sentence, specify which verison(s) the instructions were tested on, as you already do.

_(Edited: fix typo in point 2)_

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -116,50 +131,54 @@ cd build
...
+
+* For information about communicating with your `rippled` server using the `rippled` API, see the [`rippled` API reference](reference-rippled.html).
+
+* As a development best practice, you may want to build a `rippled` .deb file. For more information, see _Ubuntu Packaging Guide_: [Packaging New Software](http://packaging.ubuntu.com/html/packaging-new-software.html).
 
-For information about communicating with your `rippled` server using the `rippled` API, see the [`rippled` API reference](reference-rippled.html).
+* You may also want to install a `systemd` unit. For more information, see [systemd for Upstart Users](https://wiki.ubuntu.com/SystemdForUpstartUsers). If you are compiling `rippled` from source, you may want to take a look at the `rippled` [systemd unit file](https://github.com/ripple/rippled-package-builder/blob/staging/rpm-builder/rippled.service) that is a part of the `rippled-package-builder` and adjust it to fit your needs.
```
mDuo13: I think you can simplify the second sentence, from:

> If you are compiling `rippled` from source, you may want to take a look at the `rippled` [systemd unit file](https://github.com/ripple/rippled-package-builder/blob/staging/rpm-builder/rippled.service) that is a part of the `rippled-package-builder` and adjust it to fit your needs.

To:

> You can use the [official `rippled` systemd unit file](https://github.com/ripple/rippled-package-builder/blob/staging/rpm-builder/rippled.service) or modify it to suit your needs.

mDuo13 (changes requested):
Left a couple comments.

Also, I'm adapting these steps to build rippled 0.90.0-b4 against boost 1.65 on my Arch Linux system that already has boost 1.66.0 as a system library. <s>I'll get back to you on how that works.</s>

**Update:** It appears to have worked. I was able to compile `rippled` 0.90.0-b4 against boost 1.65, without affecting programs that depend on my system's boost 1.66.

content/tutorial-build-run-rippled-ubuntu.md
mDuo13: On second thought, maybe let's not use the phrase "supported versions of Ubuntu" because it's not clear whether we mean versions we support specifically for `rippled` or versions Canonical supports in general.

mDuo13: I change my recommended rephrase to the following:

> build a `rippled` executable from source for version 0.90.0 or higher on Ubuntu Linux 16.04 or higher.

mDuo13:
For the record, building boost 1.65 as part of these instructions (the `./b2 -j<number of jobs>` step, which I did with 4 jobs) went pretty fast on my current system (like 5 minutes with a four-core Intel Core i7-7700 @ 3.60GHz). It ended with the following summary.

```
common.copy stage/lib/libboost_wave.so.1.65.0
ln-UNIX stage/lib/libboost_wave.so
...failed updating 66 targets...
...skipped 12 targets...
...updated 1192 targets...
```

As far as I can tell, the failures were benign and didn't affect `rippled` itself—After compiling (which took ~20-30 minutes with the default scons settings, although I wasn't paying close attention) all 316599 unit tests passed (in just over 2 minutes).

jhaaaa:
@mDuo13 - thank you for sharing your results regarding `./b2 -j<number of jobs>`! I ran it with 1 job. I'll try my new build with 4, just to see.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -89,7 +89,7 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
 
 10. Build `rippled` binary executable from source code. This may take about 30 minutes, depending on your hardware specs.
 
-        scons
+        scons ***TODO: If 1.66.0, I get errors. Expect that the errors will go away after Scott D's PR is applied to rippled***
```
mDuo13: Tip: You can specify `-j <number of jobs>` here too. I think the space between the j and the number of jobs is probably optional? Not sure.

For what it's worth, on my PC, building rippled 0.90.0-b5 with 4 jobs cut the build time down to just 5 minutes.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -63,23 +63,23 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
 
       a. Download Boost 1.65.0.
 
-          wget https://dl.bintray.com/boostorg/release/1.65.0/source/boost_1_65_0.tar.gz
+          wget https://dl.bintray.com/boostorg/release/1.65.0/source/boost_1_65_0.tar.gz ***TODO: If 1.66.0: https://dl.bintray.com/boostorg/release/1.66.0/source/boost_1_66_0.tar.gz***
```
mDuo13: Looks like [as of `rippled` 0.90.0-b5, boost 1.66.0 it is](https://github.com/ripple/rippled/pull/2348#issuecomment-361460549).

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -89,7 +89,7 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
 
 10. Build `rippled` binary executable from source code. This may take about 30 minutes, depending on your hardware specs.
```
mDuo13: Extra step you may want to put in between 9 and 10 here. If you previously built, or (more importantly) tried and failed to build `rippled`, you should delete the `build/` directory to start clean. Otherwise, you may get unexpected behavior like a `rippled` executable that crashes due to a segmentation fault (segfault).

If this is your first time building `rippled`, you won't have a `build/` directory and can skip that step.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -63,23 +63,23 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
...
+          wget https://dl.bintray.com/boostorg/release/1.65.0/source/boost_1_65_0.tar.gz ***TODO: If 1.66.0: https://dl.bintray.com/boostorg/release/1.66.0/source/boost_1_66_0.tar.gz***
 
-      b. Untar `boost_1_65_0.tar.gz`.
+      b. Untar `boost_1_65_0.tar.gz`. ***TODO: If 1.66.0, boost_1_66_0.tar.gz***
 
-          tar xvzf boost_1_65_0.tar.gz
+          tar xvzf boost_1_65_0.tar.gz ***TODO: If 1.66.0, boost_1_66_0.tar.gz***
 
-      c. In the new `boost_1_65_0` directory, run:
+      c. In the new `boost_1_65_0` directory, run: ***TODO: If 1.66.0, boost_1_66_0***
```
mDuo13: Nit: When the instructions are this granular, it's best to provide examples of every command line you need to enter, even when they're super-basic things like `cd boost_1_66_0`. Otherwise, people skimming just the code examples will miss it.

Also, if you could explain what the bootstrap script does, that'd be nice. If not, no big deal.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -63,23 +63,23 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
 
       a. Download Boost 1.65.0.
```
mDuo13: Don't forget to update this line to 1.66.0 also

mDuo13 (approved):
Instructions worked great for me (on Arch Linux) and I suspect they'll work well on many flavors of Linux.

Provisionally approved contingent on changing all those "TODO" statements to boost 1.66.0.

mDuo13:
Oh, also, you should rebase onto the latest master to fix the broken links. Ping me if you want a refresher of how to do that (it can be tricky)

content/tutorial-build-run-rippled-ubuntu.md
jhaaaa: I ran ./b2 -j 4 -- with the space between -j and 4. The job finished with the space included, so I added it in the doc. However, the time it took didn’t get shorter for me. I wonder if it is something about it running on AWS?

I ran scons with a -j value twice -- and it hung at the same location each time. When I removed the -j it ran as usual. I left the -j value out of the doc, just because I wasn’t able to get it to work in my workflow -- though I do trust that it should work and that it worked for you!

jhaaaa:
@mDuo13 Thank you for the offer to help with the rebase! Perhaps we can find a time to do it tomorrow? It would be great to just walk through the workflow for a new WP article as well. I created it and added the URL to dactyl-config, but I left it in draft state.

content/tutorial-build-run-rippled-ubuntu.md
ripplederek-admin: Generally, you'll only see an improvement with higher values of \<j\> when you have at least \<j\> CPU cores on your server. It's quite possible your test AWS instance only has one or two CPU cores, so you wouldn't see an improvement when using -j 4.

MarkusTeufelberger: You'd also need some relatively significant amounts of RAM to compile rippled, maybe your instance ran out of memory?

MarkusTeufelberger: Alternatively you can compile in nounity mode, but that'll take a bit longer and doesn't seem to be well tested.

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -55,41 +55,53 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
...
 8. Install `wget` to be able to download Boost in the next step.
 
-        sudo apt-get install wget
+        sudo apt-get -y install wget
 
-8. Compile Boost.
+9. Compile Boost.
 
-    Starting in `rippled` 0.90, the compatible Boost version is 1.65.0. Because Boost version 1.65.0 isn't available in the Ubuntu 16.04 repos, you'll need to compile it yourself. ***TODO: Per Scott D, the recommendation will be either 1.65.0 or 1.66.0+ for 0.90.0. Per Scott D, test with 1.65.0 first; this doc reflects that test. If this PR is accepted, the recommendation will be 1.66.0+***
+    Starting in `rippled` 0.90, the compatible Boost version is 1.66.0. Because Boost version 1.66.0 isn't available in the Ubuntu 16.04 repos, you'll need to compile it yourself.
```
mDuo13: Nitpick: "you'll need to"→"you must" (avoids future tense, simpler for non-native English speakers)

content/tutorial-build-run-rippled-ubuntu.md
```diff
@@ -179,6 +182,6 @@ Watchdog: Launching child 1
```
mDuo13: Another step you may want to add, either to "Next Steps" or as an optional last step before "3. Run rippled" is to run the unit tests:

```
./rippled -u
```

If there are no test failures, then you can be fairly certain that your `rippled` executable compiled correctly.

PR #292, opened 2018-01-04
Re-sort Known Amendments

mDuo13:
While writing AmendmentBot, I discovered that some of the amendments were not _actually_ in alphabetical order in the body. This moves them to be properly alphabetical.

Additionally, this PR rearranges the order of the elements in the table to follow a similar order that can be mechanically reproduced from the details of the amendments. The primary element of the new sort order is the current status; consequently, vetoed amendments are now at the end instead of being interspersed sort of arbitrarily based on when Ripple (the company) decided to remove them.

This file is generated automatically, so it also has some whitespace changes, which should be consistent with what AmendmentBot will generate in the future.


content/reference-amendments.md
```diff
@@ -5,13 +5,13 @@ The following is a comprehensive list of all known amendments and their status o
...
-| [CryptoConditionsSuite][] | TBD        | [Planned: TBD]( "BADGE_LIGHTGREY")  |
-| [OwnerPaysFee][]          | TBD        | [Planned: TBD]( "BADGE_LIGHTGREY")  |
-| [Tickets][]               | TBD        | [Planned: TBD]( "BADGE_LIGHTGREY")  |
+| [FlowCross][]             | v0.70.0    | [Planned: TBD]( "BADGE_LIGHTGREY") |
+| [SHAMapV2][]              | v0.33.0    | [Planned: TBD]( "BADGE_LIGHTGREY") |
+| [Checks][]                | TBD        | [Planned: TBD]( "BADGE_LIGHTGREY") |
+| [DepositAuth][]           | TBD        | [Planned: TBD]( "BADGE_LIGHTGREY") |
+| [CryptoConditionsSuite][] | TBD        | [Planned: TBD]( "BADGE_LIGHTGREY") |
+| [OwnerPaysFee][]          | TBD        | [Planned: TBD]( "BADGE_LIGHTGREY") |
+| [Tickets][]               | TBD        | [Planned: TBD]( "BADGE_LIGHTGREY") |
```
jhaaaa: I may be misunderstanding the sort order - but should CryptoConditionsSuite come before DepositAuth?

content/reference-amendments.md
```diff
@@ -5,13 +5,13 @@ The following is a comprehensive list of all known amendments and their status o
...
 | [SortedDirectories][]     | v0.80.0    | [Enabled: 2017-11-14](https://xrpcharts.ripple.com/#/transactions/6E2309C156EBF94D03B83D282A3914671BF9168FB26463CFECD068C63FFFAB29 "BADGE_GREEN") |
 | [fix1201][]               | v0.80.0    | [Enabled: 2017-11-14](https://xrpcharts.ripple.com/#/transactions/B1157116DDDDA9D9B1C4A95C029AC335E05DB052CECCC5CA90118A4D46C77C5E "BADGE_GREEN") |
 | [fix1512][]               | v0.80.0    | [Enabled: 2017-11-14](https://xrpcharts.ripple.com/#/transactions/63F69F59BEFDC1D79DBF1E4060601E05960683AA784926FB74BC55074C4F6647 "BADGE_GREEN") |
```
jhaaaa: Again, sorry if I'm misunderstanding the sort order -- but should SortedDirectories come after fix1512?

content/reference-amendments.md
```diff
@@ -23,22 +23,21 @@ The following is a comprehensive list of all known amendments and their status o
 | [PayChan][]               | v0.33.0    | [Enabled: 2017-03-31](https://xrpcharts.ripple.com/#/transactions/16135C0B4AB2419B89D4FB4569B8C37FF76B9EF9CE0DD99CCACB5734445AFD7E "BADGE_GREEN") |
 | [Escrow][]                | v0.60.0    | [Enabled: 2017-03-31](https://xrpcharts.ripple.com/#/transactions/C581E0A3F3832FFFEEB13C497658F475566BD7695B0BBA531A774E6739801515 "BADGE_GREEN") |
```
jhaaaa: You know that I know that you can sort things, probably with your eyes closed LOL, so my questions are just about understanding the sort order being used. If the Status is the first sort, is the Introduced version the second sort, with the older version coming first? Or is the Name the second sort? For PayChan and Escrow above, the former is true. For FlowCross and SHAMapV2 at the top, the latter is true.

content/reference-amendments.md
mDuo13: Oh, I think I know what's going on here. The sorting criteria within "Planned" amendments is the "Introduced" version (in descending order) but that means that ones with the same "TBD" end up in non-deterministic order.

I can fix that by adding another level of sorting based on the name.

content/reference-amendments.md
jhaaaa: @mDuo13 - this one above still seems odd to me. 🤔

mDuo13: Ah, this is another interesting case. Within the "Enabled" amendments, the only sorting key is the release date, so the display order of amendments enabled on the same day isn't defined. I can add a secondary sort to order them by name.

mDuo13:
!rebuild

jhaaaa (approved):
MUCH easier to grok the amendments and their statuses with the sorting you've implemented. 👍

PR #293, opened 2018-01-10
Update example transactions

mDuo13:
- Use example.com instead of mduo13.com in AccountSet
- Use 12-drop fees for most example transactions


jhaaaa (approved):
Sorry for the delay - looks good to me!

PR #298, opened 2018-01-29
History sharding feature, with XRP Ledger Network image

jbheron:
This PR adds a concept document for History Sharding, describes the XRP Ledger Network visually with node stores and shard stores, and features sections for:

- History Sharding
- Acquiring and Sharing Shards
- Comparing the Node Store and Shard Store
- XRP Ledger Network Data Integrity
- Shard Store Configuration
- Sizing the Shard Store

xrp-ledger-network-node-and-shard-stores

Context: https://ripplelabs.atlassian.net/browse/DOC-1272


content/concept-history-sharding.md
```diff
@@ -0,0 +1,57 @@
...
+path=/Users/{username}/.rippled/db/shards/nudb
+max_size_gb=50
+# End shard support
+```
+
+**Tip:** Ripple recommends setting the `type=NuDB` to use the NuDB for sharding, instead of the `RocksDB`.
+
+For more information, reference the `[shard_db]` example in the [rippled.cfg configuration example](https://github.com/ripple/rippled/blob/master/doc/rippled-example.cfg).
+
+***TODO: Question: What reason can we state for recommending NuDB over RocksDB?***
```
miguelportilla: The primary reason would be that NuDB uses far fewer file handles per Shard than RocksDB.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,57 @@
...
+
+***TODO: Question: Does this feature modify the NuDB header format with a network identifier and range of ledgers held, per David's article?***
```
miguelportilla: No. The NuDB header format was not changed.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,57 @@
...
+
+[![XRP Ledger Network: Nodes and Shard Store Diagram](img/xrp-ledger-network-node-and-shard-stores.png)](img/xrp-ledger-network-node-and-shard-stores.png)
+
+## Acquiring and Sharing Shard Stores
+
+Acquiring shards begins after synchronizing with the network and performing ledger backfilling. During this time of lower network activity, the shard database may be asked to select a shard to acquire. If a shard is selected, the ledger acquire process begins by retrieving the sequence of the last ledger in the shard and working backwards toward the first. Shards continue to be acquired until reaching the maximum allocated disk space for shards, at which point the current shard may replace an older shard.
+
+Servers advertise to each other the groups of historical ledgers they have available, and servers maintain a map of the portions of the network they see. Periodically, for example every 256 ledgers, servers make a policy decision whether to start a new current ledger database. If they do, they also decide whether to keep or remove older databases. If they have sufficient disk space, they acquire a new range of ledgers or expand one they currently hold.
```
miguelportilla: The current implementation randomly selects a new shard to acquire if they have sufficient disk space and does not give the current shard special consideration. The logic to specifically consider the current shard with a policy will be implemented in a future version.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Ledger Data Integrity
+
+The history of all ledgers is shared by servers agreeing to keep particular ranges of historical ledgers. This makes it possible for servers to confirm that they have all the data they agreed to maintain, and trivializes the work required to produce proof trees or ledger deltas. Sharding also makes it possible to challenge servers to demonstrate they hold the data they claim to have, and verify their proofs.
+
+When a `rippled` server has too much recent history in the node store, the server has two options. It can discard the shard from recent storage, or it can move the shared from recent storage to historical storage and eject an existing shard. This decision is made following the purge logic for the shard feature.
```
miguelportilla: `the shared from recent` should be `the shard from recent`

miguelportilla: Suggestion: `history in the node store` can be confusing here. It might be a good idea to call it `shard store`.

miguelportilla:
Suggestion: `The primary persistent datastore of recent observed or acquired ledgers, local to each rippled server`

content/concept-history-sharding.md
jbheron: Oh, interesting, I was referring to the `node store` here, and perhaps incorrectly, @miguelportilla ? Specifically, I was describing what I understood to be the process of balancing the two stores (node and shard). Sounds like this logic is constrained to the balancing and maintenance of the `shard store`. Is that accurate?

miguelportilla: Yes, the logic is constrained only to the balancing and maintenance of the shard store.

jbheron: Wonderful, thanks @miguelportilla!

content/concept-history-sharding.md
bachase: RocksDB also requires memory that scales with the size of data its storing, whereas NuDB is meant to have constant memory requirements relative to the size of data it stores. For shard store, users likely want to store many shards, hence RocksDB may require excessive memory overhead.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,57 @@
+# History Sharding
+
+Rippled sharding is a feature that manages storing and sharing of distributed historical data, as shards. As servers operate, they naturally produce a database containing the nodes from the ledgers they witnessed or acquired during network runtime.
```
bachase: Consider re-ordering the content here slightly. It might help motivate the existence of shards before having to define them:


> As servers operate, they naturally produce a database containing the data for the ledgers they witnessed or acquired during network runtime. This ledger data is stored in the node store, but the online delete logic rotates these databases when their size exceeds configured space limitations. 

> Rippled sharding is a feature that manages storing and sharing of distributed historical data as shards, which individually contain data for a segment of ledger history. In this way, Rippled sharding provides a middle ground between running a rippled server with full-history and one with only recent ledger data. History shards do not replace the node store, but implement a reliable path towards distributed ledger history across the rippled network.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,57 @@
...
+
+Ledger data is stored by the node store, but the online delete logic rotates these databases when their size exceeds configured space limitations. In this way, Rippled sharding provides a middle ground between running a rippled server with full-history and one with only recent ledger data. History shards do not replace the node store, but implement a reliable path towards distributed ledger history across the rippled network.
+
+[![XRP Ledger Network: Nodes and Shard Store Diagram](img/xrp-ledger-network-node-and-shard-stores.png)](img/xrp-ledger-network-node-and-shard-stores.png)
+
+## Acquiring and Sharing Shard Stores
+
+Acquiring shards begins after synchronizing with the network and performing ledger backfilling. During this time of lower network activity, the shard database may be asked to select a shard by the acquire ledger process. If a shard is selected, the ledger acquire process begins by retrieving the sequence of the last ledger in the shard and working backwards toward the first. Shards continue to be acquired until reaching the maximum allocated disk space for shards, at which point the current shard may replace an older shard.
```
bachase: > at which point the current shard may replace an older shard.

Is that true @miguelportilla ? Does that mean if you are trying to acquire shard, but run out of space before you finish it, you throw it away and try another one?

content/concept-history-sharding.md
```diff
@@ -0,0 +1,57 @@
...
+
+Servers advertise to each other the groups of historical ledgers they have available, and servers maintain a map of the portions of the network they see. Periodically, for example every 256 ledgers, servers make a policy decision whether to start a new current ledger database. If they do, they also decide whether to keep or remove older databases. If they have sufficient disk space, they acquire a new range of ledgers or expand one they currently hold.
```
bachase: > map of the portions of the network they see.

Is this a map of what peers it knows about? Is that independent of sharding?

content/concept-history-sharding.md
```diff
@@ -0,0 +1,57 @@
...
+
+## XRP Ledger Network Data Integrity
+
+The history of all ledgers is shared by servers agreeing to keep particular ranges of historical ledgers. This makes it possible for servers to confirm that they have all the data they agreed to maintain, and trivializes the work required to produce proof trees or ledger deltas. Sharding also makes it possible to challenge servers to demonstrate they hold the data they claim to have, and verify their proofs.
```
bachase: I'd only caution that it is not currently trivial to do ledger proofs. This will certainly help, but I don't think we have good tools (yet) to do it easily.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,57 @@
...
+
+When a `rippled` server has too much recent history in the node store, the server has two options. It can discard the shard from recent storage, or it can move the shared from recent storage to historical storage and eject an existing shard. This decision is made following the purge logic for the shard feature.
+
+When a `rippled` server has more space than the configured allocation of shard store size, it assigns storage to begin storing shards or to store more shards. To increase the probability for an even distribution of the history of the network's ledgers, shards are selected randomly among the shards not currently held, with no preferences. Once chosen, the selected shard is fetched, validated, and committed to long term storage.
```
bachase: > to begin storing shards or to store more shards.

I'm having trouble parsing that distinction.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,57 @@
...
+
+## XRP Ledger Network Data Integrity
+
+The history of all ledgers is shared by servers agreeing to keep particular ranges of historical ledgers. This makes it possible for servers to confirm that they have all the data they agreed to maintain, and trivializes the work required to produce proof trees or ledger deltas. Sharding also makes it possible to challenge servers to demonstrate they hold the data they claim to have, and verify their proofs.
+
+When a `rippled` server has too much recent history in the node store, the server has two options. It can discard the shard from recent storage, or it can move the shared from recent storage to historical storage and eject an existing shard. This decision is made following the purge logic for the shard feature.
```
bachase: Is the purge logic what determines whether the extra data from the node store gets moved to the shard store? Is it configurable? (@miguelportilla )

content/concept-history-sharding.md
```diff
@@ -0,0 +1,57 @@
...
+For more information, reference the `[shard_db]` example in the [rippled.cfg configuration example](https://github.com/ripple/rippled/blob/master/doc/rippled-example.cfg).
+
+***TODO: Question: What reason can we state for recommending NuDB over RocksDB?***
+
+***TODO: Question: Does this feature modify the NuDB header format with a network identifier and range of ledgers held, per David's article?***
+
+***TODO: Followup Question: If so, and if we should document the header and any relevant concerns here, what might those be?***
+
+### Sizing the Shard Store
+Determining a suitable value for the shard store will involve carefully considering different aspects of the process. Although redundant, it is entirely possible to hold full history in the node store and the shard store. Servers can shrink the range of ledgers they hold by dropping a database and then re-acquiring whatever ledgers they decide they do want to hold.
```
bachase: Is the `dropping a database and then re-acquiring` in reference to the node store, shard store, or both?

content/concept-history-sharding.md
```diff
@@ -0,0 +1,57 @@
...
+
+The following concerns should each be considered in determining shard store sizing:
+
+- Each shard of the node store is 2^14 ledgers (at time of publishing, Q1 2018), which may not be optimal.
```
bachase: Is it possible to replace this with more of a recipe for configuration to make it more actionable?  Or if not a general recipe, suggested scenarios (e.g. full-history, minimal node store + larger shard store, medium node store + medium shard store)?

content/concept-history-sharding.md
bachase: Oh, do you mean storing the *first* shard or storing *>= second*?

bachase (commented):
Some of my comments might need @miguelportilla to weigh in.

I like the picture too. Consider having a few of the nodes without the shard store layer to indicate it is optional.

content/concept-history-sharding.md
jbheron: I think so! I'll admit, that's from something David wrote, and I had to read it a couple of times to come up with the same conclusion you described. Sounds like it could be made more clear.

jbheron:
@bachase, thanks for the 👍  on the image. Do you think noting whether nodes operate as validators is helpful in visualizing the network, or detracts from the model of disambiguating node stores from shard stores? I got thinking about it how it could be improved and wasn't sure whether it helped or not?

bachase:
@jbheron Good question. I'd lean towards not indicating validator vs not validator to keep it about the different store types. However, I think adding text that both validators and tracking nodes can run shard stores is a good idea. Perhaps that could go in the last section with a note that we don't recommend validators run shard stores due to the added overhead that isn't essential for the validation process. That is, its better to run two nodes--one a validator and one for managing your ledger history.

content/concept-history-sharding.md
jbheron: It is meant to refer to the shard store, but your question makes me realize they would presumably not choose the ledgers to subsequently acquire, as the acquire process offers only random consideration. So, they'd get a new random distribution of history shards.  I'll update with that in mind.

content/concept-history-sharding.md
miguelportilla: @bachase A shard being acquired is not thrown away if the system runs out of space. Instead, progress will be halted until space becomes available. Logic dealing with the current shard replacing an older shard has not been implemented yet.  `current shard` meaning the latest complete shard, consisting of the newest 16384 ledgers.

content/concept-history-sharding.md
miguelportilla: @bachase Yes and it is independent of sharding. That function occurs in `RCLConsensus::Adaptor::notify` which notifies other peers via the Overlay. 'map' is inaccurate, as only a first and last ledger index is advertised.

@jbheron Is this paragraph referring the shard store or the node store?

content/concept-history-sharding.md
miguelportilla: @bachase  This logic has not been implemented and not planned. Because of their size, moving a shard into the node store or vice versa would be problematic.

@jbheron I suggest removing this paragraph.

content/concept-history-sharding.md
miguelportilla: "When a `rippled` server has more space than the configured allocation of shard store size, it assigns storage to begin storing shards or to store more shards."

This should be simplified. The conditions for acquiring and storing a shard are that there is enough free disk space available to store a complete shard and that the complete shard size does not exceed the remaining free space of the maximum space allocated in the config file (max_size_gb).

content/concept-history-sharding.md
mDuo13: Style nit: Always `rippled` (lowercase, code-font)

Although in this particular case, I would start it with "History sharding" instead of "Rippled sharding"

content/concept-history-sharding.md
jbheron: @miguelportilla, thank you for the clarification. It is intended to reference the shard store, and comes from David's original notes in confluence about the feature. I see from your PR that there are some details here that are confusing. I'm going to rework this section a little.

content/concept-history-sharding.md
mDuo13: Suggested rewrite of the first sentence:

> Historical sharding distributes the transaction history of the XRP Ledger into segments, called _shards_, across servers in the XRP Ledger network. Using this feature, individual `rippled` servers can contribute to storing historical data without needing to store the entire (multiple terabyte) history.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,57 @@
+# History Sharding
+
+Rippled sharding is a feature that manages storing and sharing of distributed historical data, as shards. As servers operate, they naturally produce a database containing the nodes from the ledgers they witnessed or acquired during network runtime.
+
+Ledger data is stored by the node store, but the online delete logic rotates these databases when their size exceeds configured space limitations. In this way, Rippled sharding provides a middle ground between running a rippled server with full-history and one with only recent ledger data. History shards do not replace the node store, but implement a reliable path towards distributed ledger history across the rippled network.
```
mDuo13: Several style nitpicks:

- Passive→Active in the first sentence. "Each `rippled` server stores ledger data in its node store, but..." instead of "Ledger data is stored..., but..."
- Change "Rippled sharding" to "history sharding".
- Strike "In this way," from the second sentence.
- Instead of "rippled network" in the last sentence, say "XRP Ledger network"

content/concept-history-sharding.md
```diff
@@ -0,0 +1,57 @@
...
+
+[![XRP Ledger Network: Nodes and Shard Store Diagram](img/xrp-ledger-network-node-and-shard-stores.png)](img/xrp-ledger-network-node-and-shard-stores.png)
```
mDuo13: In the diagram's Shard Store section, I'd recommend changing:

> a proportional percentage of the distributed, historical database

To:

> a segment of the distributed historical database

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Acquiring and Sharing Shard Stores
+
+Acquiring shards begins after synchronizing with the network and performing ledger backfilling. During this time of lower network activity, the shard database may be asked to select a shard by the acquire ledger process. To increase the probability for an even distribution of the history of the network's ledgers, shards are randomly selected for acquisition, and the current shard is given no special consideration. If a shard is selected, the ledger acquire process begins by fetching the sequence of the last ledger in the shard and working backwards toward the first. If the system runs out of space before completely acquiring the shard, progress is halted until space becomes available. Shards continue to be acquired randomly until reaching the maximum allocated disk space for shards.
```
mDuo13: > performing ledger backfilling

Let's be more specific:

> backfilling ledger history to the configured number of recent ledgers

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
+# History Sharding
+
+As servers operate, they naturally produce a database containing the nodes from the ledgers they witnessed or acquired during network runtime. This ledger data is stored by the `node store`, but the online delete logic rotates these databases when their size exceeds configured space limitations.
+
+Rippled sharding is a feature that manages storing and sharing of distributed historical data as shards, which individually contain data for a segment of ledger history. In this way, Rippled sharding provides a middle ground between running a rippled server with full-history and one with only recent ledger data. A `shard store` does not replace a `node store`, but implements a reliable path towards distributed ledger history across the XRP Ledger Network.
+
+[![XRP Ledger Network: Nodes and Shard Store Diagram](img/xrp-ledger-network-node-and-shard-stores.png)](img/xrp-ledger-network-node-and-shard-stores.png)
+
+## Acquiring and Sharing Shard Stores
```
mDuo13: "Shard Stores"→"History Shards"

content/concept-history-sharding.md
mDuo13: Also, let's begin with a statement that `rippled` servers only acquire shards of historical data if they are configured to store history shards.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
...
+
+Acquiring shards begins after synchronizing with the network and performing ledger backfilling. During this time of lower network activity, the shard database may be asked to select a shard by the acquire ledger process. To increase the probability for an even distribution of the history of the network's ledgers, shards are randomly selected for acquisition, and the current shard is given no special consideration. If a shard is selected, the ledger acquire process begins by fetching the sequence of the last ledger in the shard and working backwards toward the first. If the system runs out of space before completely acquiring the shard, progress is halted until space becomes available. Shards continue to be acquired randomly until reaching the maximum allocated disk space for shards.
+
+Servers share historical data when peer `rippled` servers request selected historical ledgers if those `rippled` servers have that data available. From the peer responses, the historical data for that shard is assembled and stored until the ledger data for that shard is complete. If there is sufficient disk space, the `rippled` server acquires additional shards to add to the `shard store`
```
mDuo13: - I'd strike "if those `rippled` servers have that data available" from the first sentence. I think it makes more sense without that clause.

- Also, fix passive voice:
    
    > From the peer responses, the historical data for that shard is assembled and stored
    
    To:
    
    > A `rippled` server configured with a shard database assembles historical data from a range of ledgers by requesting those ledgers from its peers. The shard is complete when it contains all the ledgers in a specific range. If the `rippled`'s shard store has sufficient disk space after completing a shard, it acquires additional shards.

- Let's not put "shard store" or "node store" in code-font.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Comparing the Node Store and Shard Store
+
+Like the `node store`, the `shard store` derives from the base class `Database`. The common interface facilitates replacing the `node store` when storing or fetching from various places in the project. For example, the inbound ledger process uses synchronization filters to save data received from peers to either store. Specifying the database object for the filter allows `rippled` to decide which store receives the data. Similarly, a `Family` tells `SHAMaps` which database to use when fetching or storing. The derived stores themselves rely on these commonalities when copying ledgers from one store to another or validating their state and transaction trees.
```
mDuo13: Let's not delve into the specifics of the class hierarchy. Instead, I suggest:

> The shard store shares code and a common interface with the node store where `rippled` stores its recent ledger history.

mDuo13: What are synchronization filters?

mDuo13: Honestly I think this chunk of documentation might be best moved to a section more clearly aimed at `rippled` contributors rather than people just running, administering, and interfacing with `rippled`. I suspect that most or all of this info is not relevant for people who aren't reading or contributing to the `rippled` source code.

Maybe a file in the rippled repository is the right place for this info.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
...
+
+## XRP Ledger Network Data Integrity
+
+The history of all ledgers is shared by servers agreeing to keep particular ranges of historical ledgers. This makes it possible for servers to confirm that they have all the data they agreed to maintain, and produce proof trees or ledger deltas. Sharding also makes it possible to challenge servers to demonstrate they hold the data they claim to have, and verify their proofs.
+
+## Shard Configuration
+Rippled can be configured to begin storing ledger history in shards by adding the `shard_db` section to the `rippled.cfg` file.
```
mDuo13: Another "Rippled"→`rippled` instance

mDuo13: Also, passive voice. Suggested change:

> To configure your `rippled` to store shards of ledger history, add a `shard_db` section to your `rippled.cfg` file.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Shard Configuration Example
+The following snippet from an example `rippled.cfg` file shows the configuration fields for adding sharding to a `rippled` server:
+
+```
+# Begin Shard support
+[shard_db]
+type=NuDB
+path=/Users/{username}/.rippled/db/shards/nudb
```
mDuo13: Let's change this path to match the example node store path:

```
path=/var/lib/rippled/db/shards/nudb
```

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
...
+max_size_gb=50
+# End shard support
```
mDuo13: I suggest removing the "Begin shard support" / "end shard support" comments from the example.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+**Tip:** Ripple recommends using NuDB for sharding (`type=NuDB`). NuDB uses fewer file handles per shard than RocksDB, while RocksDB uses memory that scales with the size of data its storing, which, with many shards may require excessive memory overhead.
```
mDuo13: I suggest splitting up the "NuDB uses..., while RocksDB uses..., which..." sentence. Also, some copy-edits (wrong "it's", moving "with many shards")

> NuDB uses fewer file handles and less memory per shard than RocksDB. When storing many shards, RocksDB may require excessive memory to store the shard database.

mDuo13: Also, to be crystal clear, let's say "Ripple recommends using NuDB for the shard store" instead of "...for sharding."

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
...
+
+For more information, reference the `[shard_db]` example in the [rippled.cfg configuration example](https://github.com/ripple/rippled/blob/master/doc/rippled-example.cfg).
```
mDuo13: "reference the"→"see the"

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Sizing the Shard Store
+Determining a suitable value for the shard store will involve carefully considering different aspects of the process. Although redundant, it is entirely possible to hold full history in the node store and the shard store. Servers can shrink the range of ledgers they hold by dropping the shard database and then re-acquiring ledgers through the acquisition process.
```
mDuo13: - "suitable value" → "suitable size"
- "different aspects of the process": which process?
- > Servers can shrink the range of ledgers they hold by dropping the shard database and then re-acquiring ledgers through the acquisition process.
    
    I'm not sure I understand this. Are you talking about if you want to stop using the shard store? Or is this backwards, and you mean if you're going to start using the shard store, you can configure your server to have less recent history and use some of that space to store a shard of older history?

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
...
+
+The following concerns should each be considered in determining shard store sizing:
```
mDuo13: PV. "should each be considered in determining shard store sizing"→"you should consider each... when deciding what size your shard store should be"

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
...
+
+- Each shard of the node store is 2^14 ledgers (at time of publishing, Q1 2018), which may not be optimal.
```
mDuo13: Strike "which may not be optimal." Also, generally, you don't need to add "at time of publishing" for things like this—this is web documentation; we can change it later if the number of ledgers in a shard changes. That'll be part of a `rippled` release, so we'd definitely have fair warning.

It would be useful to add estimates / ranges of:

1. How much disk space a single shard occupies.
2. How long a time period one shard represents.

_Those_ may be something that would justify an "at time of publishing" since that changes dynamically due to factors outside of our control.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
...
+- An effective configuration might limit the node store only to recent history.
```
mDuo13: I suggest you change this to:

> You should leave room for the node store to keep recent history.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
...
+- The node store history size should at minimum be twice the ledgers per shard, due to the fact that the current shard may be chosen to be stored and it would be wasteful to reacquire that data.
+- The unit of work performed when the server decides to retain the current shard.
+- The time to acquire, number of file handles, and memory cache usage is directly affected by sizing.
+
+**Tip:** While both validator and tracking (or stock) `rippled` servers can run shard stores, Ripple recommends adding history sharding only for non-validator nodes to reduce overhead for validator nodes. If you run a validator node and want to manage ledger history with sharding, run a separate `rippled` server with a sharding.
```
mDuo13: Avoid the phrase "validator node". Instead, say "validating server" or just "validator".

content/concept-history-sharding.md
```diff
@@ -0,0 +1,53 @@
...
+For more information, reference the `[shard_db]` example in the [rippled.cfg configuration example](https://github.com/ripple/rippled/blob/master/doc/rippled-example.cfg).
+
+### Sizing the Shard Store
+Determining a suitable value for the shard store will involve carefully considering different aspects of the process. Although redundant, it is entirely possible to hold full history in the node store and the shard store. Servers can shrink the range of ledgers they hold by dropping the shard database and then re-acquiring ledgers through the acquisition process.
+
+The following concerns should each be considered in determining shard store sizing:
+
+- Each shard of the node store is 2^14 ledgers (at time of publishing, Q1 2018), which may not be optimal.
+- An effective configuration might limit the node store only to recent history.
+- The node store history size should at minimum be twice the ledgers per shard, due to the fact that the current shard may be chosen to be stored and it would be wasteful to reacquire that data.
```
mDuo13: Passive voice. Also, when you say, "the current shard", do you mean the most recent ledger data?

mDuo13 (changes requested):
Lots of suggestions on wording and style. This took me a while, so some of them are on outdated diffs; sorry. I think some of the comments may still be relevant even if the sentence they comment on has changed a little since.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,56 @@
...
+
+Rippled sharding is a feature that manages storing and sharing of distributed historical data as shards, which individually contain data for a segment of ledger history and are stored separately, in a shard store. In this way, Rippled sharding provides a middle ground between running a rippled server with full-history and one with only recent ledger data. A shard store does not replace a node store, but implements a reliable path towards distributed ledger history across the XRP Ledger Network.
+
+[![XRP Ledger Network: Nodes and Shard Store Diagram](img/xrp-ledger-network-node-and-shard-stores.png)](img/xrp-ledger-network-node-and-shard-stores.png)
+
+## Acquiring and Sharing Shards
+
+Acquiring shards begins after synchronizing with the network and performing ledger backfilling. During this time of lower network activity, a `rippled` server set to maintain a `shard_db` randomly chooses a shard to add to its shard store. To increase the probability for an even distribution of the network ledger history, shards are randomly selected for acquisition, and the current shard is given no special consideration.
```
MarkusTeufelberger: What exactly is a "shard" anyways? A leaf node? A full SHAmap? A range of ledgers with all their objects + InnerNodes?

Sounds like the latter ("a range of ledgers and all their objects").

content/concept-history-sharding.md
```diff
@@ -0,0 +1,56 @@
...
+[shard_db]
+type=NuDB
+path=/Users/{username}/.rippled/db/shards/nudb
+max_size_gb=50
+# End shard support
+```
+
+**Tip:** Ripple recommends using NuDB for sharding (`type=NuDB`). NuDB uses fewer file handles per shard than RocksDB, while RocksDB uses memory that scales with the size of data its storing, which, with many shards may require excessive memory overhead.
+
+**Tip:** While both validator and tracking (or stock) `rippled` servers can run shard stores, Ripple recommends adding history sharding only for non-validator nodes to reduce overhead for validator nodes. If you run a validator node and want to manage ledger history with sharding, run a separate `rippled` server with a sharding.
```
MarkusTeufelberger: "with a sharding" sounds strange

content/concept-history-sharding.md
```diff
@@ -0,0 +1,56 @@
...
+
+For more information, reference the `[shard_db]` example in the [rippled.cfg configuration example](https://github.com/ripple/rippled/blob/master/doc/rippled-example.cfg).
+
+### Sizing the Shard Store
+Determining a suitable value for the shard store will involve carefully considering different aspects of the process. Although redundant, it is entirely possible to hold full history in the node store and the shard store. Servers can shrink the range of ledgers they hold by dropping the shard database and then re-acquiring ledgers through the acquisition process.
+
+The following concerns should each be considered in determining shard store sizing:
+
+- Each shard of the node store is 2^14 ledgers (at time of publishing, Q1 2018), which may not be optimal
```
MarkusTeufelberger: 16384 for those that don't know the powers of 2 by heart...

MarkusTeufelberger: Also it would be interesting to have a frame of reference for the amount of data necessary for a shard over time, is this hundreds of GB or just 3?

content/concept-history-sharding.md
```diff
@@ -0,0 +1,56 @@
...
+- An effective configuration might limit the node store only to recent history
+- The node store history size should at minimum be twice the ledgers per shard, due to the fact that the current shard may be chosen to be stored and it would be wasteful to reacquire that data
+- The unit of work performed when the server decides to retain the current shard
```
MarkusTeufelberger: Which unit?

content/concept-history-sharding.md
```diff
@@ -0,0 +1,56 @@
+# History Sharding
+
+As servers operate, they naturally produce a database containing the nodes from the ledgers they witnessed or acquired during network runtime. This ledger data is stored by the node store, but the online delete logic rotates these databases when their size exceeds configured space limitations.
```
MarkusTeufelberger: > when their size exceeds configured space limitations

Would be cool, but at the moment it is only possible to configure a number of stored ledgers there, right?

content/concept-history-sharding.md
MarkusTeufelberger: A different way to express this is that 1 shard is a bit more than half a day of data at current speeds (full state + changes)

content/concept-history-sharding.md
miguelportilla: @MarkusTeufelberger Yes, a shard is a range of ledgers stored in a database. Ledgers are stored in the node or shard store exactly the same. A ledger is comprised of a header and two SHAMaps.

content/concept-history-sharding.md
miguelportilla: Shard size depends on ledger size and early history is much smaller than current. I can't recall the exact sizes from memory but shards with early history are a couple of hundred MB where late shards near 4GB.

content/concept-history-sharding.md
miguelportilla: I believe this would apply to the possibility of swapping an older shard for the current shard. This logic is not implemented in the current version. This text should probably be removed.

content/concept-history-sharding.md
miguelportilla: @MarkusTeufelberger Yes, you are correct.

content/concept-history-sharding.md
jbheron: Fair point, I'm striking it as TMI.

content/concept-history-sharding.md
jbheron: Nice catch, thanks @MarkusTeufelberger!

content/concept-history-sharding.md
jbheron: Ah, interesting. OK, changing to:

> Each `rippled` server stores ledger data in its node store, but the online delete logic rotates these databases when the number of stored ledgers exceeds configured space limitations.

jbheron:
Thanks for the review notes, @miguelportilla, @bachase, @mDuo13, and @MarkusTeufelberger! I really appreciate them. 😸

content/img-sources/xrp-ledger-network-node-and-shard-stores.txt
```diff
@@ -0,0 +1 @@
+https://docs.google.com/drawings/d/15rIV9apyiVZo6lquys3CBSY9iyiMc0ZorMXOzYN4lW8/edit?usp=sharing
```
mDuo13, line 1: Couple more comments on the current diagram:

- I suggest removing "(validator or tracking)" from the node store header. What you really mean is "all servers" and that's apparent from the diagram, so I think it's fine to remove the parenthetical entirely.
- I still think you should revise the description of the shard store to change "proportional percentage" to "a segment of the distributed historical database"

content/concept-history-sharding.md
```diff
@@ -0,0 +1,51 @@
...
+
+Once a shard is selected, the ledger acquire process begins by fetching the sequence of the last ledger in the shard and working backwards toward the first. The retrieval process begins with the server checking for the data locally. For data that is not available, the server requests data from its peer `rippled` servers. Those servers that have the data available for the requested period respond with their history. The requesting server combines those responses to create the shard. The shard is complete when it contains all the ledgers in a specific range. If the shard store has sufficient disk space after completing a shard, it acquires additional shards.
```
mDuo13: > If the shard store has sufficient disk space after completing a shard, it acquires additional shards.

This last sentence is now redundant with the more thorough explanation in the following paragraph.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,51 @@
...
+
+Using the history sharding feature, individual `rippled` servers can contribute to storing historical data without needing to store the entire (multiple terabyte) history. A shard store does not replace a node store, but implements a reliable path towards distributed ledger history across the XRP Ledger Network.
+
+[![XRP Ledger Network: Nodes and Shard Store Diagram](img/xrp-ledger-network-node-and-shard-stores.png)](img/xrp-ledger-network-node-and-shard-stores.png)
+
+## Acquiring and Sharing History Shards
+
+Only `rippled` servers that are configured to store history shards acquire history shards. For those servers, acquiring shards begins after synchronizing with the network and backfilling ledger history to the configured number of recent ledgers. During this time of lower network activity, a `rippled` server set to maintain a `shard_db` randomly chooses a shard to add to its shard store. To increase the probability for an even distribution of the network ledger history, shards are randomly selected for acquisition, and the current shard is given no special consideration.
```
mDuo13: > Only `rippled` servers that are configured to store history shards acquire history shards.

Suggested rephrase:

> `rippled` servers acquire and store history shards only if configured to do so.

content/concept-history-sharding.md
```diff
@@ -0,0 +1,51 @@
...
+
+If a `rippled` server runs out of space before completely acquiring a shard, it stops its retrieval process until it has space available to continue. After that point, the most recently completed shard may replace an older shard. If there is sufficient disk space, the `rippled` server acquires additional randomly selected shards to add to the shard store until reaching the maximum allocated disk space for shards (`max_size_gb`).
+
+## XRP Ledger Network Data Integrity
+
+The history of all ledgers is shared by servers agreeing to keep particular ranges of historical ledgers. This makes it possible for servers to confirm that they have all the data they agreed to maintain, and produce proof trees or ledger deltas. Sharding also makes it possible to challenge servers to demonstrate they hold the data they claim to have, and verify their proofs.
```
mDuo13: > Sharding also makes it possible to challenge servers to demonstrate they hold the data they claim to have, and verify their proofs.

Is this true in the current implementation? If not, we should strike it until it's implemented. If it is, I'd like to see docs on how to do this (make a ticket in the backlog for that task if it's not trivial).

content/concept-history-sharding.md
```diff
@@ -0,0 +1,51 @@
...
+
+To configure your `rippled` to store shards of ledger history, add a `shard_db` section to your `rippled.cfg` file.
+
+### Shard Configuration Example
+The following snippet from an example `rippled.cfg` file shows the configuration fields for adding sharding to a `rippled` server:
+
+```
+[shard_db]
+type=NuDB
+path=/Users/rippled/.rippled/db/shards/nudb
```
mDuo13: `/Users/` is a macOS path, which is not a supported OS for production. Let's use a path that makes sense on Red Hat and/or Ubuntu Linux, such as `/home/rippled/.rippled/db/shards/nudb` or `/var/lib/rippled/db/shards/nudb`

content/concept-history-sharding.md
```diff
@@ -0,0 +1,51 @@
...
+max_size_gb=50
+```
+
+**Tip:** Ripple recommends using NuDB for the shard store (`type=NuDB`). NuDB uses fewer file handles per shard than RocksDB. RocksDB uses memory that scales with the size of data it stores, which may require excessive memory overhead.
+
+**Tip:** While both validator and tracking (or stock) `rippled` servers can be configured to use history shard stores, Ripple recommends adding history sharding only for non-validator nodes to reduce overhead for validator nodes. If you run a validator node and want to manage ledger history using sharding, run a separate `rippled` server with sharding enabled.
```
mDuo13: change "validator node" to "validator" here

content/concept-history-sharding.md
```diff
@@ -0,0 +1,51 @@
...
+
+For more information, reference the `[shard_db]` example in the [rippled.cfg configuration example](https://github.com/ripple/rippled/blob/master/doc/rippled-example.cfg).
+
+### Sizing the Shard Store
+Determining a suitable size for the shard store involves careful consideration. The following concerns contribute to determining shard store sizing:
+
+- Although redundant, it is possible to hold full ledger history in the node store and the history shard store
+- Each shard of the node store is 2^14 ledgers (16384), which may not be optimal
```
mDuo13: > which may not be optimal

I don't think this phrase is useful here. It feels like it's a critique of the implementation on a parameter that's hard-coded. Let's remove it.

mDuo13 (changes requested):
The edits look pretty good, but I think a few of my original comments got lost. Apologies if you were still working on them and I'm just repeating myself here

content/concept-history-sharding.md
jbheron: Nice catch!

content/concept-history-sharding.md
jbheron: As  I understand it, that is possible, but difficult, with the current tooling. As such, yeah, that makes sense. I'll make a ticket about it and strike it here.

jbheron: See https://ripplelabs.atlassian.net/browse/DOC-1292

content/concept-history-sharding.md
```diff
@@ -30,21 +30,21 @@ The following snippet from an example `rippled.cfg` file shows the configuration
...
+**Tip:** While both validator and tracking (or stock) `rippled` servers can be configured to use history shard stores, Ripple recommends adding history sharding only for non-validator nodes to reduce overhead for validator nodes. If you run a validator and want to manage ledger history using sharding, run a separate `rippled` server with sharding enabled.
 
 For more information, reference the `[shard_db]` example in the [rippled.cfg configuration example](https://github.com/ripple/rippled/blob/master/doc/rippled-example.cfg).
 
 ### Sizing the Shard Store
 Determining a suitable size for the shard store involves careful consideration. The following concerns contribute to determining shard store sizing:
 
 - Although redundant, it is possible to hold full ledger history in the node store and the history shard store
-- Each shard of the node store is 2^14 ledgers (16384), which may not be optimal
+- Each shard of the node store is 2^14 ledgers (16384)
```
mDuo13: Just noticed: I think rather than "shard of the node store is 2^14 ledgers" it should read "Each shard consists of 2^14 ledgers"

content/concept-history-sharding.md
```diff
@@ -30,21 +30,21 @@ The following snippet from an example `rippled.cfg` file shows the configuration
...
 
-**Tip:** While both validator and tracking (or stock) `rippled` servers can be configured to use history shard stores, Ripple recommends adding history sharding only for non-validator nodes to reduce overhead for validator nodes. If you run a validator node and want to manage ledger history using sharding, run a separate `rippled` server with sharding enabled.
+**Tip:** While both validator and tracking (or stock) `rippled` servers can be configured to use history shard stores, Ripple recommends adding history sharding only for non-validator nodes to reduce overhead for validator nodes. If you run a validator and want to manage ledger history using sharding, run a separate `rippled` server with sharding enabled.
 
 For more information, reference the `[shard_db]` example in the [rippled.cfg configuration example](https://github.com/ripple/rippled/blob/master/doc/rippled-example.cfg).
 
 ### Sizing the Shard Store
 Determining a suitable size for the shard store involves careful consideration. The following concerns contribute to determining shard store sizing:
```
mDuo13: > The following concerns contribute to determining shard store sizing:

Suggested rephrase:

> You should consider the following when deciding what size your shard store should be:

content/concept-history-sharding.md
mDuo13: Also, let's add a comment giving a ballpark of how much disk space and time a ledger shard consists of.

For disk size, let's say:

> A shard occupies approximately 200 MB to 4 GB based on the age of the shard. Older shards are smaller because there was less activity in the XRP Ledger at the time.

As for time, my calculations tell me the oldest ledgers shards occupy about 85 hours' worth of time, and the most recent shards occupy about 16-17 hours. So how about this?

> A shard represents about 10 to 100 hours of time, depending on the rate at which new ledgers are closed. As of early 2018, recent shards tend to represent about 15-20 hours.

mDuo13 (changes requested):
This is looking really good now. Just a couple things left:

1) Add a "New in `rippled` 0.90.0" message to the top of the file. When we actually release 0.90.0, we can change this to a badge that links the relevant release notes, as we do in several other places.

2. As discussed on Slack, change all instances of "node store" to "ledger store"

3. Left a couple more comments on the sizing section

content/concept-history-sharding.md
jbheron: Going to say 'contains', as they actually consist of headers and two SHAMaps -- hope that makes sense.

jbheron: Great!

content/concept-history-sharding.md
```diff
@@ -0,0 +1,55 @@
...
+
+# History Sharding
+
+As servers operate, they naturally produce a database containing data about the ledgers they witnessed or acquired during network runtime. Each `rippled` server stores that ledger data in its ledger store, but the online delete logic rotates these databases when the number of stored ledgers exceeds configured space limitations.
+
+Historical sharding distributes the transaction history of the XRP Ledger into segments, called shards, across servers in the XRP Ledger network. A shard is a range of ledgers, consisting of a header and two SHAMaps. A `rippled` server stores ledgers in both the ledger store and the shard store in the same way.
+
+Using the history sharding feature, individual `rippled` servers can contribute to storing historical data without needing to store the entire (multiple terabyte) history. A shard store does not replace a ledger store, but implements a reliable path towards distributed ledger history across the XRP Ledger Network.
+
+[![XRP Ledger Network: Nodes and Shard Store Diagram](img/xrp-ledger-network-node-and-shard-stores.png)](img/xrp-ledger-network-node-and-shard-stores.png)
```
mDuo13: Don't forget to change the diagram alt-text (and optionally, filename)

mDuo13 (approved):

PR #301, opened 2018-02-07
Amendments: how to tell if rippled is amendment blocked

jhaaaa:
* Adds a few tips for verifying that your `rippled` server is amendment blocked.
* Adds a tip for finding out which amendments are causing the block.
* Will add a bit more details about how and why a `rippled` server becomes amendment blocked.

### DIFF
--- a/content/concept-amendments.md
+++ b/content/concept-amendments.md
@@ -86,16 +86,16 @@ When an amendment gets enabled for the network after the voting process, servers
 * Do not participate in the consensus process
 * Do not vote on future amendments
 
-Becoming amendment blocked is a security feature to protect backend applications. Rather than guessing and maybe misinterpreting a ledger after new rules have been applied, `rippled` reports that it does not know the state of the ledger because it does not know how the amendment works.
+Becoming amendment blocked is a security feature to protect applications that depend on XRP Ledger data. Rather than guessing and maybe misinterpreting a ledger after new rules have been applied, `rippled` reports that it does not know the state of the ledger because it does not know how the amendment works.
 
-The amendments that a `rippled` server is configured to vote for or against have no impact on whether the server becomes amendment blocked. A `rippled` server always follows the set of amendments enabled by the rest of the network, to the extent possible. A server only becomes amendment blocked if the enabled amendment is not included in the amendment definitions compiled into the server's source code -- in other words, if the amendment is newer than the server.
+The amendments that a `rippled` server is configured to vote for or against have no impact on whether the server becomes amendment blocked. A `rippled` server always follows the set of amendments enabled by the rest of the network, to the extent possible. A server only becomes amendment blocked if an enabled amendment is not included in the amendment definitions compiled into the server's source code -- in other words, if the amendment is newer than the server.
 
 If your server is amendment blocked, you must [upgrade to a new version](tutorial-rippled-setup.html#updating-rippled) to sync with the network.
 
 
 #### How to Tell If Your `rippled` Server Is Amendment Blocked
 
-One of the first signs that your `rippled` server is amendment blocked is an `amendmentBlocked` error that is returned when you submit a transaction. Here's an example `amendmentBlocked` error:
+One of the first signs that your `rippled` server is amendment blocked is an `amendmentBlocked` error that is returned [when you submit a transaction](reference-rippled.html#submit). Here's an example `amendmentBlocked` error:
 
 ```
 {
@@ -215,8 +215,6 @@ In this example, conflicts with the following features are causing your `rippled
 * `F64E1EABBE79D55B3BB82020516CEC2C582A98A6BFE20FBE9BB6A0D233418064`
 
 To look up which `rippled` version supports these features, see [Known Amendments](reference-amendments.html).
-
-***TODO: Question: Do we think this is the workflow users will use to figure this out? If yes, then it might be nice for the user if the table at the top of the Known Amendments page would list amendment IDs. I know that it makes the table a little uglier, but the feature response doesn't always provide the short name and always provides the ID. The way the page is currently structured, the user will need to ctrl+f for the ID, note the short name, and then scroll back up to the table at the top to note the version. They'll need to do this for each amendment.***
 
 
 ## Testing Amendments
--- a/content/concept-amendments.md
+++ b/content/concept-amendments.md
@@ -112,7 +112,7 @@ One of the first signs that your `rippled` server is amendment blocked is an `am
 }
 ```
 
-The following `rippled` error also indicates that your server is amendment blocked:
+The following `rippled` log message also indicates that your server is amendment blocked:
 
 ```
 2018-Feb-12 19:38:30 LedgerMaster:ERR One or more unsupported amendments activated: server blocked.
--- a/content/concept-amendments.md
+++ b/content/concept-amendments.md
@@ -86,11 +86,120 @@ When an amendment gets enabled for the network after the voting process, servers
 * Do not participate in the consensus process
 * Do not vote on future amendments
 
-Becoming amendment blocked is a security feature to protect backend applications. Rather than guessing and maybe misinterpreting a ledger after new rules have applied, `rippled` reports that it does not know the state of the ledger because it does not know how the amendment works.
+Becoming amendment blocked is a security feature to protect backend applications. Rather than guessing and maybe misinterpreting a ledger after new rules have been applied, `rippled` reports that it does not know the state of the ledger because it does not know how the amendment works.
 
 The amendments that a `rippled` server is configured to vote for or against have no impact on whether the server becomes amendment blocked. A `rippled` server always follows the set of amendments enabled by the rest of the network, to the extent possible. A server only becomes amendment blocked if the enabled amendment is not included in the amendment definitions compiled into the server's source code -- in other words, if the amendment is newer than the server.
 
 If your server is amendment blocked, you must [upgrade to a new version](tutorial-rippled-setup.html#updating-rippled) to sync with the network.
+
+
+#### How to Tell If Your `rippled` Server is Amendment Blocked
+
+***TODO: Question: Okay location for this doc? It is building upon a concept, but is tutorial-ish.***
+
+One of the first signs that your `rippled` server is amendment blocked is that an `amendmentBlocked` error is returned when you submit a transaction. Here's an example `amendmentBlocked` error:
+
+```
+{
+   "result":{
+      "error":"amendmentBlocked",
+      "error_code":14,
+      "error_message":"Amendment blocked, need upgrade.",
+      "request":{
+         "command":"submit",
+         "tx_blob":"479H0KQ4LUUXIHL48WCVN0C9VD7HWSX0MG1UPYNXK6PI9HLGBU2U10K3HPFJSROFEG5VD749WDPHWSHXXO72BOSY2G8TWUDOJNLRTR9LTT8PSOB9NNZ485EY2RD9D80FLDFRBVMP1RKMELILD7I922D6TBCAZK30CSV6KDEDUMYABE0XB9EH8C4LE98LMU91I9ZV2APETJD4AYFEN0VNMIT1XQ122Y2OOXO45GJ737HHM5XX88RY7CXHVWJ5JJ7NYW6T1EEBW9UE0NLB2497YBP9V1XVAEK8JJYVRVW0L03ZDXFY8BBHP6UBU7ZNR0JU9GJQPNHG0DK86S4LLYDN0BTCF4KWV2J4DEB6DAX4BDLNPT87MM75G70DFE9W0R6HRNWCH0X075WHAXPSH7S3CSNXPPA6PDO6UA1RCCZOVZ99H7968Q37HACMD8EZ8SU81V4KNRXM46N520S4FVZNSJHA"
+      },
+      "status":"error"
+   }
+}
+```
+
+***TODO: Is this a universal error? Should it be added here? https://ripple.com/build/rippled-apis/#universal-errors***
+
+You can verify that your `rippled` server is amendment blocked using the [`server_info`](reference-rippled.html#server-info) command. In the response, look for `result.info.amendment_blocked`. If `amendment_blocked` is set to `true`, your server is amendment blocked. For example:
+
+```
+{
+  "result": {
+    "info": {
+      "amendment_blocked": true,
+      "build_version": "0.90.0-b2",
+      "complete_ledgers": "6015234-6022198,6022744-6025498,6026905-6029106,6030549-6031191,6031472-6032443,6033090-6033095,6033339-6033537,6034791-6034950,6034980-6035798,6036466,6036516-6036703,6037478-6037671,6038085-6040415,6042691-6046229,6046413-6053861",
+...
+      "validation_quorum": 4,
+      "validator_list_expires": "2018-Jan-24 00:00:00"
+    },
+    "status": "success"
+  }
+}
+```
+
+If your server is not amendment blocked, the `amendment_blocked` field is not returned in the response.
+
+***TODO: Question: Update the `server_info` doc to include an `amendment_blocked` field descr? https://ripple.com/build/rippled-apis/#server-info. Does this amendment_blocked also display in `server_state`? https://ripple.com/build/rippled-apis/#server-state. If yes, also add descr there?***
+
+To find out which amendments are blocking your `rippled` server, use the [`feature`](reference-rippled.html#feature) admin command and look for features (amendments) that have `"enabled" : true` and `"supported" : false`. These values for a feature mean that the amendment is currently enabled (required) in the latest ledger, but your server does not know how to apply the amendment. For example:
+
+***TODO: Update `features` doc to talk about the implications of when `"enabled" : true` and `"supported" : false`. Useful? https://ripple.com/build/rippled-apis/#feature***
+
+```
+{
+   "id":1,
+   "result":{
+      "features":{
+         "B4D44CC3111ADD964E846FC57760C8B50FFCD5A82C86A72756F6B058DDDF96AD":{
+            "enabled":true,
+            "supported":false,
+            "vetoed":false
+         },
+         "6C92211186613F9647A89DFFBAB8F94C99D4C7E956D495270789128569177DA1":{
+            "enabled":true,
+            "supported":false,
+            "vetoed":false
+         },
+         "B9E739B8296B4A1BB29BE990B17D66E21B62A300A909F25AC55C22D6C72E1F9D":{
+            "enabled":true,
+            "supported":false,
+            "vetoed":false
+         },
+         "1D3463A5891F9E589C5AE839FFAC4A917CE96197098A1EF22304E1BC5B98A454":{
+            "enabled":true,
+            "supported":false,
+            "vetoed":false
+         },
+         "CC5ABAE4F3EC92E94A59B1908C2BE82D2228B6485C00AFF8F22DF930D89C194E":{
+            "enabled":true,
+            "supported":false,
+            "vetoed":false
+         },
+...
+         "08DE7D96082187F6E6578530258C77FAABABE4C20474BDB82F04B021F1A68647":{
+            "enabled":false,
+            "name":"PayChan",
+            "supported":true,
+            "vetoed":false
+         }
+      },
+      "status":"success"
+   }
+}
+```
+
+In this example, conflicts with the following features are causing the `rippled` server to be amendment blocked:
+
+* `B4D44CC3111ADD964E846FC57760C8B50FFCD5A82C86A72756F6B058DDDF96AD`
+
+* `6C92211186613F9647A89DFFBAB8F94C99D4C7E956D495270789128569177DA1`
+
+* `B9E739B8296B4A1BB29BE990B17D66E21B62A300A909F25AC55C22D6C72E1F9D`
+
+* `1D3463A5891F9E589C5AE839FFAC4A917CE96197098A1EF22304E1BC5B98A454`
+
+* `CC5ABAE4F3EC92E94A59B1908C2BE82D2228B6485C00AFF8F22DF930D89C194E`
+
+***TODO: Still need to build out an explanation of how these features became the ones that are amendment blocking the server. Working on amendment blocking my rippled server.***
+
+[Upgrade to a new `rippled` version](tutorial-rippled-setup.html#updating-rippled) to unblock your server and enable it to sync with the network again. ***TODO: Question: Do these instructions need a refresh? Need to take a look.***
 
 
 ## Testing Amendments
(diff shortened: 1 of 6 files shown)


content/concept-amendments.md
```diff
@@ -86,13 +86,122 @@ When an amendment gets enabled for the network after the voting process, servers
...
+Becoming amendment blocked is a security feature to protect backend applications. Rather than guessing and maybe misinterpreting a ledger after new rules have been applied, `rippled` reports that it does not know the state of the ledger because it does not know how the amendment works.
 
 The amendments that a `rippled` server is configured to vote for or against have no impact on whether the server becomes amendment blocked. A `rippled` server always follows the set of amendments enabled by the rest of the network, to the extent possible. A server only becomes amendment blocked if the enabled amendment is not included in the amendment definitions compiled into the server's source code -- in other words, if the amendment is newer than the server.
 
 If your server is amendment blocked, you must [upgrade to a new version](tutorial-rippled-setup.html#updating-rippled) to sync with the network.
 
 
+#### How to Tell If Your `rippled` Server is Amendment Blocked
+
+***TODO: Question: Okay location for this doc? It is building upon a concept, but is tutorial-ish.***
```
mDuo13: It's probably fine to put it here for now.

Long-term, I think this'll fit into section VI.3 of the rippled doc plan ("The rippled Server - Advanced Configuration")

Nitpick: capitalize "is" or don't capitalize "if". I think the former is proper style, although I'm no expert.

content/concept-amendments.md
```diff
@@ -86,13 +86,122 @@ When an amendment gets enabled for the network after the voting process, servers
 * Do not participate in the consensus process
 * Do not vote on future amendments
 
-Becoming amendment blocked is a security feature to protect backend applications. Rather than guessing and maybe misinterpreting a ledger after new rules have applied, `rippled` reports that it does not know the state of the ledger because it does not know how the amendment works.
+Becoming amendment blocked is a security feature to protect backend applications. Rather than guessing and maybe misinterpreting a ledger after new rules have been applied, `rippled` reports that it does not know the state of the ledger because it does not know how the amendment works.
```
mDuo13: Is this change necessary? Rules can "apply", active voice, can't they?

content/concept-amendments.md
```diff
@@ -86,13 +86,122 @@ When an amendment gets enabled for the network after the voting process, servers
...
+      "request":{
+         "command":"submit",
+         "tx_blob":"479H0KQ4LUUXIHL48WCVN0C9VD7HWSX0MG1UPYNXK6PI9HLGBU2U10K3HPFJSROFEG5VD749WDPHWSHXXO72BOSY2G8TWUDOJNLRTR9LTT8PSOB9NNZ485EY2RD9D80FLDFRBVMP1RKMELILD7I922D6TBCAZK30CSV6KDEDUMYABE0XB9EH8C4LE98LMU91I9ZV2APETJD4AYFEN0VNMIT1XQ122Y2OOXO45GJ737HHM5XX88RY7CXHVWJ5JJ7NYW6T1EEBW9UE0NLB2497YBP9V1XVAEK8JJYVRVW0L03ZDXFY8BBHP6UBU7ZNR0JU9GJQPNHG0DK86S4LLYDN0BTCF4KWV2J4DEB6DAX4BDLNPT87MM75G70DFE9W0R6HRNWCH0X075WHAXPSH7S3CSNXPPA6PDO6UA1RCCZOVZ99H7968Q37HACMD8EZ8SU81V4KNRXM46N520S4FVZNSJHA"
+      },
+      "status":"error"
+   }
+}
+```
+
+***TODO: Is this a universal error? Should it be added here? https://ripple.com/build/rippled-apis/#universal-errors***
```
mDuo13: Good question. I'd like a developer to confirm.

content/concept-amendments.md
```diff
@@ -86,13 +86,122 @@ When an amendment gets enabled for the network after the voting process, servers
...
+...
+      "validation_quorum": 4,
+      "validator_list_expires": "2018-Jan-24 00:00:00"
+    },
+    "status": "success"
+  }
+}
+```
+
+If your server is not amendment blocked, the `amendment_blocked` field is not returned in the response.
```
mDuo13: Add a caution that versions older than 0.80.0 (per RIPD-1479?) don't include this field even if your server is amendment blocked.

content/concept-amendments.md
```diff
@@ -86,13 +86,122 @@ When an amendment gets enabled for the network after the voting process, servers
...
+
+***TODO: Question: Update the `server_info` doc to include an `amendment_blocked` field descr? https://ripple.com/build/rippled-apis/#server-info. Does this amendment_blocked also display in `server_state`? https://ripple.com/build/rippled-apis/#server-state. If yes, also add descr there?***
```
mDuo13: Respectively:

1. Yes.
2. Yes. (Confirmed [in the code](https://github.com/ripple/rippled/blob/3e9e0c4b53aa29775b02f5a092c79587cebef8d0/src/ripple/app/misc/NetworkOPs.cpp#L2141-L2142)—server_info and server_state use the same underlying function and there's no special code to change how this field displays.)
3. Yes.

content/concept-amendments.md
```diff
@@ -86,13 +86,122 @@ When an amendment gets enabled for the network after the voting process, servers
...
+
+To find out which amendments are blocking your `rippled` server, use the [`feature`](reference-rippled.html#feature) admin command and look for features (amendments) that have `"enabled" : true` and `"supported" : false`. These values for a feature mean that the amendment is currently enabled (required) in the latest ledger, but your server does not know how to apply the amendment. For example:
+
+***TODO: Update `features` doc to talk about the implications of when `"enabled" : true` and `"supported" : false`. Useful? https://ripple.com/build/rippled-apis/#feature***
```
mDuo13: Yes, with a link back to this page.

content/concept-amendments.md
```diff
@@ -86,13 +86,122 @@ When an amendment gets enabled for the network after the voting process, servers
...
+
+* `6C92211186613F9647A89DFFBAB8F94C99D4C7E956D495270789128569177DA1`
+
+* `B9E739B8296B4A1BB29BE990B17D66E21B62A300A909F25AC55C22D6C72E1F9D`
+
+* `1D3463A5891F9E589C5AE839FFAC4A917CE96197098A1EF22304E1BC5B98A454`
+
+* `CC5ABAE4F3EC92E94A59B1908C2BE82D2228B6485C00AFF8F22DF930D89C194E`
+
+***TODO: Still need to build out an explanation of how these features became the ones that are amendment blocking the server. Working on amendment blocking my rippled server.***
```
mDuo13: The typical case is you're running a version of `rippled` that's older than the one used by trusted validators.

content/concept-amendments.md
```diff
@@ -86,13 +86,122 @@ When an amendment gets enabled for the network after the voting process, servers
...
+
+[Upgrade to a new `rippled` version](tutorial-rippled-setup.html#updating-rippled) to unblock your server and enable it to sync with the network again. ***TODO: Question: Do these instructions need a refresh? Need to take a look.***
```
mDuo13: They shouldn't. Even if they do, that's a separate task.

mDuo13 (changes requested):
Answered the questions I could, picked some nits, looks good overall.

If you're having trouble actually amendment-blocking your server, that's not surprising. The code for (more explicitly) amendment-blocking your server is new in 0.80.0 and the newest amendment currently enabled on the production XRP Ledger is from 0.80.0. Your best bet might be to connect a v0.80.0 server to the test net where some 0.90.0 amendments are already turned on.

content/concept-amendments.md
ryangyoung: I think the question might be what's applying the rules to what? My interpretation is that In the original, the rules are the agent applying themselves to the ledger. In the amended version, `rippled` has applied the rules to the ledger. FWIW, the amended version makes more intuitive sense to me as a (maybe not very smart) reader.

content/concept-amendments.md
ryangyoung: You always want to upgrade to latest version, right? Would there be a case where you _don't_ want to use a certain amendment? (I can't think of why you'd want to do that, just asking.)

ryangyoung (commented):
LGTM.

content/concept-amendments.md
mDuo13: It's certainly not guaranteed that you would want to use a certain amendment. Amendments can change _anything_ about the ledger's transaction processing, so the changes could be distasteful or unprofitable for you. Alternatively, you may find a bug in an amendment's code.

If the rest of the network approves an amendment, your only choices are to accept that and move on, or fork away from the network from a point before the amendment became enabled. Something akin to the latter is what created the Bitcoin Cash and Ethereum Classic forks. (Bitcoin and Ethereum protocols don't have a formalized amendment process in the way the XRP Ledger does, but that doesn't stop them from changing their rules. It just means that changes to their behavior are far less predictable and servers that aren't upgraded automatically fork off rather than becoming amendment blocked).

ryangyoung: So that means it sort of doesn't matter _which_ amendments aren't supported, right? If _any_ of them are `supported: false`, you have to upgrade to the latest `rippled` version. Else, the decision point described. 

I guess I'm pointing this out because the way the amendments are specified now sort of makes it seem like you could pick and choose which ones you want to support (emphasis on "sort of"). That is, in a way it doesn't matter which specific ones are causing the block; you've got to take all of them if you want to get unblocked.

content/concept-amendments.md
ryangyoung: Here's what the _Microsoft Manual of Style says about title capitalization:

>* Capitalize all nouns, verbs (including is and other forms of be), adverbs (including than and when), adjectives (including this and that), and pronouns (including its).
>* Capitalize the first and last words, regardless of their parts of speech (“The Text to Look For”).
>*  Capitalize prepositions that are part of a verb phrase (“Backing Up Your Disk”).
>*  Do not capitalize articles (a, an, the) unless the article is the  first word in the title.
>*  Do not capitalize coordinate conjunctions (and, but, for, nor, or, yet, so) unless the conjunction is the first word in the title.
>*  Do not capitalize prepositions of four or fewer letters unless the preposition is the first or the last word in the title.
>*  Do not capitalize to in an infinitive phrase (“How to Format Your Hard Disk”) unless the phrase is the first word in the title.
>*  Capitalize the second part of a hyphenated compound if it would be capitalized without the hyphen. Always capitalize the second part of a hyphenated compound if it is the last word of a heading or title.

content/concept-amendments.md
mDuo13: If any amendments are both enabled _and_ unsupported, you must upgrade or fork.

For amendments that are unsupported and _not_ enabled, you can "veto" them (configure your server not to vote in favor of them) and hope that other trusted validators do the same. You're not amendment blocked until those amendments become enabled.

content/concept-amendments.md
jhaaaa: I am no expert either, but from what I've read, 'if" is a subordinating conjunction and should be capitalized -- per AP and Chicago, for example. Sound okay? What is the next style guide in line when Microsoft Manual of Style doesn't specify something?

ryangyoung: The IBM Style Guide. AP and Chicago deviate can pretty far from tech writing standards. AP's stance on serial commas, for instance: yuck.

mDuo13: The Microsoft one is more specific than most other style guides I saw when it comes to title case. It seems like most style guides—contrary to what I would've expected—suggest you should capitalize "Is" though.

jhaaaa: IBM Style Guide on subordinating conjunctions: In headline-style capitalization, capitalize the initial letter of the following words ... subordinating conjunctions, such as after, although, because, before, how, if, than, that, though, until, when, where, whether, and while.

content/concept-amendments.md
jhaaaa: @mDuo13 and @ryangyoung - Yes, thank you for this thread! 🙏  This is exactly what I was trying to suss out in trying to amendment block my server -- and then I was going to see what it takes to get it out of that state. 

Got it - so if you are amendment blocked, it may be interesting to know which features are blocking you -- but regardless, you cannot cherry pick which amendments ("enabled" and "unsupported") you want to take to unblock yourself. You just need to upgrade to the version of `rippled` that will unblock you. 

And just to be sure that I fully understand -- this may not always be the latest version of `rippled`, is that right? Meaning, can there be a `rippled` version that does not amendment block previous versions? If this is true, you could have a scenario in which upgrading to a version before the latest version can unblock you. Do we think there are folks operating servers who would have a use case for needing to know how to do this? Meaning, per the doc, the requirement would be to upgrade to the version that will unblock you (I can write a doc on how to determine this), but the recommendation would be to just upgrade to the latest version. 

For amendments that are "not enabled" and "unsupported", you do still have a choice in terms of being able to veto specific amendments, but this discussion is not about getting out of an amendment blocked state, but rather is important in the context of voting, correct?

content/concept-amendments.md
jhaaaa: @mDuo13 - Checked with Brandon W - he's thinking that this is not a universal error: It is currently only returned for the subset of commands that are designated with `NEEDS_CURRENT_LEDGER` and  `NEEDS_CLOSED_LEDGER`. https://github.com/ripple/rippled/blob/c3a6b1600fffd5e7c22d790b29e09fe0e09fcfff/src/ripple/rpc/impl/RPCHandler.cpp#L160-L165
https://github.com/ripple/rippled/blob/1c44c4a43eb9e46d8270892a9e49466ec7cc3efb/src/ripple/rpc/impl/Handler.cpp#L100

content/concept-amendments.md
mDuo13: > Can there be a rippled version that does not amendment block previous versions?

Yes, absolutely. I mean, new versions themselves don't automatically amendment-block previous versions. Different versions of `rippled` can coexist as long as no enabled amendments are only known by some versions.

> Do we think there are folks operating servers who would have a use case for needing to know how to [upgrade to not the newest version]?

Some people like to thoroughly read the code or wait for others to test a new version.  Those people may end up a version behind everyone else, which is probably OK depending on the version and why it was released. (If it has security fixes, they should really upgrade ASAP.)

> Meaning, per the doc, the requirement would be to upgrade to the version that will unblock you (I can write a doc on how to determine this), but the recommendation would be to just upgrade to the latest version.

Yes.

> For amendments that are "not enabled" and "unsupported", you do still have a choice in terms of being able to veto specific amendments, but this discussion is not about getting out of an amendment blocked state, but rather is important in the context of voting, correct?

Yes. This is described in the [amendment voting documentation](https://ripple.com/build/amendments/#amendment-voting). To "veto" an amendment means to configure your server not to vote in favor of that amendment. (Anyone not voting in favor of an amendment is effectively voting against it.)

content/concept-amendments.md
```diff
@@ -86,13 +86,139 @@ When an amendment gets enabled for the network after the voting process, servers
...
+      "request":{
+         "command":"submit",
+         "tx_blob":"479H0KQ4LUUXIHL48WCVN0C9VD7HWSX0MG1UPYNXK6PI9HLGBU2U10K3HPFJSROFEG5VD749WDPHWSHXXO72BOSY2G8TWUDOJNLRTR9LTT8PSOB9NNZ485EY2RD9D80FLDFRBVMP1RKMELILD7I922D6TBCAZK30CSV6KDEDUMYABE0XB9EH8C4LE98LMU91I9ZV2APETJD4AYFEN0VNMIT1XQ122Y2OOXO45GJ737HHM5XX88RY7CXHVWJ5JJ7NYW6T1EEBW9UE0NLB2497YBP9V1XVAEK8JJYVRVW0L03ZDXFY8BBHP6UBU7ZNR0JU9GJQPNHG0DK86S4LLYDN0BTCF4KWV2J4DEB6DAX4BDLNPT87MM75G70DFE9W0R6HRNWCH0X075WHAXPSH7S3CSNXPPA6PDO6UA1RCCZOVZ99H7968Q37HACMD8EZ8SU81V4KNRXM46N520S4FVZNSJHA"
+      },
+      "status":"error"
+   }
+}
+```
+
+The following `rippled` error also indicates that your server is amendment blocked:
```
mDuo13: > `rippled` error

I suggest the phrase "log message" instead since that indicates where you'd see this message.

content/concept-amendments.md
```diff
@@ -86,13 +86,139 @@ When an amendment gets enabled for the network after the voting process, servers
...
+
+* `157D2D480E006395B76F948E3E07A45A05FE10230D88A7993C71F97AE4B1F2D1`
+
+* `67A34F2CF55BFC0F93AACD5B281413176FEE195269FA6D95219A2DF738671172`
+
+* `F64E1EABBE79D55B3BB82020516CEC2C582A98A6BFE20FBE9BB6A0D233418064`
+
+To look up which `rippled` version supports these features, see [Known Amendments](reference-amendments.html).
+
+***TODO: Question: Do we think this is the workflow users will use to figure this out? If yes, then it might be nice for the user if the table at the top of the Known Amendments page would list amendment IDs. I know that it makes the table a little uglier, but the feature response doesn't always provide the short name and always provides the ID. The way the page is currently structured, the user will need to ctrl+f for the ID, note the short name, and then scroll back up to the table at the top to note the version. They'll need to do this for each amendment.***
```
mDuo13: Yeah, I was thinking about changing the Known Amendments table to list them like that. It quickly starts to run off the side of the page on some screens, though. That's something to address in a different task.

content/rippled-api-methods/server_info.md
```diff
@@ -280,6 +280,7 @@ The `info` object may have some arrangement of the following fields:
 
 | `Field`                             | Type                      | Description |
 |:------------------------------------|:--------------------------|:-----------|
+| `amendment_blocked`                 | Boolean                   | If the server is [amendment blocked](concept-amendments.html#amendment-blocked), this field displays and is set to `true`. If the server is older than 0.80.0, this field does not display even if the server is amendment blocked. If the server is not amendment blocked, this field does not display. |
```
mDuo13: Add a _(May be omitted)_ marker, and rephrase to match the "closed_ledger" phrasing a little more closely, and use a badge for the new feature. Namely:

> If `true`, this server is [amendment blocked] (concept-amendments.html#amendment-blocked). If the server is not amendment blocked, the response omits this field. [New in: rippled 0.80.0][] |

(You'll also have to add 0.80.0 to `snippets/rippled_versions.md`)

content/rippled-api-methods/server_state.md
```diff
@@ -250,6 +250,7 @@ The `state` object may have some arrangement of the following fields:
 
 | `Field`                          | Type             | Description            |
 |:---------------------------------|:-----------------|:-----------------------|
+| `amendment_blocked`              | Boolean          | If the server is [amendment blocked](concept-amendments.html#amendment-blocked), this field displays and is set to `true`. If the server is older than 0.80.0, this field does not display even if the server is amendment blocked. If the server is not amendment blocked, this field does not display. |
```
mDuo13: Edit to match the recommendations from server_info

mDuo13 (changes requested):
Left a couple more suggestions, very minor at this point.

content/concept-amendments.md
```diff
@@ -86,13 +86,139 @@ When an amendment gets enabled for the network after the voting process, servers
 * Do not participate in the consensus process
 * Do not vote on future amendments
 
-Becoming amendment blocked is a security feature to protect backend applications. Rather than guessing and maybe misinterpreting a ledger after new rules have applied, `rippled` reports that it does not know the state of the ledger because it does not know how the amendment works.
+Becoming amendment blocked is a security feature to protect backend applications. Rather than guessing and maybe misinterpreting a ledger after new rules have been applied, `rippled` reports that it does not know the state of the ledger because it does not know how the amendment works.
```
mDuo13: After re-reading this in context, I think I would suggest changing "backend applications" to "applications that depend on XRP Ledger data". My rationale is that "backend" is a fairly vaguely defined term which people tend to use in varying ways, and also because JavaScript web-apps (typically considered "frontend" code) can connect to `rippled` directly via WebSocket, but amendment blocking is meant to protect them, too.

content/concept-amendments.md
```diff
@@ -86,13 +86,139 @@ When an amendment gets enabled for the network after the voting process, servers
...
 
 The amendments that a `rippled` server is configured to vote for or against have no impact on whether the server becomes amendment blocked. A `rippled` server always follows the set of amendments enabled by the rest of the network, to the extent possible. A server only becomes amendment blocked if the enabled amendment is not included in the amendment definitions compiled into the server's source code -- in other words, if the amendment is newer than the server.
```
mDuo13: > A server only becomes amendment blocked if the enabled amendment

I suggest changing this to "if _an_ enabled amendment"

(as usual, apologies for asking you to fix things that you didn't break)

content/concept-amendments.md
```diff
@@ -86,13 +86,139 @@ When an amendment gets enabled for the network after the voting process, servers
...
 
 If your server is amendment blocked, you must [upgrade to a new version](tutorial-rippled-setup.html#updating-rippled) to sync with the network.
 
 
+#### How to Tell If Your `rippled` Server Is Amendment Blocked
+
+One of the first signs that your `rippled` server is amendment blocked is an `amendmentBlocked` error that is returned when you submit a transaction. Here's an example `amendmentBlocked` error:
```
mDuo13: Make "when you submit a transaction" a link to the submit command documentation. (`reference-rippled.html#submit`)

Also, that reminds me, the submit command docs need "amendmentBlocked" in the "Possible Errors" table. Can you cover that in this PR, too? 🙇

content/concept-amendments.md
```diff
@@ -86,13 +86,139 @@ When an amendment gets enabled for the network after the voting process, servers
...
+
+* `157D2D480E006395B76F948E3E07A45A05FE10230D88A7993C71F97AE4B1F2D1`
+
+* `67A34F2CF55BFC0F93AACD5B281413176FEE195269FA6D95219A2DF738671172`
+
+* `F64E1EABBE79D55B3BB82020516CEC2C582A98A6BFE20FBE9BB6A0D233418064`
+
+To look up which `rippled` version supports these features, see [Known Amendments](reference-amendments.html).
+
+***TODO: Question: Do we think this is the workflow users will use to figure this out? If yes, then it might be nice for the user if the table at the top of the Known Amendments page would list amendment IDs. I know that it makes the table a little uglier, but the feature response doesn't always provide the short name and always provides the ID. The way the page is currently structured, the user will need to ctrl+f for the ID, note the short name, and then scroll back up to the table at the top to note the version. They'll need to do this for each amendment.***
```
mDuo13: Don't forget to remove this TODO before merging.

mDuo13 (changes requested):
Found one missing piece—the `amendmentBlocked` error response in the submit command reference docs—and one "TODO" still there. Other than that, looks fantastic.

content/concept-amendments.md
jhaaaa: 🙏

mDuo13 (approved):
_and_ you sorted the possible errors from submit? 💯

PR #302, opened 2018-02-08
Update tutorial-rippled-setup.md - more validators

intelliot:
There are now more than 5 core validators.

The default validator list is in the example validators.txt file: https://github.com/ripple/rippled/blob/release/doc/validators-example.txt


mDuo13 (approved):
👍  good catch

PR #307, opened 2018-02-15
Ports Capacity Planning recommendations to /rippled-setup

jbheron:
This PR adds Mark Travis' [Capacity Planning](https://ripplelabs.atlassian.net/wiki/spaces/~mtravis/pages/205914715/rippled+Capacity+Planning+for+3rd+Parties) documentation to the Dev Portal, which fills the gap left from documenting only the minimum requirements.

As a snippet, this addition will be maintainable for any IA changes during migration to the new Dev Portal.

The PR also changes some other items:
- Changes the *System Requirements* heading to *Minimum System Requirements* and adds a link to the new recommendation. 
- Adds a caveat about sizing for Validators
- Adds links to the new section in the *Additional Configuration* section

For context: https://ripplelabs.atlassian.net/browse/DOC-1262

### DIFF
--- a/content/snippets/capacity-planning.md
+++ b/content/snippets/capacity-planning.md
@@ -1,16 +1,16 @@
 # Capacity Planning
 
 
-Even the most minimally functional `rippled` server must contain the latest completed Ledgers to submit transactions to the network and verify the integrity of the XRP Ledger. Beyond these, consider the following server requirements:
+Even the most minimally functional `rippled` server must contain the most recently validated Ledger versions to submit transactions to the network and verify the integrity of the XRP Ledger. Beyond these, consider the following server requirements:
 
 - Handling ever-increasing transaction volume
 - Servicing transaction reporting information to clients
 - Maintaining varying amounts of historical data
 
-Planning for the server resources required to run `rippled` relies on understanding the impact of two factors:
+Capacity planning for `rippled` involves two factors:
 
 - The [configuration settings](#configuration-settings) that affect resource utilization
-- The commodity [network and hardware](#network-and-hardware) requirements to achieve scalable architecture with consistent performance across the XRP Ledger Network
+- The [network and hardware] (#network and hardware) requirements to achieve consistent, good performance across the XRP Ledger network
 
 ## Configuration Settings
 
--- a/content/snippets/capacity-planning.md
+++ b/content/snippets/capacity-planning.md
@@ -1,7 +1,7 @@
 # Capacity Planning
 
 
-Even the most minimally functional `rippled` server must contain the most recently validated Ledger versions to submit transactions to the network and verify the integrity of the XRP Ledger. Beyond these, consider the following server requirements:
+Even the most minimally functional `rippled` server must contain the most recently validated ledger versions to submit transactions to the network and verify the integrity of the XRP Ledger. Beyond these, consider the following server requirements:
 
 - Handling ever-increasing transaction volume
 - Servicing transaction reporting information to clients
@@ -18,7 +18,16 @@ Ripple recommends the following guidelines to improve performance. You can set t
 
 ### Node Size
 
-The `node_size` parameter determines the size of database caches. Ripple recommends you always set it to `huge` for production servers. Increased cache size requires less disk I/O and allows `rippled` to improve performance. The trade-off to improving performance better is that memory requirements increase.
+The `node_size` parameter determines the size of database caches. Ripple recommends you always set it to `huge` for production servers. Increased cache size requires less disk I/O and allows `rippled` to improve performance. The trade-off to improving performance is that memory requirements increase.
+
+The `node_size` parameter determines the size of database caches. Larger database caches decrease disk I/O requirements at a cost of higher memory requirements. Ripple recommends you always use the largest database cache your available memory can support. For production servers with the recommended 32 GB of RAM, set `node_size` to `huge`. For `node_size` guidelines, see the following table:
+
+| Available RAM for `rippled` | Recommended `node_size` value | Notes                              |
+|:----------------------------|:------------------------------|:-----------------------------------|
+| < 8 GB                      | `tiny`                        | Not recommended                    |
+| 8 GB                        | `low`                         |                                    |
+| 16 GB                       | `medium`                      |                                    |
+| 32 GB                       | `huge`                        | Recommended for production servers |
 
 ### Node DB Type
 
@@ -27,7 +36,7 @@ The `type` field in the `node_db` section of the `rippled.cfg` file sets the typ
 #### RocksDB vs NuDB
 RocksDB requires approximately one-third less disk storage than NuDB and provides a corresponding improvement in I/O latency. However, this comes at a cost of increased memory utilization as storage size grows. NuDB, on the other hand, has nearly constant performance and memory footprint regardless of storage.
 
-For `rippled` servers that operate as validators, which keep only a few days' worth of data or less, Ripple recommends using RocksDB. For all other uses, Ripple recommends users configure NuDB for the Ledger Store. NuDB has no performance-related configuration options. But tuning parameters are available for RocksDB that have been used to help achieve maximum tested transaction processing throughput, and are as follows:
+For `rippled` servers that operate as validators, which keep only a few days' worth of data or less, Ripple recommends using RocksDB. For all other uses, Ripple recommends users configure NuDB for the Ledger Store. NuDB has no performance-related configuration settings. However, tuning parameters are available for RocksDB that have been used to  achieve the maximum tested transaction processing throughput, shown here:
 
 ```
 [node_db]
@@ -42,7 +51,7 @@ path={path_to_ledger_store}
 
 ### Historical Data
 
-The amount of historical data kept online, as set in the `online_delete` and  `advisory_delete` field, is a major contributor to required storage space. Currently, about 12GB of data is stored per day. However, you should expect this amount to grow as transaction volume increases across the XRP Ledger network.
+The amount of historical data kept online, as set in the `online_delete` and  `advisory_delete` fields, is a major contributor to required storage space. Currently, about 12GB of data is stored per day. However, you can expect this amount to grow as transaction volume increases across the XRP Ledger network.
 
 ### Log Level
 
@@ -68,10 +77,11 @@ For best performance in enterprise production environments, Ripple recommends ru
 
 #### SSD Storage
 
-SSD Storage should support several thousand both read and write IOPS. Maximum read IOPS come from heavily used servers such as those behind https://s1.ripple.com and https://s2.ripple.com, which service transaction and ledger history retrieval requests. This is sometimes over 10,000 reads per second! Maximum write IOPS have come from performance testing in which up to 7,000 writes/s have been observed.
+SSD Storage should support several thousand both read and write IOPS.
+The maximum reads and writes per second that Ripple engineers have observed are over 10,000 reads per second (in heavily-used public server clusters), and over 7,000 writes per second (in dedicated performance testing).
 
 #### CPU Utilization and Virtualization
-Ripple perfmance engineering finds that bare metal servers acheive maximum throughput. However, it is likely that hypervisors would cause minimal degradation in performance.
+Ripple performance engineering finds that bare metal servers achieve maximum throughput. However, it is likely that hypervisors cause minimal degradation in performance.
 
 #### Network
 
@@ -79,18 +89,19 @@ Any enterprise or carrier-class data center should have plenty of network bandwi
 
 #### Storage
 
-Ripple recommends estimating storage sizing at roughly 12GB per day of data kept online with NuDB. RockDB requires around 8GB per day. However, future storage requirements are not known and extra capacity should be available to account for future growth. Currently, a server with all XRP Ledger history requires 6.8TB.
+Ripple recommends estimating storage sizing at roughly 12GB per day of data kept online with NuDB. RocksDB requires around 8GB per day. However, the data per day changes with activity in the network. You should provision extra capacity to prepare for future growth. Currently, a server with all XRP Ledger history requires 6.8TB.
 
 <!-- {# ***TODO: Update the dated storage consideration above, as needed. ***#} -->
-<!-- {# ***TODO: Create historic metrics that a user can use to derive what will be required. For ex, a chart with 1TB in 2014, 3TB in 2015, 7TB in 2018 ***#} -->
+<!-- {# ***TODO: DOC-1331 tracks: Create historic metrics that a user can use to derive what will be required. For ex, a chart with 1TB in 2014, 3TB in 2015, 7TB in 2018 ***#} -->
 
 #### Memory
 
-Memory requirements are mainly a function of the `node_size` configuration option and the amount of client traffic retrieving historical data. As mentioned, production servers should maximize performance and set this parameter to `huge`. A smaller memory footprint can be achieved by setting the parameter lower, but that should only be used for testing. With a `node_size` of `medium`, a `rippled` server can be reasonably stable in a test Linux system with as little as 8GB of RAM.
+Memory requirements are mainly a function of the `node_size` configuration setting and the amount of client traffic retrieving historical data. As mentioned, production servers should maximize performance and set this parameter to `huge`.
+You can set the `node_size` parameter lower to use less memory, but you should only do this for testing. With a `node_size` of `medium`, a `rippled` server can be reasonably stable in a test Linux system with as little as 8GB of RAM.
 
 #### Amazon Web Servives
 
-Amazon Web Services (AWS) is a popular virtualized hosting environment, and you can run `rippled`  in that environment. However, using EBS storage is not recommended, because the maximum number of IOPS (5,000) may not be able to keep up with the heaviest demands, yet it is very expensive.
+Amazon Web Services (AWS) is a popular virtualized hosting environment, and you can run `rippled`  in that environment. However, using Elastic Block Storage (EBS) is not recommended, because the maximum number of IOPS (5,000) may not be able to keep up with the heaviest demands, yet it is very expensive.
 
 AWS instance stores (`ephemeral` storage) do not have these constraints. Therefore, Ripple recommends deploying `rippled` servers with host types such as `M3` that have instance storage. The `database_path` and `node_db` path should each reside on instance storage.
 
--- a/content/snippets/capacity-planning.md
+++ b/content/snippets/capacity-planning.md
@@ -1,6 +1,6 @@
 # Capacity Planning
 
-This section describe some of the challenges related to scaling `rippled` servers for testing and production deployments. Additionally, it describes how hardware and configuration settings relate to those challenges. Finally, this section offers recommendations to assist you in properly setting up `rippled` to meet the goals for the use-case of your deployment.
+This section describes some of the challenges related to scaling `rippled` servers for testing and production deployments. Additionally, it describes how hardware and configuration settings relate to those challenges. Finally, this section offers recommendations to assist you in properly setting up `rippled` to meet the goals for the use-case of your deployment.
 
 Even the most minimally functional `rippled` server must contain the most recently validated ledger versions to submit transactions to the network and verify the integrity of the XRP Ledger. Beyond these requirements, consider the following possible business requirements:
 
@@ -56,7 +56,9 @@ path={path_to_ledger_store}
 
 ### Historical Data
 
-The amount of historical data that a `rippled` server keeps online is a major contributor to required storage space. Currently, about 12GB of data is stored per day. You can expect this amount to grow as transaction volume increases across the XRP Ledger network. You can control how much data you keep with the `online_delete` and `advisory_delete` fields.
+The amount of historical data that a `rippled` server keeps online is a major contributor to required storage space. Currently, a `rippled` server stores about 12GB of data per day (2018-03-01). You can expect this amount to grow as transaction volume increases across the XRP Ledger network. You can control how much data you keep with the `online_delete` and `advisory_delete` fields.
+
+Online deletion enables pruning of `rippled` ledgers from databases with no disruption of service. Without online deletion, those databases grow without bounds. Freeing disk space requires stopping the process and manually removing database files. Online deletion only removes records that are not part of the current ledgers.
 
 <!-- {# ***TODO***: Add link to online_delete section, when complete, per https://ripplelabs.atlassian.net/browse/DOC-1313  #} -->
 
@@ -68,7 +70,7 @@ The default `rippled.cfg` file sets the logging verbosity to `warning`. This set
 
 ## Network and Hardware
 
-Each `rippled` server in the XRP Ledger network performs all of the transaction processing work of the network. It is unknown how soon volumes will approach maximum network capacity. Therefore, the baseline hardware for production `rippled` servers should be similar to that used in Ripple's [performance testing](https://ripple.com/dev-blog/demonstrably-scalable-blockchain/).
+Each `rippled` server in the XRP Ledger network performs all of the transaction processing work of the network. It is unknown when volumes will approach maximum network capacity. Therefore, the baseline hardware for production `rippled` servers should be similar to that used in Ripple's [performance testing](https://ripple.com/dev-blog/demonstrably-scalable-blockchain/).
 
 ### Recommendation
 
@@ -84,18 +86,18 @@ For best performance in enterprise production environments, Ripple recommends ru
 
 #### SSD Storage
 
-SSD Storage should support several thousand of both read and write IOPS. The maximum reads and writes per second that Ripple engineers have observed are over 10,000 reads per second (in heavily-used public server clusters), and over 7,000 writes per second (in dedicated performance testing).
+SSD storage should support several thousand of both read and write IOPS. The maximum reads and writes per second that Ripple engineers have observed are over 10,000 reads per second (in heavily-used public server clusters), and over 7,000 writes per second (in dedicated performance testing).
 
 #### CPU Utilization and Virtualization
 Ripple performance engineering has determined that bare metal servers achieve maximum throughput. However, it is likely that hypervisors cause minimal degradation in performance.
 
 #### Network
 
-Any enterprise or carrier-class data center should have substantial network bandwidth to support running `rippled` servers. The minimum requirements are roughly 2Mbps transmit and 2Mbps receive for current transaction volumes. However, these can burst up to 100MBps transmissions when serving up historical ledger and transaction reports. When a `rippled` server initially starts up, it can burst to well over 20Mbps receive.
+Any enterprise or carrier-class data center should have substantial network bandwidth to support running `rippled` servers. The minimum requirements are roughly 2Mbps transmit and 2Mbps receive for current transaction volumes. However, these can burst up to 100MBps transmissions when serving historical ledger and transaction reports. When a `rippled` server initially starts up, it can burst to over 20Mbps receive.
 
 #### Storage
 
-Ripple recommends estimating storage sizing at roughly 12GB per day of data kept online with NuDB. RocksDB requires around 8GB per day. However, the data per day changes with activity in the network. You should provision extra capacity to prepare for future growth. Currently, a server with all XRP Ledger history requires 6.8TB.
+Ripple recommends estimating storage sizing at roughly 12GB per day of data kept online with NuDB. RocksDB requires around 8GB per day. However, the data per day changes with activity in the network. You should provision extra capacity to prepare for future growth. Currently, a server with all XRP Ledger history requires 6.8TB (2018-03-01).
 
 <!-- {# ***TODO: Update the dated storage consideration above, as needed. ***#} -->
 <!-- {# ***TODO: DOC-1331 tracks: Create historic metrics that a user can use to derive what will be required. For ex, a chart with 1TB in 2014, 3TB in 2015, 7TB in 2018 ***#} -->
--- a/content/snippets/capacity-planning.md
+++ b/content/snippets/capacity-planning.md
@@ -2,7 +2,7 @@
 
 This section describes some of the challenges related to scaling `rippled` servers for testing and production deployments. Additionally, it describes how hardware and configuration settings relate to those challenges. Finally, this section offers recommendations to assist you in properly setting up `rippled` to meet the goals for the use-case of your deployment.
 
-Even the most minimally functional `rippled` server must contain the most recently validated ledger versions to submit transactions to the network and verify the integrity of the XRP Ledger. Beyond these requirements, consider the following possible business requirements:
+Even the most minimally functional `rippled` server must contain the most recently validated ledger versions to submit transactions to the network and verify the integrity of the XRP Ledger. Beyond these requirements, consider the following possible business needs:
 
 - Handling ever-increasing transaction volume
 - Servicing transaction reporting information to clients
@@ -57,7 +57,7 @@ path={path_to_ledger_store}
 
 ### Historical Data
 
-The amount of historical data that a `rippled` server keeps online is a major contributor to required storage space. Currently, a `rippled` server stores about 12GB of data per day (2018-03-01). You can expect this amount to grow as transaction volume increases across the XRP Ledger network. You can control how much data you keep with the `online_delete` and `advisory_delete` fields.
+The amount of historical data that a `rippled` server keeps online is a major contributor to required storage space. At the time of writing (2018-03-01), a `rippled` server stores about 12GB of data per day. You can expect this amount to grow as transaction volume increases across the XRP Ledger network. You can control how much data you keep with the `online_delete` and `advisory_delete` fields.
 
 Online deletion enables pruning of `rippled` ledgers from databases with no disruption of service. Without online deletion, those databases grow without bounds. Freeing disk space requires stopping the process and manually removing database files. Online deletion only removes records that are not part of the current ledgers.
 
@@ -98,7 +98,7 @@ Any enterprise or carrier-class data center should have substantial network band
 
 #### Storage
 
-Ripple recommends estimating storage sizing at roughly 12GB per day of data kept online with NuDB. RocksDB requires around 8GB per day. However, the data per day changes with activity in the network. You should provision extra capacity to prepare for future growth. Currently, a server with all XRP Ledger history requires 6.8TB (2018-03-01).
+Ripple recommends estimating storage sizing at roughly 12GB per day of data kept online with NuDB. RocksDB requires around 8GB per day. However, the data per day changes with activity in the network. You should provision extra capacity to prepare for future growth. At the time of writing (2018-03-01), a server with all XRP Ledger history requires 6.8TB.
 
 <!-- {# ***TODO: Update the dated storage consideration above, as needed. ***#} -->
 <!-- {# ***TODO: DOC-1331 tracks: Create historic metrics that a user can use to derive what will be required. For ex, a chart with 1TB in 2014, 3TB in 2015, 7TB in 2018 ***#} -->
@@ -110,7 +110,7 @@ You can set the `node_size` parameter lower to use less memory, but you should o
 
 #### Amazon Web Servives
 
-Amazon Web Services (AWS) is a popular virtualized hosting environment. You can run `rippled`  in that environment. Ripple does not recommend using Elastic Block Storage (EBS), because the maximum number of IOPS (5,000) may not be able to keep up with the heaviest demands, but it is very expensive.
+Amazon Web Services (AWS) is a popular virtualized hosting environment. You can run rippled in AWS, but Ripple does not recommend using Elastic Block Storage (EBS). Elastic Block Storage's maximum number of IOPS (5,000) is insufficient for `rippled`'s heaviest loads, despite being very expensive.
 
 AWS instance stores (`ephemeral` storage) do not have these constraints. Therefore, Ripple recommends deploying `rippled` servers with host types such as `M3` that have instance storage. The `database_path` and `node_db` path should each reside on instance storage.
 
--- a/content/snippets/capacity-planning.md
+++ b/content/snippets/capacity-planning.md
@@ -14,20 +14,20 @@ Capacity planning for `rippled` involves two factors:
 
 ## Configuration Settings
 
-Ripple recommends the following guidelines to improve performance, based on how our performance engineers evaluate the configuration properties listed. You can set the following parameters in the `rippled.cfg` file to improve performance for your `rippled` server.
+Ripple recommends the following guidelines to improve performance. You can set the following parameters in the `rippled.cfg` file to improve performance for your `rippled` server.
 
 ### Node Size
 
-The `node_size` parameter determines the size of database caches. For production systems, Ripple recommends that you always set it to `huge`. Increased cache size requires less disk I/O and allows `rippled` to improve performance. The trade-off to improving performance better is that memory requirements increase.
+The `node_size` parameter determines the size of database caches. For production systems, Ripple recommends that for production servers, you always set it to `huge`. Increased cache size requires less disk I/O and allows `rippled` to improve performance. The trade-off to improving performance better is that memory requirements increase.
 
 ### Node DB Type
 
 The `type` field in the `node_db` section of the `rippled.cfg` file sets the type of key-value store that `rippled` uses to persist the XRP Ledger in the Ledger Store. You can set the value to either `rocksdb` or `nudb`. For setting the store that `rippled` uses to persist history shards, if you are participating in history sharding, see [History Sharding](XREF: concept-history-sharding.md#shard-store-configuration).
 
 #### RocksDB vs NuDB
-RocksDB requires approximately one-third less disk storage than NuDB and provides a corresponding improvement in I/O latency. However, this comes at a cost of increased memory utilization as storage size grows. Nudb, on the other hand, has nearly constant performance and memory footprint regardless of storage.
+RocksDB requires approximately one-third less disk storage than NuDB and provides a corresponding improvement in I/O latency. However, this comes at a cost of increased memory utilization as storage size grows. NuDB, on the other hand, has nearly constant performance and memory footprint regardless of storage.
 
-For `rippled` servers that operate as validators, which keep only a few days' worth of data or less, Ripple recommends using RocksDB. For all other uses, Ripple recommends users configure NuDB for the Ledger Store. Nudb has no performance-related configuration options. But tuning parameters are available for RocksDB that have been used to help achieve maximum tested transaction processing throughput, and are as follows:
+For `rippled` servers that operate as validators, which keep only a few days' worth of data or less, Ripple recommends using RocksDB. For all other uses, Ripple recommends users configure NuDB for the Ledger Store. NuDB has no performance-related configuration options. But tuning parameters are available for RocksDB that have been used to help achieve maximum tested transaction processing throughput, and are as follows:
 
 ```
 [node_db]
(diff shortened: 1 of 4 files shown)


content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
+# Capacity Planning
+
+
+Even the most minimally functional `rippled` server must contain the latest completed Ledgers to submit transactions to the network and verify the integrity of the XRP Ledger. Beyond these, consider the following server requirements:
```
mDuo13: Instead of:

> the latest completed Ledgers

The following is more consistent with the terms we use in the docs:

> the most recently validated ledger versions

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+
+- Handling ever-increasing transaction volume
+- Servicing transaction reporting information to clients
+- Maintaining varying amounts of historical data
+
+Planning for the server resources required to run `rippled` relies on understanding the impact of two factors:
```
mDuo13: I feel like this sentence is really wordy. It can probably be slimmed down, like:

> Capacity planning for `rippled` involves two factors:

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+
+- The [configuration settings](#configuration-settings) that affect resource utilization
+- The commodity [network and hardware](#network-and-hardware) requirements to achieve scalable architecture with consistent performance across the XRP Ledger Network
```
mDuo13: I would scratch "commodity" from this sentence. Not that it's wrong, but I think it's a bit of a red herring—it's not important whether or not the hardware is commodity.

I would also scratch "scalable architecture with" to simplify the sentence further, then add "good" to rule out the "well, it's consistently bad" possibility. Ultimately:

> - The [network and hardware] (#network and hardware) requirements to achieve consistent, good performance across the XRP Ledger network

mDuo13 (commented):
(Posting partial review now, will finish reviewing later)

jbheron:
@dncohen, thought you might want to see where this will go in the docs.

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+- Maintaining varying amounts of historical data
+
+Capacity planning for `rippled` involves two factors:
+
+- The [configuration settings](#configuration-settings) that affect resource utilization
+- The [network and hardware] (#network and hardware) requirements to achieve consistent, good performance across the XRP Ledger network
+
+## Configuration Settings
+
+Ripple recommends the following guidelines to improve performance, based on how our performance engineers evaluate the configuration properties listed. You can set the following parameters in the `rippled.cfg` file to improve performance for your `rippled` server.
```
mDuo13: > based on how our performance engineers evaluate the configuration properties listed.

I suggest removing this entire clause. Let's cut to the point. People can probably figure out that we came up with these guidelines by having an engineer test stuff.

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Node Size
+
+The `node_size` parameter determines the size of database caches. For production systems, Ripple recommends that you always set it to `huge`. Increased cache size requires less disk I/O and allows `rippled` to improve performance. The trade-off to improving performance better is that memory requirements increase.
```
mDuo13: This contradicts some of the recommendations that I've read elsewhere. You should set only set it to `huge` if you have a lot of RAM (I think huge is supposed to correspond to 32GB?). You can run a server that works OK with less RAM if the setting is smaller. (@miguelportilla's recommendation was 4GB tiny / 8GB low / 16GB medium / 32GB large or huge)

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+
+The `type` field in the `node_db` section of the `rippled.cfg` file sets the type of key-value store that `rippled` uses to persist the XRP Ledger in the Ledger Store. You can set the value to either `rocksdb` or `nudb`. For setting the store that `rippled` uses to persist history shards, if you are participating in history sharding, see [History Sharding](XREF: concept-history-sharding.md#shard-store-configuration).
+
+#### RocksDB vs NuDB
+RocksDB requires approximately one-third less disk storage than NuDB and provides a corresponding improvement in I/O latency. However, this comes at a cost of increased memory utilization as storage size grows. Nudb, on the other hand, has nearly constant performance and memory footprint regardless of storage.
+
+For `rippled` servers that operate as validators, which keep only a few days' worth of data or less, Ripple recommends using RocksDB. For all other uses, Ripple recommends users configure NuDB for the Ledger Store. Nudb has no performance-related configuration options. But tuning parameters are available for RocksDB that have been used to help achieve maximum tested transaction processing throughput, and are as follows:
```
mDuo13: There's an instance of "Nudb" that should be re-capitalized here.

Also, some of our recent testing suggested that RocksDB had better performance than NuDB almost across the board—approximately ⅔ as many IOPS used, only 2% more RAM needed, comparable CPU. Perhaps more importantly, when RocksDB gets oversaturated, `rippled` does a better job of keeping sync with the network's most recently validated ledger whereas a `rippled` server with NuDB as its ledger store falls out of sync when it gets oversaturated with transactions.

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+file_size_mb=64
+file_size_mult=2
+filter_bits=12
+cache_mb=512
+path={path_to_ledger_store}
+```
+
+### Historical Data
+
+The amount of historical data kept online, as set in the `online_delete` and  `advisory_delete` field, is a major contributor to required storage space. Currently, about 12GB of data is stored per day. However, you should expect this amount to grow as transaction volume increases across the XRP Ledger Network.
```
mDuo13: In "XRP Ledger network" don't capitalize "network"

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Log Level
+
+The default logging verbosity for `rippled` in the `log_level` field is set to `debug` and requires several GB or more per day, depending on transaction volumes and client activity. Setting this to a lower level, such as warning, greatly reduces the space and I/O write requirements. The trade-off to a lower setting is decreased visibility for trouble-shooting problems.
```
mDuo13: This is a tricky case where the default config file sets the log level to a non-default value. I suggest changing this to:

> The default `rippled.cfg` file configures the logging verbosity to `warning`. This setting greatly reduces disk space and I/O requirements over more verbose logging. However, more verbose logging may provide more insight when troubleshooting.
>
> *Caution:* If you omit the `log_level` command from the `[rpc_startup]` stanza, `rippled` falls back to `debug` level logging, which requires several more GB of disk space per day, depending on transaction volumes and client activity. 

([Reference on default config log level](https://github.com/ripple/rippled/blob/develop/doc/rippled-example.cfg#L1107))

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+
+- Operating System: Ubuntu 16.04+
+- CPU: Intel Xeon 3+ GHz processor with 4 cores and hyperthreading enabled
+- Disk: SSD
+- RAM: 32 GB RAM
+- Network: Enterprise data center network with a gigabit network interface on the host
+
+#### SSD Storage
+
+SSD Storage should support several thousand both read and write IOPS. Maximum read IOPS come from heavily used servers such as those behind https://s1.ripple.com and https://s2.ripple.com, which service transaction and ledger history retrieval requests. This is sometimes over 10,000 reads per second! Maximum write IOPS have come from performance testing in which up to 7,000 writes/s have been observed.
```
mDuo13: > Maximum read IOPS come from heavily used servers such as those behind https://s1.ripple.com and https://s2.ripple.com, which service transaction and ledger history retrieval requests. This is sometimes over 10,000 reads per second! Maximum write IOPS have come from performance testing in which up to 7,000 writes/s have been observed.

Sounds like you copied a sentence from @mtrippled's methodology? Maybe better to summarize it for users:

> The maximum reads and writes per second Ripple has observed are over 10,000 reads per second (in heavily-used public server clusters), and over 7,000 writes per second (in dedicated performance testing).

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### CPU Utilization and Virtualization
+Ripple perfmance engineering finds that bare metal servers acheive maximum throughput. However, it is likely that hypervisors would cause minimal degradation in performance.
```
mDuo13: typo: perfmance

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### Network
+
+Any enterprise or carrier-class data center should have plenty of network bandwidth to support running `rippled` servers. The minimum requirements are roughly 2Mbps transmit and 2Mbps receive for current transaction volumes. However, these can burst up to 100MBps transmissions when serving up historical ledger and transaction reports. When a `rippled` server initially starts up, it can burst to well over 20Mbps receive.
+
+#### Storage
+
+Ripple recommends estimating storage sizing at roughly 12GB per day of data kept online with NuDB. RockDB requires around 8GB per day. However, future storage requirements are not known and extra capacity should be available to account for future growth. Currently, a server with all XRP Ledger history requires 6.8TB.
```
mDuo13: > However, future storage requirements are not known and extra capacity should be available to account for future growth.

Let's make this more active:

> However, the data per day changes with activity in the network. You should provision extra capacity to prepare for future growth.

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+
+<!-- {# ***TODO: Update the dated storage consideration above, as needed. ***#} -->
+<!-- {# ***TODO: Create historic metrics that a user can use to derive what will be required. For ex, a chart with 1TB in 2014, 3TB in 2015, 7TB in 2018 ***#} -->
```
mDuo13: I like this. Do we have a JIRA task to do this work for the years the XRP Ledger has existed so far?

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### Memory
+
+Memory requirements are mainly a function of the `node_size` configuration option and the amount of client traffic retrieving historical data. As mentioned, production servers should maximize performance and set this parameter to `huge`. A smaller memory footprint can be achieved by setting the parameter lower, but that should only be used for testing. With a `node_size` of `medium`, a `rippled` server can be reasonably stable in a test Linux system with as little as 8GB of RAM.
```
mDuo13: > A smaller memory footprint can be achieved by setting the parameter lower, but that should only be used for testing.

PV. Suggested change:

> You can set the `node_size` parameter lower to use less memory, but you should only do this for testing.

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### Amazon Web Servives
+
+Amazon Web Services (AWS) is a popular virtualized hosting environment, and you can run `rippled`  in that environment. However, using EBS storage is not recommended, because the maximum number of IOPS (5,000) may not be able to keep up with the heaviest demands, yet it is very expensive.
+
+AWS instance stores (`ephemeral` storage) do not have these constraints. Therefore, Ripple recommends deploying `rippled` servers with host types such as `M3` that have instance storage. The `database_path` and `node_db` path should each reside on instance storage.
```
mDuo13: Is it "instance stores" or "instance storage"?

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+
+**Caution:** AWS instance storage is not guaranteed to provide durability in the event of hard drive failure. Further, data will be lost when the instance is stopped and restarted (but not when just rebooted).
```
mDuo13: I suspect this may not _actually_ be a "caution" situation. You should add a comment explaining why this ephemeral storage is OK for `rippled`'s databases:

> AWS instance storage is not guaranteed in the event of hard drive failure. Further, data is lost whenever the instance is stopped and restarted (although not if it is directly rebooted). However, this is usually acceptable for a `rippled` server because any individual server doesn't store any critical, unique data. At most, a `rippled` server may need to generate a new unique server ID when it reconnects to the network, re-download the configured amount of ledger history from its peer servers, and it may lose track of some unconfirmed transactions that were queued at the time it shut down.

mDuo13: For reference, when I say "unique server ID" I'm referring to the cryptographic keys that every server uses to identify itself and sign its communications in peer-to-peer communications. (The public key of this keypair is the `pubkey_node` from server_info.) The server generates this key when it starts up and re-generates it if it doesn't have one. It's generally not important whether a server changes IDs this way. The only exception would be if you have your servers configured for [clustering](https://ripple.com/build/rippled-setup/#clustering) in which case the other members of the cluster won't recognize the server by its new ID. You can recognize this by hard-coding the server's ID in the config file using the `[node_seed]` stanza.

content/tutorial-rippled-setup.md
```diff
@@ -75,9 +77,9 @@ A `rippled` server should run comfortably on commodity hardware, to make it inex
 - Disk: Minimum 50GB SSD recommended (500+ IOPS, more is better) for the database partition
 - RAM: 4+GB
```
mDuo13: Let's update the minimum requirements to 8GB, to match the comments in the capacity planning doc.

content/tutorial-rippled-setup.md
```diff
@@ -57,16 +58,17 @@ There are several properties that define a good validator. The more of these pro
...
 Production `rippled` instances can [use Ripple's binary executable](#installation-on-centosred-hat-with-yum), available from the Ripple [yum](https://en.wikipedia.org/wiki/Yellowdog_Updater,_Modified) repository.
 
-## System Requirements
+## Minimum System Requirements
 
-A `rippled` server should run comfortably on commodity hardware, to make it inexpensive to participate in the network. At present, we recommend the following:
+A `rippled` server should run comfortably on commodity hardware, to make it inexpensive to participate in the network. At present, we recommend the following mimimum requirements:
 
 - Operating System:
     - Production: CentOS or RedHat Enterprise Linux (latest release) or Ubuntu (15.04+) supported
```
mDuo13: Sort of unrelated, but let's update the minimum supported version of Ubuntu to 16.04 since 15.04 is now long past end-of-life.

content/tutorial-rippled-setup.md
```diff
@@ -75,9 +77,9 @@ A `rippled` server should run comfortably on commodity hardware, to make it inex
 - Disk: Minimum 50GB SSD recommended (500+ IOPS, more is better) for the database partition
 - RAM: 4+GB
 
-Amazon EC2's m3.large VM size may be appropriate depending on your workload. (Validating servers need more resources.)
+Amazon EC2's `m3.large` VM size may be appropriate depending on your workload. A fast network connection is preferable. Any increase in a server's client-handling load will increase resources needs.
```
mDuo13: instead of "will increase" let's say, "increases" here

content/tutorial-rippled-setup.md
```diff
@@ -75,9 +77,9 @@ A `rippled` server should run comfortably on commodity hardware, to make it inex
...
 
-Naturally, a fast network connection is preferable.
+**Tip:** For recommendation beyond the minimum requirements, see [Capacity Planning](#capacity-planning).
```
mDuo13: Let's say, "recommendations" plural here

mDuo13 (changes requested):
I'd like to see engineers weigh in on the RocksDB vs. NuDB thing. In the meantime, I suggested a number of simplifications and one complication (re: ephemeral storage)

content/snippets/capacity-planning.md
jbheron: Will add the caveat 'for production servers', as much of this document is geared to that scope.

content/snippets/capacity-planning.md
jbheron: There were actually two! Another on line 28 -- thanks.

jbheron: I can't speak to the technical recommendation. @mtrippled, Rome's concern ☝️ ?

content/snippets/capacity-planning.md
jbheron: Great, that explanation (your second comment) makes sense, but I don't think we need to get into server ID's here, rather just manage the data loss exposure. I'll take a middle approach to that with my update.

jbheron: > **Caution:** AWS instance storage is not guaranteed to provide durability in the event of hard drive failure. Further, data will be lost when the instance is stopped and restarted (but not when just rebooted). This loss can be acceptable for a `rippled` server because an individual server can usually re-acquire that data from its peer servers.

content/tutorial-rippled-setup.md
jbheron: The capacity planning does not describe the minimum settings, but the production deployment settings. I think we should describe both, and more, and those will be better organized in the documentation when we make hardware/configuration suggestions for the different persona/use cases (as part of the Info Architecture upgrades and dev portal migration). In the mean time, I thought this struck a healthy middle ground of having a minimum requirements section and capacity planning recommendations. What do you think, @mDuo13 ?

mDuo13: I think that what I've read recently is that 4GB is not good enough for the minimum requirements anymore given growth in the ledger size and transaction volume

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,95 @@
...
+- The [configuration settings](#configuration-settings) that affect resource utilization
+- The [network and hardware] (#network and hardware) requirements to achieve consistent, good performance across the XRP Ledger network
+
+## Configuration Settings
+
+Ripple recommends the following guidelines to improve performance. You can set the following parameters in the `rippled.cfg` file to improve performance for your `rippled` server.
+
+### Node Size
+
+The `node_size` parameter determines the size of database caches. For production systems, Ripple recommends that for production servers, you always set it to `huge`. Increased cache size requires less disk I/O and allows `rippled` to improve performance. The trade-off to improving performance better is that memory requirements increase.
```
mtrippled: Redundancy: "For production systems, Ripple recommends that for production..."

content/snippets/capacity-planning.md
mtrippled: Either, actually: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html
"store" is singular for instance storage.

content/tutorial-rippled-setup.md
```diff
@@ -57,28 +58,28 @@ There are several properties that define a good validator. The more of these pro
...
+A `rippled` server should run comfortably on commodity hardware, to make it inexpensive to participate in the network. At present, we recommend the following mimimum requirements:
 
 - Operating System:
-    - Production: CentOS or RedHat Enterprise Linux (latest release) or Ubuntu (15.04+) supported
+    - Production: CentOS or RedHat Enterprise Linux (latest release) or Ubuntu (16.04+) supported
     - Development: Mac OS X, Windows (64-bit), or most Linux distributions
 - CPU: 64-bit x86_64, 2+ cores
 - Disk: Minimum 50GB SSD recommended (500+ IOPS, more is better) for the database partition
-- RAM: 4+GB
+- RAM: 8GB
```
mtrippled: For testing, 8GB+
For production, 32GB

mtrippled:
I'm OK with this, but I can't seem to find where to submit my approval through the github ui. :+1:

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,97 @@
+# Capacity Planning
+
+
+Even the most minimally functional `rippled` server must contain the most recently validated Ledger versions to submit transactions to the network and verify the integrity of the XRP Ledger. Beyond these, consider the following server requirements:
```
mDuo13: > validated Ledger versions

"ledger" should be lowercase here

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,97 @@
...
+- The [configuration settings](#configuration-settings) that affect resource utilization
+- The [network and hardware] (#network and hardware) requirements to achieve consistent, good performance across the XRP Ledger network
+
+## Configuration Settings
+
+Ripple recommends the following guidelines to improve performance. You can set the following parameters in the `rippled.cfg` file to improve performance for your `rippled` server.
+
+### Node Size
+
+The `node_size` parameter determines the size of database caches. Ripple recommends you always set it to `huge` for production servers. Increased cache size requires less disk I/O and allows `rippled` to improve performance. The trade-off to improving performance better is that memory requirements increase.
```
mDuo13: > The trade-off to improving performance better

The "better" feels superfluous here. Maybe rephrase the whole paragraph and add a table:

> The `node_size` parameter determines the size of database caches. Larger database caches decrease disk I/O requirements at a cost of higher memory requirements. Ripple recommends you always use the largest database cache your available memory can support. For production servers with the recommended 32 GB of RAM, set `node_size` to `huge`. For development servers, see the following table:
> 
> | Available RAM for `rippled` | Recommended `node_size` value | Notes |
> |---|---|---|
> | < 8 GB | `tiny` | *Not recommended!* |
> | 8 GB | `low` | |
> | 16 GB | `medium` | |
> | 32 GB | `huge` | **Recommended for production servers.** |

content/snippets/capacity-planning.md
mDuo13: This typo ("perfmance" in the second word) is still there.

mDuo13 (changes requested):

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,97 @@
...
+file_size_mb=64
+file_size_mult=2
+filter_bits=12
+cache_mb=512
+path={path_to_ledger_store}
+```
+
+### Historical Data
+
+The amount of historical data kept online, as set in the `online_delete` and  `advisory_delete` field, is a major contributor to required storage space. Currently, about 12GB of data is stored per day. However, you should expect this amount to grow as transaction volume increases across the XRP Ledger network.
```
jhaaaa: change "field" to "fields"

content/snippets/capacity-planning.md
jhaaaa: change "acheive" to "achieve"

content/snippets/capacity-planning.md
jhaaaa: change "RockDB" to "RocksDB"

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,97 @@
...
+<!-- {# ***TODO: Update the dated storage consideration above, as needed. ***#} -->
+<!-- {# ***TODO: Create historic metrics that a user can use to derive what will be required. For ex, a chart with 1TB in 2014, 3TB in 2015, 7TB in 2018 ***#} -->
+
+#### Memory
+
+Memory requirements are mainly a function of the `node_size` configuration option and the amount of client traffic retrieving historical data. As mentioned, production servers should maximize performance and set this parameter to `huge`. A smaller memory footprint can be achieved by setting the parameter lower, but that should only be used for testing. With a `node_size` of `medium`, a `rippled` server can be reasonably stable in a test Linux system with as little as 8GB of RAM.
+
+#### Amazon Web Servives
+
+Amazon Web Services (AWS) is a popular virtualized hosting environment, and you can run `rippled`  in that environment. However, using EBS storage is not recommended, because the maximum number of IOPS (5,000) may not be able to keep up with the heaviest demands, yet it is very expensive.
```
jhaaaa: mega-nit 😸 : Since we spell out AWS, consider spelling out EBS - Elastic Block Store. Add "Amazon" to the front, if we think folks won't know that EBS is an AWS product. Basically, if we assume that folks don't know what AWS is an acronym for, then we have to assume they don't know what EBS is an acronym for.

content/snippets/capacity-planning.md
jhaaaa: Received some feedback in the IA interviews that folks want to know more about how to use the `online_delete` feature -- beyond what is currently covered in the `rippled.cfg` inline doc. Is there anything more to say on this topic in the context of Capacity Planning? Is our recommendation to stick with the default `2000`? If you leave `advisory_delete` set to `0`, it sounds like the `online_delete` can happen without an external admin command - does that mean that once you hit 2,000 stored ledgers, the oldest ledgers will automatically be deleted from the Ledger Store? Perhaps this full discussion is outside the scope of this doc -- but thought I'd mention it since I heard a request for more on this topic.

mtrippled: It's documented more extensively here, but the big red warning at the top is off-putting: https://wiki.ripple.com/Online_deletion

ryangyoung: @mtrippled we're trying to deprecate that wiki, hence the red banner ;)

@jhaaaa @mtrippled we have a few stories for improving our config documentation beyond the example file in the `rippled` repo. I've added a specific bullet point for `online_delete` recommendations to this one: https://ripplelabs.atlassian.net/browse/DOC-1313 (Ripple employee access only). If we need to, we'll create a separate story just for `online_delete` configuration suggestions.

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,97 @@
+# Capacity Planning
+
+
```
ryangyoung: Who is the intended audience of this document? What problem is it trying to solve? Consider adding a brief (1-3 sentences) introduction of the topic before jumping into the technical details.

content/snippets/capacity-planning.md
ryangyoung: > Beyond these

Perhaps: 

> Beyond these requirements, you should also consider the following parameters when setting up your `rippled` server:

nit: Those "following" bullet points are not strictly speaking requirements because they all may not apply to every user. Though, of course, that depends on how we define the audience in the intro I recommend adding. (And that may all change with the IA information @jhaaaa is gathering.)

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,97 @@
...
+Even the most minimally functional `rippled` server must contain the most recently validated Ledger versions to submit transactions to the network and verify the integrity of the XRP Ledger. Beyond these, consider the following server requirements:
+
+- Handling ever-increasing transaction volume
+- Servicing transaction reporting information to clients
+- Maintaining varying amounts of historical data
+
+Capacity planning for `rippled` involves two factors:
```
ryangyoung: Not super important, but maybe the framing should be that the above three points are your business requirements (what you're trying to do) and these two factors are the technical `rippled` requirements. That is, here are things that you determine based on the goal you're trying to achieve, and here are the things that affect how `rippled`could help or hinder you in achieving that goal.

content/snippets/capacity-planning.md
ryangyoung: With the table, we could probably even drop those last two sentences in favor of something like:

> See the following table for recommended settings. 

That way we're not saying the same thing twice in two different places.

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,97 @@
...
+
+Ripple recommends the following guidelines to improve performance. You can set the following parameters in the `rippled.cfg` file to improve performance for your `rippled` server.
+
+### Node Size
+
+The `node_size` parameter determines the size of database caches. Ripple recommends you always set it to `huge` for production servers. Increased cache size requires less disk I/O and allows `rippled` to improve performance. The trade-off to improving performance better is that memory requirements increase.
+
+### Node DB Type
+
+The `type` field in the `node_db` section of the `rippled.cfg` file sets the type of key-value store that `rippled` uses to persist the XRP Ledger in the Ledger Store. You can set the value to either `rocksdb` or `nudb`. For setting the store that `rippled` uses to persist history shards, if you are participating in history sharding, see [History Sharding](XREF: concept-history-sharding.md#shard-store-configuration).
```
ryangyoung: > For setting the store that `rippled` uses to persist history shards, if you are participating in history sharding, see [History Sharding](XREF: concept-history-sharding.md#shard-store-configuration).

Suggest making that new paragraph and rephrasing to something like:

> `rippled` offers a history sharding feature that allows you to store a randomized range of ledgers in a separate shard store. For more information about how to use this feature, see see [History Sharding](XREF: concept-history-sharding.md#shard-store-configuration).

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,97 @@
...
+
+#### RocksDB vs NuDB
+RocksDB requires approximately one-third less disk storage than NuDB and provides a corresponding improvement in I/O latency. However, this comes at a cost of increased memory utilization as storage size grows. NuDB, on the other hand, has nearly constant performance and memory footprint regardless of storage.
+
+For `rippled` servers that operate as validators, which keep only a few days' worth of data or less, Ripple recommends using RocksDB. For all other uses, Ripple recommends users configure NuDB for the Ledger Store. NuDB has no performance-related configuration options. But tuning parameters are available for RocksDB that have been used to help achieve maximum tested transaction processing throughput, and are as follows:
```
ryangyoung: > Ripple recommends users configure NuDB for the Ledger Store. NuDB has no performance-related configuration options.

nit: To keep this parallel to the previous sentence, how about:

> ...Ripple recommends using NuDB for the Ledger Store. NuDB has no performance-related configuration options.

A couple of other small nits:

* Move " , which keep only a few days' worth of data or less," to the end of the sentence to improve readability. 
* Start a new paragraph for the config part and make the two options more parallel:

> NuDB has no performance-related configuration options. RocksDB has performance-related configuration options. You can use these to achieve the maximum tested transaction processing throughput. Here is an example of the recommended configuration:

content/snippets/capacity-planning.md
ryangyoung: > The amount of historical data kept online, as set in the `online_delete` and  `advisory_delete` field, is a major contributor to required storage space.

Suggest breaking this up a little, like so:

> The amount of historical data kept online is a major contributor to required storage space. You can control how much data you keep with the `online_delete` and  `advisory_delete` fields.

Or maybe move the second proposed sentence to the end of the paragraph after the problem is stated.

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,97 @@
...
+path={path_to_ledger_store}
+```
+
+### Historical Data
+
+The amount of historical data kept online, as set in the `online_delete` and  `advisory_delete` field, is a major contributor to required storage space. Currently, about 12GB of data is stored per day. However, you should expect this amount to grow as transaction volume increases across the XRP Ledger network.
+
+### Log Level
+
+The default `rippled.cfg` file configures the logging verbosity to `warning`. This setting greatly reduces disk space and I/O requirements over more verbose logging. However, more verbose logging provides increased visibility when troubleshooting.
```
ryangyoung: > configures

Maybe "sets"?

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,97 @@
...
+
+**Note:** If you omit the `log_level` command from the `[rpc_startup]` stanza, `rippled` falls back to `debug` level logging, which requires several more GB of disk space per day, depending on transaction volumes and client activity.
```
ryangyoung: Almost seems like this warrants a "caution" maybe?

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Network and Hardware
+
+Each `rippled` server in the XRP Ledger network performs all of the transaction processing work of the network. It is unknown how soon volumes will approach maximum network capacity. Therefore, the baseline hardware for production `rippled` servers should be similar to that used in Ripple's [performance testing](https://ripple.com/dev-blog/demonstrably-scalable-blockchain/).
+
+### Recommendation
+
+For best performance in enterprise production environments, Ripple recommends running `rippled` on bare metal, with the following characteristics:
```
ryangyoung: > Ripple recommends running `rippled` on bare metal, with the following characteristics:

Super nit: I think you should omit the comma after "metal". (That's an intuitive suggestion, so could be wrong about that.)

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,97 @@
...
+### Node Size
+
+The `node_size` parameter determines the size of database caches. Ripple recommends you always set it to `huge` for production servers. Increased cache size requires less disk I/O and allows `rippled` to improve performance. The trade-off to improving performance better is that memory requirements increase.
+
+### Node DB Type
+
+The `type` field in the `node_db` section of the `rippled.cfg` file sets the type of key-value store that `rippled` uses to persist the XRP Ledger in the Ledger Store. You can set the value to either `rocksdb` or `nudb`. For setting the store that `rippled` uses to persist history shards, if you are participating in history sharding, see [History Sharding](XREF: concept-history-sharding.md#shard-store-configuration).
+
+#### RocksDB vs NuDB
+RocksDB requires approximately one-third less disk storage than NuDB and provides a corresponding improvement in I/O latency. However, this comes at a cost of increased memory utilization as storage size grows. NuDB, on the other hand, has nearly constant performance and memory footprint regardless of storage.
```
ryangyoung: Maybe add a link to the storage section from here or the "Node DB Type" section?

content/snippets/capacity-planning.md
ryangyoung: > configuration option

I think we use "configuration option" and "setting" interchangeably. Not a huge deal, but we should try to be consistent (if we're not already; I didn't double check).

content/snippets/capacity-planning.md
ryangyoung: > Amazon Web Services (AWS) is a popular virtualized hosting environment, and you can run `rippled`  in that environment. 

Why not make these two sentences?

ryangyoung: nit: drop the "However"

ryangyoung (commented):
Suggested a few places where some rephrasing would probably increase clarity/readability.

content/snippets/capacity-planning.md
jbheron: We do now, @mDuo13, see https://ripplelabs.atlassian.net/browse/DOC-1331 (Ripple only).

content/snippets/capacity-planning.md
jbheron: Great idea. I'm thinking:
> This section describe some of the challenges related to scaling `rippled` servers for testing and production deployments. Additionally, it describes how hardware and configuration settings relate to those challenges. Finally, this section offers recommendations to assist you in properly setting up `rippled` to meet the goals for the use-case of your deployment.

content/snippets/capacity-planning.md
jbheron: If we haven't already done so, we should probably codify what warrants a `Caution:` vs. `Tip:` vs `Note:` vs `Warning:`.

jbheron: I tend to think of `Caution:` as a callout for possible data loss or some other undesirable business outcome. @ryangyoung

content/snippets/capacity-planning.md
```diff
@@ -1,33 +1,47 @@
...
 
-Even the most minimally functional `rippled` server must contain the most recently validated Ledger versions to submit transactions to the network and verify the integrity of the XRP Ledger. Beyond these, consider the following server requirements:
+Even the most minimally functional `rippled` server must contain the most recently validated ledger versions to submit transactions to the network and verify the integrity of the XRP Ledger. Beyond these requirements, consider the following possible business requirements:
 
 - Handling ever-increasing transaction volume
 - Servicing transaction reporting information to clients
 - Maintaining varying amounts of historical data
 
-Capacity planning for `rippled` involves two factors:
+How you meet your `rippled` capacity planning needs depends on how you address these technical factors:
```
mDuo13: > How you meet your `rippled` capacity planning needs depends on how you address these technical factors:

This sentence is a mouthful! How about this instead:

> To meet your `rippled` capacity requirements, you must address these technical factors:

content/snippets/capacity-planning.md
```diff
@@ -1,33 +1,47 @@
...
+|:----------------------------|:------------------|:-----------------------------------|
+| < 8GB                       | `tiny`            | Not recommended                    |
+| 8GB                         | `low`             |                                    |
+| 16GB                        | `medium`          |                                    |
+| 32GB                        | `huge`            | Recommended for production servers |
 
 ### Node DB Type
 
-The `type` field in the `node_db` section of the `rippled.cfg` file sets the type of key-value store that `rippled` uses to persist the XRP Ledger in the Ledger Store. You can set the value to either `rocksdb` or `nudb`. For setting the store that `rippled` uses to persist history shards, if you are participating in history sharding, see [History Sharding](XREF: concept-history-sharding.md#shard-store-configuration).
+The `type` field in the `node_db` section of the `rippled.cfg` file sets the type of key-value store that `rippled` uses to persist the XRP Ledger in the Ledger Store. You can set the value to either `rocksdb` or `nudb`.
```
mDuo13: Let's make "Ledger Store" lower-case to match with the usage in the sharding doc and with the "shard store" phrase below.

content/snippets/capacity-planning.md
```diff
@@ -1,33 +1,47 @@
...
+
+`rippled` offers a history sharding feature that allows you to store a randomized range of ledgers in a separate shard store. For more information about how to use this feature, see [History Sharding](XREF: concept-history-sharding.md#shard-store-configuration).
```
mDuo13: This paragraph could use something to link it to this section. Maybe something like this after the first sentence?

> You may want to configure the shard store to use a different type of key-value store than the ledger store.

content/snippets/capacity-planning.md
```diff
@@ -1,33 +1,47 @@
...
+RocksDB requires approximately one-third less disk storage than NuDB and provides a corresponding improvement in I/O latency. However, this comes at a cost of increased memory utilization as storage size grows. NuDB, on the other hand, has nearly constant performance and memory footprint regardless of [storage](#storage).
 
-For `rippled` servers that operate as validators, which keep only a few days' worth of data or less, Ripple recommends using RocksDB. For all other uses, Ripple recommends users configure NuDB for the Ledger Store. NuDB has no performance-related configuration options. But tuning parameters are available for RocksDB that have been used to help achieve maximum tested transaction processing throughput, and are as follows:
+`rippled` servers that operate as validators keep only a few days' worth of data or less. Ripple recommends using RocksDB for validators. For all other uses, Ripple recommends using NuDB for the Ledger Store.
```
mDuo13: > `rippled` servers that operate as validators keep only a few days' worth of data or less.

Not strictly true. How about "should keep" for clarity here?

Also, another instance of "Ledger Store" that can be lowercased.

content/snippets/capacity-planning.md
```diff
@@ -1,33 +1,47 @@
...
+
+NuDB has no performance-related configuration options. RocksDB has performance-related configuration options. You can use these to achieve the maximum tested transaction processing throughput. Here is an example of the recommended configuration:
```
mDuo13: Let's start with RocksDB since that's the point of this paragraph. Suggested rephrase:

> RocksDB has performance-related configuration options you can modify to achieve maximum transaction processing throughput. (NuDB does not have performance-related configuration options.) Here is an example of the recommended configuration for a `rippled` server using RockDB:

mDuo13 (approved):
Left a few nitpicky comments, but overall, this looks really good now. Excited to get this published.

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,117 @@
+# Capacity Planning
+
+This section describes some of the challenges related to scaling `rippled` servers for testing and production deployments. Additionally, it describes how hardware and configuration settings relate to those challenges. Finally, this section offers recommendations to assist you in properly setting up `rippled` to meet the goals for the use-case of your deployment.
+
+Even the most minimally functional `rippled` server must contain the most recently validated ledger versions to submit transactions to the network and verify the integrity of the XRP Ledger. Beyond these requirements, consider the following possible business requirements:
```
ryangyoung: > Beyond these requirements, consider the following possible business requirements:

Maybe use "business needs" to avoid using "requirements" twice?

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,117 @@
...
+file_size_mb=64
+file_size_mult=2
+filter_bits=12
+cache_mb=512
+path={path_to_ledger_store}
+```
+
+### Historical Data
+
+The amount of historical data that a `rippled` server keeps online is a major contributor to required storage space. Currently, a `rippled` server stores about 12GB of data per day (2018-03-01). You can expect this amount to grow as transaction volume increases across the XRP Ledger network. You can control how much data you keep with the `online_delete` and `advisory_delete` fields.
```
ryangyoung: > Currently, a `rippled` server stores about 12GB of data per day (2018-03-01).

Move the date to after currently to make it clear that the date refers to what is "current" rather than the data stored on that day. That is:

> At the time of writing (2018-03-01), a `rippled` server stores about 12GB of data per day.

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,117 @@
...
+
+Online deletion enables pruning of `rippled` ledgers from databases with no disruption of service. Without online deletion, those databases grow without bounds. Freeing disk space requires stopping the process and manually removing database files. Online deletion only removes records that are not part of the current ledgers.
+
+<!-- {# ***TODO***: Add link to online_delete section, when complete, per https://ripplelabs.atlassian.net/browse/DOC-1313  #} -->
```
ryangyoung: 👍

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,117 @@
...
+#### CPU Utilization and Virtualization
+Ripple performance engineering has determined that bare metal servers achieve maximum throughput. However, it is likely that hypervisors cause minimal degradation in performance.
+
+#### Network
+
+Any enterprise or carrier-class data center should have substantial network bandwidth to support running `rippled` servers. The minimum requirements are roughly 2Mbps transmit and 2Mbps receive for current transaction volumes. However, these can burst up to 100MBps transmissions when serving historical ledger and transaction reports. When a `rippled` server initially starts up, it can burst to over 20Mbps receive.
+
+#### Storage
+
+Ripple recommends estimating storage sizing at roughly 12GB per day of data kept online with NuDB. RocksDB requires around 8GB per day. However, the data per day changes with activity in the network. You should provision extra capacity to prepare for future growth. Currently, a server with all XRP Ledger history requires 6.8TB (2018-03-01).
```
ryangyoung: > Currently, a server with all XRP Ledger history requires 6.8TB (2018-03-01).

Similar suggestion as above:

> At the time of writing (2018-03-01), a server with all XRP Ledger history requires 6.8TB.

content/snippets/capacity-planning.md
```diff
@@ -0,0 +1,117 @@
...
+<!-- {# ***TODO: DOC-1331 tracks: Create historic metrics that a user can use to derive what will be required. For ex, a chart with 1TB in 2014, 3TB in 2015, 7TB in 2018 ***#} -->
+
+#### Memory
+
+Memory requirements are mainly a function of the `node_size` configuration setting and the amount of client traffic retrieving historical data. As mentioned, production servers should maximize performance and set this parameter to `huge`.
+You can set the `node_size` parameter lower to use less memory, but you should only do this for testing. With a `node_size` of `medium`, a `rippled` server can be reasonably stable in a test Linux system with as little as 8GB of RAM.
+
+#### Amazon Web Servives
+
+Amazon Web Services (AWS) is a popular virtualized hosting environment. You can run `rippled`  in that environment. Ripple does not recommend using Elastic Block Storage (EBS), because the maximum number of IOPS (5,000) may not be able to keep up with the heaviest demands, but it is very expensive.
```
ryangyoung: > but it is very expensive.

Does that mean the preceding phrase should be "because the maximum number of IOPS (5,000) **may be able to** keep up with the heaviest demands"?

ryangyoung (approved):
Suggested a few more minor changes, but otherwise this looks great!

ryangyoung (approved):
Suggested a few more minor changes, but otherwise this looks great!

content/snippets/capacity-planning.md
mDuo13: It doesn't. EBS is expensive _and_ potentially insufficient.

A previous version of this sentence used "yet" instead of "but"—an interesting case where the two conjunctions aren't synonymous and "yet" has a meaning that conveys the intended meaning better than "but" does. That said, I think it's best to rewrite around the issue completely.

Here's my suggested rewrite:

> You can run `rippled`  in AWS, but Ripple does not recommend using Elastic Block Storage (EBS). Elastic Block Storage's maximum number of IOPS (5,000) is insufficient for `rippled`'s heaviest loads, despite being very expensive.

PR #312, opened 2018-02-20
Set Boost version to 1.64.0

jhaaaa:
Update tutorial to reflect that the recommended Boost version for rippled 0.90.0 is Boost 1.64.0. Previously, the tutorial used Boost 1.66.0.


mDuo13 (approved):
:+1:

PR #313, opened 2018-02-21
Reorganize page order and categories per XRP Ledger doc plan

mDuo13:
Reorganizes doc ordering and categories to match our [XRP Ledger Doc plan](https://ripplelabs.atlassian.net/wiki/spaces/DOCS/pages/110715686/XRP+Ledger+Enterprise+Documentation+Plan) (Ripple employees only). Updates the templates to look and work more like the templates from the Ripple Client Portal.

Also:
- Splits up the rippled reference into:
    - rippled concept page (also taking some content from the rippled setup page)
    - rippled APIs introduction page
    - rippled API conventions page
    - public rippled API methods page
    - admin rippled API methods page
- Splits up the transactions reference into:
    - transactions concept page
    - transaction format reference
    - transaction results reference
- Adds a "Related Projects" block to the index page
- Moves the cryptographic keys page to match filename conventions for the rest of the docs
- Removes vestigial/redundant lines from the dactyl-config.yml file
- Edits the Freeze page so the headings follow our style rules

Fixes DOC-1308 and DOC-1281.


ryangyoung:
Overall, this is a definitive improvement over the existing organizational model of the docs. However, I'm concerned that this somewhat duplicates the IA work @jhaaaa is doing to answer the overarching question of who the audience(s) is. Without understanding who is reading the content, it's difficult to say whether any organizational scheme or model is serving them. To prevent more duplicative work, I suggest that this be put on hold until we understand a little more about who we're trying to communicate to. 

That said, here are some general observations and questions:

* "XRP Tasks" is very broad as a category. It also doesn't address some very basic tasks that I imagine some newcomers would want to accomplish, like how to send XRP? How to trade XRP? "Reliable Transaction Submission" sort of covers how to send XRP, but it starts with conceptual information about the ledger, workflows, and best practices. Not until the second to last section do we get to a usable example of how to submit a transaction. Also, the first listed topic in this category is RippleAPI Beginner's Guide. That almost seems like redirect. Do I need to know about RippleAPI to do something with XRP? The Regular Key article seems like it should be the first topic. The other topics seem advanced, especially for newcomers: what if I don't want to multi-sign or put XRP in escrow?

* How did you determine which articles should be included in the "Technical Overview" section? That name also seems a little broad since it could include the `rippled` server, reserves, transaction cost, and probably some others. Who is this section for? What problem is it trying to solve? If I work at an exchange and am trying to set up our initial integration, I probably don't care a ton about the consensus process. 

* Any type of user will need to figure out how they want to handle the `rippled` server first, right? If you're testing some calls, the public servers are fine. If you're an exchange, you'll need to set up your own, etc etc. Therefore, maybe moving the server section to an earlier or more prominent position might make sense. 

* Can some of these articles be reused within multiple categories? For example, exchanges will need to set up a server (or two). They'll want to set up issuing and operational addresses. They'll probably also want to set up multi-sign, and so on. I understand that some of this may be achieved through linking in some articles already, but if the goal is a better user experience it might make sense to provide that path at a higher level than require readers to dive into an article that may or may not be what they want to do first. 

* I wonder if things that we've de-emphasized, like demurrage and the gateway model should be in a top-level overview like this.

jhaaaa:
So much to love about this update! 💯 I put a few comments and suggestions for tweaks on a Confluence page - just b/c it was easier to provide visuals there: https://ripplelabs.atlassian.net/wiki/x/bgDMDg.

mDuo13:
The "Edit on GitHub" links for the new pages are broken because it looks for them on the master branch. Trying to think of a way to work around that. (Would be easy if the Dactyl link checker supported wildcard known-broken links, but it doesn't.)

dactyl-config.yml
```diff
@@ -124,341 +124,344 @@ targets:
...
         targets:
             - local
             - ripple.com
 
-    -   name: Transaction Format
-        category: References
-        html: reference-transaction-format.html
-        md: reference-transaction-format.md
+    -   md: concept-transactions.md
+        category: Technical Overview
```
ryangyoung, line 153: Sort of a nit, but why call this "Technical Overview"? It's in the technical documentation site/section so I think the "technical" could be assumed, right? 

Along similar lines, I think the other category names could be shortened. That would have a couple of benefits: improved brevity and clarity, decreased ambiguity (especially w/r/t "XRP Tasks"), and more usable space on the landing page and sidebar. Therewith, here are my suggestions for new names:

1. Technical Overview --> Overview (I would also be OK with XRP Ledger Overview; that's wordier but more relevant)  
2. Reference Material --> Reference ("Reference Material" sounds like the dustiest section of the library)
3. About XRP (or XRP Overview to be parallel with the XRP Ledger overview)
4. The rippled Server --> Using the rippled Server OR Working with rippled (side note: I can understand why the `rippled-reference.html` is in this section, but, as a user, I'd probably expect that to be in the...Reference section. This might be a case for listing the same topic twice?)
5. XRP Tasks --> Working with XRP OR Using XRP
6. The Exchange Business Model -->  Exchanges
7. The Gateway Business Model --> Gateways
8. Issued Currencies --> Working with Issued Currencies
9.  API Tools (I'm OK with this)

ryangyoung (commented):
For the midterm, with the understanding that this will be adapted per @jhaaaa's IA plans, I'm OK with this proposed structure. There are obviously still a lot of gaps to fill in and improvements to be made, but this is probably the best organization we could do for now, absent the aforementioned IA overlay. 

The only slightly substantive comments I have are about changing the names of some of the categories.

mDuo13:
Replaced by #319 and #321.

PR #315, opened 2018-02-27
Add fix1513 and update status of Checks and DepositAuth

mDuo13:
Updates the Known Amendments page:

- updates the source code link for Feature.cpp
- adds the fix1513 amendment, which was mistakenly omitted from the rippled 0.90.0 release announcement. The status of this amendment is set to "Planned" on the expectation that it will lose a majority and *not* be enabled on March 9 as currently scheduled
- Adds the announced expected date DepositAuth and Checks will become enabled.
- Makes the amount of whitespace between amendment descriptions consistently 2 lines


jhaaaa (approved):

ximinez (approved):
:+1:

PR #316, opened 2018-02-28
Use Google Form for validator domain verification

wilsonianb:
Replaces validators@ripple.com with Google Form/Sheet in order to allow automated validator domain signature verification.


mDuo13 (approved):
:+1:

PR #317, opened 2018-03-02
account_objects - consistent description for `validated`

intelliot:
The sentence seemed to be missing the word `a` so I just replaced it with the exact same description we use for `account_tx`'s `validated` response field.

https://ripple.com/build/rippled-apis/#account-tx


mDuo13:
Changed the description in this one to "response" instead of "request" since that's more precise. We should also edit the other instance to match...

Thanks for this catch though.

PR #318, opened 2018-03-07
Edits capacity planning doc from rippled team feedback

jbheron:
After publishing the Capacity Planning section of the `rippled` setup page, some of the `rippled` team spotted a couple of areas that could be improved:

* 500+ IOPS `=>` 1000 IOPS 
* an unset `log_level` defaults to `warning`, not `debug`, as we had documented

### DIFF
--- a/content/snippets/capacity-planning.md
+++ b/content/snippets/capacity-planning.md
@@ -67,7 +67,7 @@ Online deletion enables pruning of `rippled` ledgers from databases without any 
 
 The default `rippled.cfg` file sets the logging verbosity to `warning`. This setting greatly reduces disk space and I/O requirements over more verbose logging. However, more verbose logging provides increased visibility for troubleshooting.
 
-**Note:** If you omit the `log_level` command from the `[rpc_startup]` stanza, `rippled` falls back to `warning` level logging, which requires several more GB of disk space per day, depending on transaction volumes and client activity.
+**Caution:** If you omit the `log_level` command from the `[rpc_startup]` stanza, `rippled` writes logs to disk at the `debug` level and outputs `warning` level  logs to the console. `debug` level logging requires several more GB of disk space per day than `warning` level, depending on transaction volumes and client activity.
 
 ## Network and Hardware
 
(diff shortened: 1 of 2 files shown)


jbheron:
@mtrippled, could you take a look at these changes to see if they improve our recommendations?

content/snippets/capacity-planning.md
```diff
@@ -67,7 +67,7 @@ Online deletion enables pruning of `rippled` ledgers from databases without any
 
 The default `rippled.cfg` file sets the logging verbosity to `warning`. This setting greatly reduces disk space and I/O requirements over more verbose logging. However, more verbose logging provides increased visibility for troubleshooting.
 
-**Note:** If you omit the `log_level` command from the `[rpc_startup]` stanza, `rippled` falls back to `debug` level logging, which requires several more GB of disk space per day, depending on transaction volumes and client activity.
```
mDuo13, line 70: This isn't right.

I don't know for sure what the default is, but look at the full sentence in context of the previous one. The `warning` level isn't going to generate more logs than the `warning` level.

mDuo13 (changes requested):

content/snippets/capacity-planning.md
ryangyoung: Maybe the feedback was about the callout itself? That is, changing `Note` to `Warning`? @nbougalis

mDuo13: Based on the talk with the team, the following statement is correct:

> If you omit the `log_level` command from the `[rpc_startup]` stanza, `rippled` writes logs to disk at the `debug` level, which requires several more GB of disk space per day, depending on transaction volumes and client activity.

Because the tricky part is, by default, what it writes to disk and what it prints to stdout are different levels. >_<

jbheron: I like this suggestion, but I think we may as well document the full effect of omitting the `log_level` setting, which gives the `warning` level in the console, as well as upgrade that Note to a Caution. Something like:

> **Caution:** If you omit the `log_level` command from the `[rpc_startup]` stanza, `rippled` writes logs to disk at the `debug` level and outputs `warning` level  logs to the console. `debug` level logging requires several more GB of disk space per day than `warning` level, depending on transaction volumes and client activity.

ryangyoung:
Looks like this is good to go. Any reason not to merge it now?

PR #320, opened 2018-03-08
Improve page-toc sidebar width ~1440px window size

mDuo13:


ryangyoung (approved):
I tested this out and the squeezed in-page TOC issue I was seeing doesn't seem to be happening anymore. 

On a somewhat unrelated note, developers.ripple.com has a lot of unused white space on the left; on the local build from this PR there's less of it. I wonder if it can be even further reduced so that the body text is wider and even easier to read.

mDuo13:
It's possible to use up more or even all of the whitespace. The challenge is making sure that, depending on the width of the window, it doesn't make the line widths so long that it's _harder_ to read. (And of course, making sure that whatever is occupying that space still looks/works nicely on screens with less space to spare.)

We should get metrics on the site so we can measure which window sizes are most commonly used and optimize the experience for those.

PR #322, opened 2018-03-14
Update channel API methods with 0.90.0 fixes

mDuo13:
account_channels validated and ledger_* fields (RIPD-1466)
channel_verify public_key in hex format (RIPD-1467)

(Both addressed by https://github.com/ripple/rippled/pull/2192 )

Resolves DOC-1277

### DIFF
--- a/content/rippled-api-methods/account_channels.md
+++ b/content/rippled-api-methods/account_channels.md
@@ -83,7 +83,10 @@ An example of a successful response:
         "public_key_hex": "023693F15967AE357D0327974AD46FE3C127113B1110D6044FD41E723689F81CC6",
         "settle_delay": 86400
       }
-    ]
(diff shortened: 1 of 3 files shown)


mDuo13:
Aside: the API responses are "fudged"—the channels in question no longer exist in the production ledger version listed, you'll get an empty array of channels instead of the non-empty one demonstrated. This is probably not a big deal, as the point is to be an example of what can show up in what formats, which is still accurate.

jhaaaa (approved):
LGTM. Thoughtful, detailed updates to the code and to the doc.

seelabs (approved):
:+1:

mDuo13:
The only broken link is unrelated to this PR, will have to fix that but not needed here

PR #323, opened 2018-03-15
Add API tools and links

mDuo13:
Achieves parity with the current set of API tools and links on ripple.com/build, by adding:

- RPC tool
- ripple.txt validator
- Test Net faucet
- Dev Blog link
- Ripple.com link
- GitHub org link

Also:
- fixes an issue @jhaaaa noted where the link in the page-table-of-contents to the top header didn't work
- sets the "related resources" links to open in new tabs/windows


jhaaaa:
- The 1st heading links in the sidebar work - thank you!

- Great to have all of the **API Tools** available. One nit - on the **XRP Test Net Faucet** page, the **Test Net Servers** sidebar runs off the visible page when I look at it on my laptop screen. 

- Thank you for making the **Related Projects** links open in a new tab. Should the **Edit on GitHub** link in the header and all of the links in the **Ripple Resources** nav section open in new tabs as well?

- In the **Ripple Resources** nav section, a few nits for your consideration:

   - Consider changing **Official Website** to **ripple.com**. For me, **Official Website** implies that the Dev Portal is the “unofficial” site. This link text will also tell users that it leads to the corporate/marketing site.

   - Consider changing **GitHub Organization** to **Open Source on GitHub** -- just to give the user a little more info about where the link leads. Hoping that the _Ripple_ Resources section heading will tell folks that this is _Ripple’s_ open source on GitHub.

   - Considering changing the order of links to 

      1. Dev Blog
      2. Open Source on GitHub
      3. ripple.com
   
      Thinking that this will offer the resources in an order that matches the spirit of dev priorities: info about code first, marketing last.

mDuo13:
- Rebased onto master to fix the broken link.
- Fix missing sidebar borders caused by removing the "!important" line in ripple.css. As a bonus, the fix doesn't use !important.
- Changed the XRP Test Net Faucet page so it won't overflow. I had to do some custom styling for this page because the URLs are just too long to fit in a normally-sized sidebar and they won't word-break properly.
- Edit in GitHub link opens in new tab/window now also.
    - Leaving the "Ripple Resources" links in the sidebar navigation as-is, so they're consistent with other links in that sidebar. (All such links open in the same tab.)
- Changed the "Ripple Resources" section as follows:
    - Renamed "Official Website" to "Corporate Website" and moved it to last.
    - Moved "Dev Blog" to first
    - Renamed "GitHub Organization" to "Open Source on GitHub"

jhaaaa:
Thank you for the updates -- looks GREAT. I apologize for the delay, I ran into some git newbie issues with getting the latest version of the branch to build locally.

Just wanted to check in on a few things 🤓 :

- For the links in the Ripple Resources section, do we want to open them in a new browser tab? I've been taught to open links to a different domain (i.e. a site that uses a different navigational structure) in a new tab -- but I've been holding on to this for a long time. 😂  I know that there is a different perspective, saying that all links should open in the current tab, regardless - because that is what users expect to happen. When we open links in a new tab, this rule says that we are actually disrupting the user experience. I think the same could be said for sending a user to a different site when they don't know they are being sent elsewhere. 🤔 The solution I am ultimately happiest with is opening the link in a new tab and also displaying an icon next to the link indicating that it will open in a new tab -- but some may call that overkill 💅 .......

- When I am on an API Tool page - the Edit on GitHub button in top right gives me a 404. Do we want to hide the Edit on GitHub button when the content accessed can't be edited on GitHub?

mDuo13:
- On the first point (links in new tabs or not), I tend to agree with both schools of thought a little bit, but the more important rule in my experience is that items within the same context should behave alike. So, from that perspective, the failure isn't whether or not the links in question open in a new tab but the fact that there are inter-site and intra-site links presented in the same sidebar context in the first place. Two out of three related resources do make themselves fairly clear that they're not on the same site ("on" and "Website") so it's probably clear enough, and I feel fairly strongly that as long as they're in the same sidebar as the other links, they should behave similarly by opening in the same tab.
    - TL;DR no change for now, but consider moving the related resources to a separate type of navigation in a future template update
- It's not the fact that they're API tools that's causing the 404s, it's the fact that they're not currently available in the default branch. The "Edit on GitHub" links for the new pages _do_ work—they go to the template for the tool, if you build the site with the correct information. (If you look at the Jenkins build results, none of those links resulted in 404s, because the Jenkins build actually uses the pull request's fork information to define a base URL for the links. That's the "fork-aware" stuff from https://github.com/ripple/ripple-dev-portal/pull/319 .) To build fork-aware "Edit on GitHub" links locally, you need to pass in some vars like this:

        dactyl_build --vars '{"github_forkurl": "https://github.com/mduo13/ripple-dev-portal", "github_branch": "doc1327"}'

    Optionally, for PR builds, also add the pull request info:

        dactyl_build --vars '{"github_forkurl": "https://github.com/mduo13/ripple-dev-portal", "github_branch": "doc1327", "github_pr_id": "323", "is_pr_build": true}'

    (You can write the vars info to a `something.json` file and pass the filename instead of passing the JSON on the commandline if that's easier.)

Finally: the broken link is from the ripple-lib 0.19.0 docs. It'll be fixed in master as soon as PR#327 is merged.

PR #324, opened 2018-03-16
Fix ledger command response fields

intelliot:
The documentation of the `ledger` command was missing some response fields.

- `accounts` is actually called `accountState`.
- `close_flags` and `parent_close_time` were missing. I copied the descriptions for these fields from the RippleAPI docs.
- Added non-deprecated equivalents for some fields listed as deprecated.


mDuo13:
:+1: Thanks.

PR #321, opened 2018-03-12
Divide rippled references into multiple pages

mDuo13:
(This and #319 together replace #313.)

Splits up the rippled reference into:

   - rippled concept page (also taking some content from the rippled setup page)
   - rippled APIs introduction page
   - rippled API conventions page
   - public rippled API methods page
   - admin rippled API methods page

Splits up the transactions reference into:

   - transactions concept page
   - transaction format reference
   - transaction results reference

PR #325, opened 2018-03-17
Add Checks concept article

ryangyoung:


content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+XRPL Checks also solve a problem that is unique to the XRP Ledger: they allow users to reject unwanted payments or accept only a portion of a payment. This is useful for institutions that need to be careful about accepting payments for compliance reasons.
+
+Checks potentially enable many other use cases. Ripple encourages the community to find new and creative applications for Checks.
+
+
+### Use Case: Payment Authorization
+
+**Problem:** To comply with regulations like [BSA, KYC, AML, and CFT](https://ripple.com/build/gateway-guide/#gateway-compliance), financial institutions must provide documentation about the source of funds they receive. Such regulations seek to prevent the illicit transfer of funds by requiring institutions to disclose the source and destination of all payments processed by the institution. Because of the nature of the XRP Ledger, anyone could potentially send XRP (and, under the right circumstances, issued currencies) to an institution's account on the XRP Ledger. Dealing with such unwanted payments adds significant overhead cost to these institutions' compliance departments, including potential fines or penalties.
```
jbheron: Perhaps strike 'overhead', as synonymous with cost:
> **Problem:** To comply with regulations like [BSA, KYC, AML, and CFT](https://ripple.com/build/gateway-guide/#gateway-compliance), financial institutions must provide documentation about the source of funds they receive. Such regulations seek to prevent the illicit transfer of funds by requiring institutions to disclose the source and destination of all payments processed by the institution. Because of the nature of the XRP Ledger, anyone could potentially send XRP (and, under the right circumstances, issued currencies) to an institution's account on the XRP Ledger. Dealing with such unwanted payments adds significant cost to these institutions' compliance departments, including potential fines or penalties.

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+<!--{# Diagram sources: https://docs.google.com/drawings/d/1Ez8OZVB2TLH-b_kSFOAgfYqXlEQt4KaUBW6F3TJAv_Q/edit #}-->
+
+
+[![Check flow diagram (successful cashing)](img/checks-happy_path.png)](img/checks-happy_path.png)
+
+**Step 1:** To create a Check, the sender submits a [CheckCreate](https://ripple.com/build/transactions/#checkcreate) transaction and specifies the receiver (`Destination`), expiration date (`Expiration`), and maximum amount that may be debited from the sender's account (`SendMax`).
+
+
+**Step 2:** After the CheckCreate transaction is processed, a [Check object](https://ripple.com/build/ledger-format/#check) (`tlCheck`) is created on the XRP Ledger. This object contains the properties of the Check as defined by the transaction that created it. The object can only be modified by the sender (by canceling it with a [CheckCancel](https://ripple.com/build/transactions/#checkcancel) transaction) or receiver (by canceling it or cashing it) before the expiration time is reached. After the expiration time, anyone may cancel the Check.
```
jbheron: Some passive voice here may be improved:

> **Step 2:** After the CheckCreate transaction is processed, the XRP Ledger creates a [Check object](https://ripple.com/build/ledger-format/#check) (`tlCheck`). This object contains the properties of the Check as defined by the transaction that created it. The object can only be modified by the sender (by canceling it with a [CheckCancel](https://ripple.com/build/transactions/#checkcancel) transaction) or receiver (by canceling it or cashing it) before its expiration time passes. After the expiration time, anyone may cancel the Check.

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
+# Checks
+
+In the XRP Ledger (XRPL), a Check is similar to a personal paper check. Like traditional checks, XRPL Checks start with the sender of the funds creating a Check that specifies an amount and receiver. The receiver cashes the check to pull the funds from the sender's account into the receiver's account. No money moves until the receiver cashes the Check. Because funds are not put on hold when the Check is created, cashing a Check can fail if the sender doesn't have enough funds when the receiver tries to cash it, just like traditional checks. If there's a failure cashing the check, the sender can retry until the check expires.
```
mDuo13: Let's not use "XRPL". I know "XRP Ledger" is more letters and more syllables, but I'd rather not force people to learn yet another acronym. P.S.: God, _please_ may "zerple" never catch on.

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+Unlike traditional checks, XRPL checks have expiration dates after which they may no longer be cashed. This is to prevent an unusable Check object from cluttering the Ledger. If the receiver doesn't successfully cash the Check before it expires, the Check object remains in the XRP Ledger until someone cancels it. Anyone may cancel the Check after it expires. Only the sender and receiver can cancel the Check before it expires or is cashed. The Check object is removed from the Ledger when the sender cashes the check or someone cancels it.
```
mDuo13: "XRPL checks" → "XRP Ledger Checks" (caps).

The capital C is extra important here so they aren't confused with the invariant checker's checks (totally different meaning of the word "checks").

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+**Step 3:** To cash the check, the receiver submits a [CheckCash](https://ripple.com/build/transactions/#checkcash) transaction. The receiver has two options for cashing the check:
+
+* `Amount` — The receiver can use this option to specify an exact amount to cash. This may be useful for cases where the sender has padded the check to cover possible [transfer fees](https://ripple.com/build/transfer-fees/) and the receiver can only accept the exact amount on an invoice or other contract.
+
+* `DeliverMin` — The receiver can use this option to specify the minimum amount they are willing to receive from the check. If the receiver uses this option, `rippled` attempts to deliver as much as possible and will deliver at least this amount. The transaction fails if the sender doesn't have at least this amount in their account.
+
+If the sender has enough funds to cover the Check and the expiration date has not passed, the funds are debited from the sender's account and credited to the receiver's account, and the Check object is is destroyed.
```
jbheron: I'd suggest referring to only one of 'expiration time' or 'expiration date' to describe `Expiration` throughout the document: 

- Three uses of 'expiration date'
- Four uses of 'expiration time'

content/concept-checks.md
mDuo13: > Unlike traditional checks, XRPL checks have expiration dates after which they may no longer be cashed.

1. Some paper checks _do_ have expiration dates. Meanwhile, the expiration on XRP Ledger Checks is also optional. It's probably best not to say this at all.
2. Unusuble Checks _can_ clutter the ledger until someone goes out of their way to cancel them. I think the real reason Checks have an expiration time is so you can set expectations.

mDuo13: > The Check object is removed from the Ledger when the sender cashes the check or someone cancels it.

To be absolutely clear, let's say "when the sender _successfully_ cashes the Check" to distinguish from failed attempts to cash it.

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+Unlike traditional checks, XRPL checks have expiration dates after which they may no longer be cashed. This is to prevent an unusable Check object from cluttering the Ledger. If the receiver doesn't successfully cash the Check before it expires, the Check object remains in the XRP Ledger until someone cancels it. Anyone may cancel the Check after it expires. Only the sender and receiver can cancel the Check before it expires or is cashed. The Check object is removed from the Ledger when the sender cashes the check or someone cancels it.
+
+Checks are similar to [Escrow](https://ripple.com/build/escrow/#escrow) and [Payment Channels](https://ripple.com/build/payment-channels-tutorial/), two other asynchronous transfer methods available on the XRP Ledger. There are two important differences between those methods and Checks:
```
mDuo13: Only Payment Channels are _asynchronous_ in the sense that you can send money without syncing up to XRP Ledger progress every time.

Both Checks and Payment Channels are multi-part ways of sending and receiving money, which sets them apart from Payments which are atomic. (I wish I had better words than "multi-part" and "atomic" for this since "atomic" has other implications and "multi-part" is pretty vague.)

jbheron (approved):
Looks good! 👍 As mentioned, a totally interesting topic that ports and expands a practical banking process. Great coverage. I nitpicked a few copy items.

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+* You can send issued currency with Checks. You can only use XRP with Payment Channels and Escrow.
```
mDuo13: > You can only use XRP with Payment Channels and Escrow.

Suggested rephrase:

> Payment Channels and Escrow can only send XRP.

(To make it clear you _can_ use XRP in other ways.)

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+* Checks do not tie up any funds. The XRP involved in Payment Channels and Escrow cannot be spent after until it is released by the sender (Payment Channels), an expiration, or crypto-condition (Escrow).
```
mDuo13: > until it is released by the sender (Payment Channels)

Not wrong, but maybe a little imprecise.

> until it is redeemed with a claim provided from the sender (Payment Channels)

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+
+***Note:*** The [Checks](https://ripple.com/build/known-amendments/#checks) amendment changes the [OfferCreate transaction](https://ripple.com/build/transactions/#offercreate) to return `tecEXPIRED` when trying to create an Offer whose expiration time is in the past. Without this amendment, an OfferCreate whose expiration time is in the past returns `tesSUCCESS` but does not create or execute an Offer.
```
mDuo13: Maybe just add a note that this isn't specifically related to Checks except that it's a feature that was enabled by the same amendment (because it uses an error code that didn't exist prior to the introduction of Checks)?

Or omit this note. An argument could be made that it's more distracting than useful.

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Why Checks?
+
+Traditional checks allow people to transfer balances without immediately exchanging physical currency. XRPL Checks allow people to exchange funds asynchronously using a process that is familiar to and accepted by the banking industry.
```
mDuo13: Nit: Use "Paper checks" instead of "Traditional checks" because "traditions" have a different connotation and change over time.

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+XRPL Checks also solve a problem that is unique to the XRP Ledger: they allow users to reject unwanted payments or accept only a portion of a payment. This is useful for institutions that need to be careful about accepting payments for compliance reasons.
```
mDuo13: > they allow users to reject unwanted payments or accept only a portion of a payment.

Strictly speaking, Checks don't let you reject unwanted payments. They only provide a mechanism for accepting payments if you're rejecting simple payments. Suggested rewrite:

> They allow users to accept all or part of a payment, even when blocking direct payments with [DepositAuth] (concept-depositauth.html).

content/concept-checks.md
mDuo13: Use `tutorial-gateway-guide.html` instead of `https://ripple.com/build/gateway-guide/`. (In fact, categorically, use the HTML filenames when linking other pages in the Dev Portal. The link substitution filter will change them appropriately when building the ripple.com target.)

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+**Solution:** Institutions can enable [Deposit Authorization](https://ripple.com/build/deposit-authorization/#deposit-authorization) on their XRP Ledger accounts by [setting the `asfDepositAuth` flag in an `AccountSet` transaction](https://ripple.com/build/transactions/#accountset-flags). This makes the account unable to accept Payment transactions. The only way for accounts with Deposit Authorization set to receive funds is through Escrow, Payment Channels, or Checks. Checks are the most straightforward, familiar, and flexible way to transfer funds if Deposit Authorization is enabled.
```
mDuo13: The links should be:

- `concept-deposit-auth.html` (the `#deposit-authorization` anchor is extraneous)
- `reference-transaction-format.html#accountset-flags`

mDuo13: > This makes the account unable to accept Payment transactions.

Suggest "receive" instead of "accept" since it doesn't require action to receive Payments normally.

mDuo13: > The only way for accounts with Deposit Authorization set to receive funds is through...

When discussing flags, use "enabled" instead of "set" where possible. It's less ambiguous. Even then, the word order in this sentence can be a bit tricky ("enabled to receive funds" isn't supposed to be a single phrase, but you can misread it that way), so here's a suggested rephrase:

> Accounts with Deposit Authorization enabled can only receive funds through...

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+## Usage
+
+Checks typically have the lifecycle described below.
+
+<!--{# Diagram sources: https://docs.google.com/drawings/d/1Ez8OZVB2TLH-b_kSFOAgfYqXlEQt4KaUBW6F3TJAv_Q/edit #}-->
+
+
+[![Check flow diagram (successful cashing)](img/checks-happy_path.png)](img/checks-happy_path.png)
+
+**Step 1:** To create a Check, the sender submits a [CheckCreate](https://ripple.com/build/transactions/#checkcreate) transaction and specifies the receiver (`Destination`), expiration date (`Expiration`), and maximum amount that may be debited from the sender's account (`SendMax`).
```
mDuo13: **Tip:** If you include the `snippets/tx-type-links.md` snippet, you can use the predefined reference links for transaction types. Those links will go to `reference-transaction-format.html` automatically and get substituted on ripple.com as necessary.

The reference links defined in that snippet are as follows (for any transaction type as long as it's listed in the [first line of the snippet](https://github.com/ripple/ripple-dev-portal/blob/master/content/snippets/tx-type-links.md)):
-  `[CheckCreate][]`
- `[CheckCreate transaction][]`
- `[CheckCreate transactions][]`

Incidentally, for pseudo-transactions, it's:

- `[EnableAmendment][]`
- `[EnableAmendment pseudo-transaction][]`
- `[EnableAmendment pseudo-transactions][]`

Note to self: Add this information to the README or something for the repo...

content/concept-checks.md
mDuo13: - `https://ripple.com/build/ledger-format/` → `reference-ledger-format.html`
- `[CheckCancel](https://ripple.com/build/transactions/#checkcancel) transaction` → `[CheckCancel transaction][]` (with the tx-type-links include)

content/concept-checks.md
mDuo13: Nit: the `Expiration` is optional

content/concept-checks.md
ryangyoung: I agree that "overhead" and "cost" are synonymous, but they're not the same thing. I think specifying the _type_ of cost here makes the problem clearer. Open to arguments against that, though!

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+
+**Step 2:** After the CheckCreate transaction is processed, a [Check object](https://ripple.com/build/ledger-format/#check) (`tlCheck`) is created on the XRP Ledger. This object contains the properties of the Check as defined by the transaction that created it. The object can only be modified by the sender (by canceling it with a [CheckCancel](https://ripple.com/build/transactions/#checkcancel) transaction) or receiver (by canceling it or cashing it) before the expiration time is reached. After the expiration time, anyone may cancel the Check.
+
+**Step 3:** To cash the check, the receiver submits a [CheckCash](https://ripple.com/build/transactions/#checkcash) transaction. The receiver has two options for cashing the check:
```
mDuo13: You can use the `[CheckCash transaction][]` style link here too

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+* `Amount` — The receiver can use this option to specify an exact amount to cash. This may be useful for cases where the sender has padded the check to cover possible [transfer fees](https://ripple.com/build/transfer-fees/) and the receiver can only accept the exact amount on an invoice or other contract.
```
mDuo13: Nit:  instead of "can only accept the exact amount on" I suggest "prefers to accept only the amount of"

mDuo13: - `https://ripple.com/build/transfer-fees/` → `concept-transfer-fees.html`

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+* `DeliverMin` — The receiver can use this option to specify the minimum amount they are willing to receive from the check. If the receiver uses this option, `rippled` attempts to deliver as much as possible and will deliver at least this amount. The transaction fails if the sender doesn't have at least this amount in their account.
```
mDuo13: > willing to receive from the check

caps on "Check"

content/concept-checks.md
ryangyoung: Looks like the docs favor "expiration time" over "expiration date" so went with the former.

content/concept-checks.md
mDuo13: > The transaction fails if the sender doesn't have at least this amount in their account.

Not exactly. The transaction fails if _the amount that can be credited to the receiver_ is not at least this amount. The distinction is relevant in case of, for example, transfer fees.

content/concept-checks.md
jbheron: Hmm, I think the truth of it is that my lack of understanding how expenses are categorized keeps me from agreeing that fees and penalties add to overhead cost. I'd encourage the specificity, as long as it's accurate. Rather, I think fees and penalties would be categorized differently, as they don't recur. I believe that overhead describes things like labor  and rent.

content/concept-checks.md
mDuo13: You should explicitly cover the other case. For example:

> If cashing the Check fails, the Check object remains in the ledger. The receiver can try to cash the Check again after this. The receiver can wait for the sender to have more funds available, cash the Check for a lesser amount right away, or give up and cancel the Check.

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+<!--{# Diagram sources: https://docs.google.com/drawings/d/1JOgI3H5tpV1yasYe5WLrdxVXLhcQu0bhPfN0mzzS1YQ/edit #}-->
+
+
+[![Check flow diagram (expiration)](img/checks-expiration.png)](img/checks-expiration.png)
+
+
+All Checks start the same way, so **Steps 1 and 2** are the same.
+
+**Step 3a:** If the Check expires before the receiver can cash it, the Check can no longer be cashed but remains in the Ledger.
```
mDuo13: lowercase "ledger" here

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+**Step 4a:** After a Check expires, anyone may cancel it by submitting a [CheckCancel](https://ripple.com/build/transactions/#checkcancel) transaction. That transaction removes the Check from the Ledger.
```
mDuo13: lowercase "ledger" here too

and use a reference link for `[CheckCancel transaction][]`

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+**Step 3a:** If the Check expires before the receiver can cash it, the Check can no longer be cashed but remains in the Ledger.
+
+**Step 4a:** After a Check expires, anyone may cancel it by submitting a [CheckCancel](https://ripple.com/build/transactions/#checkcancel) transaction. That transaction removes the Check from the Ledger.  
+
+
+
+## Availability of Checks
+
+Checks were enabled in `rippled` v0.90.0 with the  ["Checks" amendment](https://ripple.com/build/known-amendments/#checks).
```
mDuo13: - `https://ripple.com/build/known-amendments/#checks` → `reference-amendments.html#checks`

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+You can check the status of the Checks amendment using the [`feature` command](https://ripple.com/build/rippled-apis/#feature).
```
mDuo13: > Checks were enabled in `rippled` v0.90.0

Not precisely correct. When/if Checks are enabled depends primarily on which ledger chain you're on. For the production XRP Ledger, Checks have been enabled since some time around March 15 (Known Amendments page needs to be updated). On the Test Net, they've been enabled a while longer. If you start a new genesis ledger with v0.90.0 (in stand-alone mode for example), Checks are not enabled by default, nor are any other amendments, until at least ledger index 2, and it's possible to veto them and create a ledger chain that doesn't ever have Checks even using rippled 0.90.0. (See the section after the caution in [New Genesis Ledger](https://ripple.com/build/stand-alone-mode/#new-genesis-ledger) for some explanation.) That said, you need at least `rippled` v0.90.0 to properly process Checks, so here's a suggested rephrase:

- Checks require `rippled` version 0.90.0 or higher and have been enabled on the production XRP Ledger since 2018-03-15.
- When not using the production XRP Ledger, you can check the status of the Checks amendment using the [`feature` command].

mDuo13: Also, `https://ripple.com/build/rippled-apis/#feature` → `reference-rippled.html#feature` (after PR #321 it'll be `reference-rippled-api-admin.html#feature` instead)

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+**Solution:** Institutions can enable [Deposit Authorization](https://ripple.com/build/deposit-authorization/#deposit-authorization) on their XRP Ledger accounts by [setting the `asfDepositAuth` flag in an `AccountSet` transaction](https://ripple.com/build/transactions/#accountset-flags). This makes the account unable to accept Payment transactions. The only way for accounts with Deposit Authorization set to receive funds is through Escrow, Payment Channels, or Checks. Checks are the most straightforward, familiar, and flexible way to transfer funds if Deposit Authorization is enabled.
+
+
+## Usage
+
+Checks typically have the lifecycle described below.
+
+<!--{# Diagram sources: https://docs.google.com/drawings/d/1Ez8OZVB2TLH-b_kSFOAgfYqXlEQt4KaUBW6F3TJAv_Q/edit #}-->
```
mDuo13, line 39: Nitpick with the diagram: in step 3, it looks like one of the lines slightly overlaps Barb's name.

content/concept-checks.md
```diff
@@ -0,0 +1,97 @@
...
+
+If the sender has enough funds to cover the Check and the expiration date has not passed, the funds are debited from the sender's account and credited to the receiver's account, and the Check object is is destroyed.
+
+
+
+#### Expiration Case
+
+In the case of expirations, Checks have the lifecycle described below.
+
+<!--{# Diagram sources: https://docs.google.com/drawings/d/1JOgI3H5tpV1yasYe5WLrdxVXLhcQu0bhPfN0mzzS1YQ/edit #}-->
```
mDuo13, line 63: Diagram suggestion: there's a clock icon in the Ripple presentation template you can use without needing an additional credit. You might also be able to copy-paste the "poof" graphic from the Escrow presentation.

Not that it's bad to give credit where due, but IMO it's better to use assets we have full permission over so there aren't any credits to distract from reading the diagram.

mDuo13 (changes requested):
Left lots of style suggestions and nitpicks on the nitty-gritty details. The broad strokes are covered well and most of the content is very elegant and lucid.

This did make me realize that there are a lot of unwritten style guidelines I've been following in the XRP Ledger docs without fully realizing it. I should probably find time to add those to our official style guide somewhere.

content/concept-checks.md
mDuo13: Right. Strictly speaking, a "date" doesn't have hour/minute/second data. The XRP Ledger always uses to-the-second times, so "expiration time" is correct.

content/concept-checks.md
ryangyoung: Maybe we should've gone with "Cheques" for the sake of disambiguation.

intelliot: Reminds me of "tennis rackets" vs. "tennis racquets". I still prefer Checks, though - cheques would have looked strange, to Americans at least.

content/concept-checks.md
ryangyoung: Got it. Revised (for now) to:

> Because they are both executed in multiple steps, Checks and Payment Channels allow you to send funds without syncing to the XRP Ledger for each step. There are two important differences between the other two methods and Checks:

content/concept-checks.md
ryangyoung: Revised to:

> With Payment Channels and Escrow, you can only send XRP.

content/concept-checks.md
ryangyoung: Revised to:

> The XRP involved in Payment Channels and Escrow cannot be spent until it is redeemed with a claim provided by the sender (Payment Channels), or released by an expiration or crypto-condition (Escrow).

content/concept-checks.md
ryangyoung: Revised to:

> ***Note:*** The [Checks](https://ripple.com/build/known-amendments/#checks) amendment changes the expiration behavior of the [OfferCreate transaction](https://ripple.com/build/transactions/#offercreate). For more information, see [OfferCreate](reference-transaction-format.html#expiration).

content/concept-checks.md
mDuo13: ... Sorry, that's still not quite right. You still have to sync up with the XRP Ledger for every step of cashing a Check. It's really only Payment Channels that are different in that regard.

content/concept-checks.md
ryangyoung: I was mostly joking @intelliot 😃  Searching through the source code for "checks" just yielded more results than I would have liked. Agree that we should actually stick with "Checks" despite that.

content/concept-checks.md
ryangyoung: Oh yeah, that's true. I removed that line entirely since it doesn't seem to add much, if any, value now.

content/concept-checks.md
```diff
@@ -1,33 +1,33 @@
...
+XRP Ledger Checks have expiration times after which they may no longer be cashed. If the receiver doesn't successfully cash the Check before it expires, the Check object remains in the XRP Ledger until someone cancels it. Anyone may cancel the Check after it expires. Only the sender and receiver can cancel the Check before it expires or is cashed. The Check object is removed from the Ledger when the sender successfully cashes the check or someone cancels it.
 
-Checks are similar to [Escrow](https://ripple.com/build/escrow/#escrow) and [Payment Channels](https://ripple.com/build/payment-channels-tutorial/). Because they are both executed in multiple steps, Checks and Payment Channels allow you to send funds without syncing to the XRP Ledger for each step. There are two important differences between the other two methods and Checks:
+Checks are similar to [Escrow](concept-escrow.html) and [Payment Channels](tutorial-paychan.html). But there are two important differences between those methods and Checks:
```
mDuo13: Super-nit: I suggest either connecting these sentences or dropping the "But" from the second sentence.

Super-nit 2: Instead of saying "methods", let's say "features" to avoid using the same word we use to refer to API methods.

content/concept-checks.md
```diff
@@ -1,33 +1,33 @@
...
 
 * You can send issued currency with Checks. With Payment Channels and Escrow, you can only send XRP.
 
 * Checks do not tie up any funds. The XRP involved in Payment Channels and Escrow cannot be spent until it is redeemed with a claim provided by the sender (Payment Channels), or released by an expiration or crypto-condition (Escrow).
 
 
-***Note:*** The [Checks](https://ripple.com/build/known-amendments/#checks) amendment changes the expiration behavior of the [OfferCreate transaction](https://ripple.com/build/transactions/#offercreate). For more information, see [OfferCreate](reference-transaction-format.html#expiration).
+***Note:*** The [Checks](reference-amendments.html#checks) amendment changes the expiration behavior of the [OfferCreate transaction](reference-transaction-format.html#offercreate). For more information, see [OfferCreate](reference-transaction-format.html#expiration).
```
mDuo13: Yet another super nit: "[Checks amendment]" would be a better link than "[Checks] amendment"

content/concept-checks.md
```diff
@@ -39,16 +39,16 @@ Checks typically have the lifecycle described below.
...
 
 
-**Step 2:** After the CheckCreate transaction is processed, a [Check object](https://ripple.com/build/ledger-format/#check) (`tlCheck`) is created on the XRP Ledger. This object contains the properties of the Check as defined by the transaction that created it. The object can only be modified by the sender (by canceling it with a [CheckCancel](https://ripple.com/build/transactions/#checkcancel) transaction) or receiver (by canceling it or cashing it) before the expiration time passes. After the expiration time, anyone may cancel the Check.
+**Step 2:** After the CheckCreate transaction is processed, a [Check object](reference-ledger-format.html#check) (`tlCheck`) is created on the XRP Ledger. This object contains the properties of the Check as defined by the transaction that created it. The object can only be modified by the sender (by canceling it with a [CheckCancel](reference-transaction-format.html#checkcancel) transaction) or receiver (by canceling it or cashing it) before the expiration time passes. After the expiration time, anyone may cancel the Check.
```
mDuo13: It's not `tlCheck`, it's `ltCHECK` (the "lt" stands for "ledger type").

That said, I don't think it's important for users to know what the code calls a check type internally, so you could probably just drop that parenthetical entirely.

mDuo13 (changes requested):
I left a couple really nitpicky comments, but that's not why I'm requesting changes.

I'd like to see this doc use the tx-type-links snippet, both for consistency with some other docs and to make it easier to move transaction type links around later.

My other feedback is regarding the diagrams.

1. In both diagrams, the XRP amount boxes seem to move left and right kind of arbitrarily from step to step. (It's especially obvious in Barb's "500 XRP" box from step 2 to 3a from the expiration diagram.) Can you align them so both Alice's and Barb's XRP balance boxes are in the same place from step to step? Related, the spacing between the triskellion and Alice's "1000 XRP" within the box is a little inconsistent.
2. In both diagrams, rather than saying "tlCheck" the center should just say "Check object" (as noted elsewhere, I don't think it's necessary to use the exact internal name, but if we're going to be precise, it's lt, not tl)
3. In several steps of both diagrams the center column shows "100 XRP" in the Check. That _would_ be OK except that it's too similar to the Escrow diagram which is trying to convey an important distinction from this case. When an Escrow object is created in the ledger, it _actually holds that XRP_. The same is _not_ true for Checks; the XRP noted in a check is just a promised amount. Here's how I propose to resolve that:
    - Remove the triskellion from the center Check object's text box. In general, for these types of diagrams, we can use the triskellion to represent where the XRP is actually at.
    - Change the first line of the Check object's text box from "100 XRP" to "SendMax: 100 XRP". That would also help make sense of the case where you cash a Check for less than the stated amount.
4. In the happy path diagram, I think there should not be a stopwatch icon in step 3's CheckCash transaction.
5. Is it possible to change the stopwatch icon to white when it's in a dark blue transaction box? On my screen, there's not much contrast with the background color of the transaction boxes so it's a little hard to make out.

_(Edited—merged points 4 & 1, then added 2 new points about the stopwatches.)_

mDuo13:
I should also mention that the majority of the text is awesome and I'm anxious to see this published ASAP.

ryangyoung:
Thanks for the great feedback, @mDuo13. I updated the diagram as you suggested and made the other changes to the text you spotted.

content/concept-checks.md
```diff
@@ -0,0 +1,103 @@
+# Checks
+
+In the XRP Ledger, a Check is similar to a personal paper check. Like traditional checks, XRP Ledger Checks start with the sender of the funds creating a Check that specifies an amount and receiver. The receiver cashes the check to pull the funds from the sender's account into the receiver's account. No money moves until the receiver cashes the Check. Because funds are not put on hold when the Check is created, cashing a Check can fail if the sender doesn't have enough funds when the receiver tries to cash it, just like traditional checks. If there's a failure cashing the check, the sender can retry until the check expires.
```
jbheron: I think it may best serve the reader to begin with a description or definition, if possible, before moving onto analogies. Something like the following, but better:

> In the XRP Ledger, Checks are payments that allow senders to initiate either XRP or issued currency payments that earmark funds, and receivers to choose if and when they accept those payments. A Check is similar to a personal paper check...

jbheron:
Added a suggestion as a single comment.

content/concept-checks.md
```diff
@@ -0,0 +1,103 @@
...
+
+**Step 3a:** If the Check expires before the receiver can cash it, the Check can no longer be cashed but remains in the ledger.
+
+**Step 4a:** After a Check expires, anyone may cancel it by submitting a [CheckCancel][] transaction. That transaction removes the Check from the ledger.  
+
+
+
+## Availability of Checks
+
+Checks require `rippled` v0.90.0 or later and have been enabled on the production XRP Ledger since 2018-03-15.
```
mDuo13: Just found out today that Checks haven't been enabled yet on the production ledger. That explains why we couldn't find the EnableAmendment pseudo-tx!

The plan is to start voting in favor of Checks soon (maybe today?), but they won't be enabled for at least 2 more weeks.

content/concept-checks.md
jbheron: Also, here's from the 0.90.0 [releases](https://github.com/ripple/rippled/releases/tag/0.90.0) page:

> The `Checks` amendment, which allows users to create deferred payments that can be cancelled or cashed by their intended recipients.

ryangyoung: Revised to: 

> The Checks feature in the XRP Ledger allows users to create deferred payments that can be canceled or cashed by the intended recipients. Like personal paper checks, XRP Ledger Checks start

content/concept-checks.md
```diff
@@ -76,7 +76,10 @@ All Checks start the same way, so **Steps 1 and 2** are the same.
 
 ## Availability of Checks
 
-Checks require `rippled` v0.90.0 or later and have been enabled on the production XRP Ledger since 2018-03-15.
+Checks require `rippled` v0.90.0 or later. The Checks amendment was enabled for voting on the production XRP Ledger on 2018-03-22 in this [EnableAmendment pseudo-transaction](https://ripple.com/build/transactions/#enableamendment): [`EC0D90F114D4F3C599FEABCAF47DCF0CAC29A272D1EF6C9BEF7F420819540F04`](https://xrpcharts.ripple.com/#/transactions/EC0D90F114D4F3C599FEABCAF47DCF0CAC29A272D1EF6C9BEF7F420819540F04).
```
mDuo13: Let's not say "enabled for voting", because you don't need a transaction for a transaction to become open for voting. What we really mean is it gained support of a majority of trusted validators. If you look at the Flags of the raw transaction, you'll see the value 65536, also known as [`tfGotMajority`](https://ripple.com/build/transactions/#enableamendment-flags). In other words, the comment just below this is slightly misleading because this _is_ the GotMajority transaction.

mDuo13 (approved):
The revised diagrams and links look good.

PR #329, opened 2018-04-03
Checks Tutorials

mDuo13:
Includes currently-viable instructions for many Checks tasks, although several may need to change slightly as rippled and ripple-lib are updated. (Those changes are noted as TODOs in comments.)

Does not include a comprehensive suite of examples (JSON-RPC, WebSocket, etc.) for every step because that's a _lot_ of work. I've tried to include at least RippleAPI and Commandline examples for most steps and could probably add JSON-RPC easily enough (since the responses are basically just the Commandline ones). I don't have a commandline example for account_objects and maybe should (it's a slightly inconvenient one since you have to use the `json` method instead of the native `account_objects` commandline syntax to get  filtering by type)

Includes minor changes to the Regular Keys tutorial (pulling out some snippets for reuse) and moves the Checks concept's place in the Dactyl config file to match the Enterprise doc plan outline (even though this will probably change again later per the information architecture).


content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+
+| Field             | Value                     | Description                  |
+|:------------------|:--------------------------|:-----------------------------|
+| `TransactionType` | String                    | The value `CheckCreate` indicates this is a CheckCreate transaction. |
+| `Account`         | String (Address)          | The address of the sender who is creating the Check. (In other words, your address.) |
+| `Destination`     | String (Address)          | The address of the intended recipient who can cash the Check. |
+| `SendMax`         | String or Object (Amount) | The maximum amount the sender can be debited when this Check gets cashed. For XRP, use a string representing drops of XRP. For issued currencies, use an object with `currency`, `issuer`, and `value` fields. See [Specifying Currency Amounts][] for details. If you want the recipient to be able to cash the Check for an exact amount of a non-XRP currency with a [transfer fee](concept-transfer-fees.html), remember to include an extra percentage to pay for the transfer fee. |
```
scottschurr: It's all good and I particularly like the last sentence. :+1:

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+
+{% include 'snippets/checkcash-prereqs.md' %}
+
+### {{cash_exact_n.next()}}. Prepare the CheckCash transaction
+
+Figure out the values of the [CheckCash transaction][] fields. To cash a check for an exact amount, the following fields are the bare minimum; everything else is either optional or can be [auto-filled](reference-transaction-format.html#auto-fillable-fields) when signing:
+
+| Field             | Value                     | Description                  |
+|:------------------|:--------------------------|:-----------------------------|
+| `TransactionType` | String                    | The value `CheckCancel` indicates this is a CheckCancel transaction. |
```
scottschurr: Copy-and-paste-o.  Should be `CheckCash` in two places.

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+| `Account`         | String (Address)          | The address of the sender who is canceling the Check. (In other words, your address.) |
```
scottschurr: canceling -> cashing

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+| `CheckID`         | String                    | The ID of the Check object in the ledger to cancel. You can get this information by looking up the metadata of the CheckCreate transaction using the [`tx` method](reference-rippled.html#tx) or by looking for Checks using the [`account_objects` method](reference-rippled.html#account-objects). |
```
scottschurr: cancel -> cash

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+| `Amount`          | String or Object (Amount) | The amount to redeem from the Check. For XRP, this must be a string specifying drops of XRP. For issued currencies, this is an object with `currency`, `issuer`, and `value` fields. The `currency` and `issuer` fields must match the corresponding fields in the Check object, and the `value` must be less than or equal to the amount in the Check object. For more information on specifying currency amounts, see [Specifying Currency Amounts][]. |
```
scottschurr: Seems like it's worth mentioning somewhere (not necessarily here) that, when cashing a Check, specifying the Amount from the Check object may not be successful.  That could happen because the currency issuer has a transfer fee.  It could also happen if the writer of the Check has insufficient funds.

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+
+{% include 'snippets/checkcash-prereqs.md' %}
+
+### {{cash_flex_n.next()}}. Prepare the CheckCash transaction
+
+Figure out the values of the [CheckCash transaction][] fields. To cash a check for a flexible amount, the following fields are the bare minimum; everything else is either optional or can be [auto-filled](reference-transaction-format.html#auto-fillable-fields) when signing:
+
+| Field             | Value                     | Description                  |
+|:------------------|:--------------------------|:-----------------------------|
+| `TransactionType` | String                    | The value `CheckCancel` indicates this is a CheckCancel transaction. |
```
scottschurr: CheckCancel -> CheckCash twice

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+| `Account`         | String (Address)          | The address of the sender who is canceling the Check. (In other words, your address.) |
```
scottschurr: canceling -> cashing

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+| `CheckID`         | String                    | The ID of the Check object in the ledger to cancel. You can get this information by looking up the metadata of the CheckCreate transaction using the [`tx` method](reference-rippled.html#tx) or by looking for Checks using the [`account_objects` method](reference-rippled.html#account-objects). |
```
scottschurr: cancel -> cash

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+| `DeliverMin`      | String or Object (Amount) | A minimum amount to receive from the Check. If you cannot receive at least this much, cashing the Check fails, leaving the Check in the ledger so you can try again. For XRP, this must be a string specifying drops of XRP. For issued currencies, this is an object with `currency`, `issuer`, and `value` fields. The `currency` and `issuer` fields must match the corresponding fields in the Check object, and the `value` must be less than or equal to the amount in the Check object. For more information on specifying currency amounts, see [Specifying Currency Amounts][]. |
```
scottschurr: It's all good, but I especially like the second sentence.

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+
+<!-- MULTICODE_BLOCK_END -->
+
+#### Handling Errors
+
+If cashing the Check failed with a `tec`-class code, look up the code in the [Full Transaction Response List](reference-transaction-format.html#full-transaction-response-list) and respond accordingly. Some common possibilities for CheckCash transactions:
+
+| Result Code | Meaning | How to Respond |
+|-------------|---------|----------------|
+| `tecEXPIRED` | The Check has expired. | Cancel the Check and ask the sender to create a new Check with a later Expiration time. |
```
scottschurr: Nice table!

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+
+If the Check was cashed for a flexible `DeliverMin` amount and succeeded, you can assume that the Check was cashed for at least the `DeliverMin` amount. To get the exact amount delivered, check the transaction metadata. <!--{# TODO: Update if RIPD-1623 adds a delivered_amount field. #}--> The metadata's `AffectedNodes` array contains one or two objects that reflect the change in balances from cashing the Check, depending on the type of currency.
+
+- For XRP, the `AccountRoot` object of the Check's sender has its XRP `Balance` field debited. The `AccountRoot` object of the Check's recipient (the one who sent the CheckCash transaction) has its XRP `Balance` credited for at least the `DeliverMin` of the CheckCash transaction minus the [transaction cost](concept-transaction-cost.html) of sending the transaction.
+
+    For example, the following `ModifiedNode` shows that the the account rGPnRH1EBpHeTF2QG8DCAgM7z5pb75LAis, the Check's recipient and the sender of this CheckCash transaction, had its XRP balance change from `9999999970` drops to `10099999960` drops, meaning the recipient was credited a _net_ of 99.99999 XRP as a result of processing the transaction.
```
scottschurr: "the the" -> "the"

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+              "PreviousFields": {
+                 "Balance": "9999999970",
+                 "Sequence": 4
+              },
+              "PreviousTxnID": "0283465F0D21BE6B1E91ABDE17266C24C1B4915BAAA9A88CC098A98D5ECD3E9E",
+              "PreviousTxnLgrSeq": 8005334
+            }
+          }
+
+    The net amount of 99.99999 XRP includes deducting the transaction cost is destroyed to pay for sending this CheckCash transaction. The following transaction instructions (excerpted) show that the transaction cost (the `Fee` field) was 10 drops of XRP. By adding this to the net balance change, we conclude that the recipient, rGPnRH1EBpHeTF2QG8DCAgM7z5pb75LAis, was credited a _gross_ amount of exactly 100 XRP for cashing the Check.
```
scottschurr: The first sentence reads funny, "...deducting the transaction cost is destroyed...".  Maybe "... cost _that_ is ..."?

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+recipient_address = "rBXsgNkPcDN2runsvWmwxk3Lh97zdgo9za"
+account_objects_response = get_account_objects({
+    account: recipient_address,
+    ledger_index: "validated",
+    type: "check"
+})
 
+for (i=0; i < account_objects_response.account_objects.length; i++) {
+  check_object = account_objects_response.account_objects[i]
+  if (check_object.Account == recipient_address) {
```
scottschurr: Should this be `check_object.Destination == recipient_address`?

content/tutorial-checks.md
scottschurr: In the `DeliverMin` field description, below, you say something like, "If you cannot receive this much, cashing the Check fails, leaving the Check in the ledger so you can try again."  I think that would be a great addition here as well.

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+
+{% include 'snippets/wait-for-validation.md' %}
+
+### {{cash_exact_n.next()}}. Confirm final result
+
+Use the [`tx` method](reference-rippled.html#tx) with the CheckCash transaction's identifying hash to check its status. Look for a `"TransactionResult": "tesSUCCESS"` field in the transaction's metadata, indicating that the transaction succeeded, and the field `"validated": true` in the result, indicating that this result is final.
+
+If the check was cashed for an exact `Amount` and succeeded, you can assume that the recipient was credited for exactly that amount (with possible rounding for very large or very small amounts of issued currencies).
+
+If cashing the Check failed, the Check remains in the ledger so you can try cashing again later. You may want to [cash the Check for a flexible amount](#cash-a-check-for-a-flexible-amount) instead.
```
scottschurr: I like this sentence.  I wonder if there's a good place for a comparable sentence in the **Confirm final result** section of **Cash a Check for a flexible amount**?

scottschurr (approved):
Nice work.  I did note a small number of copy-and-paste-os the you will probably want to address. :+1:

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
-# Checks Tutorial (draft)
+# Checks Tutorials
 
-***TODO: This whole page is just example requests and responses for now, not a finished page.***
+Checks in the XRP Ledger are similar to paper personal checks. This tutorial steps through the following processes for using Checks:
```
ryangyoung: This might be a good place to add a link the conceptual article.

ryangyoung: This, or within the tasks themselves, might also be a good place to provide a brief use-case like statement. _Why_ would I want to do these things?

ryangyoung: A bit of a stretch and something that would involve more work, but it would be really cool to see all of these tasks tied together by a sort of pseudo-narrative that explains the "why" for each task. For example, BobCo did some crypto consulting work for AliceCo and to fulfill the invoice they received, AliceCo wants to send a check to BobCo's account at Acme Credit Union, which has deposit auth enabled. (It would be _really_ cool to tie the narrative to one of our known use cases, but that's probably even more of a stretch.)

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
 
-<!--{# Accounts used in this example:
-rUn84CUYbNjRoTQ6mSW7BVJPSVJNLb1QLo snkuWqxoqt6aeykTbkEWrTMJHrWGM (as the sender)
-rfkE1aSy9G8Upk4JssnwBxhEv5p4mn2KTy sn2Zh1tRZyodU9qNy9tMnQr9UbBss (as the dest.)
+## Send a Check
 
-both need to be funded and the sender needs enough XRP to send and pay the reserves
-Reminder: don't use these addresses for real XRP because the secrets are now public
-#}-->
+Sending a Check involves sending a [CheckCreate transaction][], which creates a [Check object in the ledger](reference-ledger-format.html#check). Most of the fields of a CheckCreate transaction are similar to the fields of a (push) [Payment][], because a Check is like writing permission for an intended recipient to pull a payment from you. You can create a check for XRP or an issued currency.
```
ryangyoung: This is great, but, again, it would be great to see at least a brief explanation of why I'd want to do this. (Also, this is mostly covered in the concept article, though without a few of the details here. So I'd question if we want to reiterate that content here. In other words, if I wanted to understand how it works, the concept is where I'd look.)

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
 
-## Create check
+{% set send_n = cycler(* range(1,99)) %}
 
-CheckCreate tx rUn84→rfkE1
+### Prerequisites
```
ryangyoung: Are these prerequisites for all of the tasks in this topic? If so, I'd suggest moving this section up a level since you need these things do the rest of the steps, not just sending a check.

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+
+To send a Check with this tutorial, you need the following:
+
+- The **address** and **secret key** of a funded account to send the Check from.
+    - You can use the [XRP Ledger Test Net Faucet](https://ripple.com/build/xrp-test-net/) to get a funded address and secret with 10,000 Test Net XRP.
```
ryangyoung: This is great. Would it also be worth linking to `wallet_create` and whatever other methods you'd need to do the same on the production network... oh, when Checks is enabled there. Never mind.

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+- The **address** of a funded account to receive the Check.
+- A secure way to sign transactions, such as [RippleAPI][] or your own [`rippled` server](tutorial-rippled-setup.html).
+- A client library that can connect to a `rippled` server, such as [RippleAPI][] or any HTTP or WebSocket library.
+    - For more information, see [Connecting to `rippled`](reference-rippled.html#connecting-to-rippled).
+
+### {{send_n.next()}}. Prepare the CheckCreate transaction
+
+Decide how much money the Check is for and who can cash it. Figure out the values of the [CheckCreate transaction][] fields. The following fields are the bare minimum; everything else is either optional or can be [auto-filled](reference-transaction-format.html#auto-fillable-fields) when signing:
```
ryangyoung: Especially for a tutorial, I think it would help the user to provide an arbitrary amount (whatever's in the example) and receiver (e.g. BobCo) here. Basically, another appeal for the use case/narrative thread.

ryangyoung: The following table is, I think, already in the reference. I know this version is listing only the bare minimum fields, but I wonder if we or want to recreate it here. One goal of tutorials, I think, is to try to make things _seem_ easier (without sacrificing accuracy, of course).

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+
+| Field             | Value                     | Description                  |
+|:------------------|:--------------------------|:-----------------------------|
+| `TransactionType` | String                    | The value `CheckCreate` indicates this is a CheckCreate transaction. |
```
ryangyoung: If we do keep the table here, I think the description fields should be more targeted toward the user. As a user, I'd ask: so, is `CheckCreate` what I should use here, or what? The existing language seems better suited to reference material.

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+| `Account`         | String (Address)          | The address of the sender who is creating the Check. (In other words, your address.) |
```
ryangyoung: Another place where some example value would really help a user new to Checks.

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+| `Destination`     | String (Address)          | The address of the intended recipient who can cash the Check. |
```
ryangyoung: Again, an example value (like BobCo) would be great here.

content/tutorial-checks.md
ryangyoung: Agree with Scott on both points. I think an example scenario would really clarify that caveat, too.

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+| `SendMax`         | String or Object (Amount) | The maximum amount the sender can be debited when this Check gets cashed. For XRP, use a string representing drops of XRP. For issued currencies, use an object with `currency`, `issuer`, and `value` fields. See [Specifying Currency Amounts][] for details. If you want the recipient to be able to cash the Check for an exact amount of a non-XRP currency with a [transfer fee](concept-transfer-fees.html), remember to include an extra percentage to pay for the transfer fee. |
+
+If you are using [RippleAPI](reference-rippleapi.html), you can use the `prepareCheckCreate()` helper method.
+
+**Note:** RippleAPI supports Checks in versions 0.19.0 and up.
+
+#### Example CheckCreate Preparation
```
ryangyoung: Nit: since this is a tutorial, how about rephrasing this to something like  "Example: Prepare CheckCreate Transaction"

ryangyoung: Ah, I see. This makes more sense in the generated doc.

content/tutorial-checks.md
```diff
@@ -1,588 +1,928 @@
...
+
+```json
+{% include 'code_samples/checks/cli/sign-create-resp.txt' %}
 ```
 
-## wait for validation / ledger_accept
+<!-- MULTICODE_BLOCK_END -->
 
-## tx on checkcancel
+### {{send_n.next()}}. Submit the signed transaction
```
ryangyoung: To beat what may at this point already be a dead horse, the text in the include reads more like a reference than a tutorial. Again, as a user, I might ask something like, "OK, cool. What do _I_ do now?"

ryangyoung (changes requested):
This content is great and, I know, the result of a lot of hard and sometimes frustrating work. It's wonderful that we have working examples for all of the permutations of sending and cashing checks. This is a framework that we can modify and build on, and, in some ways can serve as a model. 

That said...a lot of the material reads very similarly to a reference. It's awesome that there are specific example requests and responses for each possible permutation. But, because a lot of the examples cover so much information and caveats, I worry that a user who _just_ wants to figure out how to send a check will end up lost in some of the details that may be better housed in the reference. I think this can be addressed by two hopefully not too onerous changes:

* Split up each tutorial into its own file. Currently, it's very long. I think a better user experience would be to have each tutorial (send, look up checks, cash checks for exact amount, etc) on a separate, short page. That provides at least an illusion of simplicity. And the more flexible nesting planned for Dactyl will make it easier to discover the possible next steps. 

* Add an example scenario that provides some kind of use case (tied if possible to one of our top-level personas and/or use cases). That way, it's clear to a user that, "hey, I could use this to fulfill an invoice!" (or whatever). A scenario that unfolds across each separate tutorial would be especially helpful (i.e., Alice prepares checkCreate tx to Bob for widget delivery, Alice submits CheckCreate transaction, Bob looks up checks to him, Alice looks up status of checks, Bob cashes check for exact amount, Bob cashes another check from Alice for a flexible amount).

content/tutorial-checks.md
mDuo13: Yeah, I waffled between "cost destroyed" and "cost that is destroyed" and ended up at a completely unreasonable compromise. ;)

content/tutorial-checks.md
mDuo13: The equivalent sentence when you're already cashing for a flexible amount is in the "Handling Errors" section:

> If cashing the Check failed with a `tec`-class code, look up the code in the [Full Transaction Response List] and respond accordingly.

content/snippets/checkcash-prereqs.md
```diff
@@ -0,0 +1,10 @@
+The prerequisites for cashing a check are the same whether you are cashing it for an exact amount or a flexible amount.
+
+- You need the ID of a Check object currently in the ledger.
```
ryangyoung, line 3: Nit: for readability, add at least one line between each item in the list.

content/tutorial-checks.md
```diff
@@ -1,588 +1,932 @@
...
-rfkE1aSy9G8Upk4JssnwBxhEv5p4mn2KTy sn2Zh1tRZyodU9qNy9tMnQr9UbBss (as the dest.)
+## Send a Check
 
-both need to be funded and the sender needs enough XRP to send and pay the reserves
-Reminder: don't use these addresses for real XRP because the secrets are now public
-#}-->
+Sending a Check is like writing permission for an intended recipient to pull a payment from you. The outcome of this process is a [Check object in the ledger](reference-ledger-format.html#check) which the recipient can cash later.
 
-## Create check
+In many cases, you want to send a [Payment][] instead of a Check, since that delivers the money directly to the recipient in one step. However, if your intended recipient uses [DepositAuth](concept-depositauth.html), you cannot send them Payments directly, so a Check is a good alternative.
```
ryangyoung: 🙌

ryangyoung (approved):

ryangyoung:
I haven't built this locally, but it looks like each section of the tutorial was broken up into its own section, which should improve the user experience. That addresses my first major point. 

The second major point, about adding some kind of through-line scenario or use case was somewhat addressed with the addition of some language about why a user would want to send or cancel a check. That's great. I think we could extend that even further, though, and for that, I added https://ripplelabs.atlassian.net/browse/DOC-1380 (Ripple employees only).  But as far as I'm concerned, this is good to publish for now. We can add the use case stuff later.

PR #331, opened 2018-04-04
Fix typo and clarify amendments in genesis ledgers

mDuo13:
Started this to fix a "all The genesis ledger" typo but ended up adding a heading so we can link to this section specifically, and clarified the amendment stuff a little.


ryangyoung (approved):
Took a quick look; didn't see any glaring issues. Adding the header was a good idea!

PR #333, opened 2018-04-04
Clarify transfer fee precision

mDuo13:
The old docs said 9 digits of precision, which is confusing because `TransferRate` values are 10 digits long. (The trick is, you can't change the leading 1 except in the case of the max value `2000000000`.) It's really 9 digits after the decimal point when expressed as a decimal (1.n), which is 7 digits after the decimal if expressed as a percentage (n%).

This change attempts to clear that up without getting too complicated.

Resolves DOC-415


ryangyoung (approved):
👍

PR #337, opened 2018-04-13
New Dev Portal Templates (in-progress)

mDuo13:
For now, use `dactyl_build -c 5layers.yml` to build the version that has the new hierarchical structure. The plan is to build out the new config and eventually swap it in for the existing `dactyl-config.yml` file.

- [x] breadcrumbs down to 5 layers of nesting for docs
- [x] placeholder templates for some of our planned landing pages
- [x] alternate config that demonstrates the 5-layer nesting. 
- [x] Category title links to first page in category, with the expander/collapser being a separate button
- [x] Change expander/collapser to use "+/-"  and be to the left of the category name
- [x] top navigation by "funnel"
- [x] category pages have a "doc_type" header above the side nav
- [x] search bar (uses Google—needs Google to index the site before it'll work)
- [x] new footer
- [x] new Home page
    - <s>"You're in Great Company" section</s>
   - [x] "Built on the XRP Ledger" section
   - [ ] Non-placeholder graphics?
- [x] "Docs" landing page
    - [x] drop "popular tools" from Docs landing
    - [x] add Doc Index to docs landing (below the fold) and drop index from home landing
- [x] Sidebars that work outside of the "Docs" funnel
- [x] "Use Case" page templates
- <s>"Related Resources" box-</s>
- [x] Change home landing call-to-action verbs
- [x] full fluid layout
    - [x] 1-column pages
    - [x] 3-column pages
    - [x] tools with sidebar nav (move to right in 3-col layout?)
    - [x] breakpoint for right column on narrow views
- [x] Extra "supercategory" level for rippled API reference
    - [x] sidebar nav support
    - [x] breadcrumbs support
    - [x] page-children support
- [x] full "Docs" index support in page-children
- [x] API Reference landing page
- [ ] Landing design updates
    - [x] New splash at top
    - [x] Redone "XRP Ledger Features" section
    - [ ] Update "Built on XRP Ledger" section
        - [ ] button-links for every card (just "Contact" for the not-yet-written ones)
        - [ ] align icons (text below icons maybe?)
        - [x] move "Have an idea?" up above modules
    - [ ] Dividing lines between some modules?
    - [ ] Make "References" look good under the new rules / design space
- [ ] General visual polish
    - [x] optimize line length by font size
    - [ ] Test Net sidebar cleanup
    - [ ] ripple.txt Validator input box sizing
    - [x] Make modules cards with visible edges again
    - [x] Remove bullet points from link lists in cards
        - [ ] fix link hierarchy in some cards (e.g. References box on Docs home)
    - [ ] Only one link to known amendments
    - [ ] Use dates for "New and Updated Docs"
    - [ ] For cards with one action, make the whole card a link w/ hover effect but no header links
        - [x] make whole card a link
        - [ ] add hover effect
    - [ ] smaller icons and fonts for XRP features
    - [x] "Related Resources" - align body text (maybe by moving "login required" up a line)
    - [x] "external link" icon smaller everywhere
    - [x] Reduce colors in footer (match ripple.com?)
    - [x] match footer padding to landing section padding
    - [x] move search button inside search input box (!?)
    - [x] avoid center-aligning things
        - [ ] fix card images centering undesirably
    - [ ] make entire left nav collapsible(??)
    - [ ] make 3-column layout sidebars wider, collapse sooner
    - [x] give right sidebar more space to the right (avoid two side-by-side scrollbars)
    - [x] fix the header spilling into two rows at 993-1076px wide (shrink search?)
    - [ ] fix breadcrumb spacing inconsistencies (refs have too much, concepts not enough??)
    - [ ] remove bg-color hover in left nav
    - [ ] change font colors/weights of left nav to more consistent hierarchy
        - [ ] maybe invert link color/black pattern for sidebar
    - [x] re-add underlines to content links
    - [x] use Roboto Mono for code


jhaaaa:
This is fantastic -- it is EVERYTHING! Thank you for figuring out a way to handle this UX. Wanted to check on just a couple of things -- please forgive if some of the review feedback is not appropriate for this PR. Please forgive if I've gone down the 🐰 hole!

1. We need a style to handle when a category item in the left nav does not have children. I think this is rare to non-existent in the **Docs** area, but it is present in **Dev Tools** and **News**. Suggested style starting point here: https://invis.io/YZGMZX5TPXF#/291604706_Selected-NoChildren. In this _example_, note that **Introduction** has no children and is the current page. Note that **The Exchange Business Model** has no children and is not the current page. 

2. This may be coming later, but wanted to mention this sooner than later, just in case. 🙌  The expand/collapse icon in the left nav should point down when the section is collapsed and point up when the section is expanded. Basically, the arrow should point in the direction of UI movement initiated when a user clicks the arrow. For example: https://projects.invisionapp.com/share/YZGMZX5TPXF#/screens/291604707

3. This may also be coming later, but just calling it out just in case. For the **Concepts** area, the top of the left nav needs to display the title of the landing page and provide access to it once you’ve navigated into one of its children. For example, see the navigation from landing-concepts.html to concept-accounts.html. Once I am on concept-accounts.html, the only way to get back to landing-concepts.html is via the breadcrumbs. The title in the left nav also provides a way for the user to orient themselves once they've been clicking links moving around in the docs -- I can look there and get a reality check of what area I'm in. See https://invis.io/YZGMZX5TPXF#/288519100_Concepts-Landing for placement of this title. The equivalent functionality in the client portal docs is the **Administration Guide** link (minus the << link) at the top of the left nav here: https://clients-staging-private.ripple.com/docs/xcurrent/v3.4.0/administration-guide/xc-revision-summary-admin-guide.html

4. Again, may be coming later -- but please see reference-rippled-public.html. For this area, I think the category should actually be **rippled API Reference**. The subcategories are then Public rippled Methods, Admin rippled Methods, Transaction Formats, Ledger Data Formats, and API Conventions. See the mockup here: https://invis.io/YZGMZX5TPXF#/288519125_Rippled-API-Landing

Thank you again for creating this for us -- we're so lucky to have your expertise on the team, Rome! 😅

mDuo13:
First: Make sure you're building `dactyl_build -c 5layers.yml` to get the version that has the full hierarchy (although not all the pages).

Point-by-point:

1. I was planning on having those other funnels having different sidebar code anyway, but that's a good point on "The Exchange Business Model". I can probably do the style up so it looks OK with no children. Rather than having a box with a dot, I'll just have no expander/collapser button in that case.

2. Your recommendation that "the arrow should point to the direction of UI movement" didn't match with my experience, so I did some research on the conventions of accordion/tree UI buttons. What I found is that the arrow should point sideways when collapsed, and down when expanded. (Sources [[1]](https://www.nngroup.com/articles/tree-testing/), [[2]](https://ux.stackexchange.com/questions/5735/which-way-should-arrows-point-in-a-collapsible-accordion-left-down-down-up), [[3]](https://www.viget.com/articles/testing-accordion-menu-designs-iconography/).)

    The studies I could find were for interfaces where clicking the category text was the same as expanding it, where ours has slightly different actions associated with those actions. They determined that putting the icon left of the text led to faster task time, possibly because when the icon was on the right people thought that the actions might be different. In either case, it's very consistent that an accordion or tree layout has the collapsed arrow point sideways (usually to the right, even when it's on the right of the text, which is backwards of what we currently have) if it uses arrows. Using +/- instead is also common and +/× was an interesting recommendation that supposedly has at least one user test favoring it.

    I think we're OK having the arrow to the right of the text (since in our case they really are different actions), but I would also consider moving it to the left and changing to +/- instead.

3. This is planned, just not done yet.

4. That's too many levels of hierarchy. We changed this in the IA meeting and the consensus file Ryan saved from that doesn't have a "rippled API Reference" grouping.

mDuo13:
Oh, UI problem we should probably change is that the "in this page" table and the left nav shouldn't use arrows/chevrons differently. (Currently, the "in this page" nav uses them as unclickable bullets; I think maybe we should remove them completely.)

mDuo13:
Rebased onto master to dodge the reliable transaction submission conflicts. I think the next step is to merge this (even though it's all broken)

PR #338, opened 2018-04-13
Improve reliable transaction submission pseudo-code & add ledger gap section

mDuo13:
- Changes tabs to spaces
- Adds comments and spacing clarifying major cases in the pseudo-code
- Fixes pointless/misleading resubmit suggestion when outcome is final but unknown due to ledger gap
- Adds a section clarifying what to do in case of ledger gaps

### DIFF
--- a/content/tutorial-reliable-transaction-submission.md
+++ b/content/tutorial-reliable-transaction-submission.md
@@ -104,23 +104,35 @@ On restart, or the determination of a new last validated ledger (pseudocode):
 For each persisted transaction without validated result:
     Query transaction by hash
     If (result appears in validated ledger)
-		Persist the final result
-		If (result code is tesSUCCESS)
-			Application may act based on successful transaction
-		Else
-			Application may act based on failure
-			Maybe resubmit with new LastLedgerSequence and Fee
-	Else if (LastLedgerSequence > newest validated ledger)
-        Wait for more ledgers to be validated
-	Else
-		If (server has contiguous ledger history up to and
-	        including the ledger identified by LastLedgerSequence)
-			The transaction failed
-			Submit a new transaction, if appropriate for application
-		Else
-			Repeat submission of original transaction
+      # Outcome is final
+      Persist the final result
+      If (result code is tesSUCCESS)
+        Application may act based on successful transaction
+      Else
+        Application may act based on failure
+        Maybe resubmit with new LastLedgerSequence and Fee
+
+    Else if (LastLedgerSequence > newest validated ledger)
+      # Outcome is not yet final
+      Wait for more ledgers to be validated
+
+    Else
+      If (server has contiguous ledger history up to and
+            including the ledger identified by LastLedgerSequence)
+        The transaction failed
+        Submit a new transaction, if appropriate for application
+      Else
+        # Outcome is final, but not known due to a ledger gap
+        Wait to acquire contiguous ledger history
 ```
 
+#### Ledger Gaps
+
+If your server does not have contiguous ledger history from when the transaction was originally submitted up to and including the ledger identified by LastLedgerSequence, you may not know the final outcome of the transaction. (If it was included in one of the ledger versions your server is missing, you do not know whether it succeeded or failed.)
+
+Your `rippled` server should automatically acquire the missing ledger versions when it has spare resources (CPU/RAM/disk IO) to do so, unless the ledgers are older than its configured amount of history to store. Depending on the size of the gap and the resource usage of your server, acquiring missing ledgers should take a few minutes. You can also manually request your server to acquire historical ledger versions using the [`ledger_request` command](reference-rippled.html#ledger-request).
+
+Alternatively, you can look up the status of the transaction using a different `rippled` server that already has the needed ledger history, such as Ripple's full-history servers at `s2.ripple.com`. Only use a server you trust for this purpose. A malicious server could be programmed to provide false information about the status and outcome of a transaction.
 
 
 ## Technical Application
--- a/content/tutorial-reliable-transaction-submission.md
+++ b/content/tutorial-reliable-transaction-submission.md
@@ -89,6 +89,7 @@ The submission process:
     - Transaction hash
     - `LastLedgerSequence`
     - Sender address and sequence number
+    - Most recently-validated ledger index at the time of submission
     - Application-specific data, as needed
 3. Submit the transaction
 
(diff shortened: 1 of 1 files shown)


content/tutorial-reliable-transaction-submission.md
```diff
@@ -104,23 +104,35 @@ On restart, or the determination of a new last validated ledger (pseudocode):
...
+      If (server has contiguous ledger history up to and
+            including the ledger identified by LastLedgerSequence)
+        The transaction failed
+        Submit a new transaction, if appropriate for application
+      Else
+        # Outcome is final, but not known due to a ledger gap
+        Wait to acquire contiguous ledger history
 ```
 
+#### Ledger Gaps
```
ryangyoung: 👍

content/tutorial-reliable-transaction-submission.md
```diff
@@ -104,23 +104,35 @@ On restart, or the determination of a new last validated ledger (pseudocode):
...
-			Submit a new transaction, if appropriate for application
-		Else
-			Repeat submission of original transaction
+      # Outcome is final
+      Persist the final result
+      If (result code is tesSUCCESS)
+        Application may act based on successful transaction
+      Else
+        Application may act based on failure
+        Maybe resubmit with new LastLedgerSequence and Fee
```
ryangyoung: Is there any clarity we can add about when you should or shouldn't resubmit so that we could remove the "maybe"?

content/tutorial-reliable-transaction-submission.md
```diff
@@ -104,23 +104,35 @@ On restart, or the determination of a new last validated ledger (pseudocode):
...
+        The transaction failed
+        Submit a new transaction, if appropriate for application
+      Else
+        # Outcome is final, but not known due to a ledger gap
+        Wait to acquire contiguous ledger history
 ```
 
+#### Ledger Gaps
+
+If your server does not have contiguous ledger history from when the transaction was originally submitted up to and including the ledger identified by LastLedgerSequence, you may not know the final outcome of the transaction. (If it was included in one of the ledger versions your server is missing, you do not know whether it succeeded or failed.)
```
ryangyoung: How about "continuous" work instead of "contiguous"? "Contiguous" seems like it might take readers a second longer to parse than necessary.

ryangyoung: > If it was included in one of the ledger versions your server is missing, you do not know whether it succeeded or failed.

Suggest changing "it" to "the transaction" to be totally clear.

Also, I'm having trouble understanding how the second sentence is different than the first. The second sentence is clearer to me. (That might just be me.)

content/tutorial-reliable-transaction-submission.md
```diff
@@ -104,23 +104,35 @@ On restart, or the determination of a new last validated ledger (pseudocode):
...
+
+Your `rippled` server should automatically acquire the missing ledger versions when it has spare resources (CPU/RAM/disk IO) to do so, unless the ledgers are older than its configured amount of history to store. Depending on the size of the gap and the resource usage of your server, acquiring missing ledgers should take a few minutes. You can also manually request your server to acquire historical ledger versions using the [`ledger_request` command](reference-rippled.html#ledger-request).
```
ryangyoung: Maybe add a link or a mention of where you can change that configuration. Maybe here? https://github.com/ripple/rippled/blob/develop/cfg/rippled-example.cfg#L581

ryangyoung (commented):
Suggested another term, a minor wording change, and adding a link, but otherwise looks good to me. The new section is very helpful!

content/tutorial-reliable-transaction-submission.md
mDuo13: We can possibly tighten up the language, but whether or not you want to try sending a transaction again if it fails is entirely up to your business logic. I'll think on this.

content/tutorial-reliable-transaction-submission.md
mDuo13: The pseudo-code also uses "contiguous". We can probably change both, though.

I can try to rephrase this paragraph again. I was trying to convey the difference between the following 3 cases:

- If the transaction is in a validated ledger you have, you know its outcome (succeed or fail) even if you have a ledger gap, because you only need that one ledger that has the transaction in question.
- If you have _all_ the ledgers the transaction _could_ be in and it isn't in any of them, you know it failed.
- If you don't have a ledger with the transaction in it, but the transaction could be in a ledger you don't have, you don't know the outcome of the transaction.

ryangyoung: Got it. 

In the first case, where would you get the validated ledger? From s2 or the Data API or something like that?

Anyway, the way it makes sense to me is:

* You know the outcome of the transaction because:
    * You have all of the possible the ledgers.
    * You have the ledger with the transaction.

* You don't know the outcome of a transaction because:
    * You don't have all of the possible ledgers.
    * You don't have the ledger with the transaction (which is pretty much the same as not having all the possible ledgers, right?)

mDuo13: The first case is meant to describe how you could have a gap in _the range the transaction could have been in_, but it doesn't matter because you have _the one ledger the transaction actually ended up in_. So we're still talking about your own server's ledger store.

And while the way it makes sense to you is not wrong, there's a small difference between "not having all the possible ledgers" and "not having the ledger with the transaction". It's the difference between proving a positive ("it's right here") versus proving a negative ("it's not _anywhere_").

content/tutorial-reliable-transaction-submission.md
```diff
@@ -89,6 +89,7 @@ The submission process:
     - Transaction hash
     - `LastLedgerSequence`
     - Sender address and sequence number
+    - Most recently-validated ledger index at the time of submission
```
intelliot: Suggestion:

- Latest validated ledger index at a time prior to submission

content/tutorial-reliable-transaction-submission.md
```diff
@@ -104,23 +105,49 @@ On restart, or the determination of a new last validated ledger (pseudocode):
...
-	Else if (LastLedgerSequence > newest validated ledger)
-        Wait for more ledgers to be validated
-	Else
-		If (server has contiguous ledger history up to and
-	        including the ledger identified by LastLedgerSequence)
-			The transaction failed
-			Submit a new transaction, if appropriate for application
-		Else
-			Repeat submission of original transaction
+      # Outcome is final
```
intelliot: Consistently using 4 spaces for indentation and/or using {curly braces} might make this more readable.

content/tutorial-reliable-transaction-submission.md
```diff
@@ -104,23 +105,49 @@ On restart, or the determination of a new last validated ledger (pseudocode):
...
+
+The difference between the two transaction failure cases (labeled (1) and (2) in the pseudo-code) is whether the transaction was included in a validated ledger. In both cases, you should decide carefully how to process the failure.
+
+- In failure case (1), the transaction was included in a ledger and destroyed the [XRP transaction cost](concept-transaction-cost.html), but did nothing else. This could be caused by a lack of liquidity, improperly specified [paths](concept-paths.html), or other circumstances. For many such failures, immediately retrying with a similar transaction is likely to have the same result. You may get different results if you wait for circumstances to change.
+
+- In failure case (2), the transaction was not included in a validated ledger, so it did nothing at all, not even destroy the transaction cost. This could be the result of the transaction cost being too low for the current load on the XRP Ledger, the `LastLedgerSequence` being too soon, or simply a coincidence. In contrast to failure case (1), it is more likely that a new transaction is likely to succeed if you change only the `LastLedgerSequence` and possibly the `Fee` and submit again.
```
intelliot: Suggestion:

or simply a coincidence -> or an unreliable server and/or network connection

I guess it could also be due to another transaction, consuming the same account sequence number, getting into the ledger first.

intelliot (approved):
Solid improvement over the existing doc. Just a few suggestions.

content/tutorial-reliable-transaction-submission.md
mDuo13: There are lots of things that could go wrong, but I agree "simply a coincidence" is too hand-wavey. Changed to:

> or it could be due to other conditions such as an unstable network connection.

I don't want to exclude other things, but that's probably the biggest and most understandable one.

PR #345, opened 2018-05-09
Add transaction formats to reorg

mDuo13:
Made a couple changes compared to the mind map:

- Pseudo-transactions get their own subcategory (as they used to)
- Added a page for Transaction Metadata


ryangyoung (approved):
This looks good to me. I didn't read everything but I scanned headings and the config entries and nothing jumped out at me. I also built this locally and, from a quick scan, it looks good.

PR #346, opened 2018-05-09
Reorg ledger data formats

mDuo13:


jhaaaa:
LGTM!

Just mentioning a couple of things as a reality check:

- The [Internal Type][] column heading link for field table will be fixed when we move into linking fixing mode, correct?

- For Category and SubCategory landing pages, not just in this section but throughout - some have just the context-setting overview of child pages, and some have content AND the context-setting overview of child pages that displays after the content. Sometimes the content itself provides the overviews in some form. As a part of the UI content work I did, I'll open a ticket for me to review all of these pages after the full migration to tidy this up - does that sound okay? Ideally, we'll primarily use the autogeneration from blurbs functionality that you built.

jhaaaa (approved):

mDuo13:
Yeah, I think the "Internal Type" thing should get fixed when #345 is merged because that one adds it to the common links?

And you're right about the landing page inconsistency. It would be nice to figure out exactly how we want that to work and get it working for more things. Gotta add the page-toc sidebar to the landing template for the landing pages that have more than trivial content though.

PR #348, opened 2018-05-09
Reorg 'The rippled Server' concepts

mDuo13:
Some of this content is a _little_ out of place in "Concepts", but I'm trying not to get too deep into slicing and dicing content up at this point.


jhaaaa (approved):
LGTM

PR #343, opened 2018-05-09
Reorg admin methods

ryangyoung:
Updates config file to match new structure.

PR #350, opened 2018-05-10
reorg public methods

jhaaaa:


mDuo13 (changes requested):
Almost the exact same feedback as I had on the first draft of #343 (which is probably a sign my instructions were lacking).

- The method files need to be updated to change their header levels. They should start at h1. Currently they start at h2. Be sure to also promote the other headers. You can do this with a regular expression find/replace in Atom:

    1. In the find-in-projecct box (Ctrl+Shift+F or similar), turn on the regular expression option (".*" icon)
    2. In the find box, enter `^## ` (lines starting with two # followed by a space)
    2. In the replace box, enter `# ` (start with # and a space)
    3. In the file/directory pattern box, enter `content/references/rippled-api/public-rippled-methods/*`
    4. Find and replace all
    5. Change the "find" and "replace" boxes to add another # to each (so you're turning H3's to H2's now) and repeat until you don't find any headers. (Be sure to do this from low-level headers first so you don't end up promoting the same header multiple times.)

- For the subcategory landings where the content is just an opening description and a list of children, the landing will work better if you add blurbs to the landing and its child pages. (If this seems like too much work for now, feel free to open a JIRA task to follow up on it later.)
    - You can pull blurbs for the methods from https://github.com/ripple/ripple-dev-portal/blob/master/content/references/rippled-api/rippled-old-monolith.md#list-of-admin-commands 
    - The `rippled-old-monolith.md` file has text that introduces the subcategories in between all the includes. (I actually even made some updated text [in the failed PR to reorg it before](https://github.com/ripple/ripple-dev-portal/pull/313/files#diff-011ffcf96d40f49e25a8e38dd867c701R56). Maybe you can use that to inspire the blurbs for the subcategory landing pages.

- You should delete content from the old monolith file as you get the equivalent content placed into the new structure. This helps us keep track of what content doesn't have a place in the new organization yet.

dactyl-config.yml
```diff
@@ -543,18 +544,106 @@ pages:
...
+            - local
+
+    -   name: Ledger Methods
         html: ledger-methods.html
         funnel: Docs
         doc_type: References
         supercategory: rippled API
         category: Public rippled Methods
         subcategory: Ledger Methods
+        blurb: Each `rippled` server keeps a complete copy of the XRP Ledger's current state, which contains all the accounts, transactions, offers, and other data in the network in an optimized tree format. As transactions and offers are proposed, each server incorporates them into its current copy of the ledger, closes it periodically, and (if configured) participates in advancing the globally-validated version. After the network reaches consensus, that ledger version is validated and becomes permanently immutable. Any transactions that were not included in one ledger version become candidates to be included in the next validated version.
```
mDuo13: The backticks around `rippled` don't work here, so we should remove them.

dactyl-config.yml
```diff
@@ -612,16 +811,189 @@ pages:
...
+            - local
+
+    -   name: Server Info Methods
+        html: server-info-methods.html
+        funnel: Docs
+        doc_type: References
+        supercategory: rippled API
+        category: Public rippled Methods
+        subcategory: Server Info Methods
+        blurb: Learn about methods that enable you to retrieve information about the current state of the `rippled` server. These may be useful for monitoring the health of the server, or in preparing to make requests to other API methods.
```
mDuo13: broken backticks to be removed here too

mDuo13 (approved):
caught two broken things that should be fixed, but otherwise looks fantastic!

PR #360, opened 2018-05-11
Reorg payment system basics category

mDuo13:
Also went slightly outside the "Payment System Basics" category when I realized there was a section on multi-signing left over in the transactions monolith that would merge nicely with the multi-signing placeholder article I set up earlier.

Didn't include the metadata page since that's under the transactions reference instead. (Which I think is a better place for it, probably.)

Moved "Transaction Cost" under "Transaction Basics" since that makes sense to me.


jhaaaa (approved):
LGTM!

PR #376, opened 2018-05-12
New dev portal style update

jbheron:
The goal here was to mostly tighten up the layout. Additionally, to align a few elements with what we're already doing on ripple.com. 

Nothing here is drastic, just iterating.

There are no changes to the bootstrap file.

CAVEAT: @mDuo13, or whomever review this PR, I (still) can't get my `-local` target to build, so I'm really flying blind, working off of the staging site and dev tools (_#dontrefresh!_ ). I've checked my work, but this needs to be built for a sanity check.

### DIFF
--- a/assets/css/devportal.css
+++ b/assets/css/devportal.css
@@ -463,7 +463,6 @@ a.current {
 
 .landing .card .level-1 a,
 .landing .card .curated-links li a {
-  padding: 12px;
   display: block;
   margin: 4px;
   font-weight: 200;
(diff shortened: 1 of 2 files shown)


mDuo13:
Two things:

1. Some of the changes here, such as lightening up the font, are at odds with the feedback we got from design and marketing. I think the ripple.com site design you're imitating is still sort of based on an older generation of style guide whereas the new dev portal styles are following the newer one. See #377 for a PR implementing the changes directly from design.

2. It mostly builds OK, but there are some minor glitches like icons not being centered in their cards and shading for the headers not filling the space.

Thank you for the effort but I'm not comfortable merging these changes at this time.

jbheron:
@mDuo13, I couldn't actually disagree more about this. I don't really care about the font color, other than to remove to totally garish orange choice that looks cartoonish-- and of course, that's not the main point here. The scale of the cards is almost laughably large, as are the icons within them.

It's unfortunate that the site will go out how it is, especially given that the changes this PR requests are major improvements to the amateurish look that need _something_ done to them. I understand we're pretty under resourced in this case, and it's last second. Sure, nothing's perfect, but the way the layout has been scaled is really bad in its current condition. It's too bad you can't see that.

jbheron:
Ugh, this is really annoying. I didn't see DOC-1544. @mDuo13 Thanks for getting the updates, that was my only goal-- didn't think you'd have time. Please excuse the 🔥

PR #382, opened 2018-05-15
Add new dev portal logo

ryangyoung:
Replaces current logo with new one provided by the design team.


mDuo13:
Looks good. I think ideally I'd like the text to be a little bigger, but that can come later if others agree.

The broken link is unrelated to this change—it's caused by the config file moving in the rippled 1.0 release that came out this morning.

mDuo13 (approved):

PR #383, opened 2018-05-15
fix issued currencies landing

jhaaaa:
- Moved content on issued-currencies.md to child page: issued-currencies-overview.md
- Removed issued-currencies.md file to be able to autogen landing page content from dactyl_config blurbs


mDuo13 (approved):
:+1: 
Built. Looks great!

PR #384, opened 2018-05-15
fix rippled server landing

jhaaaa:
- Moved content on the-rippled-server.md to rippled-server-modes.md.
- Deleted the-rippled-server.md so that The rippled Server landing page can be autogenerated from dactyl_config blurbs.
- Moved content from stand-alone-mode.md to rippled-server-mode.md. Wasn't expecting to do this, but it just made sense. Any thoughts?
- Deleted stand-alone-mode.md.

Looking at the structure of The rippled Server concepts section now - do we need an overview page for the `rippled` server? We basically have a one-sentence overview of the rippled server on the autogen landing page: rippled is the core peer-to-peer server that manages the XRP Ledger. Okay?


mDuo13:
> Check is failing due to broken link: https://github.com/ripple/rippled/blob/master/doc/rippled-example.cfg. Is this a new issue?

Yeah, it looks like rippled config moved to https://github.com/ripple/rippled/blob/master/cfg/rippled-example.cfg —maybe with the 1.0 release that came out today?

mDuo13:
The rearrangement here tentatively looks good, but I'll take a closer look and come back to it later.

mDuo13 (approved):
:100: 
Excellent reorganization.

PR #385, opened 2018-05-15
Fix Use Simple XRP Payments landing page

jhaaaa:
- Delete use-simple-xrp-payments.md
- Update blurb for Use Simple XRP Payments in dactyl_config


mDuo13 (approved):
:+1:

PR #392, opened 2018-05-16
Fix transaction results landing page

jhaaaa:
- Adjusted order of transaction codes in table at top of transaction-results.md
- Added blurbs for child pages in dactyl_config - even if not being used now, may be useful in the future
- Added blurb for transaction-metadata.md to display on transaction-formats.md

### DIFF
--- a/content/references/rippled-api/transaction-formats/transaction-results/transaction-results-overview.md
+++ b/content/references/rippled-api/transaction-formats/transaction-results/transaction-results-overview.md
@@ -1,4 +1,4 @@
-# Transaction Results
+# Transaction Results Overview
 
 [[Source]<br>](https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/TER.h "Source")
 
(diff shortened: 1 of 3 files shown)


dactyl-config.yml
```diff
@@ -2169,6 +2181,7 @@ pages:
         supercategory: rippled API
         category: Transaction Formats
         subcategory: Transaction Results
+        blurb: tec codes indicate that the transaction failed, but it was applied to a ledger to apply the transaction cost.
```
mDuo13: Suggestion: Instead of "applied to a ledger to apply the transaction cost" how about "applied to a ledger to deduct the transaction cost"? or maybe "included in a ledger to apply the transaction cost"?

(Yes, I realize the subpar original wording comes straight from the tec-codes file; in other words, it's my own fault. Sorry.)

content/references/rippled-api/transaction-formats/transaction-results/transaction-results-overview.md
```diff
@@ -1,17 +1,17 @@
-# Transaction Results
+# Transaction Results Overview
 
 [[Source]<br>](https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/TER.h "Source")
 
 The `rippled` server summarizes transaction results with result codes, which appear in fields such as `engine_result` and `meta.TransactionResult`. These codes are grouped into several categories of with different prefixes:
 
 | Category              | Prefix                  | Description                |
 |:----------------------|:------------------------|:---------------------------|
+| Claimed cost only     | [tec](tec-codes.html)   | The transaction did not achieve its intended purpose, but the [transaction cost](transaction-cost.html) was destroyed. This result is only final in a validated ledger. |
```
mDuo13: I think the original ordering here was from least likely to most likely to succeed, or something along those lines? It wasn't exactly useful, so the alphabetical ordering you've switched it to is an improvement. :+1:

mDuo13 (changes requested):
This is one case where I'm not 100% convinced that adding a separate "overview" page is helpful. 

With how many pages link to `transaction-results.html`, I think I'd rather that file display the "Overview" contents, complete with the provisional/final codes warning.

I'm willing to be convinced otherwise, especially since we so rarely put important content on landing pages, but given that this is a "sub-page" rather than a category page, I feel like the issue is less pressing.

jhaaaa:
Thank you for the thoughtful review, @mDuo13! I've made the changes we discussed. Ready for review at your convenience.

mDuo13 (approved):

PR #393, opened 2018-05-17
Fix ledger data format and obj type landings

jhaaaa:
This may not be what you envisioned @mDuo13 - but wanted to get an iteration in front of you and I'm happy to refine.

1. Moved content on ledger-data-format.md to Concepts > Payment System Basics > ledgers.md
2. Deleted ledger-data-format.md
3. Did some reformatting of ledger-object-types.md just to help with readability and to make it look more like the autogen landing pages

Note: #1 above broke 19 links. Before I fix them - maybe you can weigh in with feedback on the approach? 😹  Thank you!


content/references/rippled-api/ledger-data-formats/ledger-object-types/ledger-object-types.md
```diff
@@ -1,17 +1,47 @@
...
-* [**LedgerHashes** - Lists of prior ledger versions' hashes for history lookup.](ledgerhashes.html)
-* [**Offer** - An offer to exchange currencies, known in finance as an _order_.](offer.html)
-* [**PayChannel** - A channel for asynchronous XRP payments.](paychannel.html)
-* [**RippleState** - Links two accounts, tracking the balance of one currency between them. The concept of a _trust line_ is really an abstraction of this object type.](ripplestate.html)
-* [**SignerList** - A list of addresses for multi-signing transactions.](signerlist.html)
-
-Each ledger object consists of several fields. In the peer protocol that `rippled` servers use to communicate with each other, ledger objects are represented in their raw binary format. In the [`rippled` API](rippled-api.html), ledger objects are represented as JSON objects.
+Following are the different kinds of objects that can appear in the ledger's state tree. Each ledger object consists of several fields. In the peer protocol that `rippled` servers use to communicate with each other, ledger objects are represented in their raw binary format. In the [`rippled` API](rippled-api.html), ledger objects are represented as JSON objects.
```
mDuo13: Personal pet-peeve: I hate the sentence structure, "Following are..." Furthermore, I don't think "Each ledger object consists of several fields" is necessary here; almost by definition, an object consists of several fields.

Here's a pitch for a dramatically revised intro:

> Each ledger's state tree consists of a set of _ledger objects_, which collectively represent all settings, balances, and relationships in the shared ledger.

Also, maybe break the line about how the objects are represented as binary or JSON to a separate paragraph?

content/references/rippled-api/ledger-data-formats/ledger-object-types/ledger-object-types.md
```diff
@@ -1,17 +1,47 @@
...
+
+* [**AccountRoot**](accountroot.html)
+
+    The settings, XRP balance, and other metadata for one account.
+
+* [**Amendments**](amendments.html)
+
+    Singleton object with status of enabled and pending amendments.
```
mDuo13: You can summon the blurb from the config file¹ instead of moving it to the Markdown:
```
{{ (pages|getattr("html","eq","amendments.html")|first).blurb }}
```

¹if you uncomment it there

content/concepts/payment-system-basics/ledgers.md
```diff
@@ -31,6 +31,6 @@ All objects in a ledger' state tree have a unique ID. This field is returned as
 ![Diagram: rippled uses SHA-512Half to generate IDs for ledger objects. The space key prevents IDs for different object types from colliding.](img/ledger-indexes.png)
```
mDuo13: Pitch: the "Object IDs" section should be its own page in the Ledger Reference section, because it defines the generic form of all the IDs that are defined in more detail on the "Ledger Object Types" pages. 

In other words, it's highly relevant within those pages and not very useful in the "Concepts" area.

mDuo13 (changes requested):
I like moving the content to a "Ledgers" concept page, but I think we should revise this approach just slightly:

- Leave the blurbs in the config. Maybe the "Ledger Object Types" landing should be generated automatically (even though it means we have to give up on some formatting). But even if we decide we want to keep the formatting, we can pull in the blurbs using the syntax I suggested below.
- We should add "See Also" sections to make it easier to navigate between these areas
- As described below, I think we should move "Object IDs" back into the reference section.

content/references/rippled-api/ledger-data-formats/ledger-object-types/ledger-object-types.md
jhaaaa: Done - I went with the approach of autogenerating the ledger-object-types page, so I was unable to insert the suggested line break. 

If upon review you think it really needs the line break, I can go with manually building and using the summoning snippet you provided below!

content/concepts/payment-system-basics/ledgers.md
jhaaaa: Done. Does the location of the Ledger Object IDs page in the hierarchy work for you?

jhaaaa:
As suggested, added "See Also" sections to Ledgers, Ledger Header, and Ledger Object ID pages.

mDuo13 (approved):

mDuo13:
The revised version looks good.

Looking forward, I feel like this one has kind of exposed the auto-generated landings' limitations, because I'd especially like some of the landing pages to link to the ledger basics concept in their opening blurbs, but we can't easily do that unless we get the blurbs to be processed as Markdown first. I've created a ticket (DOC-1577) to track that.

PR #395, opened 2018-05-17
Fixes to Public Methods landing page

jhaaaa:
- Added method category headings and overviews to public-rippled-methods.html.
- Moved method descriptions out of links for readability - I think it makes it easier to read?
- Shortened overview text for method categories. In particular, please take a look at the overview for Ledger Methods.

At first, I was going to update the page to include the method subcategories and their overviews only, but saw that we'd then lose the single-page view of all public methods and what they do. Decided to provide both. Overkill? This is basically the left nav, but with more context -- which I have to admit, I like.

Open to feedback. Whatever structure we use for this page, I'll replicate for the Admin Methods category.

How do these landing pages impact search results? If they're having a negative impact, perhaps we can remove them from search? They are provided primarily to help folks who are manually navigating the site.

### DIFF
--- a/dactyl-config.yml
+++ b/dactyl-config.yml
@@ -1025,7 +1025,7 @@ pages:
         supercategory: rippled API
         category: Public rippled Methods
         blurb: Get data from the XRP Ledger and submit transactions using these public API methods.
-        template: template-landing-children.html
+        # template: template-landing-children.html
         targets:
             - local
 
@@ -1036,8 +1036,8 @@ pages:
         supercategory: rippled API
         category: Public rippled Methods
         subcategory: Account Methods
-        # blurb: An Account in the XRP Ledger represents a holder of XRP and a sender of transactions. Accounts can send and receive XRP and issued assets, participate in the decentralized exchange, and change their own settings. Creating an account involves generating keys and then receiving XRP from another account.
         template: template-landing-children.html
+        blurb: An account in the XRP Ledger represents a holder of XRP and a sender of transactions. Use these methods to work with account info.
         targets:
             - local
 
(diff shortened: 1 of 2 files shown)


dactyl-config.yml
```diff
@@ -1025,7 +1025,7 @@ pages:
         supercategory: rippled API
         category: Public rippled Methods
         blurb: Get data from the XRP Ledger and submit transactions using these public API methods.
-        template: template-landing-children.html
+        # template: template-landing-children.html
```
mDuo13: Let's remove rather than comment this out

mDuo13:
Addressing all of your question/comment bullets and paragraphs in order:

1. That's cool, although the old format sort of served the purpose of being an alphabetical list of all methods—an alternate view, in case you sort of remember or can guess the method name but don't know how it's categorized. Maybe we can better serve that by having a page with an alphabetical list of _all_ methods, public and admin. That's outside the scope of this task, though.

2. No strong opinion on this. The old ones have a bigger click target, but the new ones are, I guess, easier to read?

3. :+1: to the shorter description of the ledger methods category. The old one was definitely filling the gap for the ledger concept that didn't exist at the time.

4. I like having a one-page list of all the public methods, too.

5. This is good. In the future, we can probably automate more of this by pulling the method description blurbs from the pages themselves. That might be a unique template. Let's come back to that.

6. The landing pages don't seem to have a negative impact on search results. If anything, they may have a positive impact since they're adding to the number of links to the child pages. If you search for an exact method name, you'll pretty much always get that method page as the first result because both the filename and the page's only h1 match exactly, and smart search engines emphasize those things. If you search for related or adjacent terms, it's more of a tossup what shows up first but the method page always seems to be listed in the results. I was going to suggest adding meta keywords to the site to increase the chances of a hit, but apparently [Google ignores those anyway](https://webmasters.googleblog.com/2009/09/google-does-not-use-keywords-meta-tag.html).

mDuo13 (approved):

jhaaaa:
Thank you for such a thoughtful and thorough review. I appreciate your time and expertise!

For #1, I created https://ripplelabs.atlassian.net/browse/DOC-1579. I love this idea.

PR #396, opened 2018-05-19
Typo in escrow.md

mikemilano:
Looks like the sentence was refactored from something else, and `able to` was left in.


mikemilano:
On second thought, maybe it was just awkward to me the first time reading it. It might be fine as is.

mDuo13:
I think both versions are fine, but your revised version is a little more succinct, which is better. Thanks!

PR #398, opened 2018-05-21
Fix rippled API References landing page

jhaaaa:
- Deleted rippled-api.md
- Updated dactyl_config to remove rippled-api.md


mDuo13 (approved):
:+1: 
Love to see removal of source text while keeping the same or better user experience in the output.

PR #399, opened 2018-05-22
Improve table styling slightly

mDuo13:
- Turns off mid-word line-breaking of code text in table cells (which currently happens in Chrome and not Firefox anyway)
- Slightly reduces font size of table cells to reduce chance of tables spilling into the right column (and therefore under the right nav). Doesn't fix this problem entirely (still visible at, say, 996px wide on accountroot.html)


ryangyoung (approved):
Looks good on FF, Chrome, Safari, and Brave.

PR #402, opened 2018-05-22
Fix admin api methods landing page

jhaaaa:
- Modeled after Public API Methods landing page: https://github.com/ripple/ripple-dev-portal/commit/ed5c439e4fed433db6fcfbcb28ae7558391b5472
- In dactyl_config, corrected html value for Server Control Methods page


mDuo13 (approved):
:+1:

PR #406, opened 2018-05-24
Missing Optional flag for hotwallet param

mikemilano:
The `hotwallet` parameter of `gateway-balances()` didn't seem like the type of parameter that needs to be required. I tested the API w/o it and it returned successfully.


mDuo13 (approved):
Good catch. Thanks.

PR #408, opened 2018-05-30
Refactored request example to use ledger_index

mikemilano:
Websocket and JSON-RPC requests had the `ledger` parameter set when it doesn't exist in the parameter list. I vaguely remember that might be a legacy parameter name.

I've updated the parameters to use `ledger_index` instead based on the [Specifying Ledgers](https://developers.ripple.com/basic-data-types.html#specifying-ledgers) documentation.


mikemilano:
On second thought, it may be better to just remove the parameter from the request since it is optional.

I'm only certain that the parameter shouldn't be `ledger`. I'm working on assumptions beyond that.

PR #409, opened 2018-05-30
Removed ledger parameter from request in account_lines

mikemilano:
The `ledger` parameter is now `ledger_index`, which is optional and defaults to current. This PR removes the `ledger` param from the example requests.

I had opened a PR to simply change `ledger` to `ledger_index`, however the conventions of the API docs seem to only use required parameters when they are not set to the default value, so I closed that PR and created this one instead.


mDuo13 (approved):
Thanks for helping with the tricky work of finding and fixing old usage and outdated stuff!

PR #410, opened 2018-05-30
Removed ledger param from account_objects

mikemilano:
Similar to PR #409 except the deprecated `ledger` param was used in the websocket request example, while `ledger_index` was used in the JSON example.

This PR removes those parameters on both examples for consistency since the default current ledger_index is being requested. I also removed the word `Optional` from the `Type` column since the convention seems to be to indicate that in the description field.

Note: The Commandline request example has `[ledger_index]`, however the syntax suggests it's an optional argument as opposed to a flagged option (i.e. `--ledger_index=current`), so it seems proper to leave that as is.


mDuo13 (approved):
:+1:

PR #411, opened 2018-05-30
Removed deprecated legacy param from account_offers

mikemilano:
Sorry for all the PRs, I'm editing directly in github's UI. I checked ahead and this is the last one of this type in the Account API section. I'm writing a wrapper for the API so the changes are trickling in as I get through each method document.

This PR includes the same changes reported in #410, except its for `account_offers` this time.
- Removed deprecated `ledger` parameter is used in the websocket request
- Removed `ledger_index` parameter from the json request since it includes the default value (for consistency)
- Removed the word `Optional` from the Type column since the convention seems to be to indicate that in the Description area.


mDuo13 (approved):
:+1:

PR #412, opened 2018-05-31
Fixed invalid JSON response in noripple_check

mikemilano:
A trailing comma was left in the last item of the `result.problems` array in the JSON-RPC example response of noripple_check.

I'm actually creating json files with the examples to mock responses in unit tests for the library I'm writing, so my tests found it. I wouldn't have noticed otherwise. I'm grateful we have example responses to review and copy!


mDuo13:
Thanks as usual!

PR #422, opened 2018-06-18
Removed duplicate message from checkcash.md

mikemilano:
It looks like the field intro was optimized to use a snippet, but the original intro text was left in, duplicating the field intro message.


mDuo13:
Thanks for catching these. :+1:

PR #424, opened 2018-06-19
Update known amendments with fix1571 live etc.

mDuo13:
- Updates the status of fix1571 (which just went live) as well as fix1543 and fix1623 which are expected in the next couple days
- Gets rid of the distinction between the "status" (from the top table) and the "detailed status" (from the description area). This makes it easier for AmendmentBot to update the status as it changes.
- This update has been generated by a script and manually touched up afterwards.

### DIFF
--- a/content/news/known-amendments.md
+++ b/content/news/known-amendments.md
@@ -5,15 +5,15 @@ The following is a comprehensive list of all known amendments and their status o
 
 | Name                      | Introduced | Status                              |
 |:--------------------------|:-----------|:------------------------------------|
-| [fix1543][]               | v1.0.0     | [Planned: TBD]( "BADGE_LIGHTGREY") |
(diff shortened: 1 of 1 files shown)


content/news/known-amendments.md
```diff
@@ -391,7 +391,7 @@ Sorts the entries in [DirectoryNode ledger objects](directorynode.html) and fixe
 
 | Amendment ID                                                     | Status    |
 |:-----------------------------------------------------------------|:----------|
-| DA1BD556B42D85EA9C84066D028D355B52416734D3283F85E216EA5DA6DB7E13 | Enabled on TestNet; not intended for production. |
```
mDuo13, line 394: The TestNet has been reset so it is no longer true that SusPay is enabled on the TestNet.

jhaaaa (approved):
LGTM!

PR #425, opened 2018-06-20
Freeze: Even more explicit wording that XRP can't be frozen

mDuo13:
No one can freeze XRP.


ryangyoung:
👍 Looks as clear as it can get.

PR #428, opened 2018-06-26
Add rippled commandline reference

mDuo13:
Adds a (hopefully) complete reference of the commandline options for `rippled`, informed by the source, info from the experts, and the [old wiki page](https://wiki.ripple.com/Rippled).

Many of the commandline options are rather obscure and date back to early days when the XRP Ledger was much more experimental. After some discussion with Nik, I have labeled several such options as **DEPRECATED** in this reference. The plan is to remove or reorganize the least relevant options in a coming version of `rippled`, but we don't have a timeline for that yet so I plan on publishing this as-is first. In that sense, this PR also serves as a proposal for which commandline options should be considered deprecated.

Staged here: https://developers-staging.ripple.com/commandline-usage.html (Ripple VPN may be required to view.)


MarkusTeufelberger:
Ripple VPN _is_ required to view. ;-)

content/concepts/the-rippled-server/rippled-server-modes.md
```diff
@@ -58,6 +58,10 @@ When you run `rippled` in stand-alone mode, you have to tell it what ledger vers
 * Create a [new genesis ledger](start-a-new-genesis-ledger-in-stand-alone-mode.html) from scratch.
 * [Load an existing ledger version](load-a-saved-ledger-in-stand-alone-mode.html) from disk.
 
-**Caution:** In stand-alone mode, you must [manually advance the ledger](advance-the-ledger-in-stand-alone-mode.html).
+**Caution:** In stand-alone mode, you must [manually advance the
```
MarkusTeufelberger: missing words

MarkusTeufelberger:
Starting rippled only with the `--conf` flag causes it to never sync with the network, either `--net` or `--load` are necessary. Both are marked deprecated, what's the recommended way to start a server that has no node store or a server that already has a node store, but I don't know for sure if it is complete or which is the latest ledger it knows (yes, I could check the `ledger.db` file)?

content/references/rippled-api/commandline-usage.md
```diff
@@ -0,0 +1,149 @@
...
+
+Unit testing runs tests built into the `rippled` source code to confirm that the executable performs as expected. This includes things like testing built-in data types and transaction processing routines.
+
+If unit testing reports a failure, that generally indicates one of the following:
+
+- A problem occurred when compiling `rippled` and it is not functioning as intended
+- The source code for `rippled` contains a bug
+- A unit test has a bug or has not been updated to account for new behavior
+
+While running unit tests, you can also provide the following options:
```
MarkusTeufelberger: misses the `--unittest-arg arg`, `--unittest-log`, `--unittest-jobs arg` and the upcoming `--unittest-ipv4` (or `--unittest-ipv6`) flags.

mDuo13:
> Starting rippled only with the --conf flag causes it to never sync with the network, either --net or --load are necessary.

That's not intended behavior. I asked around the `rippled` team and nobody seems to be able to reproduce. If you can clarify the situation that's causing that to happen to you, you should file a bug against rippled.

content/references/rippled-api/commandline-usage.md
mDuo13: Added a task to the backlog to add the IPv4/6 flags when 1.1.0 is closer to ready.  The other ones are added in 3334e3c

content/references/rippled-api/commandline-usage.md
```diff
@@ -0,0 +1,166 @@
...
+- **Stand-Alone Mode** - Use the `-a` or `--standalone` option. Like daemon mode, except it does not connect to other servers. You can use this mode to test transaction processing or other features.
+- **Client Mode** - Specify an API method name to connect to another `rippled` server as a JSON-RPC client, then exit. You can use this to look up server status and ledger data if the executable is already running in another process.
+- **Other Usage** - Each of the following commands causes the `rippled` executable to print some information, then exit:
+    - **Help** - Use `-h` or `--help` to print a usage statement.
+    - **Unit Tests** - Use `-u` or `--unittest` to run unit tests and print a summary of results. This can be helpful to confirm that you have compiled `rippled` successfully.
+    - **Version statement** - Use `--version` to have `rippled` print its version number, then exit.
+
+## Generic Options
+
+These options are apply to most modes:
```
scottschurr: I think the "are" doesn't belong here?

content/references/rippled-api/commandline-usage.md
```diff
@@ -0,0 +1,166 @@
...
+
+| Option          | Description                                                |
+|:----------------|:-----------------------------------------------------------|
+| `--conf {FILE}` | Use `{FILE}` as the configuration file instead of looking for config files in the default locations. If not specified, `rippled` first checks the local working directory for a `rippled.cfg` file. On Linux, if that file is not found, `rippled` next checks for `$XDG_CONFIG_HOME/ripple/ripple.cfg`. (Typically, `$XDG_CONFIG_HOME` maps to `$HOME/.config`.) |
```
scottschurr: A note on how the `Option` column is being rendered in the preview.  The preview is rendering this as
```
--
conf
{FILE}
```
Is there any way to get all of that showing on a single line?  This is a general problem throughout many of the tables in this file.  It may require you re-thinking the ways the tables are laid out.

content/references/rippled-api/commandline-usage.md
```diff
@@ -0,0 +1,166 @@
...
+| `--nodetoshard`     | Before fully starting, copy any complete [history shards](history-sharding.html) from the ledger store into the shard store, up to the shard store's configured maximum disk space. Uses large amounts of CPU and I/O. Caution: this command copies data (instead of moving it), so you must have enough disk space to store the data in both the shard store and the ledger store. <!--{# Task for writing a tutorial to use this: DOC-1639 #}--> |
+| `--quorum {QUORUM}` | This option is intended for bootstrapping [test networks](parallel-networks.html). Override the minimum quorum for validation by requiring a agreement of `{QUORUM}` trusted validators. By default, the quorum for validation is automatically set to a safe number of trusted validators based on how many there are. If some validators are not online, this option can allow progress with a lower than normal quorum. **Warning:** If you set the quorum manually, it may be too low to prevent your server from diverging from the rest of the network. Only use this option if you have a deep understanding of consensus and have a need to use a non-standard configuration. |
```
scottschurr: Perhaps "**an** agreement"?

content/references/rippled-api/commandline-usage.md
```diff
@@ -0,0 +1,166 @@
...
+|:----------------------|:-----------------------------------------------------|
+| `--ledger {LEDGER}`   | Load the ledger version identified by `{LEDGER}` (either a ledger hash or a ledger index) as the initial ledger. The specified ledger version must be in the server's ledger store. |
+| `--ledgerfile {FILE}` | Load the ledger version from the specified `{FILE}`, which must contain a complete ledger in JSON format. For an example of such a file, see the provided [`ledger-file.json`]({{target.github_forkurl}}/blob/{{target.github_branch}}/content/_code-samples/rippled-cli/ledger-file.json). |
+| `--load`              | **DEPRECATED** Intended for debugging. Only load the initial ledger from the ledger store on disk. |
+| `--replay`            | Intended for debugging. Use with `--ledger` to replay a ledger close. Your server must have the ledger in question and its direct ancestor already in the ledger store. Using the previous ledger as a base, the server processes all the transactions in the specified ledger, resulting in a re-creation of the specified ledger. With a debugger, you can add breakpoints to analyze specific transaction processing logic. |
```
scottschurr, line 69: Looks great.  :+1:

content/references/rippled-api/commandline-usage.md
```diff
@@ -0,0 +1,166 @@
...
+
+```bash
+rippled --unittest=print
+```
+
+The `print` unit test is a special case that prints a list of available tests with their packages.
+
+#### Manual Unit Tests
+
+Certain unit tests are classified as "manual" because they take too long to run. These tests are marked with `|M|` in the output of the `print` unit test. Manual tests do not run by default when you run all unit tests or a package of unit tests. You can run manual tests individually by specifying the name of the test. For example:
```
scottschurr: Perhaps "take too long to run" -> "run for a long time"?

content/references/rippled-api/commandline-usage.md
```diff
@@ -0,0 +1,166 @@
...
+
+
+## Unit Tests
+
+```bash
+rippled --unittest [OPTIONS]
+rippled -u [OPTIONS]
+```
+
+Unit testing runs tests built into the `rippled` source code to confirm that the executable performs as expected. This includes things like testing built-in data types and transaction processing routines.
```
scottschurr: Is it worth saying explicitly that this option runs the specified unit tests, reports a summary, and then quits?  Your call.

scottschurr (approved):
Looks great!  I left a few nits for you to think about, but I'm good with it.

MarkusTeufelberger:
Ok, I tried to re-create the problem with using neither `--net` nor `--load` and it just takes quite long until the server finally syncs, so I'm sorry that I gave a false alarm. I still hope these stay available, because otherwise it clutters the `ledger.db` file with some random early ledgers from these startups.

content/references/rippled-api/commandline-usage.md
mDuo13: Added a line of CSS that should stop this from happening in Chrome. (For what it's worth, this is a browser difference between Chrome and Firefox.)

mDuo13:
The plan is to investigate the code and remove any options that do nothing so there are no flags for "do the default behavior" or flags that are functionally equivalent to each other. It's possible that some options may be consolidated or replaced with new options.

MarkusTeufelberger:
The default behavior will start a new ledger while syncing with peers and after a (long) while switch to the network ledger. In the meantime the new ledgers are being generated and also end up in local databases.

Both `--load` (first loads a full ledger from disk instead of starting at 0) and `--net` (first fetches a ledger from the network instead of starting at 0) work differently and are more useful in my opinion than the default behavior. They both might need better/more descriptive names though.

content/references/rippled-api/commandline-usage.md
scottschurr: FWIW, the browser I'm using is Safari.

PR #430, opened 2018-06-29
Updated freeze feature description and XRP related images.

paleorbglow:
Updated the freeze feature description and a few XRP related images.


content/concepts/issued-currencies/freezes.md
```diff
@@ -1,6 +1,10 @@
...
+XRP is not an issued currency. XRP is the only native asset on the XRP Ledger and is required in order to conduct transactions on the XRP Ledger. XRP is counterparty free, meaning that when someone holds XRP, they are not holding a liability, they are holding the actual currency, XRP. Due to this fact, XRP CANNOT be frozen by any entity or individual.
+
+All non-XRP currencies can be represented in the XRP Ledger as issued currencies. These issued currencies (sometimes called "issuances" or "IOUs") are tracked in accounting relationships, called "trust lines," between addresses. Issued currencies are typically considered as liabilities from one perspective and assets from the other, so the balance of a trust line is negative or positive depending on which side you view it from. Any address may freely issue (non-XRP) currencies, limited only by how much other addresses are willing to hold.
+
+In certain cases, to meet regulatory requirements, or while investigating suspicious activity, an exchange or gateway may want to quickly freeze non-XRP issued currency balances to temporarily prevent unauthorized withdrawals.
```
mDuo13: >  to temporarily prevent unauthorized withdrawals

I recommend removing this phrase, because:

- Freezing a trust line / currency doesn't prevent sending directly to the issuer, which is the typical way you initiate a withdrawal from a gateway. This phrase may mistakenly give the impression that freezing is enough to prevent withdrawals.
- It's also possible you may want to freeze for other reasons.

It makes me sad that this page has to spend the first two paragraphs reiterating common information that is covered by the category overview, but given the weight of the FUD about freezing I guess it's necessary.

mDuo13 (changes requested):
Seeing the XRP "X" all over the place feels weird. It especially highlights some of the (too) many we've used the Ripple triskellion in the past—to refer to:

- the XRP currency
- `rippled` servers
- non-XRP currency balances issued in the XRP Ledger

That last one, which appears in the gateway balance sheet diagrams, may be weird enough to warrant something other than a straight substitution of the triskellion for the XRP symbol...

PR #431, opened 2018-07-03
Emphasize no one can freeze XRP

ryangyoung:


jhaaaa (approved):
Looks great - thank you!

PR #437, opened 2018-07-24
Known Amendments: add fix1515 & correct other statuses

mDuo13:
- Adds a description for the new fix1515 amendment
- Changes the status of the commented-out amendments from "Planned" to "In Development". Also changes the not-commented-out but also not-complete CryptoConditionsSuite amendment to "In Development".
- Minor whitespace churn. Sorry.

### DIFF
--- a/content/news/known-amendments.md
+++ b/content/news/known-amendments.md
@@ -222,9 +222,13 @@ With this amendment, the new `STAmountCalcSwitchovers` code applies, which may c
 |:-----------------------------------------------------------------|:----------|
 | 5D08145F0A4983F23AFFFF514E83FAD355C5ABFBB6CAB76FB5BC8519FF5F33BE | Planned   |
 
-Fixes a bug that can occur when a cross-currency payment attempts to consume a large number of offers at the same exchange rate. (Also applies to offers if [FlowCross][] is enabled.) Without the amendment, payment processing gives up on using particular order books if processing the order book would require consuming over 2000 offers.
+Changes how Payment transactions consume offers to remove a minor difference in how payment processing and offer processing consume liquidity. (Also affects how OfferCreate transactions are processed if [FlowCross][] is enabled.)
 
-With this amendment, if a transaction processes over 1000 offers at the same exchange rate, the payment correctly consumes the liquidity from those offers, then moves on to other paths or exchange rates to finish processing the transaction.
+Without the amendment, payment processing gives up on using particular order books if the transaction would consume over 2000 offers at the same exchange rate. In this case, the payment does not use the liquidity from those offers, and does not consider that order book's remaining liquidity when attempting to complete the payment.
+
+With this amendment, if any transaction processes over 1000 offers at the same exchange rate, the transaction consumes the liquidity from the first 1000 offers, then does not consider that order book's remaining liquidity when attempting to complete the payment.
+
+In both cases, transaction processing can still complete by using liquidity from other paths or exchange rates.
 
 
 ## fix1523
--- a/content/news/known-amendments.md
+++ b/content/news/known-amendments.md
@@ -5,13 +5,14 @@ The following is a comprehensive list of all known amendments and their status o
 
 | Name                      | Introduced | Status                              |
 |:--------------------------|:-----------|:------------------------------------|
-| [DepositPreauth][]        | v1.1.0     | [Planned: TBD]( "BADGE_LIGHTGREY") |
-| [Checks][]                | v0.90.0    | [Planned: TBD]( "BADGE_LIGHTGREY") |
-| [FlowCross][]             | v0.70.0    | [Planned: TBD]( "BADGE_LIGHTGREY") |
-| [CryptoConditionsSuite][] | TBD        | [Planned: TBD]( "BADGE_LIGHTGREY") |
-| [OwnerPaysFee][]          | TBD        | [Planned: TBD]( "BADGE_LIGHTGREY") |
-| [SHAMapV2][]              | TBD        | [Planned: TBD]( "BADGE_LIGHTGREY") |
-| [Tickets][]               | TBD        | [Planned: TBD]( "BADGE_LIGHTGREY") |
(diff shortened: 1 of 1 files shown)


content/news/known-amendments.md
```diff
@@ -5,13 +5,14 @@ The following is a comprehensive list of all known amendments and their status o
 
 | Name                      | Introduced | Status                              |
```
ryangyoung, line 6: What would it take to make these tables [sortable](https://www.kryogenix.org/code/browser/sorttable/)? I can see that they're sorted by introduced version. And that the in-page TOC lists them in alphabetical order.  It's a little disorienting. Sortable tables seem like they'd go some way to helping with that. There may be other ways to handle it.

content/news/known-amendments.md
```diff
@@ -5,13 +5,14 @@ The following is a comprehensive list of all known amendments and their status o
```
ryangyoung, line 5: Speaking of sorting--if we leave the table static for now, we might want to explain here how we're sorting it, just to be friendly.

ryangyoung (approved):
Looks good! Had some comments about how the table is sorted but nothing that should hold up merging and publishing this.

content/news/known-amendments.md
```diff
@@ -214,6 +215,18 @@ Fixes a bug that resulted in transaction processing not using new `STAmountCalcS
...
+[fix1515]: #fix1515
+
+| Amendment ID                                                     | Status    |
+|:-----------------------------------------------------------------|:----------|
+| 5D08145F0A4983F23AFFFF514E83FAD355C5ABFBB6CAB76FB5BC8519FF5F33BE | Planned   |
+
+Fixes a bug that can occur when a cross-currency payment attempts to consume a large number of offers at the same exchange rate. (Also applies to offers if [FlowCross][] is enabled.) Without the amendment, payment processing gives up on using particular order books if processing the order book would require consuming over 2000 offers.
+
+With this amendment, if a transaction processes over 1000 offers at the same exchange rate, the payment correctly consumes the liquidity from those offers, then moves on to other paths or exchange rates to finish processing the transaction.
+
```
seelabs: I wouldn't characterize the current behavior as a bug. It's fine behavior as long as offer crossing would clear the offers. When FlowCross activates, we want to make sure the offers would still clear. Instead of having special behavior in payments that checked if it was in "offer crossing mode" or not, we decided to change the behavior of payments so it matched offer crossing more closely.

So I'd rephrase this to emphasize that we're aligning the behavior of offer crossing and payments. Maybe something like:
```
Removes a difference in how payment processing and offer crossing consume
liquidity. Without this fix, payment processing gives up on using particular
order books if processing the order book would require consuming over 2000
offers at the same exchange rate. It would not use the liquidity from those
offer and the path containing those offers would not be considered for new
liquidity.

With this amendment, if a transaction processes would require consuming over
1000 offers at the same exchange rate, the payment will consume the liquidity
from the first 1000 of those offers. After the liquidity is consumed the path
containing those offers would not be considered for new liquidity.
```

content/news/known-amendments.md
mDuo13: Making the tables sortable would be a matter of (1) adding some javascript to the templates, and (2) maybe adding something to make sure the javascript applied to the correct tables. Possibly making every table in the content body is an option, in which case (2) wouldn't be that hard.

The table itself is sorted in what attempts to be "chronological" order for things that don't always have an obvious chronology. In practice, this means the table is sorted primarily by status, with secondary sorts depending on the status and a tertiary sort by name. The long descriptions outside the table are in alphabetical order.  The sorting criteria is complicated enough that I think describing it would be more distracting than useful.

seelabs (approved):
:+1:

content/news/known-amendments.md
ryangyoung: > The sorting criteria is complicated enough that I think describing it would be more distracting than useful.

I have to vehemently disagree with that reasoning. If it's so complicated that we can't even explain, that's a problem for users. It had been a while since I'd seen this page when reviewed it and it took me a lot more time than I would've liked to figure out where this amendment was. The double presentation of the list (in the page and side toc) and the idiosyncratic sorting did it not make it a very intuitive experience (somewhat the opposite, in fact). I looked to the description for some logic as to the sorting and didn't get anything, which increased my frustration. 

I know this is, overall, a fairly minor point but I don't think we should get in the habit of deferring or omitting explanations because something is too complicated. And, actually, "sorted in this order: status, chronologically (by release version), and alphabetically (by amendment name)" doesn't seem all that complicated.

mDuo13: That's because the sort order is more complicated than that.

The [actual sort order](https://github.com/ripple/amendmentbot/blob/master/src/amendment.py#L56-L76) is:

1. Sort by status in the following order: (No status), "Planned", "Open for Voting", "Expected", "Enabled", "Vetoed"
    - (Aside: it's weird that "In Development" isn't in the AmendmentBot source there. Not sure how it's working without sorting those, then.)
2. Secondary sort depends on the status:
    - For "Planned", "Open for Voting", and "Vetoed" statuses, sort descending by introduced version. (slightly undesirably, this means "TBD" comes after "vX.Y.Z". That's probably something to change, actually)
    - For "Expected" status, sort descending by date expected (that is, later dates first)
    - For "Enabled" status, sort descending by date enabled (most recently enabled first)
3. Tertiary sort is ascending by name

I don't mind adding a line like, "sorted in chronological order (most recent first)" to the description, but digging that deeply into the details of the sorting implementation, especially when it's likely to change subtly, is completely besides the point of the document.

The document has two use cases:

1. You want to know what's currently going on with amendments. Did that amendment you need get enabled? What's in voting right now? What's coming soon?
    - The table of known amendments serves this need by listing the most recent amendments at the top, most active. If you scroll down through the table, you get essentially a history of prior changes in the order they occurred, with the failed potential changes at the bottom.
2. You want to know what a specific amendment does.
    - The "In this page" table, listed by chronological order, serves this need. Amendments don't move around within the page content after they're added, and regardless of what happened to the amendment you can find it by name.

PR #441, opened 2018-08-02
Ignore initial result of tx submission (RLJS-607)

intelliot:
The initial result of transaction submission should always be ignored. Transactions that succeed here can ultimately fail, and transactions that fail here can ultimately succeed.

Also fixes a minor typo.


mDuo13:
Good catch. It can be informative to know the immediate result, but unless you properly understand that it's a tentative result, it's better not to know!

Oh, but maybe `tem` codes should still actually be printed, since they're generally final?

intelliot:
I'm told that an amendment can theoretically make even `tem` transactions later succeed.

I would be fine with printing the response code, but it might not be important enough to include in this example.

PR #444, opened 2018-08-07
Add dedicated Data API target

mDuo13:
Creates a "data-api-only" target for easier exporting of the Data API
reference to be used in the upstream repository. This target contains
link substitutions so that the links work on an external site (like
GitHub's README view).

I constructed this while testing changes to Dactyl v0.7.3. It would've been pretty handy for the last couple of Data API changes I had to make.


mDuo13:
Rebased onto master to get the fix from #447.

PR #451, opened 2018-08-24
chore: update capacity-planning.md

gnanderson:
`low` seems to be an invalid value causing an unhandled exception:

    Terminating thread rippled: main: unhandled N5beast14BadLexicalCastE 'std::bad_cast'

It appears correct value is `small`...


mDuo13 (approved):
Good catch, thanks.

PR #454, opened 2018-08-31
Update Data API docs for v2.3.5

mDuo13:
As far as I know, the Data API v2.3.5 only removes some old functionality without adding any.

These changes sync the Dev Portal docs up with the Data API changes https://github.com/ripple/rippled-historical-database/commit/5d4a8e58e80929734c7eab475a4e8c006648cedb , https://github.com/ripple/rippled-historical-database/commit/a8fca2e7e7ef1da60be7b9308a1434c898eccddd , and https://github.com/ripple/rippled-historical-database/commit/36efceef0afde17235a435763f856ce7f18bf24b .


jhaaaa (approved):
Looks great! 

The link checker will catch this 🤓, but I think the `[Get Issued Value - `GET /v2/network/issued_value`](#get-issued-value)` anchor link needs to be removed.

PR #445, opened 2018-08-15
Introductory consensus article

mDuo13:
- Adds a new, simpler article introducing consensus at a high level for the intro category
- Moves the existing consensus article to the "Consensus Network" category
- Adds a FAQ-like article describing how Consensus handles various challenges and attack types.

Reviewers: please review the introduction to consensus article closely with an eye for things that can be removed to make the article more approachable at an introductory level, and omissions that may lead to misunderstandings later.

PR #458, opened 2018-09-22
Consensus diagrams

mDuo13:
- Revises the diagrams uses in the "Consensus" document to match the style for the recent "Intro to Consensus" article.
- Revises the Consensus document itself to better use the diagrams and to fix some outdated style from the previous article.
- Adds a "Transaction Queue" article with a diagram of how the queue interacts with consensus. This article also takes over some of the queue-related content that previously existed in the Transaction Cost page.

<s>This PR currently contains an unused HTML version of the queue+consensus diagram and some images only used in various versions of that diagram. (The initial thought was that making the diagram HTML would be more accessible and flexible, but in practice, it didn't fit well on screen at almost any screen width. I concluded that a simple image with the text reproduced in the document was more flexible and would display better on a variety of viewing devices.) **Reviewers: I'd appreciate your feedback on whether the unused diagram is useful scrap material or should be removed.**</s>

Edit: I have removed the unused HTML diagram and its related images.


mDuo13:
Preview of these docs staged here so you can see the diagrams in context:
_(Note: Ripple VPN is required to view staging.)_
https://developers-staging.ripple.com/consensus.html
https://developers-staging.ripple.com/transaction-queue.html

EDIT: Updated with changes through 1177df2.

content/concepts/consensus-network/consensus.md
```diff
@@ -6,35 +6,36 @@ This article provides a high level overview of the XRP Ledger, the information i
...
+A ledger version has two identifiers. One identifier is its _ledger index_, also called a _sequence number_. Ledger versions are numbered incrementally. For example, if the current ledger version has ledger index of 100, the previous has ledger index 99 and the next has ledger index 101. The other identifier is a _ledger hash_, which is a digital fingerprint of the ledger's contents.
 
-User level changes to the ledger are the results of transactions. Examples of [transactions](transaction-formats.html) include payments, changes to account settings or trust lines, and offers to trade. Each transaction authorizes changes to the ledger and is cryptographically signed by an account owner. Transactions are the only way to authorize changes to an account.
+As servers propose transactions to apply to the ledger, they may create several candidate ledger versions with slightly different contents. These candidate ledger versions have the same ledger index but different ledger hashes. Of the many candidates, only one can become validated. All the other candidate ledger versions are discarded. Thus, there is exactly exactly one validated ledger hash for each ledger index in history.
```
ximinez: Duplicated word: `exactly exactly`

content/concepts/consensus-network/consensus.md
```diff
@@ -6,35 +6,36 @@ This article provides a high level overview of the XRP Ledger, the information i
...
 
-A ledger instance also contains a set of transactions and metadata about those transactions. The transactions are those that have been applied to the prior ledger to create the new instance. The metadata records precisely the effects of the transaction on the ledger.
+User level changes to the ledger are the results of transactions. Examples of [transactions](transaction-formats.html) include payments, changes to account settings or trust lines, and offers to trade. Each transaction authorizes changes to the ledger and is cryptographically signed by an account owner. Transactions are the only way to authorize changes to an account, or to change anything else in the ledger.
```
ximinez: This reads a little awkwardly to me: `Each transaction authorizes changes to the ledger`. I'm having some noun/verb confusion with "changes". I think it may be clearer as `Each transaction authorizes one or more changes to the ledger`

content/concepts/consensus-network/consensus.md
```diff
@@ -54,61 +55,82 @@ Important: Some [`rippled` APIs](rippled-api.html) provide provisional results,
...
+
+#### Compare Results
+
+Validators each relay their results in the form of a signed message containing the hash of the ledger version they calculated. These messages, called _validations_, allow each server to compare the ledger it computed with those of its peers.
+
+[![Figure 8: Ledger is Validated When Supermajority of Peers Calculate the Same Result Result](img/consensus-declare-validation.png)](img/consensus-declare-validation.png)
+
+_Figure 8: Ledger is Validated When Supermajority of Peers Calculate the Same Result — Servers compare their calculated ledger with the hashes received from chosen validators. If not in agreement, the server must re-calculate or retrieve the correct ledger._
+
+Servers in the network recognize a ledger instance as validated when a supermajority of the peers have signed and broadcast the same validation hash <a href="#footnote_8" id="footnote_from_8"><sup>8</sup></a>. Going forward, transactions are applied to this updated and now validated ledger with sequence number N+1.
```
ximinez: Should `supermajority of the peers` instead say `supermajority of chosen validators` in this and the last couple of paragraphs? Because when I see "peers", I think "the nodes I'm connected to", and I don't care about their validations (unless they happen to be a chosen validator).

content/concepts/consensus-network/consensus.md
```diff
@@ -54,61 +55,82 @@ Important: Some [`rippled` APIs](rippled-api.html) provide provisional results,
...
 
-[![Figure 8: Ledger is Validated When Supermajority of Peers Calculate the Same Result Result](img/consensus-declare-validation.png)](img/consensus-declare-validation.png)
+    [Canonical order](https://github.com/ripple/rippled/blob/8429dd67e60ba360da591bfa905b58a35638fda1/src/ripple/app/misc/CanonicalTXSet.cpp#L25-L36) is not the order the transactions were received, because servers may receive the same transactions in different order. To prevent participants from competing over transaction ordering, canonical order is hard to manipulate.
+
+3. Process each transaction according to its instructions, in order. Update the ledger's state data accordingly.
 
-_Figure 8: Ledger is Validated When Supermajority of Peers Calculate the Same Result — Nodes compare their calculated ledger with the hashes received from chosen validators. If not in agreement, the node must re-calculate or retrieve the correct ledger._
+    If the transaction cannot be successfully executed, destroy the XRP transaction cost only. Include the transaction with a [`tec`-class result code](tec-codes.html).
```
ximinez: `tec` results also cause the account's sequence number to increment.

content/concepts/consensus-network/consensus.md
```diff
@@ -150,24 +172,22 @@ Best practices for applications submitting transactions include:
 
 ## End Notes
 
-<a href="#footnote_from_1" id="footnote_1"><sup>1</sup></a> – A ledger instance can also be uniquely identified by its hash, which is a digital fingerprint of its contents.
-
 <a href="#footnote_from_2" id="footnote_2"><sup>2</sup></a> – Transactions with **tec** result codes are included in ledgers and not do the requested action. The rationale for this is that multiple transactions may be submitted in sequence, with the order of processing determined by a sequence number associated with an account (not to be confused with the ledger sequence number). To prevent a hold on account sequence numbers which would block later transactions, the transaction is processed to consume the sequence number. Additionally, transactions which are distributed to the network must claim a fee to prevent network abuse.
```
ximinez: Footnotes start from number 2.

ximinez (changes requested):
I'm not done yet, but I wanted to share some issues I've found so far.

content/concepts/consensus-network/consensus.md
mDuo13: Linked to a (rewritten) footnote on tec codes and modified the other case that said something similar.

content/concepts/consensus-network/transaction-queue.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Transaction Queue and Consensus
+
+The transaction queue plays an important role in selecting the transactions that are included or excluded from a given ledger version in the consensus process. The following steps describe how the transaction queue relates to the [consensus process](consensus.html).
+
+[![Transaction queue and consensus diagram](img/consensus-with-queue.png)](img/consensus-with-queue.png)
+
+1. **Consensus Round 1** - Each validator proposes a set of transactions to be included in the next ledger version. Each also keeps a queue of candidate transactions not currently proposed.
+
+2. **Consensus Round 2** - If a validator removes a transaction from its proposal in later rounds, it adds that transaction to its queue.
```
ximinez, line 13: This is a little bit oversimplified, but probably appropriate for this level. At the code level, the transaction stays in the open/candidate ledger until consensus is finished, then any transactions in the open ledger that didn't get validated are reprocessed through the queue, potentially getting added if they don't pay enough.

content/concepts/consensus-network/transaction-queue.md
```diff
@@ -0,0 +1,69 @@
...
+
+3. **Consensus Round N** - The consensus process continues until enough servers agree on a transaction set.
+
+4. **Validation** - Servers confirm that they built the same resulting ledger and declare it validated.
+
+5. **Building the Next Proposal** - Each validator prepares its proposal for the next ledger version, starting with queued transactions.
+
+6. **Adding to the Queue** - If the next proposed ledger is already full, incoming transactions are queued for a later ledger version.
```
ximinez: This step leaves out a very important part of the design philosophy for fee escalation and the queue, which is that if a transaction is important enough, there is always a fee that a transaction can pay to bypass the queue and get directly into the open ledger. The ledger never really gets "full", but it gets to a point where transactions aren't cheap. I'm not sure how to better phrase this, but I'll give it a shot... (Feel free to rewrite as appropriate.)

> When the next proposed ledger contains enough transactions, incoming transactions that don't pay the escalated fee are queued for a later ledger version.

> If the next proposed ledger contains enough transactions, incoming transactions that pay enough to be valid, but not enough to meet the escalated cost may be queued for a later ledger version.

Ugh. This is why I'm a code monkey.

content/concepts/consensus-network/transaction-queue.md
```diff
@@ -0,0 +1,69 @@
...
+* If a transaction affects how the sending address authorizes transactions, no other transactions from the same address can be queued behind it. [New in: rippled 0.32.0][]
+* If the transaction includes a `LastLedgerSequence` field, the value of that field must be at least **the current ledger index + 2**.
+
+### Fee Averaging
+
+[New in: rippled 0.33.0][]
+
+If a sending address has one or more transactions queued, that sender can "push" the existing queued transactions into the open ledger by submitting a new transaction with a high enough transaction cost to pay for all of them. Specifically, the new transaction must increase the total transaction cost of the queued transactions from the same sending address, including the new transaction, to cover the [open ledger cost](transaction-cost.html#open-ledger-cost) of each transaction as it gets added to the ledger. The total must include the increased open ledger cost for each new transaction. The transactions must still follow the other [queuing restrictions](#queuing-restrictions) and the sending address must have enough XRP to pay the transaction costs of all the queued transactions.
```
ximinez: `Specifically, the new transaction must *pay* the total transaction cost of the queued transactions...`

content/concepts/consensus-network/transaction-queue.md
```diff
@@ -0,0 +1,69 @@
...
+
+This feature helps you work around a particular situation. If you submitted one or more transactions with a low cost, which got queued, you cannot send new transactions from the same address unless you do one of the following:
+
+* Wait for the queued transactions to be included in a validated ledger, _or_
+* Wait for the queued transactions to be permanently invalidated if the transactions have the [`LastLedgerSequence` field](reliable-transaction-submission.html#lastledgersequence) set, _or_
+* [Cancel the queued transactions](cancel-or-skip-a-transaction.html) by submitting a new transaction with the same sequence number.
```
ximinez: `...and a higher cost.`

ximinez (changes requested):
I have several suggested changes for the Transaction Queue write up. Otherwise, it, and the diagram look really good.

> Reviewers: I'd appreciate your feedback on whether the unused diagram is useful scrap material or should be removed.
I am unsure of which files you are referring to that contain the unused diagram. In general, in code, it's better to remove than to comment out, especially if the unused stuff appears in source control history. So it can be found later if you need it. I think the same rule applies here.

content/concepts/consensus-network/consensus.md
```diff
@@ -6,109 +6,131 @@ This article provides a high level overview of the XRP Ledger, the information i
...
 ## Introduction
 
 The peer-to-peer XRP Ledger network provides a worldwide, shared ledger, which gives applications authoritative information about the state of its contents. This state information includes:
 
 - settings for each [account](accounts.html)
-- balances between accounts (trust lines)
+- balances of XRP and [issued currencies](issued-currencies.html)
 - offers in the distributed exchange
 - network settings, such as [transaction costs](transaction-cost.html) and [reserve](reserves.html) amounts
 - a time stamp
```
gbarr01:
```
< time stamp
---
> timestamp
```

content/concepts/consensus-network/consensus.md
```diff
@@ -6,109 +6,131 @@ This article provides a high level overview of the XRP Ledger, the information i
...
 
-[![Figure 5: Validators Propose Transaction Sets](img/consensus-candidate-transaction-sets.png)](img/consensus-candidate-transaction-sets.png)
+[![Figure 5: Validators Propose and Revise Transaction Sets](img/consensus-rounds.png)](img/consensus-rounds.png)
 
-_Figure 5: Validators Propose Transaction Sets — At the start of consensus, nodes work with different sets of transactions. Rounds of proposals determine which transactions to apply to the ledger, and which must wait for a later round of consensus._
+_Figure 5: Validators Propose and Revise Transaction Sets — At the start of consensus, validators may have different sets of transactions. In later rounds, servers modify their proposals to match what their trusted validators proposed. This process determines which transactions the should apply to the ledger version currently being discussed, and which they should postpone for later ledger versions._
```
gbarr01:
```
< This process determines which transactions the should apply
---
> This process determines which transactions they should apply
```

content/concepts/consensus-network/consensus.md
```diff
@@ -6,109 +6,131 @@ This article provides a high level overview of the XRP Ledger, the information i
...
+
+_Figure 7: An XRP Ledger Server Calculates a Ledger Validation — Each server applies agreed-upon transactions to the previous validated ledger. Validators send their results to the entire network._
+
+#### Compare Results
+
+Validators each relay their results in the form of a signed message containing the hash of the ledger version they calculated. These messages, called _validations_, allow each server to compare the ledger it computed with those of its peers.
+
+[![Figure 8: Ledger is Validated When Supermajority of Peers Calculate the Same Result Result](img/consensus-declare-validation.png)](img/consensus-declare-validation.png)
+
+_Figure 8: Ledger is Validated When Supermajority of Peers Calculate the Same Result — Servers compare their calculated ledger with the hashes received from chosen validators. If not in agreement, the server must re-calculate or retrieve the correct ledger._
```
gbarr01:
```
< Servers compare their calculated ledger with the hashes received from chosen validators. If not in agreement, the server must re-calculate or retrieve the correct ledger.
---
> Servers compare their calculated ledgers with the hashes received from their chosen validators. If not in agreement, the servers must recalculate or retrieve the correct ledger.

or

> Each server compares its calculated ledger with the hashes received from its chosen validators. If not in agreement, the server must recalculate or retrieve the correct ledger.
```

content/concepts/consensus-network/consensus.md
```diff
@@ -6,109 +6,131 @@ This article provides a high level overview of the XRP Ledger, the information i
...
 
 [![Figure 4: Participants in the XRP Ledger Protocol](img/xrp-ledger-network.png)](img/xrp-ledger-network.png)
 
 _Figure 4: Participants in the XRP Ledger Protocol_
 
-The nodes that receive, relay and process transactions may be either tracking nodes or validating nodes. Tracking nodes’ primary functions include distributing transactions from clients and responding to queries about the ledger. Validating nodes perform the same functions as tracking nodes and additionally contribute to advancing the ledger sequence <a href="#footnote_4" id="footnote_from_4"><sup>4</sup></a>.
+The servers that receive, relay and process transactions may be either tracking servers or validators. Tracking servers’ primary functions include distributing transactions from clients and responding to queries about the ledger. Validating servers perform the same functions as tracking servers and additionally contribute to advancing the ledger sequence <a href="#footnote_3" id="from_footnote_3"><sup>3</sup></a>.
```
gbarr01: Suggestion: 
```
< Tracking servers’ primary functions
---
> The primary functions of tracking servers
```

content/concepts/consensus-network/transaction-queue.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Queuing Restrictions
+
+The `rippled` server uses a variety of heuristics to estimate which transactions are "likely to be included in a ledger." The current implementation uses the following rules to decide which transactions to queue:
+
+* Transactions must be properly-formed and [authorized](transaction-basics.html#authorizing-transactions) with valid signatures.
+* Transactions with an `AccountTxnID` field cannot be queued.
+* A single sending address can have at most 10 transactions queued at the same time. In order for a transaction to be queued, the sender must have enough XRP to pay all the XRP costs of all the sender's queued transactions including both the `Fee` fields and the sum of the XRP that each transaction could send. [New in: rippled 0.32.0][]
+* If a transaction affects how the sending address authorizes transactions, no other transactions from the same address can be queued behind it. [New in: rippled 0.32.0][]
+* If the transaction includes a `LastLedgerSequence` field, the value of that field must be at least **the current ledger index + 2**.
```
gbarr01:
```
>
If a transaction affects
If the transaction includes
---
>
If a transaction affects
If a transaction includes
```

content/concepts/consensus-network/transaction-queue.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Fee Averaging
+
+[New in: rippled 0.33.0][]
+
+If a sending address has one or more transactions queued, that sender can "push" the existing queued transactions into the open ledger by submitting a new transaction with a high enough transaction cost to pay for all of them. Specifically, the new transaction must increase the total transaction cost of the queued transactions from the same sending address, including the new transaction, to cover the [open ledger cost](transaction-cost.html#open-ledger-cost) of each transaction as it gets added to the ledger. The total must include the increased open ledger cost for each new transaction. The transactions must still follow the other [queuing restrictions](#queuing-restrictions) and the sending address must have enough XRP to pay the transaction costs of all the queued transactions.
+
+This feature helps you work around a particular situation. If you submitted one or more transactions with a low cost, which got queued, you cannot send new transactions from the same address unless you do one of the following:
```
gbarr01: Suggestion: 
```
< If you submitted one or more transactions with a low cost, which got queued
---
> If you submitted one or more transactions with a low cost that were queued
```

gbarr01:
@mDuo13 , some comments about the diagrams: 

**anatomy-of-a-ledger-complete.png**

1. "Count" in Total XRP Count should be lowercase
2. Put these on one line:
* Timestamp
* Total XRP Count
3. Move "State data" in a tad and put paren content on one line.
4. For Total XRP count, maybe use a graph that goes up?

**ledger-history.png**

> Each ledger version also contains a set of transactions and metadata about those transactions. The transactions it includes are only the ones that have been applied to the previous ledger version to create the new ledger version.

This (2nd) sentence is not reflected in the diagram. I see candidate ledgers with only new proposed transactions, and none that "have been applied to the previous ledger version". Do they get applied to the latest after the candidate has been validated? Doesn't seem so. What am I missing here?sed


**ledger-changes.png**

Should the arrow for "Hash identifies parent" be pointed in the other direction?


**consensus-rounds.png**

Where is this image? Did not see it in the list.


**consensus-calculate-validation.png**

In the 2nd section, shouldn't the green transactions be applied to the previous ledger (as the text in the diagram says)? ... shouldn't the arrow go in the opposite direction? You appear to be applying them to the new ledger.


**consensus-declare-validation.png**

```
< If ≥80% of a server's trusted validators got
---
> If ≥80% of a server's trusted validators get

< If a server got
---
> If a server gets
```

gbarr01:
@mDuo13 , which is the the unused diagram? And should I review the other documents?

content/concepts/consensus-network/transaction-queue.md
mDuo13: I left this sentence as-is, but added a parenthetical statement immediately after it:

> (Transactions that pay the [open ledger cost] can still get into the next proposed ledger even if it's "full", but the open ledger cost grows exponentially with each transaction added this way.)

content/concepts/consensus-network/transaction-queue.md
mDuo13: I rewrote this paragraph a bit to improve even more on this point:

> Specifically, the new transaction must pay a high enough transaction cost to cover the [open ledger cost] of itself and each other transaction from the same sender before it in the queue. (Keep in mind that the open ledger cost increases exponentially each time a transaction pays it.) The transactions must still follow the other [queuing restrictions]...

The "of itself and each other transaction from the same sender before it in the queue" feels like quite a mouthful, but I haven't found a way to condense it any more than that.

content/concepts/consensus-network/consensus.md
mDuo13: I didn't feel happy with multiple "primary" functions, so I changed this a little more:

> The major functions of tracking servers

mDuo13:
@gbarr01 I removed the unused diagram after thinking on it some more (I reached the same conclusion as @ximinez's recommendation above).

To answer your questions on ledger-history.png, I think you've misunderstood what it means to apply a transaction to a ledger. I'm not sure what I can clarify in the document to make this make more sense. What I can say is: in the consensus process, "Ledger version" objects are immutable. (The server does keep a mutable "current open ledger" as a tentative working copy , but the consensus process doesn't actually use the open ledger at all.) When you apply transactions to any ledger version, the result is a new ledger version whose parent is the one you used as a base. The "candidate ledger versions" you see in the diagram are possible results of applying (slightly different sets of) proposed transactions to previous ledger versions. To reiterate, a ledger version contains:

- Metadata about itself, including an identifying hash of its contents and the identifying hash of its direct predecessor ("parent ledger")
- The latest state data
- Transactions that have been newly applied since the parent ledger version
    - This also includes metadata about exactly what happened when applying each transaction (essentially, each transaction's metadata shows a "diff" of the state data as a result of processing the transaction)

mDuo13:
66f7374 contains the changes to the Markdown files. The fixes to diagrams are coming next (after I get my image editor setup in place again). Notes on the specific diagrams:

- `consensus-rounds.png` is not in the diff because it's unchanged from PR #445
- `anatomy-of-a-ledger-complete.png` - I have the fixes for this ready to export. It originally looked more like what you described, but I think screwing with the aspect ratio in Google Slides broke it.
    - I won't be changing the XRP count to a graph going up. It's a graph going down because the total supply of XRP can only decrease as XRP is destroyed.
- `ledger-changes.png` "hash identifies parent" direction—I think it sort of makes sense either way, but your suggestion is better. Changing it.
- `consensus-calculate-validation.png` - As clarified earlier, old ledgers are used as a basis to apply new transactions. The diagram is correct.
- `consensus-declare-validation.png` - I think this is one case where past tense is appropriate and important, because we are describing a state in which servers are looking back on and evaluating results that took place in the recent past from the perspective of the moment in question. [The fact that it works this way is a relevant part of how the consensus algorithm approaches an unsolvable problem.](https://developers.ripple.com/consensus-principles-and-rules.html#consensus-can-fail)

PR #464, opened 2018-10-17
Send a Conditionally-Held Escrow: update example

intelliot:
Hex should be uppercase; see https://github.com/ripple/ripple-lib/issues/953


mDuo13 (approved):
Thanks for the updates.

PR #469, opened 2018-10-27
Tutorials: Add macOS install and Test Net connect to Install rippled

jhaaaa:
- Update Install `rippled` tutorial to include steps for installing `rippled` on macOS
- Add info about how to connect `rippled` to the XRP Test Net
- Incorporate feedback from PM
- Base on install steps from QA
- Base on https://github.com/ripple/rippled/tree/develop/Builds/macos

### DIFF
--- a/content/tutorials/manage-the-rippled-server/install-rippled.md
+++ b/content/tutorials/manage-the-rippled-server/install-rippled.md
@@ -96,7 +96,7 @@ This section assumes that you are using macOS X 10.0 or higher. ***TODO: okay?**
 
 {{n.next()}}. Install [Xcode](https://developer.apple.com/download/).
 
-{{n.next()}}. Install Xcode command line tools. ***TODO: Comment from Daniel: Installing homebrew removes the need to install xcode command line tools separately. JHA: homebrew lists xcode command line tools as a requirement: https://docs.brew.sh/Installation#requirements. Do we keep or remove this step?***
+{{n.next()}}. Install Xcode command line tools. ***TODO: Comment from PM: Installing homebrew removes the need to install xcode command line tools separately. JHA: homebrew lists xcode command line tools as a requirement: https://docs.brew.sh/Installation#requirements. Do we keep or remove this step?***
 
         xcode-select --install
 
@@ -108,11 +108,11 @@ This section assumes that you are using macOS X 10.0 or higher. ***TODO: okay?**
 
         brew update
 
-{{n.next()}}. Use Homebrew to install dependencies. ***TODO: Added boost per Daniel's request: Installing Boost directly is going to frustrate people - brew can do it in one command. I ended up installing Boost with Brew with no issues. Which release to choose? JHA: I did `brew install boost` and brew installs 1.67.0 in /usr/local/Cellar/boost/1.67.0_1 -- is this okay? If we install Boost via brew, do we no longer have to compile boost: ./bootstrap.sh and ./b2 cxxflags="-std=c++14"? Comment from Daniel: Build issues with Boost not finding lib (boost-context) in my case. JHA: when does this issue come up? How can we help the user avoid or address this issue?***
+{{n.next()}}. Use Homebrew to install dependencies. ***TODO: Added boost per PM's request: Installing Boost directly is going to frustrate people - brew can do it in one command. I ended up installing Boost with Brew with no issues. Which release to choose? JHA: I did `brew install boost` and brew installs 1.67.0 in /usr/local/Cellar/boost/1.67.0_1 -- is this okay? If we install Boost via brew, do we no longer have to compile boost: ./bootstrap.sh and ./b2 cxxflags="-std=c++14"? Comment from PM: Build issues with Boost not finding lib (boost-context) in my case. JHA: when does this issue come up? How can we help the user avoid or address this issue?***
 
         brew install git cmake pkg-config protobuf openssl ninja boost
 
-{{n.next()}}. Set the `BOOST_ROOT` environment variable to point to the new directory created by the Homebrew installation of Boost. Put this environment variable in your `.bash_profile` so it's automatically set when you log in. For example, add the following line to the file. To find your Boost install directory, use `brew info boost`. ***TODO: Feedback from Daniel: Have to deal with environment variables. “rc” files is a vague statement. JHA: If you do brew install boost - you still have to set env variables, correct? I did it using .bash_profile and removed references to "rc files" - is this okay? Or do we want to tell users that they can choose to put the env var in the .bashrc file instead of .bash_profile - is there any benefit to saying this?***
+{{n.next()}}. Set the `BOOST_ROOT` environment variable to point to the new directory created by the Homebrew installation of Boost. Put this environment variable in your `.bash_profile` so it's automatically set when you log in. For example, add the following line to the file. To find your Boost install directory, use `brew info boost`. ***TODO: Feedback from PM: Have to deal with environment variables. “rc” files is a vague statement. JHA: If you do brew install boost - you still have to set env variables, correct? I did it using .bash_profile and removed references to "rc files" - is this okay? Or do we want to tell users that they can choose to put the env var in the .bashrc file instead of .bash_profile - is there any benefit to saying this?***
 
         export BOOST_ROOT=/usr/local/Cellar/boost/1.67.0_1
 
@@ -120,7 +120,7 @@ This section assumes that you are using macOS X 10.0 or higher. ***TODO: okay?**
 
         source .bashrc
 
-{{n.next()}}. Clone the `rippled` source code into your desired location and access the `rippled` directory.  ***TODO: Note from Daniel: Cloning github repo is going to require proper github setup. Should we provide some guidance or references? JHA: What is the proper github setup? I may have set this up a whlie ago and I can't remember the steps that may have been required.***
+{{n.next()}}. Clone the `rippled` source code into your desired location and access the `rippled` directory.  ***TODO: Note from PM: Cloning github repo is going to require proper github setup. Should we provide some guidance or references? JHA: What is the proper github setup? I may have set this up a whlie ago and I can't remember the steps that may have been required.***
 
         git clone git@github.com:ripple/rippled.git
         cd rippled
--- a/content/tutorials/manage-the-rippled-server/install-rippled.md
+++ b/content/tutorials/manage-the-rippled-server/install-rippled.md
@@ -98,82 +98,80 @@ At this time, we don't recommend using the macOS platform for `rippled` producti
 
 That said, macOS is suitable for many development and testing tasks. `rippled` has been tested for use with macOS High Sierra up to 10.13.
 
-{% set n = cycler(* range(1,99)) %}
+1. Install [Xcode](https://developer.apple.com/download/).
 
-{{n.next()}}. Install [Xcode](https://developer.apple.com/download/).
+0. Install Xcode command line tools.
 
-{{n.next()}}. Install Xcode command line tools.
+        $ xcode-select --install
 
-        xcode-select --install
+0. Install [Homebrew](https://brew.sh/).
 
-{{n.next()}}. Install [Homebrew](https://brew.sh/).
+        $ ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
 
-        ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+0. Update Homebrew.
 
-{{n.next()}}. Update Homebrew.
+        $ brew update
 
-        brew update
+0. Use Homebrew to install dependencies.
 
-{{n.next()}}. Use Homebrew to install dependencies.
+        $ brew install git cmake pkg-config protobuf openssl ninja
 
-        brew install git cmake pkg-config protobuf openssl ninja
+0. Install Boost 1.67.0. `rippled` 1.1.x is compatible with Boost 1.67.
 
-{{n.next()}}. Install Boost 1.67.0. Currently, we support Boost 1.67 only.
+        $ brew install boost@1.67
 
-        brew install boost@1.67
-
-{{n.next()}}. Ensure that your `BOOST_ROOT` environment variable points to the directory created by the Boost installation. To find your Boost install directory, use `brew info boost`. Put this environment variable in your `.bash_profile` file so it's automatically set when you log in. For example:
+0. Ensure that your `BOOST_ROOT` environment variable points to the directory created by the Boost installation. To find your Boost install directory, use `brew info boost`. Put this environment variable in your `.bash_profile` file so it's automatically set when you log in. For example:
 
         export BOOST_ROOT=/usr/local/Cellar/boost/1.67.0_1
 
-{{n.next()}}. If you updated your `.bash_profile` file in the previous step, be sure to source it. For example:
+0. If you updated your `.bash_profile` file in the previous step, be sure to source it. For example:
 
-        source .bash_profile
+        $ source .bash_profile
 
-{{n.next()}}. Clone the `rippled` source code into your desired location and access the `rippled` directory. To do this, you'll need to set up Git (installed earlier using Homebrew) and GitHub. For example, you'll need to create a GitHub account and set up your SSH key. For more information, see [Set up git](https://help.github.com/articles/set-up-git/).
+0. Clone the `rippled` source code into your desired location and access the `rippled` directory. To do this, you'll need to set up Git (installed earlier using Homebrew) and GitHub. For example, you'll need to create a GitHub account and set up your SSH key. For more information, see [Set up git](https://help.github.com/articles/set-up-git/).
 
-        git clone git@github.com:ripple/rippled.git
-        cd rippled
+        $ git clone git@github.com:ripple/rippled.git
+        $ cd rippled
 
-{{n.next()}}. By default, cloning puts you on the `develop` branch. Use this branch if you are doing development work and want the latest set of untested features.
+0. By default, cloning puts you on the `develop` branch. Use this branch if you are doing development work and want the latest set of untested features.
 
       If you want the latest stable release, checkout the `master` branch.
 
-        git checkout master
+        $ git checkout master
 
       If you want to test out the latest release candidate, checkout the `release` branch:
 
-        git checkout release
+        $ git checkout release
 
       Or, you can checkout one of the tagged releases listed on [GitHub](https://github.com/ripple/rippled/releases).
 
-{{n.next()}}. In the `rippled` directory you just cloned, create your build directory and access it. For example:
+0. In the `rippled` directory you just cloned, create your build directory and access it. For example:
 
-        mkdir my_build
-        cd my_build
+        $ mkdir my_build
+        $ cd my_build
 
-{{n.next()}}. Build `rippled`. This could take about 5 minutes, depending on your hardware specs.
+0. Build `rippled`. This could take about 5 minutes, depending on your hardware specs.
 
-        cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug ..
+        $ cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug ..
 
       You can set `CMAKE_BUILD_TYPE` to the `Debug` or `Release` build type. All four standard [`CMAKE_BUILD_TYPE`](https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html) values are supported.
 
-{{n.next()}}. Run the build using CMake. This could take about 10 minutes, depending on your hardware specs.
+0. Run the build using CMake. This could take about 10 minutes, depending on your hardware specs.
 
-        cmake --build . -- -j 4
+        $ cmake --build . -- -j 4
 
       **Tip:** This example uses a `-j` parameter set to `4`, which uses four processes to build in parallel. The best number of processes to use depends on how many CPU cores your hardware has available. Use `sysctl -n hw.ncpu` to get your CPU core count.
 
-{{n.next()}}. Run unit tests built into the server executable. This could take about 5 minutes, depending on your hardware specs. (optional, but recommended) <!--{# ***TODO: We should provide info about what to do if you get failures. What should the user do? PM looking into this.*** #}-->
+0. Run unit tests built into the server executable. This could take about 5 minutes, depending on your hardware specs. (optional, but recommended) <!--{# ***TODO: We should provide info about what to do if you get failures. What should the user do? PM looking into this.*** #}-->
 
-        ./rippled --unittest
+        $ ./rippled --unittest
 
-{{n.next()}}. `rippled` requires the `rippled.cfg` configuration file to run. You can find an example config file, `rippled-example.cfg` in `rippled/cfg`. Make a copy and save it as `rippled.cfg` in a location that enables you to run `rippled` as a non-root user. Access the `rippled` directory and run:
+0. `rippled` requires the `rippled.cfg` configuration file to run. You can find an example config file, `rippled-example.cfg` in `rippled/cfg`. Make a copy and save it as `rippled.cfg` in a location that enables you to run `rippled` as a non-root user. Access the `rippled` directory and run:
 
-        mkdir -p /etc/opt/ripple ***TODO: changed this from ~/.config/ripple to match guidance coming in conf-file-location.md PR -- okay?***
-        cp cfg/rippled-example.cfg /etc/opt/ripple/rippled.cfg
+        $ mkdir -p /etc/opt/ripple ***TODO: changed this from ~/.config/ripple to match guidance coming in conf-file-location.md PR -- okay?***
+        $ cp cfg/rippled-example.cfg /etc/opt/ripple/rippled.cfg
 
-{{n.next()}}. Edit `rippled.cfg` to set necessary file paths. The user you plan to run `rippled` as must have write permissions to all of the paths you specify here. ***TODO: We need to add this step to the centos and ubuntu tasks, correct? For centos and ubuntu, the config files are automatically placed in the correct directories, but the following paths still need to be manually checked/updated to ensure that there are no permission issues, correct?***
+0. Edit `rippled.cfg` to set necessary file paths. The user you plan to run `rippled` as must have write permissions to all of the paths you specify here. ***TODO: We need to add this step to the centos and ubuntu tasks, correct? For centos and ubuntu, the config files are automatically placed in the correct directories, but the following paths still need to be manually checked/updated to ensure that there are no permission issues, correct?***
 
       * Set the `[node_db]` path to the location where you want to store the ledger database.
 
@@ -183,15 +181,15 @@ That said, macOS is suitable for many development and testing tasks. `rippled` h
 
       These are the only configurations required for `rippled` to start up successfully. All other configuration is optional and can be tweaked after you have a working server. For more information, see [Additional Configurations](#additional-configuration).
 
-{{n.next()}}. `rippled` requires the `validators.txt` file to run. You can find an example validators file, `validators-example.txt`, in `rippled/cfg/`. Make a copy and save it as `validators.txt` in the same folder as your `rippled.cfg` file. Access the `rippled` directory and run:
+0. `rippled` requires the `validators.txt` file to run. You can find an example validators file, `validators-example.txt`, in `rippled/cfg/`. Make a copy and save it as `validators.txt` in the same folder as your `rippled.cfg` file. Access the `rippled` directory and run:
 
-        cp cfg/validators-example.txt /etc/opt/ripple ***TODO: changed this from ~/.config/ripple to match guidance coming in conf-file-location.md PR -- okay?***
+        $ cp cfg/validators-example.txt /etc/opt/ripple ***TODO: changed this from ~/.config/ripple to match guidance coming in conf-file-location.md PR -- okay?***
 
       **Warning:** Ripple has designed a decentralization plan with maximum safety in mind. During the transition, you should not modify the `validators.txt` file except as recommended by Ripple. Even minor modifications to your validator settings could cause your server to diverge from the rest of the network and report out of date, incomplete, or inaccurate data. Acting on such data can cause you to lose money.
 
-{{n.next()}}. Access your build directory, `my_build` for example, and start the `rippled` service.
+0. Access your build directory, `my_build` for example, and start the `rippled` service.
 
-        sudo ./rippled
+        $ sudo ./rippled
 
       Here's an excerpt of what you can expect to see in your terminal:
 
--- a/content/tutorials/manage-the-rippled-server/install-rippled.md
+++ b/content/tutorials/manage-the-rippled-server/install-rippled.md
@@ -27,6 +27,8 @@ Amazon EC2's `m3.large` VM size may be appropriate depending on your workload. A
 
 ## Installation on Ubuntu with alien
 
+<!--{# ***TODO: this doc states that ubuntu has received the highest level of QA and testing - so I moved it above centos and macOS install sections. Okay?*** #}-->
+
 This section assumes that you are using Ubuntu 15.04 or later.
 
 1. Install yum-utils and alien:
@@ -108,11 +110,31 @@ This section assumes that you are using macOS X 10.0 or higher. ***TODO: okay?**
 
         brew update
 
-{{n.next()}}. Use Homebrew to install dependencies. ***TODO: Added boost to the brew installs per PM's request: Installing Boost directly is going to frustrate people - brew can do it in one command.*** Boost 1.67.x to 1.68.x are supported. To view the Boost version that will be installed by `brew install boost`, use `brew info boost`. ***TODO: added this text to ensure that users don't install unsupported versions of Boost. For example, what if the brew install boost starts installing 1.69, but we haven't tested and verified yet? Are the supported versions correct? Do we need to list supported version for any of the other dependencies as well?***
+{{n.next()}}. Use Homebrew to install dependencies.
 
-        brew install git cmake pkg-config protobuf openssl ninja boost
+        brew install git cmake pkg-config protobuf openssl ninja
 
-{{n.next()}}. Ensure that the `BOOST_ROOT` environment variable points to the directory created by the Homebrew installation of Boost. Put this environment variable in your `.bash_profile` so it's automatically set when you log in. For example, ensure that the following line is in the file. To find your Boost install directory, use `brew info boost`. ***TODO: Feedback from PM: Have to deal with environment variables. JHA: When I did a brew install boost, I had to set the env variable -- but I think it may be because I had an existing BOOST_ROOT env var set to a location I created when I previously did a manual boost install. Maybe the brew install doesn't overwrite/update an existing BOOST_ROOT value? Changed the wording a bit to address this - so it is not that you'll always need to add the env var - just make sure it is there and up to date?***
+{{n.next()}}. Install a supported version of Boost. Supported versions include Boost 1.67.x - 1.68.x. ***TODO: correctly stated supported versions? Doc will need to keep an eye on these supported versions. In the rippled repo, we say "Boost 1.67 or later is required." <-- Based on our discussion in this doc, I don't think this is true.*** Installation using Homebrew is faster and requires fewer manual steps, but you must ensure that Homebrew installs a supported version of Boost.
+
+    Pick one of the following options to install Boost:
+
+    * To ensure that Homebrew installs a supported version of Boost, use `brew install boost@1.67` to install Boost 1.67.0. ***TODO: When the supported versions change above, we can also change this, if needed. There is no brew command to install 1.68 right now. brew install boost is still installing 1.67 and there is no brew formula boost@1.68.***
+
+    * To use Homebrew to install a different supported version of Boost, check [Homebrew Formulae - boost](https://formulae.brew.sh/formula/boost) to see if there is a formula for the version you want to install. To verify the Boost version a Homebrew formula will install, use `brew info <formula>`. For example, `brew info boost` or `brew info boost@1.67`.
+
+    * To install a supported version of Boost that does not have a defined Homebrew formula, compile Boost yourself. These compile instructions are for Boost 1.68.0.
+
+        1. Download [Boost 1.68.0](https://dl.bintray.com/boostorg/release/1.68.0/source/boost_1_68_0.tar.bz2). ***TODO: do we want to give them CLI instructions? If yes, then we need to have them use brew install wget.***
+
+        2. Extract it to a folder. Be sure to note the location. ***TODO: provide CLI instructions?***
+
+        3. In a terminal, run:
+
+              `./bootstrap.sh`
+
+              `./b2 cxxflags="-std=c++14"`
+
+{{n.next()}}. Ensure that the `BOOST_ROOT` environment variable points to the directory created by your Boost installation. Put this environment variable in your `.bash_profile` so it's automatically set when you log in. For example, ensure that the following line is in the file. To find your Boost install directory, use `brew info boost`. ***TODO: JHA clean out boost env vars and uninstall boost - then try again and see what gets populated in bash_profile....***
 
         export BOOST_ROOT=/usr/local/Cellar/boost/1.67.0_1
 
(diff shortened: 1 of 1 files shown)


jhaaaa:
Hi @manojsdoshi - if you have some time, it would be great to get your input on this tutorial about `rippled` install for macOS. I based the flow on your testing steps and on feedback from PM. Thank you for any feedback you have time to provide!

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +81,193 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+This section assumes that you are using macOS X 10.0 or higher. ***TODO: okay?***
 
-3. Download the `rippled` software package:
+{% set n = cycler(* range(1,99)) %}
 
-        $ yumdownloader --enablerepo=ripple-stable --releasever=el7 rippled
+{{n.next()}}. Install [Xcode](https://developer.apple.com/download/).
 
-4. Verify the signature on the `rippled` software package:
+{{n.next()}}. Install Xcode command line tools. ***TODO: Comment from PM: Installing homebrew removes the need to install xcode command line tools separately. JHA: homebrew lists xcode command line tools as a requirement: https://docs.brew.sh/Installation#requirements. Do we keep or remove this step?***
```
garay-daniel: Ah, maybe its dependent on some sort of installation settings where you can install both simultaneously. I think its safer bet to leave it as for now, I can look into it.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +81,193 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+        ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
 
-6. Configure the `rippled` service to start on system boot:
+{{n.next()}}. Update Homebrew.
 
-        $ sudo systemctl enable rippled.service
+        brew update
 
-7. Start the `rippled` service
+{{n.next()}}. Use Homebrew to install dependencies. ***TODO: Added boost per PM's request: Installing Boost directly is going to frustrate people - brew can do it in one command. I ended up installing Boost with Brew with no issues. Which release to choose? JHA: I did `brew install boost` and brew installs 1.67.0 in /usr/local/Cellar/boost/1.67.0_1 -- is this okay? If we install Boost via brew, do we no longer have to compile boost: ./bootstrap.sh and ./b2 cxxflags="-std=c++14"? Comment from PM: Build issues with Boost not finding lib (boost-context) in my case. JHA: when does this issue come up? How can we help the user avoid or address this issue?***
```
garay-daniel: (1) Using homebrew removes the need to compile it from scratch. You also do not need to set environment variables yourself. The only catch is brew updates to the latest version of boost, which the Rippled team must first OK in order for it to work. 1.68 is the latest release and it works fine, but 1.69 will need to be checked. We may need to periodically update to reflect this?

(2) Not sure why it didn't work, I did not try to debug it and this is why I tried to use homebrew instead.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +81,193 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+
+        brew install git cmake pkg-config protobuf openssl ninja boost
+
+{{n.next()}}. Set the `BOOST_ROOT` environment variable to point to the new directory created by the Homebrew installation of Boost. Put this environment variable in your `.bash_profile` so it's automatically set when you log in. For example, add the following line to the file. To find your Boost install directory, use `brew info boost`. ***TODO: Feedback from PM: Have to deal with environment variables. “rc” files is a vague statement. JHA: If you do brew install boost - you still have to set env variables, correct? I did it using .bash_profile and removed references to "rc files" - is this okay? Or do we want to tell users that they can choose to put the env var in the .bashrc file instead of .bash_profile - is there any benefit to saying this?***
```
garay-daniel: If you install with homebrew you do not need to set your own environment variables. I think just mentioning its bash_profile would be fine. You can use your .bashrc, but it requires also using your bash_profile as far as I know. I just thought "rc files" could be misinterpreted.

garay-daniel: You do not need to set the env variables with brew, but if you choose to install in another way you will have to do it. I think bash_profile is clear enough. I think you can use your .bashrc as well, but it requires putting something into your bash_profile too as far as I know.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +81,193 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+
+        export BOOST_ROOT=/usr/local/Cellar/boost/1.67.0_1
+
+{{n.next()}}. Source the file you added the `BOOST_ROOT` environment variable to. For example:
+
+        source .bashrc
+
+{{n.next()}}. Clone the `rippled` source code into your desired location and access the `rippled` directory.  ***TODO: Note from PM: Cloning github repo is going to require proper github setup. Should we provide some guidance or references? JHA: What is the proper github setup? I may have set this up a whlie ago and I can't remember the steps that may have been required.***
```
garay-daniel: Leave it up to you here, I imagine most will have proper github setup, but was worth mentioning. Maybe a simple line to make that clear is enough? You need a github acc, setup your ssh key, etc. 

https://help.github.com/articles/set-up-git/

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +81,193 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+
+        cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug ..
+
+      You can set `CMAKE_BUILD_TYPE` to the `Debug` or `Release` build type. All four standard [`CMAKE_BUILD_TYPE`](https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html) values are supported.
+
+{{n.next()}}. Run the build using CMake. This could take about 10 minutes, depending on your hardware specs.
+
+        cmake --build . -- -j 4
+
+      This example uses a `-j` parameter set to `4`, which uses four processes to build in parallel. The best number of processes to use depends on how many CPU cores your hardware has available. Use `sysctl -a | grep machdep.cpu | grep core_count` to get your CPU core count. ***TODO: the mac build page in the github repo goes much more in-depth about cmake build options - should we include these here? https://github.com/ripple/rippled/tree/develop/Builds/macos#generate-and-build***
```
garay-daniel: I don't think its necessary and did not use it myself. Not sure what the background for those instructions are however, maybe Rome knows more?

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +81,193 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+
+{{n.next()}}. Run unit tests built into the server executable. This could take about 5 minutes, depending on your hardware specs. (optional, but recommended) ***TODO: We should provide info about what to do if you get failures. What should the user do?***
+
+        ./rippled --unittest
+
+{{n.next()}}. Create a copy of the example config file and save it to a location that enables you to run `rippled` as a non-root user (recommended). ***TODO: unlike the centos and ubuntu installs on this page, the macos install requires that you manually put rippled.cfg and validator.txt in place. At least that is what I experienced - true?***
```
garay-daniel: Yes I had to create both.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +81,193 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+
+      You can set `CMAKE_BUILD_TYPE` to the `Debug` or `Release` build type. All four standard [`CMAKE_BUILD_TYPE`](https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html) values are supported.
+
+{{n.next()}}. Run the build using CMake. This could take about 10 minutes, depending on your hardware specs.
+
+        cmake --build . -- -j 4
+
+      This example uses a `-j` parameter set to `4`, which uses four processes to build in parallel. The best number of processes to use depends on how many CPU cores your hardware has available. Use `sysctl -a | grep machdep.cpu | grep core_count` to get your CPU core count. ***TODO: the mac build page in the github repo goes much more in-depth about cmake build options - should we include these here? https://github.com/ripple/rippled/tree/develop/Builds/macos#generate-and-build***
+
+{{n.next()}}. Run unit tests built into the server executable. This could take about 5 minutes, depending on your hardware specs. (optional, but recommended) ***TODO: We should provide info about what to do if you get failures. What should the user do?***
```
garay-daniel: Yes agreed, we need to learn what those failures are. I can ask.

content/tutorials/manage-the-rippled-server/install-rippled.md
jhaaaa: Got it. When I installed, it did not update my BOOST_ROOT env var, but that is probably because it was already set? Maybe it doesn't overwrite/update if a value exists? I've updated the text to address - let's review and see if it works for you.

content/tutorials/manage-the-rippled-server/install-rippled.md
jhaaaa: Sounds great - let's see if Manoj and Rome have feedback. I'm going to add Rome as a reviewer once I get a first round of reviews incorporated. 👍

content/tutorials/manage-the-rippled-server/install-rippled.md
jhaaaa: Fantastic - thank you!

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -26,6 +25,42 @@ Amazon EC2's `m3.large` VM size may be appropriate depending on your workload. A
 **Tip:** For recommendations beyond the minimum requirements, see [Capacity Planning](capacity-planning.html).
 
 
+## Installation on Ubuntu with alien
```
jhaaaa: We state that ubuntu has received the highest level of QA and testing - so I moved it above centos and macOS install sections. Okay?

content/tutorials/manage-the-rippled-server/install-rippled.md
mDuo13: Users will have to compile boost themselves any time the version of boost that's in the homebrew repositories is not a version that's supported by rippled. At the moment the two are compatible but I'm not convinced that's going to remain true for the long term. Perhaps we should have both instructions available and switch the recommendations periodically based on whether the easier brew install is compatible.

mDuo13: And as Daniel alluded, the compatibility between the Brew and rippled versions can break in two ways:

- brew updates Boost to a new version that's not compatible with the current release of rippled
- rippled updates to a new version that's not compatible with the version of Boost in the Brew repositories

The former is more likely than the latter since Brew tends to update quickly, but the latter is also possible if the person who's currently maintaining the Boost package for Brew changes or takes a long vacation or something, or if the new version of Boost is held up from being added to Brew because of some other incompatibilities with other Brew software that requires it. (Similar situations have happened to me on Arch Linux.)

The "compile it yourself" instructions should be fairly stable and will only change as the result of `rippled` updates, so that would be a much easier maintenance path going forward, even though it's more work for users. The "use the version from Brew" instructions are only sometimes going to be possible and if we want to support those we'll have to keep track of when new versions of Boost are added to Brew. When the Brew-based installation instructions work, they are much easier for users, so I leave it to @jhaaaa to decide whether that's worth the effort of keeping track.

content/tutorials/manage-the-rippled-server/install-rippled.md
mDuo13: I'm not sure on this, but maybe you only need to set the BOOST_ROOT environment variable if you're compiling Boost yourself. If you're able to install a compatible version of Boost via Brew (per the debate above) then the part about setting your BOOST_ROOT env var may be moot.

content/tutorials/manage-the-rippled-server/install-rippled.md
mDuo13: When I updated the Ubuntu equivalent of these instructions, I didn't want to overwhelm the reader with options and possibilities, so I left a couple of tips but didn't go into extreme detail. I also don't like relying on a series of several piped commands including a `grep` for getting your CPU count, since a minor change in the OS's output format could break that. I went with a slightly more generic message about where to get information about your processor.

These instructions can probably be based on the changes I made there: https://github.com/ripple/ripple-dev-portal/blame/master/content/tutorials/manage-the-rippled-server/build-run-rippled-ubuntu.md#L107

content/tutorials/manage-the-rippled-server/install-rippled.md
garay-daniel: I guess we should decide how tricky boost compilation is. Im not sure how often Boost updates, but thats about how often we would need to update these instructions with the new version for brew commands. You can have brew install a specific version of a formula, which is the 3rd option as well. Something like brew install boost 1.67.0 ...

content/tutorials/manage-the-rippled-server/install-rippled.md
garay-daniel: Sorry if I was not clear. You definitely must set the env variable if compiling yourself. If you also install a brew version, you would have 2 boost(s) since brew uses its directory (different from yours probably) to install. A fresh install with only brew would not need the BOOST_ROOT env variable.

content/tutorials/manage-the-rippled-server/install-rippled.md
jhaaaa: Great - I will leave out the reference to the in-depth cmake build options. 👍 

@mDuo13 - Regarding the grep - I added the command because `cat /proc/cpuinfo` didn't work for me in macOS. I see your point about it being brittle! Is there a more stable way to get this info for macOS? Or perhaps I should just leave the command out for macOS?

mDuo13: Does this command work on macOS?

```
 sysctl -n hw.ncpu 
```

jhaaaa: YES! 🙌 Thank you!

jhaaaa:
@garay-daniel and @mDuo13 - wow, thank you so much for your thoughtful and detailed feedback on this PR! 👍 I think I've addressed most of our open items with this latest push. Please take a look to ensure that I've met your requirements. You can search on `TODO` to find my comments/questions.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -26,6 +25,44 @@ Amazon EC2's `m3.large` VM size may be appropriate depending on your workload. A
...
+
+<!--{# ***TODO: this doc states that ubuntu has received the highest level of QA and testing - so I moved it above centos and macOS install sections. Okay?*** #}-->
```
manojsdoshi: I think this is fair since most of the development and testing happens on Ubuntu.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +83,222 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
 
-        $ sudo apt-get update
-        $ sudo apt-get install yum-utils alien
+At this time, we don't recommend using the macOS platform for `rippled` production use. For production, consider using the [Ubuntu platform](#installation-on-ubuntu-with-alien), which has received the highest level of quality assurance and testing.
 
-2. Install the Ripple RPM repository:
+That said, macOS is suitable for many development and testing tasks.
 
-        $ sudo rpm -Uvh https://mirrors.ripple.com/ripple-repo-el7.rpm
+This section assumes that you are using macOS X 10.0 or higher. ***TODO: okay?***
```
manojsdoshi: This makes sense since most people who are developing at Ripple are on macOS X 10.0 (most people are on macOS High Sierra – 10.13). 
Although we haven't upgraded to the latest macOS - "10.14 Mojave" so I am not sure if anything might have changed in that version. Not sure how to handle this in the docs maybe put a note stating that we have tested only till 10.13

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +83,222 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
 
-        $ sudo systemctl enable rippled.service
+        brew update
 
-7. Start the `rippled` service
+{{n.next()}}. Use Homebrew to install dependencies.
+
+        brew install git cmake pkg-config protobuf openssl ninja
+
+{{n.next()}}. Install a supported version of Boost. Supported versions include Boost 1.67.x - 1.68.x. ***TODO: correctly stated supported versions? In the rippled repo, we say "Boost 1.67 or later is required." <-- Based on our discussion in this doc, I don't think this open-ended "later" part is true, is it?***
```
manojsdoshi: With the current release, we have tested with boost 1.67 and support that. 
We can put a note that, it will work with 1.68 but not yet tested.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +83,222 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
 
 ## Postinstall
 
-It can take several minutes for `rippled` to sync with the rest of the network, during which time it outputs warnings about missing ledgers. After that, you have a fully functional stock `rippled` server that you can use for local signing and API access to the XRP Ledger.
+It can take several minutes for `rippled` to sync with the rest of the network, during which time it outputs warnings about missing ledgers.
+
+Here's an excerpt of what you can expect to see in your terminal: ***TODO: you'll see this in the CLI for the macOS install because we tell them to start the service using `sudo ./rippled`, but not for centos and ubuntu. For centos and ubuntu, we tell the user to start the service using `sudo systemctl start rippled.service`, which starts the service, but doesn't display a log of the service starting up. At least for first-time start up - would it make sense to give the user this start up command (sudo ./rippled) instead so that they can see it running? Perhaps in Postinstall we can give them the `sudo systemctl start rippled.service` command for when they want to start the service, but don't need to see the running log?***
```
manojsdoshi: For centos and ubuntu after running 'sudo systemctl start rippled' the user can run another command  'sudo systemctl status rippled' to verify is rippled has started.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +83,222 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+
+{{n.next()}}. Install a supported version of Boost. Supported versions include Boost 1.67.x - 1.68.x. ***TODO: correctly stated supported versions? In the rippled repo, we say "Boost 1.67 or later is required." <-- Based on our discussion in this doc, I don't think this open-ended "later" part is true, is it?***
+
+    Pick one of the following Boost installation methods: ***TODO: I provided the two options below because these are the most straightforward ways for a user to install one of the two supported versions of Boost. In the future, if there are more supported versions or more brew boost formalae, there may be more options to document.***
+
+    * To install Boost 1.67.0, use `brew install boost@1.67`. ***TODO: When Boost 1.67.0 is no longer supported, we'll need to change this formula. Until then, the user will be able to use this command to get a supported version of Boost. `brew install boost` also installs 1.67.0, but as discussed, we don't want to provide the command because it won't always install 1.67.0 and may cause the user to install an unsupported version. There is no brew command to install 1.68 right now.***
```
manojsdoshi: lgtm

jhaaaa:
@mDuo13 and @garay-daniel - Manoj helped get Nik's confirmation that we'll go with Boost 1.67.0 only for now. I'll update the doc to reflect this and send a note when this PR is ready for another review.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +83,222 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+
+    * If you manually compiled and installed Boost, ensure that `BOOST_ROOT` is set to the location where you installed Boost. For example:
+
+        `export BOOST_ROOT=/Users/tblee/boost/boost_1_68_0`
+
+{{n.next()}}. If you updated your `.bash_profile` file in the previous step, be sure to source it. For example:
+
+        source .bash_profile
+
+{{n.next()}}. Clone the `rippled` source code into your desired location and access the `rippled` directory. To do this, you'll need to set up Git (installed earlier using Homebrew) and GitHub. For example, you'll need to create a GitHub account and set up your SSH key. For more information, see [Set up git](https://help.github.com/articles/set-up-git/). ***TODO: does this sound okay for introducing what a user may have to do to set up git and GitHub?***
```
manojsdoshi: lgtm

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +83,222 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+
+        git checkout master
+
+      Or, you can checkout one of the tagged releases listed on [GitHub](https://github.com/ripple/rippled/releases).
+
+      If you want to test out the latest release candidate, checkout the `release` branch:
+
+        git checkout release
+
+{{n.next()}}. In the `rippled` directory you just cloned, create your build directory and access it. ***TODO: Does this "In the rippled directory you just cloned" language make sense? I just want to tell the user where to create the build directory.*** For example:
```
manojsdoshi: lgtm

content/tutorials/manage-the-rippled-server/install-rippled.md
jhaaaa: I'm going to leave out the mention of 1.68. When I put it in the doc - it seems to scream out - install me. 😹

jhaaaa:
@mDuo13, @garay-daniel, and @manojsdoshi: Here's and updated version. So close! Please search on TODO to see my outstanding comments/questions. Any feedback/guidance you have time to provide is greatly appreciated!

jhaaaa:
@mDuo13, @garay-daniel, and @manojsdoshi - added info about how to connect your `rippled` to the XRP Test Net. Add to the Additional Configuration section. Just a few more steps to review. Thank you!

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -295,9 +265,9 @@ Once you have your stock `rippled` server running, you may want to consider runn
 
 `rippled` should connect to the XRP Ledger with the default configurations. However, you can change your settings by editing the `rippled.cfg` file. For recommendations about configuration settings, see [Capacity Planning](capacity-planning.html).
 
-Changes to the `[node_db]`, `[debug_logfile]`, or `[database_path]` sections may require you to give the `rippled` user and group ownership to your new configured path: ***TODO: I added [node_db] to this list - okay? When I try the following command, `chown -R rippled:rippled /var/lib/rippled/db/rocksdb`, I get `chown: rippled: illegal group name.` Is `rippled:rippled` just an example below -- or should there be a `rippled` group that is created by the build/install process?***
+Changes to the `[node_db]`, `[debug_logfile]`, or `[database_path]` sections may require you to give the `rippled` user and group ownership to your new configured path: ***TODO: I added [node_db] to this list - okay?***
```
manojsdoshi: lgtm

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -295,9 +265,9 @@ Once you have your stock `rippled` server running, you may want to consider runn
...
 
-    $ chown -R rippled:rippled <configured path>
+    $ chown -R rippled:rippled <configured path> ***TODO: Is `rippled:rippled` just an example value? When I run this command on macOS, I get `chown: rippled: illegal group name.` When I run it on centos, I get chown: changing ownership of ... : Operation not permitted***
```
manojsdoshi: This group doesn't exist on your Mac that's why you get this error. 
When you spin up aws machines to test centos/Ubuntu (I assume that's what you are doing) this user group is created by default (OPS has this configured) hence you don't get this error. 

But you have the right idea to give permissions to the user who is going to run rippled (maybe we don't need to give an example of how to change permissions? but just say we need to get the right permission to start rippled)

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -201,7 +173,7 @@ This section assumes that you are using macOS X 10.0 or higher. ***TODO: okay?**
         mkdir -p /etc/opt/ripple ***TODO: changed this from ~/.config/ripple to match guidance coming in conf-file-location.md PR -- okay?***
         cp cfg/rippled-example.cfg /etc/opt/ripple/rippled.cfg
 
-{{n.next()}}. Edit `rippled.cfg` to set necessary file paths. The user you plan to run `rippled` as must have write permissions to all of the paths you specify here. ***TODO: I think we need to add this step to the centos and ubuntu tasks, correct? For centos and ubuntu, the config files are automatically placed in the correct directories, but the following paths still need to be manually updated to avoid permission issues, correct?***
+{{n.next()}}. Edit `rippled.cfg` to set necessary file paths. The user you plan to run `rippled` as must have write permissions to all of the paths you specify here. ***TODO: We need to add this step to the centos and ubuntu tasks, correct? For centos and ubuntu, the config files are automatically placed in the correct directories, but the following paths still need to be manually checked/updated to ensure that there are no permission issues, correct?***
```
manojsdoshi: Yup, it's a good idea for the user to verify if they have the correct permissions rather than rippled failing to start and the user opening a support ticket for us.

manojsdoshi: Maybe this can go as a tip to the user or debug steps if they cant start rippled

manojsdoshi (approved):
LGTM

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +85,234 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+At this time, we don't recommend using the macOS platform for `rippled` production use. For production, consider using the [Ubuntu platform](#installation-on-ubuntu-with-alien), which has received the highest level of quality assurance and testing.
 
-3. Download the `rippled` software package:
+That said, macOS is suitable for many development and testing tasks. `rippled` has been tested for use with macOS High Sierra up to 10.13.
 
-        $ yumdownloader --enablerepo=ripple-stable --releasever=el7 rippled
+{% set n = cycler(* range(1,99)) %}
 
-4. Verify the signature on the `rippled` software package:
+{{n.next()}}. Install [Xcode](https://developer.apple.com/download/).
```
mDuo13: I don't remember what the outcome is from our previous discussion on this topic, but I still think it's better to follow the more concise paradigm of having all steps (after the first) be numbered "0" if you want to have list steps be automatically numbered.

After all, this:

```
1. step 1
0. step 2
0. step 3
```

... is much easier to read and write than this:

```
{% set n = cycler(* range(1,99)) %}

{{n.next()}}. Step 1
{{n.next()}}. Step 2
{{n.next()}}. Step 3
```

... and the former _also_ has the advantage of working in the GitHub preview and other Markdown parsers that don't use Dactyl's pre-parser.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +85,234 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+{{n.next()}}. Update Homebrew.
 
-7. Start the `rippled` service
+        brew update
+
+{{n.next()}}. Use Homebrew to install dependencies.
+
+        brew install git cmake pkg-config protobuf openssl ninja
+
+{{n.next()}}. Install Boost 1.67.0. Currently, we support Boost 1.67 only.
```
mDuo13: Nit: "We" don't support Boost; `rippled` 1.1.x _is compatible with_ Boost 1.67.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +85,234 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+
+        brew install boost@1.67
+
+{{n.next()}}. Ensure that your `BOOST_ROOT` environment variable points to the directory created by the Boost installation. To find your Boost install directory, use `brew info boost`. Put this environment variable in your `.bash_profile` file so it's automatically set when you log in. For example:
+
+        export BOOST_ROOT=/usr/local/Cellar/boost/1.67.0_1
```
mDuo13, line 127: An oddity (which applies to the other version of these instructions) worth note: This is the only code sample in these steps which isn't (intended as) a commandline instruction. Almost coincidentally, it works as one, but it won't have quite the intended effect—it's temporary only, and the next step is pointless if you don't put this in your `.bash_profile`.

Our examples are somewhat inconsistent about this, but one commonly used convention is to start commandline examples with a `$` prompt (for non-administrative commands) or with a `#` prompt (when the commandline has root-level—not just sudo—permissions; rare). There are some downsides (you have to remove the $ when you copy-paste commands) but this is one case where the difference would help convey some useful information. We should discuss making a note about this in our style guide and being consistent about it.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +85,234 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+
+      **Tip:** This example uses a `-j` parameter set to `4`, which uses four processes to build in parallel. The best number of processes to use depends on how many CPU cores your hardware has available. Use `sysctl -n hw.ncpu` to get your CPU core count.
+
+{{n.next()}}. Run unit tests built into the server executable. This could take about 5 minutes, depending on your hardware specs. (optional, but recommended) <!--{# ***TODO: We should provide info about what to do if you get failures. What should the user do? PM looking into this.*** #}-->
+
+        ./rippled --unittest
+
+{{n.next()}}. `rippled` requires the `rippled.cfg` configuration file to run. You can find an example config file, `rippled-example.cfg` in `rippled/cfg`. Make a copy and save it as `rippled.cfg` in a location that enables you to run `rippled` as a non-root user. Access the `rippled` directory and run:
+
+        mkdir -p /etc/opt/ripple ***TODO: changed this from ~/.config/ripple to match guidance coming in conf-file-location.md PR -- okay?***
```
mDuo13: Seems fine (assuming this works on macOS; I haven't tested that). You _can_ have the config file in your home dir and owned by a non-admin user, which has pros and cons and relates to what user you plan to run `rippled` as.

- Running as root / admin is easy to set up, since you probably don't have to worry about wrong permissions
    - If you're running `rippled` as a root user, it won't look in a non-root user's home dir, so that won't work.
- Running as not-root lets you limit the damage the server can do if something unexpectedly bad happens, like going berserk due to a bug or being maliciously controlled by some external person. We do our best to make these kinds of things almost impossible, but it's a sensible (and common) precaution to take when running any server that's connected to the internet at large.
    - Typically Linux  packages for servers are configured to automate the work to add a user specifically for that server, then give it permissions to _only_ the files it wants. This probably includes read-only permissions to its config file (which are owned by and writable only as root)
    - For dev purposes, you can run the server as your own user, which is slightly less setup but still more secure than running as root. This last case is the one where you want the config file in a home directory.

content/tutorials/manage-the-rippled-server/install-rippled.md
mDuo13: Let's:

1. Keep the scope of this task from growing to modify the other installs
2. Not add steps telling people to check / edit things they don't need to edit (if they're following the directions in the intended scenario).
    - To be clear, we may need to add these steps to the macOS build & run instructions, but let's not add them to the CentOS / Ubuntu installs from packages.
3. Instead of adding steps, let's just put in a brief note with a link to [the existing docs for rippled not being able to read a config file](https://developers.ripple.com/server-wont-start.html#failed-to-open-etcoptripplerippledcfg). 
4. Improve those directions in a later ticket if necessary.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +85,234 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+
+      * Set the `[database_path]` to the location where you want to store other database data. (This includes an SQLite database with configuration data, and is typically one level above the `[node_db]` path field.)
+
+      * Set the `[debug_logfile]` to a path where `rippled` can write logging information.
+
+      These are the only configurations required for `rippled` to start up successfully. All other configuration is optional and can be tweaked after you have a working server. For more information, see [Additional Configurations](#additional-configuration).
+
+{{n.next()}}. `rippled` requires the `validators.txt` file to run. You can find an example validators file, `validators-example.txt`, in `rippled/cfg/`. Make a copy and save it as `validators.txt` in the same folder as your `rippled.cfg` file. Access the `rippled` directory and run:
+
+        cp cfg/validators-example.txt /etc/opt/ripple ***TODO: changed this from ~/.config/ripple to match guidance coming in conf-file-location.md PR -- okay?***
```
mDuo13: This needs to change the filename of the file when it moves it, to remove the `-example` part:

```
cp cfg/validators-example.txt /etc/opt/ripple/validators.txt
```

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +85,234 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+
+      **Warning:** Ripple has designed a decentralization plan with maximum safety in mind. During the transition, you should not modify the `validators.txt` file except as recommended by Ripple. Even minor modifications to your validator settings could cause your server to diverge from the rest of the network and report out of date, incomplete, or inaccurate data. Acting on such data can cause you to lose money.
+
+{{n.next()}}. Access your build directory, `my_build` for example, and start the `rippled` service.
+
+        sudo ./rippled
```
mDuo13: This runs `rippled` as root—not recommended. If you set up file paths for your config and databases to directories you can write, then `rippled` should run without root permissions.

Caution: if you've ever started rippled as root (including using `sudo`) then it probably created some database files as a root user, which means it'll fail with a permissions error if you try to run it as non-root user, _even if the user has permissions to the directories containing the databases_. To fix this, you have to (using root permissions) delete `rippled`'s database files (in particular, the ones it created when it ran as an admin user, but also possibly ones it created when run as a non-root user before that, because they may be incomplete/corrupt if you delete some database files). The easiest way to do this is `sudo rm -r` the `rippled` database directory (assuming you don't have anything else in that directory) and then re-create it without sudo / admin, as the user you intend to run `rippled` with.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +85,234 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+For next steps, see [Postinstall](#postinstall) and [Additional Configuration](#additional-configuration).
 
-        $ sudo systemctl start rippled.service
 
 ## Postinstall
 
-It can take several minutes for `rippled` to sync with the rest of the network, during which time it outputs warnings about missing ledgers. After that, you have a fully functional stock `rippled` server that you can use for local signing and API access to the XRP Ledger.
+It can take several minutes for `rippled` to sync with the rest of the network, during which time it outputs warnings about missing ledgers.
+
+<!--{# ***TODO: for the future, provide a reference for what the most common and unintuitive messages mean -- esp when signaling possible errors.*** #}-->
```
mDuo13: Link this page: https://developers.ripple.com/understanding-log-messages.html

content/tutorials/manage-the-rippled-server/install-rippled.md
mDuo13: Options:

1. Tell the user to run rippled as their own user. This is a reasonable balance of safe and easy—not as safe as running as a dedicated user like `rippled`, not as easy as running as root, but in between. Change the example command to use `$USER` (the environment variable for the current user, assuming that's also valid on macOS like it is on Linux).
2. Tell the user to create a `rippled` user for safety and provide steps for (a) how to create such a user, (b) how to give it the right permissions, and (c) how to run `rippled` as that user. This is fairly standard system admin-level stuff, which may or may not be a bit too much work for the intended audience here. We're already telling them not to run on macOS in production so taking production-level precautions is probably too much here.

I'd lean towards 1 since it's more expedient. There are third-party docs they can leverage if they want to re-create a production-like environment on macOS despite our recommendations, and we probably shouldn't provide recommendations for things we don't support anyway.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +85,234 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+
+Changes to the `[node_db]`, `[debug_logfile]`, or `[database_path]` sections may require you to give the `rippled` user and group ownership to your new configured path: ***TODO: I added [node_db] to this list - okay?***
+
+    $ chown -R rippled:rippled <configured path> ***TODO: Is `rippled:rippled` just an example value? When I run this command on macOS, I get `chown: rippled: illegal group name.` When I run it on centos, I get chown: changing ownership of ... : Operation not permitted***
+
+Restart `rippled` for any configuration changes to take effect: ***TODO: when I try this on macOS, I get `sudo: service: command not found`. I don't think the `service` command exists for macOS. For macOS, can the user just stop (ctrl+c? Not sure if that is the proper way to stop the service) and start rippled (sudo ./rippled) instead? For centos and ubuntu, we use `sudo systemctl start rippled.service` to start the service - below should we use `sudo systemctl restart rippled.service` - just to be consistent?***
```
mDuo13: The `service` command is part of systemd, which is not part of macOS. Also, running `./rippled` or `sudo ./rippled` is not running it as a service.

So yeah, Ctrl-C would be one way to stop it if you did that. 

 I'm not sure what the macOS equivalent of "setting up a systemd unit to run it as a service" is.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +85,234 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+
+    $ sudo service rippled restart
+
+### Connect Your `rippled` to the XRP Test Net
+
+Ripple has created the [XRP Test Network](https://ripple.com/build/xrp-test-net/) to provide a testing platform for the XRP Ledger. XRP Test Net funds are not real funds and are intended for testing only. You may want to connect your `rippled` server to the XRP Test Net to test out and understand `rippled` functionality before connecting to the production XRP Ledger Network. ***TODO: stated correctly?***
```
mDuo13: :+1: 

It's also about testing out that your own code interacts correctly with `rippled`.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +85,234 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
+        # https://vl.ripple.com
+        #
+        # [validator_list_keys]
+        # ED2677ABFFD1B33AC6FBC3062B71F1E8397C1505E1C42C64D11AD1B28FF73F4734
 
-        $ chown -R rippled:rippled <configured path>
+3. Restart `rippled`.
 
-Restart `rippled` for any configuration changes to take effect:
+4. To verify that your `rippled` is connected to the XRP Test Net, go to the [XRP Test Net Faucet](https://developers.ripple.com/xrp-test-net-faucet.html) and click **Generate credentials**. Make a note of the **Address** value. Make the [`account_info`](https://developers.ripple.com/account_info.html) request to your `rippled` server. ***TODO: Any ideas for a better way to this check?***
```
mDuo13: This works. You can also run `server_info` and see what ledger version your server is on (assuming the state is `full`/`proposing` and it's not still syncing). If it's close to or the same as what `s1.ripple.com`'s server_info reports, then you're on the production network. If it's much smaller (like, fewer digits) then you're on the testnet.

content/tutorials/manage-the-rippled-server/install-rippled.md
```diff
@@ -46,59 +85,234 @@ This section assumes that you are using CentOS 7 or Red Hat Enterprise Linux 7.
...
 
-        $ sudo service rippled restart
+      If the request returns account information, your `rippled` is connected to the XRP Test Net. If the request cannot find the account, your `rippled` is not connected to the XRP Test Net. Check your `rippled.cfg` and `validators.txt` configurations.
```
mDuo13, line 327: Technically this could be fooled if someone funded your testnet credentials on prod. It's not likely to happen in the timescale described here but if you tried this same test much later after creating your testnet credentials, it could happen.

mDuo13 (changes requested):
This is shaping up nicely. I left some info and hopefully we can get this in good shape without dragging it out much more, especially since the more changes that happen here the more conflicts that have to be resolved with #471.

content/tutorials/manage-the-rippled-server/install-rippled.md
jhaaaa: Thank you so much for taking the time to explain this convention to me. 👍  My work life is changed forever. I've updated this doc accordingly. It would be great to include it in the style guide.

content/tutorials/manage-the-rippled-server/install-rippled.md
jhaaaa: Great to know that issuing the `./rippled` command is not running `rippled` as a service. 😅 For now, I'll go with Ctrl-C to stop and `./rippled` to start it up again. For the future, I will look into a command for starting `rippled` as a service for macOS.

content/tutorials/manage-the-rippled-server/install-rippled.md
jhaaaa: Do you have a 🔮 in which you can see what I'm doing on my command line? 😂  Deleted my db files and it ran without sudo. Removed `sudo` here.

jhaaaa:
@manojsdoshi - thank you for all of your feedback and for your approval!

@mDuo13 and @garay-daniel - thank you for your feedback and guidance as well! I believe this commit addresses all comments/feedback. No more "TODO"s to search for 😂. When you have a moment - please take a look and provide your approval, if worthy!

mDuo13 (approved):
:+1: The revisions look great!

jhaaaa:
The build is breaking because of a broken link to https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html. The link works when I access it directly and when I access it from my locally built doc.

PR #476, opened 2018-11-16
Improve documentation of rippled hardware requirements

jhaaaa:
- Improve documentation of `[node_db]` and `[node_size]` stanzas in `rippled.cfg`
- Differentiate between memory/RAM vs storage and their settings in this doc
- Working to reduce ambiguity and increase consistency in this doc
- Still working: Once we've gotten to a place where the technical content is accurate, may be able to improve doc structure/presentation to make pros/cons and choices clearer (i.e. bullets or a table for choosing between RocksDB and NuDB)

### DIFF
--- a/content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
+++ b/content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
@@ -23,12 +23,12 @@ Each `node_size` has a corresponding requirement for available RAM. For example,
 
 To tune your server, it may be useful to start with `tiny` and increase the size to `small`, `medium`, and so on as you refine the requirements for your use case.
 
-| `node_size` value | RAM available for `rippled` | Notes                      |
-|:------------------|:----------------------------|:---------------------------|
-| < 8GB             | `tiny`                      | Not recommended for testing or production servers. If not specified in `rippled.cfg`, `tiny` is the default value. |
-| 8GB               | `small`                     | Recommended for test servers. |
-| 16GB              | `medium`                    | Note that the `rippled-example.cfg` has its `node_size` value set to `medium`. |
-| 32GB              | `huge`                      | Recommended for production servers. |
+| RAM available for `rippled` | `node_size` value | Notes                      |
+|:----------------------------|:------------------|:---------------------------|
+| < 8GB                       | `tiny`            | Not recommended for testing or production servers. This is the default value if you don't specify a value in `rippled.cfg`. |
+| 8GB                         | `small`           | Recommended for test servers. |
+| 16GB                        | `medium`          | The `rippled-example.cfg` file uses this value. |
+| 32GB                        | `huge`            | Recommended for production servers. |
 
 Although `large` is also a legal value for `[node_size]`, in practice it performs worse than `huge` in most circumstances. Ripple recommends always using `huge` instead of `large`.
 
--- a/content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
+++ b/content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
@@ -1,23 +1,12 @@
 # Capacity Planning
 
-This section describes some of the challenges related to scaling `rippled` servers for testing and production deployments. Additionally, it describes how hardware and configuration settings relate to those challenges. Finally, this section offers recommendations to assist you in properly setting up `rippled` to meet the goals for the use case of your deployment.
-
-Even the most minimally functional `rippled` server must contain the most recently validated ledger versions to submit transactions to the network and verify the integrity of the XRP Ledger. Beyond these requirements, consider the following possible business needs:
-
-- Handling ever-increasing transaction volume
-- Servicing transaction reporting information to clients
-- Maintaining varying amounts of historical data
-
-To meet your `rippled` capacity requirements, you must address these technical factors:
-
-- The [configuration settings](#configuration-settings) that affect resource utilization
-- The [network and hardware](#network-and-hardware) requirements to achieve consistent, good performance across the XRP Ledger network
+This section describes configuration, network, and hardware recommendations that you can use to tune and optimize the performance of your `rippled` server. Being aware of these considerations can help you ensure that your `rippled` server is ready to handle XRP Ledger network capacity today and in the near future.
 
 
 
 ## Configuration Settings
 
-Ripple recommends using these configuration guidelines to optimize performance of your `rippled` server.
+Ripple recommends using these configuration guidelines to optimize resource utilization and performance of your `rippled` server.
 
 You can set the following parameters in the `rippled.cfg` file used for your `rippled` server. You can access an example configuration file, `rippled-example.cfg`, in the [`cfg` directory](https://github.com/ripple/rippled/blob/develop/cfg/rippled-example.cfg) in the `rippled` GitHub repo.
 
@@ -126,6 +115,8 @@ The example `rippled-example.cfg` file sets the logging verbosity to `warning` i
 ## Network and Hardware
 
 Each `rippled` server in the XRP Ledger network performs all of the transaction processing work of the network. Therefore, the baseline hardware for production `rippled` servers should be similar to that used in Ripple's [performance testing](https://ripple.com/dev-blog/demonstrably-scalable-blockchain/).
+
+Ensuring that your `rippled` server meets these network and hardware requirements helps achieve consistent, good performance across the XRP Ledger network.
 
 
 ### Recommendation
--- a/content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
+++ b/content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
@@ -24,90 +24,83 @@ You can set the following parameters in the `rippled.cfg` file used for your `ri
 
 ### Node Size
 
-Set the `node_size` parameter to define your `rippled` server's database cache size, as well as how much RAM you want the server to be able to use. Pick the appropriate size based on your server's expected load and available memory. ***TODO: correctly stated? The `tiny`, `huge`, etc values are setting the database cache size. When we list an Available RAM value in the table below - are we saying that this is how much RAM rippled will use? Or are we saying that if you select HUGE, you need to have at least 32GB available RAM for rippled?***
+Set the `node_size` based on your server's expected load and the amount of memory you can make available to `rippled`.
 
-The `node_size` you define determines how big of an object to treat as one "node" in the ledger's underlying tree structure. 
+The node size is talking about how big of an object to treat as one "node" in the ledger's underlying tree structure. It's not directly configuring storage or RAM, but it affects how much RAM your server uses and how fast your server can fetch things from disk.
 
-It doesn't directly configure storage or RAM, but it affects how much RAM your `rippled` server uses and how fast it can fetch things from disk. For example, a larger node size decreases disk I/O requirements, but increases memory requirements.  ***TODO: Thank you for your patient explanations. I am so close to understanding this. The current doc says that node_size defines the  But one more question - and I'm sorry if I'm making this more complicated than it is - but by "node," we usually mean an instance of a rippled server. But in this case, are we talking about "node" as a unit of information or data to be retrieved or stored, the size of which we are defining based on this setting? The current doc says that this value sets the size of the database cache.***
+For example, think of node size as the size of cardboard boxes you use to ship vases from a warehouse to your store. A larger cardboard box means that you can ship more vases at once with fewer trips from the warehouse. A large node size means that you can get more `rippled` transaction data, for example, with fewer trips to disk storage.
 
-Ripple recommends you always use the largest database cache your available memory can support. See the following table for recommended settings.
+However, with a larger box, you also need to include more packing material to keep the vases from breaking during shipping, which takes up precious space in the box. Likewise, with a larger node size, you can get a lot of transaction data, but the data may also contain `rippled` start up information and ledger reports, which takes up precious RAM that can't be used for the subset of data you're actually using.
+
+A part of tuning these configurations is about choosing the right node size so that the subset of data you're actually using can be retrieved in the fewest trips to disk storage, while also ensuring that all of it, including other data you aren't using right now, can fit in your available RAM.
+
+Ripple recommends you always use the largest node size your available RAM can support. See the following table for recommended settings.
 
 #### Recommendation
 
-| Available RAM for `rippled` | `node_size` value | Notes                      |
-|:----------------------------|:------------------|:---------------------------|
-| < 8GB                       | `tiny`            | Not recommended. If not specified in `rippled.cfg`, `tiny` is the default value. |
-| 8GB                         | `small`           | Recommended for test servers. |
-| 16GB                        | `medium`          | The example `rippled-example.cfg` has its `node_size` value set to `medium`. |
-| 32GB                        | `large`           | Ripple recommends using `huge` instead. `large` uses less memory than `huge`, but increases disk access requirements, which decreases performance. ***TODO: Included this row because `large` is an option in `rippled-example.cfg` and I think we need to acknowledge it here. Is the Available RAM value correct? We say that `large` uses less memory than `huge`, but the RAM values are the same.*** |
-| 32GB                        | `huge`            | Recommended for production servers. |
+Each `node_size` has a corresponding requirement for available RAM. For example, if you set `node_size` to `huge`, you should have at least 32GB of available RAM to help ensure that `rippled` can run smoothly.
+
+To tune your server, it may be useful to start with `tiny` and increase the size to `small`, `medium`, and so on as you refine the requirements for your use case.
+
+***TODO: Small thing, but I switched the order of the node size and RAM columns because the node size column contains the value that you are setting and should be the starting point of focus.***
+
+| `node_size` value | Available RAM for `rippled` | Notes                      |
+|:------------------|:----------------------------|:---------------------------|
+| `tiny`            | < 8GB                       | Not recommended for testing or production servers. If not specified in `rippled.cfg`, `tiny` is the default value. |
+| `small`           | 8GB                         | Recommended for test servers. |
+| `medium`          | 16GB                        | Note that the `rippled-example.cfg` has its `node_size` value set to `medium`. ***TODO: The doc also says: With a `node_size` of `medium`, a `rippled` server can be reasonably stable in a test Linux system with 16GB of RAM -- is our recommendation for test servers - small or medium? Or either?*** |
+| `large`           | 32GB                        | While `large` is a legal value, Ripple recommends just using `huge` instead. ***TODO: Per Mark, `large` uses less memory, but increases disk access requirements. Since disk is slower than memory, this reduces performance. This is a great clarification - but for both large and huge, we list 32GB RAM. If we want to include this info from Mark - do we need to reduce the RAM value we list for `large` for it to make sense?*** |
+| `huge`            | 32GB                        | Recommended for production servers. |
 
 For `node_size` troubleshooting, see [Bad node_size value](server-wont-start.html#bad-node-size-value).
 
 
 ### Node DB Type
 
-The `type` field in the `[node_db]` stanza of the `rippled.cfg` file sets the type of key-value store that `rippled` uses to hold ledger data on disk. This setting does not directly configure RAM settings, but the choice of key-value store has important implications for RAM usage because of the different ways these technologies cache and index data for fast lookup.
+The `type` field in the `[node_db]` stanza of the `rippled.cfg` file sets the type of key-value store that `rippled` uses to hold the ledger store on disk.
+
+This setting does not directly configure RAM settings, but the choice of key-value store has important implications for RAM usage because of the different ways these technologies cache and index data for fast lookup.
 
 You can set the value to either `RocksDB` or `NuDB`.
 
-#### RocksDB vs NuDB
+* If you are running a `rippled` server as a production validator, Ripple recommends that you use `RocksDB` with rotational disk storage. [Learn more](#more-about-using-rocksdb)
 
-In the example `rippled-example.cfg` file, the `type` field in the `[node_db]` stanza is set to `RocksDB`, which is optimized for spinning disks. RocksDB requires approximately one-third less disk storage than NuDB and provides better I/O latency. However, the better I/O latency comes as result of the large amount of RAM RocksDB requires to store data indexes. ***TODO: in this case, if you don't set a type value, rippled will not run. there is no default value set in the code.***
+* If you are running a `rippled` server for any other production use, Ripple recommends that you use `NuDB` with SSD storage. NuDB works with SSD storage only. [Learn more](#more-about-using-nudb)
 
-NuDB, on the other hand, has nearly constant performance and memory footprint regardless of the amount of data being [stored](#storage). NuDB _requires_ a solid-state drive, but uses much less RAM than RocksDB to access a large database.
+The example `rippled-example.cfg` file has the `type` field in the `[node_db]` stanza is set to `RocksDB`. ***TODO: Just an FYI that in this case, if you don't set a type value, rippled will not run. There is no default value coded into rippled.***
 
-Validators should be configured to use RocksDB and to store no more than about 300,000 ledgers (approximately two weeks' worth of [historical data](#historical-data) in the ledger store. For other production servers, Ripple recommends using NuDB, with an amount of historical data configured based on business needs. Machines with only spinning disks (not recommended) must use RocksDB.
+#### More About Using RocksDB
 
-***TODO: When we talk about "data indexes" - is this akin to the database cache? Meaning, while storage is used for the ledger store -- memory is used for the database cache (aka data indexes?)***
+[RocksDB](https://rocksdb.org/docs/getting-started.html) is an embeddable persistent key-value store that is optimized for rotational disks.
 
-***TODO: Based on my confusion above about the difference between storage and memory -- it would be great to standardize how we talk about what is being stored. Here we call what we are storing a "few days' worth of data", as well as something called the "ledger store". In the rippled.cfg, we refer to the "persistent datastore for rippled." In other areas of this doc and inputs to this doc, we talk about "data indexes," "persisting the XRP Ledger in the ledger store", a "database cache," and a "backend data store." Perhaps when we are talking about storage we can talk about the ledger store (or a few days' worth of ledger store data) and when we are talking about memory we can talk about the database cache? Any thoughts?***
+RocksDB requires approximately one-third less disk [storage](#storage) than NuDB and provides better I/O latency. However, the better I/O latency comes as result of the large amount of RAM RocksDB requires to store data indexes.
 
-***TODO: from rome:
+Validators should be configured to use RocksDB and to store no more than about 300,000 ledgers (approximately two weeks' worth of [historical data](#historical-data)) in the ledger store.
 
-I'm not entirely sure what you're advocating regarding the terminology, but I suspect your confusion may come from trying to separate the the persistent stores (on disk) from the transient storage (in RAM) that they require for indexes and cache. "Persist" is specifically referring to storing data for the long term (that is, on disk). The ledger store handles persisting, caching, and indexing the data to whatever extent the chosen tech is designed to do. To a large extent, the amount of RAM you need is directly correlated to how much you have stored on disk. (The size of your phone book depends on how many people are listed.)
+RocksDB has performance-related configuration options that you can set in `rippled.cfg` to achieve maximum transaction processing throughput. Here is the recommended configuration for a `rippled` server using RocksDB:
 
-I agree that there are a lot of almost interchangeable terms in use here like "ledger store", "key-value store", "node backend", "ledger database", etc. but I think untangling them and reducing the number of unique terms used may be beyond our control since those terms are also used as config options and API methods, and the names used in those places are really not that distinct or consistent. I'll try to give my breakdown of terms though (my preferred terms in bold):
-
-ledger store / node store / node DB / ledger DB - The (mandatory) thing that stores current and historical ledgers (well, their contents) in a continuous history as they're produced.
-
-shard store / shard DB - A place to store chunks of historical ledgers
-
-key-value store / node backend / key-value database - The technology that actually stores the data in the ledger store or shard store. The two current choices are RocksDB and NuDB. (In the past, rippled supported other storage technologies too.)***
-
-***TODO: from ryan: Saying "ledgers from the last few days" would probably add clarity.
-
-I think the basic breakdown is:
-
-"ledger store" -- Stores hashes of ledgers
-"persistent datastore for rippled" -- Data that's stored for a medium to long-term period. As opposed to ledger storage, which can be deleted periodically in the short-term (depending on use-cases)
-"data indexes" -- How RocksDB stores hashes
-"database cache" -- same thing as ledger store?
-"backend data store" -- same thing as ledger store?
-Perhaps when we are talking about storage we can talk about the ledger store (or a few days' worth of ledger store data) and when we are talking about memory we can talk about the database cache? Any thoughts?
-
-I think this makes sense. Instead of database cache, we might want to say "key-value store", but I might be confused as to the difference there.***
-
-***TODO: from rome: Database cache and indexes are different things. Both are involved in fetching records quickly.
-
-Index: A list of all the records you have, and possibly some details about them. For example, if you frequently have to look up people with a specific hair color, you might keep an index of people with brown hair in your yellow pages, so you can do that without calling every person in the book and asking them for their hair color. The problem is, the more indexes you keep, the more huge yellow books (indexes) you have occupying your office (RAM). Depending on how many indexes of how many records you keep, at some point you may have so many books in your office that you can't even fit enough many clients in your office to hold actual meetings. That's what happens if you use RocksDB with a big database and a small amount of RAM.
-
-Cache: A convenient place where you keep a subset of things you expect to access soon. There are many different kinds of caches but for these key-value stores, it's most likely stored in RAM. So cache may contain entire records/blocks/objects/whatever that you expect to need to fetch soon, and is probably an order of magnitude faster than going all the way to disk to fetch those things. The trick is correctly predicting which things you're going to access again soon—stuff you saw recently, stuff that's sequentially after or just nearby stuff you recently looked up, etc. Going back to our yellow pages analogy, cache is almost like a waiting room space right next to your office. Cache also takes up space (as I mentioned, probably in RAM), so you don't want to dedicate so many square feet to the waiting room that you can't actually invite many people into the office proper.
-
-RocksDB has a pretty thorough and efficient system for managing both of them. NuDB takes a different approach entirely: it says, "look, we have fast elevators and everyone lives pretty close to my office, so I'll just call them as I need them and keep as much office space as I can for actual work". If your elevators are actually slow (spinning disks) then you're going to spend a lot of time waiting around for people to show up in your office.***
-
-RocksDB has performance-related configuration options you can modify to achieve maximum transaction processing throughput. (NuDB does not have performance-related configuration options.) Here is an example of the recommended configuration for a `rippled` server using RocksDB:
+***TODO: reordered to match order in rippled-example.cfg***
 
 ```
 [node_db]
 type=RocksDB
+path={path_to_ledger_store}
 open_files=512
-file_size_mb=64
-file_size_mult=2
 filter_bits=12
 cache_mb=512
-path={path_to_ledger_store}
+file_size_mb=64
+file_size_mult=2
 ```
+
+#### More About Using NuDb
+
+[NuDB](https://github.com/vinniefalco/nudb#introduction) is an append-only key-value store that is optimized for SSD drives.
+
+NuDB has nearly constant performance and memory footprints regardless of the amount of data being [stored](#storage). NuDB _requires_ a solid-state drive, but uses much less RAM than RocksDB to access a large database.
+
+Non-validator production servers should be configured to use NuDB and to store the amount of historical data required for the use case.
+
+NuDB does not have performance-related configuration options available in `rippled.cfg`.
 
 #### History Sharding
 
@@ -118,13 +111,13 @@ path={path_to_ledger_store}
 
 The amount of historical data that a `rippled` server keeps online is a major contributor to required storage space. At the time of writing (2018-10-29), a `rippled` server stores about 12GB of data per day and requires 8.4TB to store the full history of the XRP Ledger. You can expect this amount to grow as transaction volume increases across the XRP Ledger network. You can control how much data you keep with the `online_delete` and `advisory_delete` fields.
 
-Online deletion enables purging of `rippled` ledgers from databases without any disruption of service. It removes only records that are not part of the current ledgers. ***TODO: what do we mean by "current ledgers"?*** Without online deletion, those databases grow without bounds. Freeing disk space requires stopping the process and manually removing database files. For more information, see [`[node_db]`: `online_delete`](https://github.com/ripple/rippled/blob/develop/cfg/rippled-example.cfg#L832).
+Online deletion enables the purging of `rippled` ledgers from databases without any disruption of service. It removes only records that are not part of the current ledgers. ***TODO: what do we mean by "current ledgers"?*** Without online deletion, those databases grow without bounds. Freeing disk space requires stopping the process and manually removing database files. For more information, see [`[node_db]`: `online_delete`](https://github.com/ripple/rippled/blob/develop/cfg/rippled-example.cfg#L832).
 
 <!-- {# ***TODO***: Add link to online_delete section, when complete, per https://ripplelabs.atlassian.net/browse/DOC-1313  #} -->
 
 ### Log Level
 
-The default `rippled.cfg` file sets the logging verbosity to `warning` in the `[rpc_startup]` stanza. This setting greatly reduces disk space and I/O requirements over more verbose logging. However, more verbose logging provides increased visibility for troubleshooting.
+The example `rippled-example.cfg` file sets the logging verbosity to `warning` in the `[rpc_startup]` stanza. This setting greatly reduces disk space and I/O requirements over more verbose logging. However, more verbose logging provides increased visibility for troubleshooting.
 
 **Caution:** If you omit the `log_level` command from the `[rpc_startup]` stanza, `rippled` writes logs to disk at the `debug` level and outputs `warning` level logs to the console. `debug` level logging requires several more GB of disk space per day than `warning` level, depending on transaction volumes and client activity.
 
@@ -153,12 +146,22 @@ You'll get the best performance on bare metal, but virtual machines can perform 
 
 #### Storage
 
-Ripple recommends estimating storage sizing at roughly 12GB per day of data kept online with NuDB. RocksDB requires around 8GB per day. However, the data per day changes with activity in the network. You should provision extra capacity to prepare for future growth. At the time of writing (2018-10-29), a server with all XRP Ledger history requires 8.4TB.
+Here are some estimated `rippled` storage requirements:
+
+- RocksDB stores around 8GB per day
+- NuDB stores around 12GB per day
+
+The amount of data stored per day changes with activity in the network.
+
+You should provision extra storage capacity to prepare for future growth. At the time of writing (2018-10-29), a `rippled` server storing the full history of the XRP Ledger required 8.4TB.
 
 <!-- {# ***TODO: Update the dated storage consideration above, as needed. ***#} -->
 <!-- {# ***TODO: DOC-1331 tracks: Create historic metrics that a user can use to derive what will be required. For ex, a chart with 1TB in 2014, 3TB in 2015, 7TB in 2018 ***#} -->
 
-SSD storage should support several thousand of both read and write IOPS. The maximum reads and writes per second that Ripple engineers have observed are over 10,000 reads per second (in heavily-used public server clusters), and over 7,000 writes per second (in dedicated performance testing).
+SSD storage should support several thousand of both read and write IOPS. Ripple engineers observed the following maximum reads and writes per second:
+
+- Over 10,000 reads per second (in heavily-used public server clusters)
+- Over 7,000 writes per second (in dedicated performance testing)
 
 ##### Amazon Web Services
 
@@ -170,10 +173,16 @@ AWS instance stores (`ephemeral` storage) do not have these constraints. Therefo
 
 #### RAM/Memory
 
-Memory requirements are mainly a function of the `node_size` configuration setting and the amount of client traffic retrieving historical data. As mentioned, production servers should maximize performance and set this parameter to `huge`.
-
-You can set the `node_size` parameter lower to use less memory, but you should only do this for testing. With a `node_size` of `medium`, a `rippled` server can be reasonably stable in a test Linux system with 16GB of RAM.
+Memory requirements are mainly a function of the `node_size` configuration setting and the amount of client traffic retrieving historical data. For more information about memory requirements, see [Node Size](#node-size).
 
 #### Network
 
-Any enterprise or carrier-class data center should have substantial network bandwidth to support running `rippled` servers. The minimum requirements are roughly 2Mbps transmit and 2Mbps receive for current transaction volumes. However, these can burst up to 100MBps transmissions when serving historical ledger and transaction reports. When a `rippled` server initially starts up, it can burst to over 20Mbps receive.
+Any enterprise or carrier-class data center should have substantial network bandwidth to support running `rippled` servers.
+
+Here are examples of observed network bandwidth use for common `rippled` tasks:
+
+| Task                                            | Transmit/Receive           |
+|:------------------------------------------------|:---------------------------|
+| Process current transaction volumes             | 2Mbps transmit, 2 Mbps receive |
+| Serve historical ledger and transaction reports | 100Mbps transmit           |
+| Start up `rippled`                              | 20Mbps receive             |
--- a/content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
+++ b/content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
@@ -15,14 +15,6 @@ You can set the following parameters in the `rippled.cfg` file used for your `ri
 
 Set the `node_size` based on your server's expected load and the amount of memory you can make available to `rippled`.
 
-The node size is talking about how big of an object to treat as one "node" in the ledger's underlying tree structure. It's not directly configuring storage or RAM, but it affects how much RAM your server uses and how fast your server can fetch things from disk.
-
-For example, think of node size as the size of cardboard boxes you use to ship vases from a warehouse to your store. A larger cardboard box means that you can ship more vases at once with fewer trips from the warehouse. A large node size means that you can get more `rippled` transaction data, for example, with fewer trips to disk storage.
-
-However, with a larger box, you also need to include more packing material to keep the vases from breaking during shipping, which takes up precious space in the box. Likewise, with a larger node size, you can get a lot of transaction data, but the data may also contain `rippled` start up information and ledger reports, which takes up precious RAM that can't be used for the subset of data you're actually using.
-
-A part of tuning these configurations is about choosing the right node size so that the subset of data you're actually using can be retrieved in the fewest trips to disk storage, while also ensuring that all of it, including other data you aren't using right now, can fit in your available RAM.
-
 Ripple recommends you always use the largest node size your available RAM can support. See the following table for recommended settings.
 
 #### Recommendation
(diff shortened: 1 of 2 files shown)


jhaaaa:
@mDuo13 and @ryangyoung - here's a first pass at some improvements. Please search on "TODO" to find my comments/questions. Thank you for any feedback and guidance you have time to provide!

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -13,40 +13,48 @@ To meet your `rippled` capacity requirements, you must address these technical f
...
-| 16GB                        | `medium`          |                                    |
-| 32GB                        | `huge`            | Recommended for production servers |
+| Available RAM for `rippled` | `node_size` value | Notes                      |
+|:----------------------------|:------------------|:---------------------------|
+| < 8GB                       | `tiny`            | If not specified, this is the default value. Not recommended. The delivered `rippled-example.cfg` has this value set to `medium`. ***TODO: In rippled-example.cfg, we recommend starting with the default value (tiny) and working upward, if necessary. Should we remove this "Not recommended." text here because we are recommending `tiny` as a starting value? I added the info about the delivered rippled-example.cfg file setting the configuration to `medium` so folks understand that if they don't change anything - they will get `medium`. Usually, when you hear that something is a default value, you assume that is what you'll get if you don't change anything. In this case, you'll need to change the delivered value (medium) to get the default value (tiny).*** |
```
mDuo13: This is partly just my preference, but the term "delivered config file" feels awkward. I find myself picturing a physical package delivery or something separate after the initial install process. Elsewhere I've used the terms "default config file" or "example config file"—let's stick with that if you're OK with it.

mDuo13: Also: I think the text in the config file is outdated and you probably won't have `tiny` work well on the production XRP Ledger in any circumstances nowadays. Maybe on a testnet or something tiny is fine.

I'd lean towards leaving the notes simple and easy to understand: "Not recommended" is pretty clear. Even though it leaves out some nuance, that's probably OK here?

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -13,40 +13,48 @@ To meet your `rippled` capacity requirements, you must address these technical f
...
+| 8GB                         | `small`           | Recommended for test servers.   |
+| 16GB                        | `medium`          |                            |
+| 32GB                        | `large` or `huge` | Recommended for production servers. ***TODO: Do we want to say anything about the difference between `large` and `huge`? Either can be set with an available RAM of 32GB -- how are they different? Later in the doc we say we recommend `huge` for production servers.*** |
```
mDuo13: I think the previous recommendation—`huge` only—is correct. I'm not 100% sure, but I suspect that our experience has shown `huge` to work better than `large` for 32GB RAM, so we can keep it simple again here.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -13,40 +13,48 @@ To meet your `rippled` capacity requirements, you must address these technical f
...
 
+`rippled` offers a history sharding feature that allows you to store a randomized range of ledgers in a separate shard store. You may want to configure the shard store to use a different type of key-value database than the one you defined for the ledger store using the `[node_db]` stanza. For more information about how to use this feature, see [History Sharding](https://ripple.com/build/history-sharding/#shard-store-configuration).
```
mDuo13: Let's change the "History Sharding" link to point to `history-sharding.html` instead of the `https://ripple.com/build/...` thing.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -13,40 +13,48 @@ To meet your `rippled` capacity requirements, you must address these technical f
...
+The `type` field in the `[node_db]` stanza of the `rippled.cfg` file sets the type of key-value database (or key-value store) that `rippled` uses to persist the XRP Ledger in the ledger store. ***TODO: what do we mean by persisting "the XRP Ledger in the ledger store"? By calling it the XRP Ledger it makes it sound like you are storing a full-history, which is not always the case. This content doesn't explicitly mention that this configuration also applies to the database cache (memory) - should it? Sorry if this is a dumb question - but this DB selection impacts storage, as well as RAM (memory), is that right?*** You can set the value to either `RocksDB` or `NuDB`. ***TODO: Updated the example values to the values I see in rippled-example.cfg - just for consistency and faster recognition - though the config does say these are case-insensitive.***
```
mDuo13: > the type of key-value database (or key-value store)

Can we change this back, too? While this is more thorough, it makes it a little harder to read.

> ***TODO: what do we mean by persisting "the XRP Ledger in the ledger store"? By calling it the XRP Ledger it makes it sound like you are storing a full-history, which is not always the case. This content doesn't explicitly mention that this configuration also applies to the database cache (memory) - should it? Sorry if this is a dumb question - but this DB selection impacts storage, as well as RAM (memory), is that right?***

This configuration chooses a technology for the key-value store that holds ledger data on disk. It does not _directly_ configure RAM settings, but the choice of k-v store has important implications for RAM usage because of the different ways those technologies cache and index data for fast lookup.

> ***TODO: Updated the example values to the values I see in rippled-example.cfg - just for consistency and faster recognition - though the config does say these are case-insensitive.***

:+1:

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -13,40 +13,48 @@ To meet your `rippled` capacity requirements, you must address these technical f
...
+The default backend data store is RocksDB, which is optimized for spinning disks. ***TODO: I need to test this but by default do we mean that if you don't set a node_db value, does it use RocksDB? Or by default do we mean that this is the value set in rippled-example.cfg? Just making sure.*** RocksDB requires approximately one-third less disk storage than NuDB and provides better I/O latency. However, the better I/O latency comes as result of the large amount of RAM RocksDB requires to store data indexes. ***TODO: I edited this section to clear up some ambiguity that was confusing to me. I hope that what I've come up with is still accurate? Can we change "default backend data store" to "default key-value database type"? I just want to correlate the node_db stanza name with how we refer to it here in the doc. When we talk about "data indexes" - is this akin to the database cache? Meaning, while storage is used for the ledger store -- memory is used for the database cache (aka data indexes?)***
```
mDuo13: Database cache and indexes are different things. Both are involved in fetching records quickly.

- *Index*: A list of all the records you have, and possibly some details about them. For example, if you frequently have to look up people with a specific hair color, you might keep an index of people with brown hair in your yellow pages, so you can do that without calling every person in the book and asking them for their hair color. The problem is, the more indexes you keep, the more huge yellow books (indexes) you have occupying your office (RAM). Depending on how many indexes of how many records you keep, at some point you may have so many books in your office that you can't even fit enough many clients in your office to hold actual meetings. That's what happens if you use RocksDB with a big database and a small amount of RAM.

- *Cache*: A convenient place where you keep a subset of things you expect to access soon. There are many different kinds of caches but for these key-value stores, it's most likely stored in RAM. So cache may contain entire records/blocks/objects/whatever that you expect to need to fetch soon, and is probably an order of magnitude faster than going all the way to disk to fetch those things. The trick is correctly predicting which things you're going to access again soon—stuff you saw recently, stuff that's sequentially after or just nearby stuff you recently looked up, etc. Going back to our yellow pages analogy, cache is almost like a waiting room space right next to your office. Cache also takes up space (as I mentioned, probably in RAM), so you don't want to dedicate so many square feet to the waiting room that you can't actually invite many people into the office proper.

RocksDB has a pretty thorough and efficient system for managing both of them. NuDB takes a different approach entirely: it says, "look, we have fast elevators and everyone lives pretty close to my office, so I'll just call them as I need them and keep as much office space as I can for actual work". If your elevators are actually slow (spinning disks) then you're going to spend a lot of time waiting around for people to show up in your office.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -13,40 +13,48 @@ To meet your `rippled` capacity requirements, you must address these technical f
...
+
+Ripple recommends using RocksDB for validators. `rippled` servers that operate as validators should keep only a few days' worth of [historical data](#historical-data) or less. For all other uses, Ripple recommends using NuDB.
+
+***TODO: Based on my confusion above about the difference between storage and memory -- it would be great to standardize how we talk about what is being stored. Here we call what we are storing a "few days' worth of data", as well as something called the "ledger store". In the rippled.cfg, we refer to the "persistent datastore for rippled." In other areas of this doc and inputs to this doc, we talk about "data indexes," "persisting the XRP Ledger in the ledger store", a "database cache," and a "backend data store." Perhaps when we are talking about storage we can talk about the ledger store (or a few days' worth of ledger store data) and when we are talking about memory we can talk about the database cache? Any thoughts?***
```
mDuo13: "A few days' worth of data" is referring to how many ledgers you keep in the ledger store. So maybe we should rephrase this:

> Validators should be configured to use RocksDB and to store no more than about 300,000 ledgers (approximately two weeks' worth of [historical data] in the ledger store. For other production servers, Ripple recommends using NuDB, with an amount of historical data configured based on business needs. Machines with only spinning disks (not recommended) must use RocksDB.

I'm not entirely sure what you're advocating regarding the terminology, but I suspect your confusion may come from trying to separate the the persistent stores (on disk) from the transient storage (in RAM) that they require for indexes and cache. "Persist" is specifically referring to storing data for the long term (that is, on disk). The ledger store handles persisting, caching, _and_ indexing the data to whatever extent the chosen tech is designed to do. To a large extent, the amount of RAM you need is directly correlated to how much you have stored on disk. (The size of your phone book depends on how many people are listed.)

I agree that there are a lot of almost interchangeable terms in use here like "ledger store", "key-value store", "node backend", "ledger database", etc. but I think untangling them and reducing the number of unique terms used may be beyond our control since those terms are also used as config options and API methods, and the names used in those places are really not that distinct or consistent. I'll try to give my breakdown of terms though (my preferred terms in bold):

- *ledger store* / node store / node DB / ledger DB - The (mandatory) thing that stores current and historical ledgers (well, their contents) in a continuous history as they're produced.

- *shard store* / shard DB - A place to store chunks of historical ledgers

- key-value store / node backend / key-value database - The technology that actually stores the data in the ledger store or shard store. The two current choices are RocksDB and NuDB. (In the past, rippled supported other storage technologies too.)

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -59,42 +67,40 @@ path={path_to_ledger_store}
...
 
-Online deletion enables pruning of `rippled` ledgers from databases without any disruption of service. It only removes records that are not part of the current ledgers. Without online deletion, those databases grow without bounds. Freeing disk space requires stopping the process and manually removing database files.
+Online deletion enables pruning of `rippled` ledgers from databases without any disruption of service. It removes only records that are not part of the current ledgers. Without online deletion, those databases grow without bounds. Freeing disk space requires stopping the process and manually removing database files. For more information, see [`[node_db]`: `online_delete`](https://github.com/ripple/rippled/blob/develop/cfg/rippled-example.cfg#L832).
```
mDuo13: > Freeing disk space requires stopping the process and manually removing database files.

I want more clarification on this, actually. Does this mean that your disk usage won't go down even with online delete enabled, but `rippled` can potentially reuse disk space for the deleted files to prevent it from going up further? Or does this only refer to the case where you're not using online delete?

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -59,42 +67,40 @@ path={path_to_ledger_store}
...
+
 
 ### Recommendation
 
 For best performance in enterprise production environments, Ripple recommends running `rippled` on bare metal with the following characteristics:
 
 - Operating System: Ubuntu 16.04+
 - CPU: Intel Xeon 3+ GHz processor with 4 cores and hyperthreading enabled
-- Disk: SSD
+- Disk: SSD ***TODO: instead of "Disk" - refer to "Storage" instead?***
```
mDuo13: No, leave it as "Disk: SSD" here. Maybe adding some recommended specs on the disk performance (as I've done in the "System Requirements" page recently) but absolutely when talking about hardware here it should be a straightforward hardware recommendation.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -59,42 +67,40 @@ path={path_to_ledger_store}
...
 - RAM:
 	- For testing: 8GB+
 	- For production: 32GB
 - Network: Enterprise data center network with a gigabit network interface on the host
 
-#### SSD Storage
-
-SSD storage should support several thousand of both read and write IOPS. The maximum reads and writes per second that Ripple engineers have observed are over 10,000 reads per second (in heavily-used public server clusters), and over 7,000 writes per second (in dedicated performance testing).
+***TODO: reordered the sections below to match the order of the bullets above, just for ease of tracking for the writer and the reader.***
```
mDuo13: :+1:

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -59,42 +67,40 @@ path={path_to_ledger_store}
...
 
 #### CPU Utilization and Virtualization
-Ripple performance engineering has determined that bare metal servers achieve maximum throughput. However, it is likely that hypervisors cause minimal degradation in performance.
-
-#### Network
 
-Any enterprise or carrier-class data center should have substantial network bandwidth to support running `rippled` servers. The minimum requirements are roughly 2Mbps transmit and 2Mbps receive for current transaction volumes. However, these can burst up to 100MBps transmissions when serving historical ledger and transaction reports. When a `rippled` server initially starts up, it can burst to over 20Mbps receive.
+Ripple performance engineering has determined that bare metal servers achieve maximum throughput. However, it is likely that hypervisors cause minimal degradation in performance. ***TODO: What are we saying about hypervisors here? Are we saying that you can run `rippled` on a virtual machine managed by a hypervisor with minimal degradation?***
```
mDuo13: A hypervisor is, by definition, the thing that manages a virtual machine. So "a virtual machine managed by a hypervisor" is redundant.

I think this part means, "you'll get best performance on bare metal, but virtual machines are close enough to be OK as long as the host hardware has high enough specs"

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -103,15 +109,24 @@ Ripple recommends estimating storage sizing at roughly 12GB per day of data kept
...
 
-AWS instance stores (`ephemeral` storage) do not have these constraints. Therefore, Ripple recommends deploying `rippled` servers with host types such as `M3` that have instance storage. The `database_path` and `node_db` path should each reside on instance storage.
+AWS instance stores (`ephemeral` storage) do not have these constraints. ***TODO: AWS EC2 instance stores?*** Therefore, Ripple recommends deploying `rippled` servers with host types such as `M3` that have instance storage. The `database_path` and `node_db` path should each reside on instance storage.
 
 **Caution:** AWS instance storage is not guaranteed to provide durability in the event of hard drive failure. Further, data that is lost when the instance stops and restarts (but not when just rebooted). This loss can be acceptable for a `rippled` server because an individual server can usually re-acquire that data from its peer servers.
```
mDuo13: Just noticed a typo in this sentence:

> Furthermore, data <s>that</s> is lost when the instance stops and restarts (but not when just rebooted).

Can probably tightened up to be active voice, too.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -103,15 +109,24 @@ Ripple recommends estimating storage sizing at roughly 12GB per day of data kept
...
 
-#### Memory
+SSD storage should support several thousand of both read and write IOPS. The maximum reads and writes per second that Ripple engineers have observed are over 10,000 reads per second (in heavily-used public server clusters), and over 7,000 writes per second (in dedicated performance testing).
 
-Memory requirements are mainly a function of the `node_size` configuration setting and the amount of client traffic retrieving historical data. As mentioned, production servers should maximize performance and set this parameter to `huge`.
-You can set the `node_size` parameter lower to use less memory, but you should only do this for testing. With a `node_size` of `medium`, a `rippled` server can be reasonably stable in a test Linux system with as little as 8GB of RAM.
+##### Amazon Web Services
 
-#### Amazon Web Services
+***TODO: I moved this to the Storage section so you see this info about AWS in the most relevant context. Is this okay?***
```
mDuo13: :+1:

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -103,15 +109,24 @@ Ripple recommends estimating storage sizing at roughly 12GB per day of data kept
...
 
-Amazon Web Services (AWS) is a popular virtualized hosting environment. You can run rippled in AWS, but Ripple does not recommend using Elastic Block Storage (EBS). Elastic Block Storage's maximum number of IOPS (5,000) is insufficient for `rippled`'s heaviest loads, despite being very expensive.
+Amazon Web Services (AWS) is a popular virtualized hosting environment. You can run `rippled` in AWS, but Ripple does not recommend using Elastic Block Storage (EBS). ***TODO: When we say "run rippled in AWS" - do we mean AWS EC2?***  Elastic Block Storage's maximum number of IOPS (5,000) is insufficient for `rippled`'s heaviest loads, despite being very expensive.
```
mDuo13: Yes, but I don't think the clarification is strictly necessary here.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -103,15 +109,24 @@ Ripple recommends estimating storage sizing at roughly 12GB per day of data kept
...
 
-AWS instance stores (`ephemeral` storage) do not have these constraints. Therefore, Ripple recommends deploying `rippled` servers with host types such as `M3` that have instance storage. The `database_path` and `node_db` path should each reside on instance storage.
+AWS instance stores (`ephemeral` storage) do not have these constraints. ***TODO: AWS EC2 instance stores?*** Therefore, Ripple recommends deploying `rippled` servers with host types such as `M3` that have instance storage. The `database_path` and `node_db` path should each reside on instance storage.
```
mDuo13: ¯\_(ツ)_/¯

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -103,15 +109,24 @@ Ripple recommends estimating storage sizing at roughly 12GB per day of data kept
...
 
 **Caution:** AWS instance storage is not guaranteed to provide durability in the event of hard drive failure. Further, data that is lost when the instance stops and restarts (but not when just rebooted). This loss can be acceptable for a `rippled` server because an individual server can usually re-acquire that data from its peer servers.
+
+#### RAM/Memory
+
+Memory requirements are mainly a function of the `node_size` configuration setting and the amount of client traffic retrieving historical data. As mentioned, production servers should maximize performance and set this parameter to `huge`. ***TODO: In Node Size - Recommendation section above, we recommend `large` or `huge` for production. Here we say that production servers should use huge.***
```
mDuo13: Let's stick with just `huge` in both places.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -103,15 +109,24 @@ Ripple recommends estimating storage sizing at roughly 12GB per day of data kept
...
+
+You can set the `node_size` parameter lower to use less memory, but you should only do this for testing. With a `node_size` of `medium`, a `rippled` server can be reasonably stable in a test Linux system with as little as 8GB of RAM. ***TODO: In the Node Size - Recommendation section above, we match a node size of `medium` with 16GB of RAM. Here we match `medium` with 8GB of RAM. But the text here seems to be suggesting something more like a node size of `low` with 16GB of RAM, where you have more memory, but are using the node_size to use less memory than you have. How can we resolve what seem to me to be inconsistencies?***
```
mDuo13: The node size recommendations are estimates and guidelines, so it's not that surprising that two recommendations are slightly off from one another. I don't know if anyone really knows whether `medium` or `small` is better overall when you have 16GB or even 8GB of RAM. It probably depends at least in part on what you use your server to do, and on what transactions the network processes, which is pretty unpredictable.

That said, we should probably just match the recommendations here with the table.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
ryangyoung: The example config itself is confusing, as you note. For `[node_size]`, it says that the default is `tiny` [here](https://github.com/ripple/rippled/blob/develop/cfg/rippled-example.cfg#L586). But on [lines 1122-3](https://github.com/ripple/rippled/blob/develop/cfg/rippled-example.cfg#L1122), it specifies `medium`. So that needs to be fixed as a separate issue since `tiny` seems to be the default only in the sense that the comments in the example config say it is. 

I'd vote for not re-stating that here--**especially because the first part of the description seems to be incorrect.** Rather, I propose that we change the description for `tiny` to something like:

> The smallest allowable size.  Not recommended for testing or production servers. To tune your server, it may be useful to start with this value and increase the size to `small`, `medium` and so on as you refine the requirements for your use case. 

Along with that, I suggest changing the description for `medium` to something like:

> The default value provided in the [`rippled-example.cfg` file](https://github.com/ripple/rippled/blob/develop/cfg/rippled-example.cfg#L1122). If you don't specify a different value, `rippled` runs `medium` (~16GB) RAM.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
ryangyoung: > the difference between `large` and `huge`

Sounds like the question is: is there a difference? Depending on the answer to that, we should probably just be consistent in the recommendations. For example, in the other part of the doc, we can say Ripple recommends `large` or `huge`.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -13,40 +13,48 @@ To meet your `rippled` capacity requirements, you must address these technical f
...
...ce. ***TODO: Edit stated correctly? What does "use at once" mean? The amount of RAM it can use to make a single read or write call? Do we need to say "at once"? For a single rippled server, are there multiple database caches - or just a single database cache? When we talk about Node Size, we are talking about configuring memory, and not storage, correct? Just clarifying for myself bc the two seem to get a little intertwined for me in this doc. For example, when we talk about a database cache - this is stored in RAM? But when we talk about the ledger store - that is stored in "storage," is that right? What is in the database cache vs the ledger store? The available RAM configured below has nothing to do with hardware storage, correct?*** Larger database caches decrease disk I/O requirements at a cost of higher memory requirements. ***TODO: Does it matter that SSDs don't have disks? Throughout, should we just refer to "storage" instead of "disks" or "disk storage"? For example, here we would refer to "storage I/O" instead? Does it matter?*** Ripple recommends you always use the largest database cache your available memory can support. See the following table for recommended settings.
```
ryangyoung: Might be worth adding a link to the bad `node_size` troubleshooting item somewhere in this section or the next: https://developers.ripple.com/server-wont-start.html#bad-node-size-value

ryangyoung: > When we talk about Node Size, we are talking about configuring memory, and not storage, correct? 

Looks like that's correct but I'm far, far from an expert. The example config file says that the setting "Tunes the servers based on the expected load and available memory," which sounds like what you've described. 

Based on my dim understanding, you configure storage settings in the `[node_db]` stanza, with `cache_mb` and similar fields. 

> when we talk about a database cache - this is stored in RAM?

My guess (best I can do) is that the db cache is stored in your selected DB. 

> But when we talk about the ledger store - that is stored in "storage," is that right?

Yes, correct, with the caveat that this the DB cache is also stored in storage. Again just a best guess.

> What is in the database cache vs the ledger store?

It looks like the DB cache stores "transaction metadata, account states, and ledger headers." 

The "ledger store" probably refers to the number of ledgers that `rippled` stores before they get deleted. @mDuo13 undoubtedly knows more about this, especially since he's working on updating the `online_delete` docs. A bit more tangential info on that subject [here](https://developers.ripple.com/can_delete.html) (`can_delete` request) and [here](https://developers.ripple.com/server-wont-start.html#online-delete-is-less-than-ledger-history) (troubleshooting item related to `online_delete`).

> The available RAM configured below has nothing to do with hardware storage, correct?

Correct (though take this with a grain of salt, as with my other answers).

> Throughout, should we just refer to "storage" instead of "disks" or "disk storage"

I'd vote for using "storage" consistently instead of the other terms.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
ryangyoung: >  type of key-value database (or key-value store)

The docs for RocksDB and NuDB both describe them as a "key value store" (or "key/value store"). So I'd vote for using that description and not database. 

> what do we mean by persisting "the XRP Ledger in the ledger store"

According to the NuDB intro, "The most common application for NuDB is content addressable storage where a cryptographic digest of the data is used as the key."

Based on that, the idea here seems to be: the data store is _only_ a collection of key-value pairs, where the key might be something like `ledger` and the value is the cryptographic hash of the ledger. 

_If_ that's anywhere near right, then you could probably rephrase that to something like:

> In the [node_db]` stanza of the `rippled.cfg` file, the `type` field specifies the type of key-value store that `rippled` uses. A hash of each ledger, related data, is stored in the specified key-value store. 

In other words, I agree that we should not say "XRP Ledger" here since we're talking about individual ledgers, not the entire distributed Ledger.

> but this DB selection impacts storage, as well as RAM (memory), is that right?

Seems like it only affects RAM in terms of how much would be used to look up data from the data store. So not really but using one or the other might impact RAM a little bit. Not sure if we have those metrics. 

> Updated the example values to the values I see in rippled-example.cfg - just for consistency and faster recognition - though the config does say these are case-insensitive.

👍 

The `type` field in the `[node_db]` stanza of the `rippled.cfg` file sets the type of key-value database (or key-value store) that `rippled` uses to persist the XRP Ledger in the ledger store.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
ryangyoung: Suggest adding a new header for this bit about sharding, since it's a separate (but related) subject. 

Also, consider adding a reference to the `[shard_db]` stanza to clarify that you can use the same or different key-value stores for sharding and ledgers.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
ryangyoung: Suggest adding a brief description of each DB (e.g., "RocksDB is...", "NuDB is....") here and adding links to the respective docs. Perhaps also add sub-headings for each type. 

* [RocksDB](https://rocksdb.org/docs/getting-started.html)
* [NuDB](https://github.com/vinniefalco/nudb#introduction)

> if you don't set a node_db value, does it use RocksDB? Or by default do we mean that this is the value set in rippled-example.cfg? Just making sure.

"Default" in this doc seems to mean "what's defined in the config."  I'd say, use that as the working understanding.

> Can we change "default backend data store" to "default key-value database type"?

I'd vote for using "key-value store" to reflect the docs from the two DB types. 

> When we talk about "data indexes" - is this akin to the database cache? Meaning, while storage is used for the ledger store -- memory is used for the database cache (aka data indexes?)

From the [RocksDB docs](https://github.com/facebook/rocksdb/wiki/Data-Block-Hash-Index): 

> RocksDB does a binary search when performing point lookup the keys in data blocks. However, in order to find the right location where the key may reside, multiple key parsing and comparison are needed. Each binary search branching triggers CPU cache miss, causing much CPU utilization. We have seen that this binary search takes up considerable CPU in production use-cases.

It looks like this is the main difference between RocksDB and NuDB, based on the next paragraph.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
ryangyoung: > "few days' worth of data"

Saying "ledgers from the last few days" would probably add clarity. 

I think the basic breakdown is:

* "ledger store" -- Stores hashes of ledgers
* "persistent datastore for rippled" -- Data that's stored for a medium to long-term period. As opposed to ledger storage, which can be deleted periodically in the short-term (depending on use-cases)
* "data indexes" -- How RocksDB stores hashes
* "database cache" -- same thing as ledger store?
* "backend data store" -- same thing as ledger store?

> Perhaps when we are talking about storage we can talk about the ledger store (or a few days' worth of ledger store data) and when we are talking about memory we can talk about the database cache? Any thoughts?

I think this makes sense. Instead of database cache, we might want to say "key-value store", but I might be confused as to the difference there.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
ryangyoung: > It removes only records that are not part of the current ledgers. 

What does "current ledgers" mean here?

> pruning

Nit: the example config uses "purging." That seems more appropriate here. Not entirely sure if there's a substantive difference but we should try to be consistent.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -59,42 +67,40 @@ path={path_to_ledger_store}
...
+Each `rippled` server in the XRP Ledger network performs all of the transaction processing work of the network. ***TODO: Is this true? What do we mean by "transaction processing" -- do we mean validation? Do we mean storing transaction history? I ask because I don't think "each rippled server in the XRP Ledger network" participates in the same way. Depending on the answer - aren't there some servers on the network that don't participate in "transaction processing"?*** It is unknown when volumes will approach maximum network capacity. ***TODO: This is true - but do we need to say it?*** Therefore, the baseline hardware for production `rippled` servers should be similar to that used in Ripple's [performance testing](https://ripple.com/dev-blog/demonstrably-scalable-blockchain/). ***TODO: Not sure about the word "Therefore" here - it's saying that because we don't know when volumes will approach max network capacity - you could use this configuration. I'm not sure if that make sense. I may be misunderstanding what we mean.***
```
ryangyoung: > Depending on the answer - aren't there some servers on the network that don't participate in "transaction processing"?

You're definitely right that the original phrase is too broad and vague. I'll the other questions for @mDuo13 since he undoubtedly has good answers for them. For this particular question, my understanding is that you're right: not all servers "process transactions." _Every server_ is active in the network in the sense that it communicates with other servers. Only validators "process transactions," as far as I know.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
ryangyoung: > instead of "Disk" - refer to "Storage" instead?

Maybe "Storage type"?

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
ryangyoung: Makes sense to me!

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
ryangyoung: Suggest saying "`large` or `huge`" here, depending on the answer to the question of how they differ, if at all. If they are different, we should update both instances.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -13,40 +13,48 @@ To meet your `rippled` capacity requirements, you must address these technical f
 - The [configuration settings](#configuration-settings) that affect resource utilization
 - The [network and hardware](#network-and-hardware) requirements to achieve consistent, good performance across the XRP Ledger network
 
+
```
ryangyoung, line 12: I know it's outside of the scope for this PR, but the intro could probably be reworded to more closely align with the rest of the content, especially the very first paragraph.

ryangyoung (commented):
I tried to answer as many questions as possible. Those answers are mostly best guesses that should be validated by @mDuo13 or @mtrippled or @manojsdoshi or someone else who knows better than me!

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
mtrippled: Just say "huge".

ryangyoung: @mtrippled is there a difference between "large" and "huge"?

mtrippled: "huge" uses more memory but reduces disk access requirements. Since disk is slower than memory, this increases performance.

mDuo13 (changes requested):
Shoot. I thought I submitted these comments days ago.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
mDuo13: I think J-Ha already realizes this, but for the record: there are "default" values (what the server uses if the config doesn't say) and "default config" values (what the default config file provided with rippled contains). So if you don't change anything you get `medium`, but if you remove the `node_size` field from the config entirely, you get `tiny`.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
mDuo13: The node size is talking about how big of an object to treat as one "node" in the ledger's underlying tree structure. It's not directly configuring storage or RAM, but it affects how much RAM your server uses and how fast your server can fetch things from disk. 

Think of it as the size of cardboard box you use for shipping. Larger cardboard boxes (node size) means you get more data at once (fewer trips to pick up a box from the warehouse, aka disk) but you can't be as specific about which things you hold onto at the same time because there's more extra stuff you don't need in every box (therefore it requires more RAM to store the same subset of things you're actually using). So you want to choose the right size so that "the subset of stored things you're actually using" can be retrieved in the fewest trips but you can still fit all the boxes (including the extra stuff you're not using that came in the same boxes) in your office (in RAM).

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
mDuo13: Actually, it's true that **all `rippled` servers process transactions.** Every server receives the same set of transactions, applies them individually to its closed ledger, and calculates the result. Non-validating servers don't speak up about their results, but all servers do the processing. If they get the same resulting hash as the one approved by consensus, then they know they ended up with the exact same ledger as everyone else. If not, they have to throw out their results and download the approved results, but the rest of the time they save themselves the trouble by calculating it themselves. (This also serves as a check that everyone is processing transactions legitimately.)

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
mDuo13: No, disk is the correct term.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
jhaaaa: Looks like if you are deciding between large and huge, the recommendation is to always use huge. I am going to include a row for large just to acknowledge that it is a legal value listed in rippled-example.cfg - but just use huge instead.

jhaaaa:
@ryangyoung @mtrippled @mDuo13 - thank you so much for your thorough and patient review of this PR. I've incorporated your feedback -- I hope that I've done your words and expertise justice! When you have a moment, please take another look. I'm open to any and all feedback. 🙌

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -1,117 +1,179 @@
...
+For example, think of node size as the size of cardboard boxes you use to ship vases from a warehouse to your store. A larger cardboard box means that you can ship more vases at once with fewer trips from the warehouse. A large node size means that you can get more `rippled` transaction data, for example, with fewer trips to disk storage.
+
+However, with a larger box, you also need to include more packing material to keep the vases from breaking during shipping, which takes up precious space in the box. Likewise, with a larger node size, you can get a lot of transaction data, but the data may also contain `rippled` start up information and ledger reports, which takes up precious RAM that can't be used for the subset of data you're actually using.
+
+A part of tuning these configurations is about choosing the right node size so that the subset of data you're actually using can be retrieved in the fewest trips to disk storage, while also ensuring that all of it, including other data you aren't using right now, can fit in your available RAM.
+
+Ripple recommends you always use the largest node size your available RAM can support. See the following table for recommended settings.
```
mDuo13, line 18: This part is the most important, but it's after several paragraphs of analogy. I tend to think less may be more here, since people will probably skim or just stop reading. As painful as it seems, I'd cut out the analogy stuff and put it in an addendum or later section (so people who actually care can read it) while busy people can get straight to this rule and the table.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -1,117 +1,179 @@
...
-|:----------------------------|:------------------|:-----------------------------------|
-| < 8GB                       | `tiny`            | Not recommended                    |
-| 8GB                         | `small`           |                                    |
-| 16GB                        | `medium`          |                                    |
-| 32GB                        | `huge`            | Recommended for production servers |
+Each `node_size` has a corresponding requirement for available RAM. For example, if you set `node_size` to `huge`, you should have at least 32GB of available RAM to help ensure that `rippled` can run smoothly.
+
+To tune your server, it may be useful to start with `tiny` and increase the size to `small`, `medium`, and so on as you refine the requirements for your use case.
+
+***TODO: Small thing, but I switched the order of the node size and RAM columns because the node size column contains the value that you are setting and should be the starting point of focus.***
```
mDuo13: I think the new form is kind of backwards. Most people will be looking at this table thinking, "I have this much RAM, so what should my node_size be?" The thing you're looking it up by should be the leftmost column.

mDuo13: Speaking of which, I'd say, "RAM available for rippled" rather than "Available RAM for rippled" for similar reasons.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -1,117 +1,179 @@
...
+| `large`           | 32GB                        | While `large` is a legal value, Ripple recommends just using `huge` instead. ***TODO: Per Mark, `large` uses less memory, but increases disk access requirements. Since disk is slower than memory, this reduces performance. This is a great clarification - but for both large and huge, we list 32GB RAM. If we want to include this info from Mark - do we need to reduce the RAM value we list for `large` for it to make sense?*** |
+| `huge`            | 32GB                        | Recommended for production servers. |
+
+For `node_size` troubleshooting, see [Bad node_size value](server-wont-start.html#bad-node-size-value).
+
 
 ### Node DB Type
 
-The `type` field in the `node_db` section of the `rippled.cfg` file sets the type of key-value store that `rippled` uses to persist the XRP Ledger in the ledger store. You can set the value to either `rocksdb` or `nudb`.
+The `type` field in the `[node_db]` stanza of the `rippled.cfg` file sets the type of key-value store that `rippled` uses to hold the ledger store on disk.
```
mDuo13: Nit: scratch "on disk" from the end of this sentence, since the chosen db type also holds the data (cached) in RAM.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -1,117 +1,179 @@
...
 
-`rippled` offers a history sharding feature that allows you to store a randomized range of ledgers in a separate shard store. You may want to configure the shard store to use a different type of key-value store than the ledger store. For more information about how to use this feature, see [History Sharding](https://ripple.com/build/history-sharding/#shard-store-configuration).
+This setting does not directly configure RAM settings, but the choice of key-value store has important implications for RAM usage because of the different ways these technologies cache and index data for fast lookup.
 
+You can set the value to either `RocksDB` or `NuDB`.
 
-#### RocksDB vs NuDB
-RocksDB requires approximately one-third less disk storage than NuDB and provides a corresponding improvement in I/O latency. However, this comes at a cost of increased memory utilization as storage size grows. NuDB, on the other hand, has nearly constant performance and memory footprint regardless of [storage](#storage).
+* If you are running a `rippled` server as a production validator, Ripple recommends that you use `RocksDB` with rotational disk storage. [Learn more](#more-about-using-rocksdb)
```
mDuo13: This makes it seem like we recommend using rotational disk storage. We don't.

- If your server is a validator, it only needs a small amount of history, so use RocksDB for best performance
- For most cases, use NuDB because its performance is constant even with large amounts of data on disk. A fast SSD is required.
- If you are using rotational disks (not recommended) or even just a slow SSD, use RocksDB.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -1,117 +1,179 @@
...
 
-`rippled` servers that operate as validators should keep only a few days' worth of data or less. Ripple recommends using RocksDB for validators. For all other uses, Ripple recommends using NuDB for the ledger store.
+* If you are running a `rippled` server for any other production use, Ripple recommends that you use `NuDB` with SSD storage. NuDB works with SSD storage only. [Learn more](#more-about-using-nudb)
 
-RocksDB has performance-related configuration options you can modify to achieve maximum transaction processing throughput. (NuDB does not have performance-related configuration options.) Here is an example of the recommended configuration for a `rippled` server using RocksDB:
+The example `rippled-example.cfg` file has the `type` field in the `[node_db]` stanza is set to `RocksDB`. ***TODO: Just an FYI that in this case, if you don't set a type value, rippled will not run. There is no default value coded into rippled.***
```
mDuo13: :+1: except for the typo:

> has the `type` field... is set

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -1,117 +1,179 @@
...
+
+[RocksDB](https://rocksdb.org/docs/getting-started.html) is an embeddable persistent key-value store that is optimized for rotational disks.
+
+RocksDB requires approximately one-third less disk [storage](#storage) than NuDB and provides better I/O latency. However, the better I/O latency comes as result of the large amount of RAM RocksDB requires to store data indexes.
+
+Validators should be configured to use RocksDB and to store no more than about 300,000 ledgers (approximately two weeks' worth of [historical data](#historical-data)) in the ledger store.
+
+RocksDB has performance-related configuration options that you can set in `rippled.cfg` to achieve maximum transaction processing throughput. Here is the recommended configuration for a `rippled` server using RocksDB:
+
+***TODO: reordered to match order in rippled-example.cfg***
```
mDuo13: :+1:

If you can remember long enough, you can also leave these types of "just FYI" comments out of the file, then add comments inline in GitHub when you create/update the PR.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -1,117 +1,179 @@
...
 
-Online deletion enables pruning of `rippled` ledgers from databases without any disruption of service. It only removes records that are not part of the current ledgers. Without online deletion, those databases grow without bounds. Freeing disk space requires stopping the process and manually removing database files.
+Online deletion enables the purging of `rippled` ledgers from databases without any disruption of service. It removes only records that are not part of the current ledgers. ***TODO: what do we mean by "current ledgers"?*** Without online deletion, those databases grow without bounds. Freeing disk space requires stopping the process and manually removing database files. For more information, see [`[node_db]`: `online_delete`](https://github.com/ripple/rippled/blob/develop/cfg/rippled-example.cfg#L832).
```
mDuo13: Basically, "Data in current ledgers" means any data that's used by ledger versions that area new enough not to be deleted. See my work in progress on the online deletion stuff for more details: https://github.com/mDuo13/ripple-dev-portal/blob/online_delete/content/concepts/the-rippled-server/online-deletion.md#background

(I say "used by" because the same underlying data may be a part of several ledger versions and used de-duplicated.)

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -1,117 +1,179 @@
 # Capacity Planning
 
-This section describes some of the challenges related to scaling `rippled` servers for testing and production deployments. Additionally, it describes how hardware and configuration settings relate to those challenges. Finally, this section offers recommendations to assist you in properly setting up `rippled` to meet the goals for the use case of your deployment.
+This section describes configuration, network, and hardware recommendations that you can use to tune and optimize the performance of your `rippled` server. Being aware of these considerations can help you ensure that your `rippled` server is ready to handle XRP Ledger network capacity today and in the near future.
```
mDuo13, line 3: :+1: Excellent. Much more concise and purpose-driven.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -1,117 +1,179 @@
...
 ### Recommendation
 
 For best performance in enterprise production environments, Ripple recommends running `rippled` on bare metal with the following characteristics:
 
 - Operating System: Ubuntu 16.04+
 - CPU: Intel Xeon 3+ GHz processor with 4 cores and hyperthreading enabled
-- Disk: SSD
+- Disk: SSD (7000+ writes/second, 10,000+ reads/second)
 - RAM:
 	- For testing: 8GB+
```
mDuo13: Since the heading says "production environments", let's remove the bullet points and just list the production 32GB number for RAM here

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -1,117 +1,179 @@
...
-#### CPU Utilization and Virtualization
-Ripple performance engineering has determined that bare metal servers achieve maximum throughput. However, it is likely that hypervisors cause minimal degradation in performance.
+#### Storage
 
-#### Network
+Here are some estimated `rippled` storage requirements:
 
-Any enterprise or carrier-class data center should have substantial network bandwidth to support running `rippled` servers. The minimum requirements are roughly 2Mbps transmit and 2Mbps receive for current transaction volumes. However, these can burst up to 100MBps transmissions when serving historical ledger and transaction reports. When a `rippled` server initially starts up, it can burst to over 20Mbps receive.
+- RocksDB stores around 8GB per day
+- NuDB stores around 12GB per day
```
mDuo13, line 132: I didn't realize NuDB actually _stores_ more data than RocksDB. I guess it's a data compression thing?

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -1,117 +1,179 @@
...
+
+To tune your server, it may be useful to start with `tiny` and increase the size to `small`, `medium`, and so on as you refine the requirements for your use case.
+
+***TODO: Small thing, but I switched the order of the node size and RAM columns because the node size column contains the value that you are setting and should be the starting point of focus.***
+
+| `node_size` value | Available RAM for `rippled` | Notes                      |
+|:------------------|:----------------------------|:---------------------------|
+| `tiny`            | < 8GB                       | Not recommended for testing or production servers. If not specified in `rippled.cfg`, `tiny` is the default value. |
+| `small`           | 8GB                         | Recommended for test servers. |
+| `medium`          | 16GB                        | Note that the `rippled-example.cfg` has its `node_size` value set to `medium`. ***TODO: The doc also says: With a `node_size` of `medium`, a `rippled` server can be reasonably stable in a test Linux system with 16GB of RAM -- is our recommendation for test servers - small or medium? Or either?*** |
```
mDuo13: We don't have a strict recommendation for test servers. A test server is whichever server you have around to test things with.

You can say that the default config uses this value. (The RPM installation literally copies and renames the "example config" to become the default config.)

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -1,117 +1,179 @@
...
+| `large`           | 32GB                        | While `large` is a legal value, Ripple recommends just using `huge` instead. ***TODO: Per Mark, `large` uses less memory, but increases disk access requirements. Since disk is slower than memory, this reduces performance. This is a great clarification - but for both large and huge, we list 32GB RAM. If we want to include this info from Mark - do we need to reduce the RAM value we list for `large` for it to make sense?*** |
```
mDuo13: I'd leave the `large` row out of this table. Maybe you could add a note after the table stating:

> Although `large` is also a legal value for `[node_size]`, in practice it performs worse than `huge` in most circumstances. Ripple recommends always using `huge` instead of `large`.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -1,117 +1,179 @@
...
+| `huge`            | 32GB                        | Recommended for production servers. |
+
+For `node_size` troubleshooting, see [Bad node_size value](server-wont-start.html#bad-node-size-value).
```
mDuo13: I'd be a little more specific so as not to make people overestimate what they're getting at that link:

> If you set the `node_size` parameter to an invalid value, the server [fails to start](server-wont-start.html#bad-node-size-value).

Also, could you be so kind as to edit that page and add `large` to the list of valid values for node_size in that description? Even though it's not recommended, it _is_ technically valid so it should be listed.

mDuo13 (changes requested):
This is almost there, but I have a couple more suggestions I'd like to see implemented before we merge it, especially regarding the node_size table.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
ryangyoung: Agreed. This revision is wonderful.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
jhaaaa: Sounds good to me, Rome! Making the change.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
jhaaaa: That's funny because I thought just the opposite 😂! I thought that if you know you want to run a production server - you'd look at the node size and RAM to see what you need to provide. But I see what you are saying - which is that the user is actually thinking - I have 16GB of RAM available, so I need to set my node size to medium and I can't run a production server. I was thinking that people would make the amount of RAM they need available based on what kind of server they want to run. But I get it! Changing it back. 🙌

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
jhaaaa: OMG I almost teared up when I read these comments. 😂

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
jhaaaa: Oh great @mDuo13 - thank you for this clarification. 🙌  That is great to know about the RPM - that makes so much more sense to me now. But, I think I need to leave it as "Note that the `rippled-example.cfg` has its `node_size` value set to `medium`." -- just because I can't call it the default config because we are using default to mean the config that is used when no value is provided - in this case, the default config is `tiny`. Does my reasoning for keeping is as-is make sense?

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
jhaaaa: Beautiful. Thank you for writing this up for me!

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
jhaaaa: I've saved the analogy content for another doc - or perhaps just for my personal doc trove. Sometimes I mistake questions that I have - for questions that our users have. I need to be careful about over-explaining obvious concepts, which could be unpleasant for our readers. Thank you again for the reality check on this, Rome.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -31,32 +23,33 @@ Each `node_size` has a corresponding requirement for available RAM. For example,
 
 To tune your server, it may be useful to start with `tiny` and increase the size to `small`, `medium`, and so on as you refine the requirements for your use case.
 
-***TODO: Small thing, but I switched the order of the node size and RAM columns because the node size column contains the value that you are setting and should be the starting point of focus.***
-
-| `node_size` value | Available RAM for `rippled` | Notes                      |
+| `node_size` value | RAM available for `rippled` | Notes                      |
```
mDuo13: I think the column headers are still swapped?

mDuo13 (changes requested):
The changes look great, except the table headers of the node_size rec. Fix that, and you'll have my :+1: .

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -1,117 +1,168 @@
...
-| 32GB                        | `huge`            | Recommended for production servers |
+Each `node_size` has a corresponding requirement for available RAM. For example, if you set `node_size` to `huge`, you should have at least 32GB of available RAM to help ensure that `rippled` can run smoothly.
+
+To tune your server, it may be useful to start with `tiny` and increase the size to `small`, `medium`, and so on as you refine the requirements for your use case.
+
+| `node_size` value | RAM available for `rippled` | Notes                      |
+|:------------------|:----------------------------|:---------------------------|
+| < 8GB             | `tiny`                      | Not recommended for testing or production servers. If not specified in `rippled.cfg`, `tiny` is the default value. |
+| 8GB               | `small`                     | Recommended for test servers. |
+| 16GB              | `medium`                    | Note that the `rippled-example.cfg` has its `node_size` value set to `medium`. |
```
mDuo13: Nit: change the notes to active voice (also remove the word "note" as the column header already says they're notes!):

> The default `rippled.cfg` uses this value.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
jhaaaa: Facepalm. Fixing now!

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
jhaaaa: Thank you for this! Made a few changes to address -- I hope they work for you.

mDuo13 (approved):

mDuo13:
Rebase onto the latest master if you want Jenkins to :heavy_check_mark: this PR. Otherwise, looks good to me.

jhaaaa:
Jenkins failure is due to a broken link that has already been fixed in master. Tried to rebase to master to enable Jenkins to pass, but there was a git glitch keeping it from happening. Checked with Rome and we're okay to merge.

PR #481, opened 2018-11-29
Corrections to currency format information

mDuo13:
Some corrections I found while digging through the `rippled` code and writing my transaction serialization code:

- The bit counts for the standard currency code were off by 1 byte
- The issued currency amount's exponent is represented as an unsigned integer by adding 97 to it (the comments in the code [wrongly stated 142 instead of 97](https://github.com/ripple/rippled/pull/2790/).)


content/references/rippled-api/api-conventions/currency-formats.md
```diff
@@ -33,9 +33,16 @@ Issued currencies in the XRP Ledger are represented with a custom format with th
...
+1. The first (most significant) bit for an issued currency amount is `1` to indicate that it is not an XRP amount. (XRP amounts always have the most significant bit set to `0` to distinguish them from this format.)
+2. The sign bit indicates whether the amount is positive or negative. Unlike standard [two's complement](https://en.wikipedia.org/wiki/Two%27s_complement) integers, `1` indicates **positive** in the XRP Ledger format, and `0` indicates negative.
+3. The next 8 bits represent the exponent as an unsigned integer. The exponent indicates the scale (what power of 10 the significant digits should be multiplied by) in the range -96 to +80 (inclusive). However, when serializing, we add 97 to the exponent to make it possible to serialize as an unsigned integer. Thus, a serialized value of `1` indicates an exponent of `-96`, a serialized value of `177` indicates an exponent of 80, and so on.
```
seelabs, line 40: This is only true for non-native amounts. XRP does not use exponents. For XRP, the max allowed native amount on the network is 100000000000000000, which is 57 bits. So those high three bits would occupy where this paragraph says is the exponent.

content/references/rippled-api/api-conventions/currency-formats.md
```diff
@@ -33,9 +33,16 @@ Issued currencies in the XRP Ledger are represented with a custom format with th
...
+4. The remaining 54 bits represent the significant digits as an unsigned integer. When serializing, this value is normalized to the range 10<sup>15</sup> (`1000000000000000`) to 10<sup>16</sup>-1 (`9999999999999999`) inclusive, except for the special case of the value 0, whose significant digits and sig use the value `0`.  
+
```
seelabs, line 42: This is only true for non-native amounts. For XRP, it's the remaining 62 bits, not 54 bits; Also, for XRP the canonical range will be 0 to 100000000000000000. BTW, this is only 57 bits, though I would still say "62 bits" in the docs.

content/references/rippled-api/api-conventions/currency-formats.md
```diff
@@ -48,11 +55,12 @@ The standard currency mapping allocates the bits as follows:
...
 1. The first 8 bits must be `0x00`.
-2. The next 96 bits are reserved, and should be all `0`s.
+2. The next 88 bits are reserved, and should be all `0`'s.
 3. The next 24 bits represent 3 characters of ASCII.
     Ripple recommends using [ISO 4217](http://www.xe.com/iso4217.php) codes, or popular pseudo-ISO 4217 codes such as "BTC". However, any combination of the following characters is permitted: all uppercase and lowercase letters, digits, as well as the symbols `?`, `!`, `@`, `#`, `$`, `%`, `^`, `&`, `*`, `<`, `>`, `(`, `)`, `{`, `}`, `[`, `]`, and <code>&#124;</code>. The currency code `XRP` (all-uppercase) is reserved for XRP and cannot be used by issued currencies.
-4. The next 8 bits indicate the currency version. If the same currency is reissued with a different value, you can increment this value to keep the currencies separate.
-5. The next 24 bits are reserved and should be all `0`s.
+4. The next 40 bits are reserved and should be all `0`'s.
+
+In the rare case that a field specifies a currency code for XRP, the currency code should be all zeroes.
```
seelabs: I'd probably strike the work "rare". We either support the format or we don't. A currency code of all zeros is XRP.

seelabs (commented):
Left a couple comments. BTW, I'm not sure how to interpret the `uxf` files. I assume those are specifying formatting (?), so I did not review those files.

content/references/rippled-api/api-conventions/currency-formats.md
mDuo13: Yes, in context this is only describing issued currencies (this section starts on lines 29-34, "issued currency math"). The section describing XRP is above it.

content/references/rippled-api/api-conventions/currency-formats.md
mDuo13: Yeah, this section is specifically describing issued currencies.

content/references/rippled-api/api-conventions/currency-formats.md
mDuo13: It is rare that a field specifies a currency code for XRP, though. The only case I can think of is path steps. I was hoping to head off confusion about any time you specify an XRP amount that you don't actually provide a currency code there (because the "not XRP" bit of the value handles it).

mDuo13:
Sorry, I should have clarified that the `.uxf` files are the source files for the diagrams and don't need to be reviewed. (In case you're curious, I use a tool called [umlet](http://www.umlet.com/) for simple diagrams.)

content/references/rippled-api/api-conventions/currency-formats.md
seelabs: Yeah, I see where you're comming from. OK, I'm fine with keeping "rare".

content/references/rippled-api/api-conventions/currency-formats.md
seelabs: Ah, I missed that context. Thanks for clarifying.

seelabs (approved):
:+1:

content/references/rippled-api/api-conventions/currency-formats.md
```diff
@@ -33,9 +33,16 @@ Issued currencies in the XRP Ledger are represented with a custom format with th
 
 Internally, `rippled` represents numbers for issued currencies in a custom number format. This format can store a wide variety of assets, including those typically measured in very small or very large denominations. Unlike typical floating-point representations of non-whole numbers, this format uses integer math for all calculations, so it always maintains 15 decimal digits of precision. Unlike "arbitrary precision" number formats, the custom format can always be stored in a fixed size of 64 bits.
```
jhaaaa, line 34: Sorry - not really within the scope of the PR - but just curious about what we mean by "internally"? Does it mean that as a dev using the `rippled` API, you'll never encounter values in this format in the API's requests and responses, but we're telling you how `rippled` stores these values behind the scenes to show how it supports high precision?

jhaaaa (approved):
I appreciate the graphics to help explain the concepts. But I have to admit that I don't understand how this works quite yet and my questions are beyond the scope of this PR so I'll hold off on them. From a pure language review - it looks good to me!

content/references/rippled-api/api-conventions/currency-formats.md
mDuo13: The idea behind this paragraph is:

- To let people know how to match `rippled`'s math exactly when it comes to issued currency amounts, so they know what to expect or can predict the outcomes of transactions down to the lowest decimal place. Or, more likely, to convince them that it's not worth the effort because they'd have to implement a bunch of low-level number routines, but they can come reasonably close by using a "big decimal" library with 15 digits of precision.
- To inform those implementing transaction serialization/signing that the binary format fits all that stuff in 64 bits, which also serves as a justification for why the first point is the way it is.

Some of this stuff is a little extra low-level and should probably get migrated to the Serialization Format page eventually (sometime after #483 is merged).

jhaaaa: Thank you for taking the time to provide such thorough and thoughtful context for me. Got it. 🙌

mDuo13:
Rebased onto master to get a broken link fix.

PR #482, opened 2018-11-29
Revert "Ignore initial result of tx submission (RLJS-607)"

chalermchai29:
Reverts ripple/ripple-dev-portal#445


mDuo13:
Care to explain why? The initial result is provisional and can be misleading.

mDuo13:
I believe this PR has been opened in error. If that is incorrect, please clarify and reopen.

PR #483, opened 2018-12-04
Transaction Serialization (binary format) docs

### LINKED ISSUE #148: Binary blob format is not documented
Please add specs at least for transactions.

mDuo13:
Fixes #148.

Introduces documentation and Python sample code for serializing XRP Ledger transactions. (The sample code does not serialize metadata.)

## Doc review

This PR adds a new "Serialization Format" section to the API Conventions. This page describes the binary format for transactions in detail.

- Please review `serialization.md` and the diagrams it includes for accuracy and completeness.
- No other files need documentation reviews.

Staged version (Ripple VPN required): https://developers-staging.ripple.com/serialization.html

## Code Review

This PR also adds sample code in Python for serializing transactions to `bytes` (Python's built-in binary data type), and consequently, hex blobs. The code requires Python 3.5+. (It uses [bytes.hex()](https://docs.python.org/3/library/stdtypes.html#bytes.hex) in debug & print statements. )

- [`serialize.py`](https://github.com/ripple/ripple-dev-portal/pull/483/files#diff-a8412b41fb09d9abfcca7f5b468ff115) - Serializes most data types. This file does most of the work. It also works as a commandline utility for serializing a transaction from a file, stdin, or CLI option.
- [`xrpl_num.py`](https://github.com/ripple/ripple-dev-portal/pull/483/files#diff-597f6d1ba32a77dc67f904490a2db044) - Writes issued currency amounts
- [`address.py`](https://github.com/ripple/ripple-dev-portal/pull/483/files#diff-952faec795236205c88bc9cdd63acd4c) - Decodes base58 Ripple addresses (using an included base58 library, described below) and removes the one-byte address prefix.

## Files That Don't Need Review

- `base58.py` is copied from [keis/base58](https://github.com/keis/base58). The only change is that I have changed the dictionary to be the XRP Ledger's base58 dictionary. This file probably doesn't need to be thoroughly reviewed. (A glance to make sure it's not secretly malicious might be useful?)
- The file `definitions.json` is copied directly from [ripple-binary-codec](https://github.com/ripple/ripple-binary-codec/blob/master/src/enums/definitions.json) and is not changed.
- The `test-cases/` binary and JSON files (`tx1.json`, `tx1-binary.txt`, etc.) are real transactions I fetched from the network arbitrarily.
- The `.uxf` files in `_img-sources/` (`serialization-amount.uxf`, etc.) are the source files for the diagrams. I have added several for transaction serialization. Reviewers don't need to look over the XML in detail. If you find any mistakes in the output images, you can comment in the markdown where the image is included or optionally on the line of the XML with the problem.


content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
...
+### Object Fields
+[STObject]: #object-fields
+
+Some fields, such as `SignerEntry` (in [SignerListSet transactions][]), and `Memo` (in `Memos` arrays) are objects (called the "STObject" type). The serialization of objects is very similar to that of arrays, with one difference: **object members must be placed in canonical order** within the object field, where array fields have an explicit order already.
+
+The [canonical field order](#canonical-field-order) of object fields is the same as the canonical field order for all top-level fields, but the members of the object must be sorted within the object. After the last member, there is an "Object end" Field ID of `0xe1` with no contents.
+
+The following example shows the serialization format for an object (a single `Memo` object in the `Memos` array).
+
+![Object field ID, followed by the Object ID and contents of each object member in canonical order, followed by the "Object end" field ID](img/serialization-object.png)
```
ximinez, line 264: In this image (`serialization-object.png`), there is an unlabeled block between `0x7c` and `MemoType contents`

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
...
+| `0x10`    | `currency`    | [Currency Code][] | 160 bits | 2nd   |
+| `0x20`    | `issuer`      | [AccountID][]     | 160 bits | 3rd   |
+
+[Currency Code]: currency-formats.html#standard-currency-codes
+
+Some combinations are invalid; see [Path Specifications](paths.html#path-specifications) for details.
+
+The AccountIDs in the `account` and `issuer` fields are presented _without_ a variable-length encoding prefix. When the `currency` is XRP, the currency code is represented as 160 bits of zeroes.
+
+Each step is followed directly by the next step of the path. As described above, last step of a path is followed by either `0xff` (if another path follows) or `0x00` (if this ends the last path).
```
ximinez, line 293: What indicates the end of a `Path` within the `PathSet`? Is it the `0xff` or `0x00`? In that case, you might want to state that here for clarity.

content/_code-samples/tx-serialization/xrpl_num.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        Returns canonical format for zero:
+        - "Not XRP" bit = 1
+        - "Sign bit" = 1 (for positive !!)
+        - exponent ???
+        """
+        # Mantissa is all zeroes. Must be positive zero.
+        #bitval = 0b1100000001000000000000000000000000000000000000000000000000000000
+        #return bitval.to_bytes(8, byteorder="big", signed=False)
+        return (0x8000000000000000).to_bytes(8, byteorder="big", signed=False)
```
ximinez, line 63: Should the sign bit be set here (positive zero)?

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
+# Serialization Format
+[[Source]<br>](https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/impl/STObject.cpp#L696-L718 "Source")
+
+This page describes the XRP Ledger's canonical binary format for transactions and other data. This binary format is necessary to create and verify digital signatures of those transactions' contents, and is also used in other places. The [rippled APIs](rippled-api.html) typically use JSON to communicate with client applications. However, JSON is unsuitable as a format for serializing transactions for being digitally signed, because JSON can represent the same data in many different but equivalent ways.
```
jhaaaa, line 4: What a clear description of the WHY serialization vs JSON. A thing of beauty.

This may be a dumb question - but in which cases must you/should you serialize a transaction - meaning, use binary format vs JSON? I understand from the text that it is necessary for creating and verifying digital signatures of the transaction contents - is that necessary for every transaction? Or is this only for payments?

Is what we are putting into binary format the `tx_json` that you would provide in a `sign` transaction request, for example? I have a feeling I may be missing it completely.

This may be too elementary for our audience - but I feel like I need a sample JSON (before) and the sample in the resulting binary format (after). Just to situate me with where I'm starting and what the end result is after steps 1-5 below.

content/_code-samples/tx-serialization/test-cases/README.md
```diff
@@ -0,0 +1,52 @@
+# Transaction Serialization Test Cases
```
ximinez, line 1: Is there a link to this README anywhere in serialization.md or another page? Or is the user expected to find this on github?

ximinez (changes requested):

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
...
+
+Each field has an "internal" binary format used in the `rippled` source code to represent that field when signing (and in most other cases). The internal formats for all fields are defined in the source code of [`SField.cpp`](https://github.com/ripple/rippled/blob/master/src/ripple/protocol/impl/SField.cpp). (This file also includes fields other than transaction fields.) The [Transaction Format Reference](transaction-formats.html) also lists the internal formats for all transaction fields.
+
+For example, the `Flags` [common transaction field](transaction-common-fields.html) becomes a UInt32 (32-bit unsigned integer).
+
+### Definitions File
+
+The following JSON file defines the important constants you need for serializing XRP Ledger data to its binary format and deserializing it from binary:
+
+**<https://github.com/ripple/ripple-binary-codec/blob/master/src/enums/definitions.json>**
```
jhaaaa, line 64: You probably have a reason for exposing the full link here instead of making the file name a link? Just curious!

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
+# Serialization Format
+[[Source]<br>](https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/impl/STObject.cpp#L696-L718 "Source")
+
+This page describes the XRP Ledger's canonical binary format for transactions and other data. This binary format is necessary to create and verify digital signatures of those transactions' contents, and is also used in other places. The [rippled APIs](rippled-api.html) typically use JSON to communicate with client applications. However, JSON is unsuitable as a format for serializing transactions for being digitally signed, because JSON can represent the same data in many different but equivalent ways.
+
+The process of serializing a transaction from JSON or any other representation into their canonical binary format can be summarized with these steps:
+
+1. Make sure all required fields are provided, including any required but "auto-fillable" fields.
```
jhaaaa: Would it make sense to link auto-fillable to https://developers.ripple.com/transaction-common-fields.html#auto-fillable-fields? Relevant? I think you link to it deeper in the article - so it may be too soon?

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
...
+
+    **Note:** The `SigningPubKey` must also be provided at this step. When signing, you can derive this key from the secret key that is provided for signing.
+
+2. Convert each field's data into its "internal" binary format.
+3. Sort the fields in canonical order.
+4. Prefix each field with a Field ID.
+5. Concatenate the fields (including prefixes) in their sorted order.
+
```
jhaaaa, line 21: Would it be helpful to anchor link these steps to the sections below that describe how to perform them? For step 2, link "internal binary format" to "Internal Format," for step 3 link "canonical order" to "Canonical Field Order," and for step 4 link "Field ID" to "Field IDs."  Just to give the reader a heads up that further guidance on these steps is coming up?

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
...
+**Note:** The XRP Ledger uses the same serialization format to represent other types of data, such as [ledger objects](ledger-object-types.html) and processed transactions. However, only certain fields are appropriate for including in a transaction that gets signed. (For example, the `TxnSignature` field, containing the signature itself, should not be present in the binary blob that you sign.) Thus, some fields are designated as "Signing" fields, which are included in objects when those objects are signed, and "non-signing" fields, which are not.
+
+## Sample Code
+
+The serialization processes described here are implemented in multiple places and programming languages:
+
+- In C++ [in the `rippled` code base](https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/impl/STObject.cpp).
+- In JavaScript in the [`ripple-binary-codec`](https://github.com/ripple/ripple-binary-codec/) package.
+- In Python 3 [this repository's code samples section]({{target.github_forkurl}}/blob/{{target.github_branch}}/content/_code-samples/tx-serialization/serialize.py).
```
jhaaaa: nit: 

> In Python 3 [this repository's code samples section] 

to 

> In Python 3 in [this repository's code samples section]

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
...
+| `TYPES`               | Map of data types to their "type code" for constructing field IDs and sorting fields in canonical order. Codes below 1 should not appear in actual data; codes above 10000 represent special "high-level" object types such as "Transaction" that cannot be serialized inside other objects. |
+| `LEDGER_ENTRY_TYPES`  | Map of [ledger objects](ledger-object-types.html) to their data type. These appear in ledger state data, and in the "affected nodes" section of processed transactions' [metadata](transaction-metadata.html). |
+| `FIELDS`              | A sorted array of tuples representing all fields that may appear in transactions, ledger objects, or other data. The first member of each tuple is the string name of the field and the second member is an object with that field's properties. (See the "Field properties" table below for definitions of those fields.) |
+| `TRANSACTION_RESULTS` | Map of [transaction result codes](transaction-results.html) to their numeric values. Result types not included in ledgers have negative values;`tesSUCCESS` has numeric value 0; [`tec`-class codes](tec-codes.html) represent failures that are included in ledgers. |
```
jhaaaa: super nit - add a space between `negative values;` and `tesSUCCESS`

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
...
+For purposes of serializing transactions for signing and submitting, the `FIELDS`, `TYPES`, and `TRANSACTION_TYPES` fields are necessary.
+
+The field definition objects in the `FIELDS` array have the following fields:
+
+| Field            | Type    | Contents                                        |
+|:-----------------|:--------|:------------------------------------------------|
+| `nth`            | Number  | The [field code](#field-codes) of this field, for use in constructing its [Field ID](#field-ids) and ordering it with other fields of the same data type. |
+| `isVLEncoded`    | Boolean | If `true`, this field is [variable-length encoded](#variable-length-encoding). |
+| `isSerialized`   | Boolean | If `true`, this field should be encoded into serialized binary data. When this field is `false`, the field is typically reconstructed on demand rather than stored. |
+| `isSigningField` | Boolean | If `true` this field should be serialized when preparing a transaction for signing. If `false`, this field should be omitted from the data to be signed. (It may not be part of transactions at all.) |
```
jhaaaa, line 85: If this is what we mean, perhaps consider changing:

> It may not be part of transactions at all.

to

> It must not be part of transactions at all.

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
...
+
+
+## Internal Format
+
+Each field has an "internal" binary format used in the `rippled` source code to represent that field when signing (and in most other cases). The internal formats for all fields are defined in the source code of [`SField.cpp`](https://github.com/ripple/rippled/blob/master/src/ripple/protocol/impl/SField.cpp). (This file also includes fields other than transaction fields.) The [Transaction Format Reference](transaction-formats.html) also lists the internal formats for all transaction fields.
+
+For example, the `Flags` [common transaction field](transaction-common-fields.html) becomes a UInt32 (32-bit unsigned integer).
+
+### Definitions File
+
```
jhaaaa, line 61: This content is great - thank you for providing useful context and annotations to the definitions file.

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
...
+All fields in a transaction are sorted in a specific order based on the field's type first, then the field itself second. (Think of it as sorting by family name, then given name, where the family name is the field's type and the given name is the field itself.)
+
+### Field IDs
+
+[[Source - Encoding]](https://github.com/seelabs/rippled/blob/cecc0ad75849a1d50cc573188ad301ca65519a5b/src/ripple/protocol/impl/Serializer.cpp#L117-L148 "Source")
+[[Source - Decoding]](https://github.com/seelabs/rippled/blob/cecc0ad75849a1d50cc573188ad301ca65519a5b/src/ripple/protocol/impl/Serializer.cpp#L484-L509 "Source")
+
+
+When you combine the type code and sort code, you get the field's identifier, which is prefixed before the field in the final serialized blob. The size of the Field ID is one to three bytes depending on the type code and field codes it combines. See the following table:
+
```
jhaaaa, line 94: Another example of something that may be too elementary for our audience - but is something that I think might help me grasp this faster. What do you think of providing an example field ID in each cell in the table below? Is there any overlap between my brain and our audiences' brains? 🤔 😂

jhaaaa, line 94: Sorry if I'm reading too closely into this - but I'm hanging on every word to be able to grasp this info. 🤓 I'm a little confused by the difference between the field ID being a combination of the 

> type code and sort code

And then reading that the size of the Field ID depends on the 

> type code and field codes it combines

Is the field ID the combination of the type code and sort code, or the combination of the type code and field code? Or all three? I think I'm misunderstanding.

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
...
+- If the first length byte has a value of 241 to 254, then there are three length bytes.
+
+
+## Type List
+
+Transaction instructions may contain fields of any of the following types:
+
+| Type Name     | Type Code | Variable-Length? | Description                   |
+|:--------------|:----------|:-----------------|:------------------------------|
+| [AccountID][] | 8         | Yes              | The unique identifier for an [account](accounts.html). This field is variable-length encoded, but always exactly 20 bytes. |
```
jhaaaa: Just to make sure I'm following - when we say that this is always exactly 20 bytes - we're just clarifying that even if it is variable-length encoded, its length never varies and will always be exactly 20 bytes. Or is 20 bytes not referring to the thing we are saying has a variable length?

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
...
+When decoding, you can tell from the value of the first length byte whether there are 0, 1, or 2 additional length bytes:
+
+- If the first length byte has a value of 192 or less, then that's the only length byte and it contains the exact length of the field contents in bytes.
+- If the first length byte has a value of 193 to 240, then there are two length bytes.
+- If the first length byte has a value of 241 to 254, then there are three length bytes.
+
+
+## Type List
+
+Transaction instructions may contain fields of any of the following types:
```
jhaaaa, line 160: For all of the types that are not variable-length encoded -- is it useful / possible to list their bit or byte length? Most of them have a bit value listed - but some don't. Just humbly curious about why some don't list a value. Impractical or not useful?

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
...
+
+AccountIDs that appear as stand-alone fields (such as `Account` and `Destination`) are [variable-length encoded](#variable-length-encoding) despite being a fixed 160 bits in length. As a result, the length indicator for these fields is always the byte `0x14`. AccountIDs that appear as children of special fields ([Amount `issuer`][Amount] and [PathSet `account`][PathSet]) are _not_ variable-length encoded.
+
+
+### Amount Fields
+[Amount]: #amount-fields
+
+The "Amount" type is a special field type that represents an amount of currency, either XRP or an issued currency. This type consists of two sub-types:
+
```
jhaaaa, line 203: Just an FYI that the formatting of the following bolded **XRP** and **Issued Currencies** bullets is a little off. The sentences following the bullets aren't displaying on the next line, but display immediately after the bullet text.

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
...
+[STArray]: #array-fields
+
+Some transaction fields, such as `SignerEntries` (in [SignerListSet transactions][]) and [`Memos`](transaction-common-fields.html#memos-field), are arrays of objects (called the "STArray" type).
+
+Arrays contain several [object fields](#object-fields) in their native binary format in a specific order. In JSON, each array member is a JSON "wrapper" object with a single field, which is the name of the member object field. The value of that field is the ("inner") object itself.
+
+In the binary format, each member of the array has a Field ID prefix (based on the single key of the wrapper object) and contents (comprising the inner object, [serialized as an object](#object-fields)). To mark the end of an array, append an item with a "Field ID" of `0xf1` (the type code for array with field code of 1) and no contents.
+
+The following example shows the serialization format for an array (the `SignerEntries` field):
+
```
jhaaaa, line 233: A personal observation about the format of the graphic. It is incredibly effective in conveying the format of the field! The only thing that hung me up for a few seconds was the box with SignerEntries field in it. At first glance, its width seemed to indicate the segment or length of the SignerEntries field as a sub-segment of the entire width of the larger box. It is clearly a title after just a little more examination. Not sure if it would help to change that format so that it isn't connected to the larger box or maybe so that it isn't in a box? Maybe having it float disconnected is worse. I didn't read it as such in the Amount Fields section - so maybe it is a non-issue.

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -0,0 +1,285 @@
...
+[Currency Code]: currency-formats.html#standard-currency-codes
+
+Some combinations are invalid; see [Path Specifications](paths.html#path-specifications) for details.
+
+The AccountIDs in the `account` and `issuer` fields are presented _without_ a variable-length encoding prefix. When the `currency` is XRP, the currency code is represented as 160 bits of zeroes.
+
+Each step is followed directly by the next step of the path. As described above, last step of a path is followed by either `0xff` (if another path follows) or `0x00` (if this ends the last path).
+
+The following example shows the serialization format for a PathSet:
+
```
jhaaaa, line 296: In this graphic, curious about how the Type flag values for Currency (160 bits) and Issuer (160 bits) are "combined" to use a single Type flag of 0x30. I am in way over my head here, but I wasn't thinking that Type flag values could be "added" - as in 0x10 + 0x20 = 0x30 -- is that what is happening here?

jhaaaa (approved):
I reviewed `serialization.md` and all I can say is --- wow. What a thoroughly researched and written document. We're fortunate to have you tackling tough topics like this one. Feel free to ping me if you'd like me to do any follow-up reviews.

content/references/rippled-api/api-conventions/serialization.md
mDuo13: Ah yes, the length prefix for MemoType. Originally the "length prefix" label had arrows pointing to both places but the diagram got too busy.

I guess it's also a problem in the current version to leave it entirely unlabeled though. I'll try moving things around a bit more.

content/references/rippled-api/api-conventions/serialization.md
mDuo13: It's either one, depending on whether this path is the last path, as stated in this sentence...

content/_code-samples/tx-serialization/xrpl_num.py
mDuo13: You would think so, but that's not [what the code does](https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/impl/STAmount.cpp#L606) - it returns [cNotNative, which is 0x8000000000000000](https://github.com/ripple/rippled/blob/09050a860be745c88a54fbcbce1e317c72f2f51e/src/ripple/protocol/STAmount.h#L73). I think the comment in this Python code is wrong but the uncommented code below it is right.

content/_code-samples/tx-serialization/test-cases/README.md
mDuo13: The latter. The serialization page merely links the `_code-samples/tx-serialization/` folder; if you navigate to the test cases folder, this README explains them.

content/references/rippled-api/api-conventions/serialization.md
mDuo13: Not a deep reason, but I figured since it's in a repository not otherwise mentioned, providing the full URL actually sort of conveys that context?

content/references/rippled-api/api-conventions/serialization.md
mDuo13: Not what we mean.

Some "non-signing" fields are not part of transactions at all. But some are parts of transactions that merely aren't signed. (For example, you can't sign the signature for chicken-egg reasons.)

content/references/rippled-api/api-conventions/serialization.md
mDuo13: Sorry, I made up too many terms and got them mixed up a bit. It's probably easiest to think of it as that there is no such thing as a "sort code". (I'll fix this.)

It's a little confusing, but basically:

- Sort by type code, then by field code.
- Identify a field by Field ID, which is the serialized combination of type code and field code.
- Don't actually sort by Field ID directly because the variable-byte-structure of it would screw up sorting.

content/references/rippled-api/api-conventions/serialization.md
mDuo13: No, you're correct. It's "variable-length" encoded but the length never varies. I'm pretty sure it's a mistake in the early code that we're stuck with now.

EDIT: I think this makes more sense if we say it's "length-prefixed" instead of "variable-length encoded". It's still silly, but it doesn't sound _quite_ as contradictory.

content/references/rippled-api/api-conventions/serialization.md
mDuo13: I double-checked and all the fields that have fixed lengths have their bit lengths listed in the description, but I think you're right that having a separate column for them would be correct.

I think I'll change the name of "Variable-Length Encoding" to be "Length Prefix Encoding" to be more precise. (The other variable-length fields that aren't "variable-length encoded" have end markers instead of length prefixes.)

content/references/rippled-api/api-conventions/serialization.md
mDuo13: Changed all the images in this diagram to go that way. It also alleviated a little of the clutter of the diagrams.

Updated [DOC-1549](https://ripplelabs.atlassian.net/browse/DOC-1549) to update the currency format diagrams to match (since that affects #481, this PR, and whatever follow-up work DOC-1549 entails.)

content/references/rippled-api/api-conventions/serialization.md
mDuo13: The diagram has been updated to make the "length prefixes" label apply to both blocks now.

content/references/rippled-api/api-conventions/serialization.md
mDuo13: The type object has a bunch of flags that are combined with bitwise OR operations.

```
0x20 in binary:
00100000

0x10 in binary:
00010000

    00100000
OR  00010000
------------
    00110000  = 0x30 in hex

**Comment by mDuo13** [comment]
Updated the docs and diagrams with feedback from @ximinez and @jhaaaa reviews, and rebased onto master to get the changes from #481. The [staged docs](https://developers-staging.ripple.com/serialization.html) have been updated to the latest version.

File: content/references/rippled-api/api-conventions/serialization.md (hunk 4 above)
Comment (mDuo13) [reply content/references/rippled-api/api-conventions/serialization.md:4]: > I understand from the text that it is necessary for creating and verifying digital signatures of the transaction contents - is that necessary for every transaction? Or is this only for payments?

The XRP Ledger does not treat payments differently from any other transaction type. All transactions of any kind must be authorized by a digital signature, which means you have to sign a binary format of the transaction.

>Is what we are putting into binary format the tx_json that you would provide in a sign transaction request, for example?

Yes, it is. Although we also have to add the `SigningPubKey` to the transaction instructions too, that is basically also part of the `tx_json`.

> I feel like I need a sample JSON (before) and the sample in the resulting binary format (after).

Added a comparison of JSON and binary formats using some of the test cases from the sample code.

File: content/references/rippled-api/api-conventions/serialization.md (hunk 2 above)
Comment (ximinez) [reply content/references/rippled-api/api-conventions/serialization.md:293]: Ok, I must have blipped when I read it the first time, because I just didn't get that. But I think that's on me

**Review by ximinez — APPROVED** [review]

File: content/_code-samples/tx-serialization/serialize.py [hunk 19]
```diff
@@ -0,0 +1,462 @@
...
+    Some fields contain nested non-length-prefixed AccountIDs directly; those
+    call decode_address() instead of this function.
+    """
+    return vl_encode(decode_address(address))
+
+def amount_to_bytes(a):
+    """
+    Serializes an "Amount" type, which can be either XRP or an issued currency:
+    - XRP: 64 bits; 0, followed by 1 ("is positive"), followed by 62 bit UInt amount
+    - Issued Currency: 64 bits of amount, followed by
```
seelabs: Looks like the comment is incomplete. Probably should read "followed by currency and issuer account"

content/_code-samples/tx-serialization/serialize.py
```diff
@@ -0,0 +1,462 @@
...
+def field_id(field_name):
+    """
+    Returns the unique field ID for a given field name.
+    This field ID consists of the type code and field code, in 1 to 3 bytes
+    depending on whether those values are "common" (<16) or uncommon (>=16)
+    """
+    field_type_name = DEFINITIONS["FIELDS"][field_name]["type"]
+    type_code = DEFINITIONS["TYPES"][field_type_name]
+    field_code = DEFINITIONS["FIELDS"][field_name]["nth"]
+
```
seelabs, line 60: I'd assert that both the type_code and field_code are both >=0 and <= 255.

content/_code-samples/tx-serialization/serialize.py
```diff
@@ -0,0 +1,462 @@
...
+def amount_to_bytes(a):
+    """
+    Serializes an "Amount" type, which can be either XRP or an issued currency:
+    - XRP: 64 bits; 0, followed by 1 ("is positive"), followed by 62 bit UInt amount
+    - Issued Currency: 64 bits of amount, followed by
+    """
+    if type(a) == str:
+        # is XRP
+        xrp_amt = int(a)
```
seelabs, line 144: I'd range check `xrp_amt`

content/_code-samples/tx-serialization/serialize.py
```diff
@@ -0,0 +1,462 @@
+#!/bin/env python3
```
seelabs: This doesn't work on ubuntu. I think we want `/usr/bin/env` here.

seelabs (commented):
Great job no this. This was a ton of code of code for you to wade through and understand, and it's great to get it documented. Left a couple nits, but nothing serious.

BTW, I understand the python code is sample only, and not meant for production. However, I left one or two nits about range checking some things. I think those kind of asserts can help in understanding even if we're not as worried about error handling in sample code.

content/references/rippled-api/api-conventions/serialization.md
mDuo13: Added the link

content/references/rippled-api/api-conventions/serialization.md
mDuo13: Great idea! Added links to all the steps except the concatenate one (concatenating things is trivial and not unique to the XRP Ledger). That also reminded me to move a couple sections around so that some things that aren't part of the canonical ordering stuff aren't under the "Canonical Field Order" header.

content/references/rippled-api/api-conventions/serialization.md
jhaaaa: Thank you for taking the time to explain, Rome - got it! 🙌

seelabs (approved):
:+1:

content/references/rippled-api/api-conventions/serialization.md
jhaaaa: Thank you for taking the time to explain and illustrate...this is 🏅!

content/references/rippled-api/api-conventions/serialization.md
mDuo13: I changed the table up to have diagrams with examples to hopefully help this stuff make sense.

mDuo13:
I've updated the staged docs with the edits I've made per review, including new diagrams for Field IDs.

PR #484, opened 2018-12-07
Clean up issued currency format description

mDuo13:
I realized that, despite the changes from #481, this section still had a typo ("significant digits and sig") in discussing the case for zero, and there was some redundancy as well.

I think the cleaned-up version should more clearly distinguish between the math and serialization format (so that the serialization format stuff can later be moved out to the serialization page)


jhaaaa (approved):
Nice adjustments. Looks good! 👍

PR #485, opened 2018-12-07
Hashes - clarify hash prefixes, SHA-512Half size.

mDuo13:
- Fixes a typo where one old description of SHA-512Half said "bytes" when it meant "bits"
- Adds a table of hash prefixes based on https://github.com/ripple/rippled/blob/master/src/ripple/protocol/impl/HashPrefix.cpp but leaving off the types whose hashes I pretty much never expect to see.

### DIFF
--- a/content/references/rippled-api/api-conventions/basic-data-types.md
+++ b/content/references/rippled-api/api-conventions/basic-data-types.md
@@ -26,17 +26,25 @@ Each closed [Ledger](ledger-data-formats.html) has a [Ledger Index][] and a [Has
 
 In many cases, the XRP Ledger prefixes an object's binary data with a 4-byte code before calculating its hash, so that objects of different types have different hashes even if their binary formats are the same. The 4-byte codes are structured as three bytes of ASCII followed by a zero byte.
 
-The following table shows the prefixes for some hash types that appear in API requests and responses:
+Some types of hash appear in API requests and responses. Others are only calculated as the first step of signing a certain type of data, or calculating a higher-level hash. The following table shows all 4-byte hash prefixes the XRP Ledger uses:
 
-| Object Type                                                                 | Hash Prefix (Hex) | Hash Prefix (Text) |
-|:----------------------------------------------------------------------------|:------------------|:--|
-| Signed Transaction (`hash` field of transactions)                           | `0x54584E00`      | `TXN\0` |
-| Ledger Version (`ledger_hash` fields)                                       | `0x4C575200`      | `LWR\0` |
-| [Ledger state tree information (`account_state` field)](ledger-header.html) | `0x4D4C4E00`      | `MLN\0` |
-| Unsigned Transaction (For single-signing)                                   | `0x53545800`      | `STX\0` |
-| Unsigned Transaction (For multi-signing)                                    | `0x534D5400`      | `SMT\0` |
+| Object Type                           | API Fields                           | Hash Prefix (Hex) | Hash Prefix (Text) |
+|:--------------------------------------|:-------------------------------------|:------------------|:--|
+| Consensus proposal                    | N/A                                  | `0x50525000`      | `PRP\0` |
+| Ledger Version                        | `ledger_hash`                        | `0x4C575200`      | `LWR\0` |
+| Ledger state data                     | `account_state` in [ledger header][] | `0x4D4C4E00`      | `MLN\0` |
+| Ledger data inner node                | N/A                                  | `0x4D494E00`      | `MIN\0` |
+| Ledger data inner node ([SHAMapv2][]) | N/A                                  | `0x494E5200`      | `INR\0` |
+| Payment Channel Claim                 | N/A                                  | `0x434C4D00`      | `CLM\0` |
+| Signed Transaction                    | `hash` of transactions               | `0x54584E00`      | `TXN\0` |
+| Transaction with metadata             | N/A                                  | `0x534E4400`      | `SND\0` |
+| Unsigned Transaction (Single-signing) | N/A                                  | `0x53545800`      | `STX\0` |
+| Unsigned Transaction (Multi-signing)  | N/A                                  | `0x534D5400`      | `SMT\0` |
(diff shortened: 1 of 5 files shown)


content/references/rippled-api/api-conventions/basic-data-types.md
```diff
@@ -21,6 +21,24 @@ Each closed [Ledger](ledger-data-formats.html) has a [Ledger Index][] and a [Has
...
+| Object Type                                         | Hash Prefix (Hex) | Hash Prefix (Text) |
+|:----------------------------------------------------|:------------------|:---|
+| Transaction (Signed)                                | `0x54584E00`      | `TXN\0` |
+| Ledger Version (`ledger_hash`)                      | `0x4C575200`      | `LWR\0` |
+| [Ledger state tree information](ledger-header.html) | `0x4D4C4E00`      | `MLN\0` |
+| Transaction (Unsigned, for single-signing)          | `0x53545800`      | `STX\0` |
+| Transaction (Unsigned, for multi-signing)           | `0x534D5400`      | `SMT\0` |
+| Validation vote                                     | `0x56414C00`      | `VAL\0` |
+
+[Ledger objects IDs](ledger-object-ids.html) are calculated in a similar way, with 16-bit "space key" prefixes for each object type.
```
jwbusch: You use bytes to describe the size of prefixes before, but use bits here. What does "space key" mean? You might be able to omit the end of the sentence.

content/references/rippled-api/api-conventions/basic-data-types.md
```diff
@@ -21,6 +21,24 @@ Each closed [Ledger](ledger-data-formats.html) has a [Ledger Index][] and a [Has
 {% include '_snippets/data_types/hash.md' %}
 <!--{#_ #}-->
 
+### Hash Prefixes
+[[Source]<br>](https://github.com/ripple/rippled/blob/master/src/ripple/protocol/HashPrefix.h "Source")
+
+In many cases, the XRP Ledger prefixes an object's binary data with a 4-byte code before calculating its hash, so that objects of different types have different hashes even if their binary formats are the same. The 4-byte codes are structured as three bytes of ASCII followed by a zero byte.
+
+The following table shows the hash prefixes for some common types of data:
```
jwbusch: Do we have all the types vs. just some? i.e. can we just fill out this table to cover all the cases?

jwbusch (approved):

jhaaaa (approved):
Looks good to me - thank you!

content/references/rippled-api/api-conventions/basic-data-types.md
mDuo13: I waffled on this a bit, trying to strike a balance between providing everything and not providing info that users will almost certainly not need.

In the end I kept finding more cases (like signing Claims for payment channels) where users might want to know a certain prefix I had thought they wouldn't, so I went with including everything.

content/references/rippled-api/api-conventions/basic-data-types.md
```diff
@@ -21,6 +21,33 @@ Each closed [Ledger](ledger-data-formats.html) has a [Ledger Index][] and a [Has
 {% include '_snippets/data_types/hash.md' %}
 <!--{#_ #}-->
 
+### Hash Prefixes
+[[Source]<br>](https://github.com/ripple/rippled/blob/master/src/ripple/protocol/HashPrefix.h "Source")
+
+In many cases, the XRP Ledger prefixes an object's binary data with a 4-byte code before calculating its hash, so that objects of different types have different hashes even if their binary formats are the same. The 4-byte codes are structured as three bytes of ASCII followed by a zero byte.
```
nbougalis: >The 4-byte codes are structured as three bytes of ASCII followed by a zero byte.

Technically, this isn't required: the prefix is just four bytes long and, so far, we've used three ASCII characters and a zero byte, mostly because it simplfies debugging (you can just treat the thing as a string and dump it on the console or in a log file). But it could, just as well, be `0x4D454F57` (aka "MEOW").

Also note that "three bytes of ASCII" is a bit confusing. If you really want to say that, perhaps something like "current codes consist of 3 characters from the 26-character English alphabet (encoded as ASCII) followed by a 0 byte." You don't need to link to an ASCII table :smiley:

content/references/rippled-api/api-conventions/basic-data-types.md
```diff
@@ -21,6 +21,33 @@ Each closed [Ledger](ledger-data-formats.html) has a [Ledger Index][] and a [Has
...
+| Ledger state data                     | `account_state` in [ledger header][] | `0x4D4C4E00`      | `MLN\0` |
+| Ledger data inner node                | N/A                                  | `0x4D494E00`      | `MIN\0` |
+| Ledger data inner node ([SHAMapv2][]) | N/A                                  | `0x494E5200`      | `INR\0` |
+| Payment Channel Claim                 | N/A                                  | `0x434C4D00`      | `CLM\0` |
+| Signed Transaction                    | `hash` of transactions               | `0x54584E00`      | `TXN\0` |
+| Transaction with metadata             | N/A                                  | `0x534E4400`      | `SND\0` |
+| Unsigned Transaction (Single-signing) | N/A                                  | `0x53545800`      | `STX\0` |
+| Unsigned Transaction (Multi-signing)  | N/A                                  | `0x534D5400`      | `SMT\0` |
+| Validation vote                       | N/A                                  | `0x56414C00`      | `VAL\0` |
+| Validator ephemeral key update        | N/A                                  | `0x4D414E00`      | `MAN\0` |
```
nbougalis: Should we refer to this as a "Manifest"?

nbougalis (commented):
Looks good. Left some minor comments.

content/references/rippled-api/api-conventions/basic-data-types.md
mDuo13: I mean, I thought we were going to call it a ["subkey authorization" per Confluence](https://ripplelabs.atlassian.net/wiki/spaces/~wilsonianb/pages/106004557/Renaming+Manifests) but I guess maybe we missed the train on that one?

PR #486, opened 2018-12-14
Online deletion docs

mDuo13:
This PR includes the following changes (links to the staged versions, which require Ripple VPN):

- Adds [a new concept doc about online deletion](https://developers-staging.ripple.com/online-deletion.html) to replace [the dead wiki page](https://web.archive.org/web/20150422095042/https://wiki.ripple.com/Online_deletion)
- Wraps the online deletion and history sharding docs into a new subcategory called [Ledger History](https://developers-staging.ripple.com/ledger-history.html). (I moved some of the content from an earlier version of this PR into the "Ledger History" page.)
- Adds 3 new tutorials around online deletion:
    - [Changing the amount of history you store](https://developers-staging.ripple.com/configure-online-deletion.html)
    - [Using advisory delete with cron for scheduled deletions](https://developers-staging.ripple.com/configure-advisory-deletion.html)
    - [Storing full history](https://developers-staging.ripple.com/configure-full-history.html)
- Updates the Capacity Planning doc with [a table of how much disk space is required to store different amounts of history](https://developers-staging.ripple.com/capacity-planning.html#historical-data).

The `.uxf` files are source files for the diagrams, so they don't need to be directly reviewed.


content/tutorials/manage-the-rippled-server/configuration/configure-full-history.md
```diff
@@ -0,0 +1,98 @@
...
+
+    Your server can only download historical data from the peer-to-peer network if one its direct peers has the data available. The easiest way to ensure you can download full history is to peer with a server that already has full history.
+
+    **Tip:** Ripple makes a pool of full history servers publicly available. You can resolve the domain `s2.ripple.com` a few times to get the IP addresses of these servers. Ripple offers these servers as a public service, so be aware that their availability to peer with other servers is limited and you may be blocked if you abuse them.
+
+0. If you have a database dump from another full-history server to use as a basis, set the `[import_db]` stanza of your server's config file to point to the data to be imported. (Otherwise, skip this step.)
+
+        [import_db]
+      	type=RocksDB
+      	path=/tmp/full_history_dump/
```
mDuo13, line 56: The syntax for the `[import_db]` stanza was not clearly specified in the config file so I took an educated guess. Please confirm.

content/concepts/the-rippled-server/online-deletion.md
```diff
@@ -0,0 +1,126 @@
+# Online Deletion
+[[Source]<br/>](https://github.com/ripple/rippled/blob/master/src/ripple/app/misc/SHAMapStoreImp.cpp "Source")
+
+The online deletion feature lets the `rippled` server delete the server's local copy of old ledger versions to keep disk usage from rapidly growing over time. Online deletion happens automatically by default, but can be configured to run only when prompted. [New in: rippled 0.27.0][]
```
mtrippled: Online deletion is not on by default, it needs to be configured using the online_delete parameter in the [node_db] section. Then it's automatic unless configured to run only when triggered.

content/concepts/the-rippled-server/online-deletion.md
```diff
@@ -0,0 +1,126 @@
...
+The server always keeps the complete _current_ state of the ledger, with all the balances and settings it contains. The deleted data includes older transactions and versions of the ledger state that are older than the stored history.
+
+The default config file sets the `rippled` server to keep the most recent 2000 ledger versions and automatically delete older data.
+
+**Tip:** Even with online deletion, the amount of disk space required to store the same time span's worth of ledger data increases over time, because the size of individual ledger versions tends to grow over time. This growth is very slow in comparison to the accumulation of data that occurs without deleting old ledgers. For more information on disk space needs, see [Capacity Planning](capacity-planning.html).
+
+
+## Background
+
+The `rippled` server stores data from recent ledger versions in its _ledger store_. This data accumulates as the XRP Ledger network creates new ledger versions and validates them through the [consensus process](intro-to-consensus.html).
```
mtrippled: Do we call this "ledger store" or "node store"? (no opinion here either way but I think comments in source code as well as object names tend to refer to it as "node store".

content/concepts/the-rippled-server/online-deletion.md
```diff
@@ -0,0 +1,126 @@
...
+
+The XRP Ledger stores the complete state of the current ledger, with all balances, settings, and other data, in each ledger version. Unlike many cryptocurrency systems, an XRP Ledger server does not require full history to sync with the network, know the full state, and contribute to making forward progress.
+
+Inside the ledger store, ledger data is "de-duplicated", with data that doesn't change from version to version only stored once. The records themselves in the ledger store do not indicate which ledger version(s) contain them; part of the work of online deletion is identifying which records are only used by outdated ledger versions. This process is time consuming and affects the disk I/O and application cache, so it is not feasible to delete old data on every ledger close.
+
+### Fetching History
+
+When it starts, a `rippled` server's first priority is to get a complete copy of the latest validated ledger. From there, it keeps up with advances in the ledger progress. If configured to do so, the server also backfills ledger history up to a configured amount, which must be equal or less than the cutoff beyond which online deletion is configured to delete.
```
mtrippled: Default ledger_history is 256. It can be set as low as 0 (no ledger history to fetch). However, online_delete has a minimum of 256. online_delete must be equal to or greater than ledger_history. The only reason to configure ledger_history explicitly with also online_delete configured, is to fetch more history upon first initializing rippled. Otherwise, the server will always attempt to maintain at minimum the value configured for online_delete.

content/concepts/the-rippled-server/online-deletion.md
```diff
@@ -0,0 +1,126 @@
...
+
+The server can backfill history from before it became synced, as well as filling in any gaps in the history it has collected after syncing. (Gaps in ledger history can occur if a server temporarily becomes too busy to keep up with the network, loses its network connection, or suffers other temporary issues.) To backfill history, the server requests data from its peer `rippled` servers. The amount the server tries to backfill is defined by the `[ledger_history]` setting.
```
mtrippled: Actually, let me clarify the relationship further between online_delete, advisory_delete, and ledger_history.
If advisory_delete is not configured, then rippled will only seek to maintain ledger_history ledgers between online_delete intervals: if there is a gap earlier than what online_delete has been configured for, then rippled will not attempt to fill the gap.

If advisory_delete is configured, then rippled will attempt to ensure that gaps are filled between online_delete intervals.

content/concepts/the-rippled-server/online-deletion.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Online Deletion Behavior
+
+The online deletion settings configure how many ledger versions the `rippled` server should keep available in the ledger store at a time. However, the specified number is a guideline, not a hard rule:
+
+- The server may have less than the configured number of ledger versions if it has not been running for long enough or if it lost sync with the network at any time. (The server attempts to backfill at least some history; see [fetching history](#fetching-history) above for details.)
+- The server may store up to twice the configured number of ledger versions if online deletion is set to run automatically. (Each time it runs, it reduces the number of stored ledger versions to the configured number.)
+- If advisory delete is enabled, the server stores all the ledger versions that it has acquired and built until someone calls the [can_delete method][].
+
+    For example, if you call `can_delete` with a value of `now` once per day and an `online_delete` value of 2000, the server typically stores 2000 ledger versions at minimum and just over a full day's worth of ledger versions at maximum.
```
mtrippled: The server will store between 1 and 2 full days of ledgers in this case, assuming that >=2000 ledgers happen each day.

content/concepts/the-rippled-server/online-deletion.md
```diff
@@ -0,0 +1,126 @@
...
+
+With online deletion enabled and running automatically (that is, with advisory delete disabled), the total amount of ledger data stored should remain at minimum equal to the number of ledger versions the server is configured to keep, with the maximum being roughly twice that many.
+
+When online deletion runs, it does not reduce the size of the database files on disk; it only makes space within those files available to be reused for new data. To actually reduce the disk space allocated to the database files, you must restart `rippled`.
```
mtrippled: Running manual commands to vauum the SQLite databases is the only way to reduce the data size of sqlite databases. Node store size should reduce (rocksdb or nudb), but SQLITE won't shrink. Let's not advise that people manually reduce SQLite data files, it's time-consuming and risk prone if done incorrectly.

content/concepts/the-rippled-server/online-deletion.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Interrupting Online Deletion
+
+Online deletion automatically stops if the [server state](rippled-server-states.html) becomes less than `full`. If this happens, the server writes a log message with the prefix `SHAMapStore::WRN`. The server attempts to start online deletion again after the next validated ledger version after becoming fully synced.
+
+If you stop the server or it crashes while online deletion is running, online deletion resumes after the server is restarted and the server becomes fully synced.
+
+To temporarily disable online deletion, you can use the [can_delete method][] with an argument of `never`. This change persists until the server restarts or you re-enable online deletion by calling [can_delete][can_delete method] again. For more information on controlling when online deletion happens, see [Advisory Deletion](#advisory-deletion).
```
mtrippled: "can_delete never" persists indefinitely, between restarts, until can_delete sets another value.

content/concepts/the-rippled-server/online-deletion.md
```diff
@@ -0,0 +1,126 @@
...
+
+The following settings relate to online deletion:
+
+- **`online_delete`** - Specify a number of validated ledger versions to keep. The server periodically deletes any ledger versions that are older than this number. If not specified, no ledgers are deleted.
+
+    The default config file specifies 2000 for this value. This cannot be less than 256, because some events like [Fee Voting](fee-voting.html) and the [Amendment Process](amendments.html#amendment-process) update only every 256 ledgers.
+
+    **Caution:** If you run `rippled` with `online_delete` disabled, then later enable `online_delete` and restart the server, the server disregards but does not delete existing ledger history that your server already downloaded while `online_delete` was disabled. To save disk space, delete your existing history before re-starting the server with `online_delete` back on.
+
+- **`[ledger_history]`** - Specify a number of validated ledgers, equal to or less than `online_delete`. If the server does not have at least this many validated ledger versions, it attempts to backfill them by fetching the data from peers.
```
mtrippled: Actually, let me clarify the relationship further between online_delete, advisory_delete, and ledger_history.
If advisory_delete is not configured, then rippled will only seek to maintain ledger_history ledgers between online_delete intervals: if there is a gap earlier than what online_delete has been configured for, then rippled will not attempt to fill the gap.

If advisory_delete is configured, then rippled will attempt to ensure that gaps are filled between online_delete intervals.

content/concepts/the-rippled-server/online-deletion.md
```diff
@@ -0,0 +1,126 @@
...
+
+    The default for this setting is 256 ledgers.
+
+    The following diagram shows the relationship between `online_delete` and `ledger_history` settings:
+
+    ![Ledgers older than `online_delete` are automatically deleted. Ledgers newer than `ledger_history` are backfilled. Ledgers in between are kept if available but not backfilled](img/online_delete-vs-ledger_history.png)
```
mtrippled: Actually, let me clarify the relationship further between online_delete, advisory_delete, and ledger_history.
If advisory_delete is not configured, then rippled will only seek to maintain ledger_history ledgers between online_delete intervals: if there is a gap earlier than what online_delete has been configured for, then rippled will not attempt to fill the gap.

If advisory_delete is configured, then rippled will attempt to ensure that gaps are filled between online_delete intervals.

content/concepts/the-rippled-server/online-deletion.md
```diff
@@ -0,0 +1,126 @@
...
+
+    The following diagram shows how fetch_depth works:
+
+    ![Ledger versions older than fetch_depth are not served to peers](img/fetch_depth.png)
+
+For estimates of how much disk space is required to store different amounts of history, see [Capacity Planning](capacity-planning.html#historical-data).
+
+### Advisory Deletion
+
+By default, online deletion happens automatically and periodically. If the `advisory_delete` setting is enabled, online deletion only happens when an administrator triggers it using the [can_delete method][]. You can run this command with a scheduled job to trigger automatic deletion based on clock time instead of the number of ledger versions closed.
```
mtrippled: If online_delete is configured, it happens automatically. If not configured (the default), then it doesn't. advisory_delete has no meaning if it's not configured.

And can_delete can be run manually to maintain as much history is desired, and to delete only those earlier ledgers that you don't want any more.

content/concepts/the-rippled-server/online-deletion.md
```diff
@@ -0,0 +1,126 @@
...
+
+If your server is heavily used, the extra load from online deletion can cause your server to fall behind and temporarily de-sync from the consensus network. If this is the case, you can use advisory delete and schedule online deletion to happen only during off-peak times.
+
+You can also use advisory delete for other reasons, such as if you want to manually confirm that the transaction data is backed up to a separate server before deleting it.
+
+The `can_delete` API method can enable advisory deletion (with the value `never`) or disable advisory delete (with the value `always`). These setting changes persist until you restart the `rippled` server, at which point the settings in the config file override them.
```
mtrippled: They're persisted.
never: don't do online_delete
always: do online_delete as soon as the online_delete number of ledgers has been retained.
now: do online_delete at the next ledger close or as soon as the online_delete number of ledgers have been acquired, whichever is greater.
<integer> or <ledger hash>: Delete up to and including that ledger as soon as at least the online_delete number of ledgers have been acquired.

^^
What I just described is documented at: https://developers.ripple.com/can_delete.html

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
```diff
@@ -0,0 +1,96 @@
+# Configure Advisory Deletion
+
+In its default configuration, the `rippled` server automatically deletes outdated history of XRP Ledger state and transactions as new ledger versions become available. If your server uses most of its hardware resources during peak hours, you can configure the server to delete ledgers only when prompted by a command scheduled to run during off-peak hours to reduce the chances of online deletion
```
mtrippled: By "default" rippled does not do online_delete. It needs to be configured. However, the example config file shipped right now with the source code happens to configure online_delete.

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
```diff
@@ -0,0 +1,96 @@
...
+To configure advisory deletion with a daily schedule, perform the following steps:
+
+1. Enable `advisory_delete` in the `[node_db]` stanza of your `rippled`'s config file.
+
+        [node_db]
+        # Other settings unchanged ...
+      	online_delete=2000
+      	advisory_delete=1
+
+    - Set `advisory_delete` to `1` to require approval before running online deletion. (Set it to `0` to run online deletion automatically as new ledger versions become available.)
```
mtrippled: require the can_delete API call to trigger (not "approval" from anybody). Set to '0' to run online deletion once the configured number of ledgers are acquired.

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
```diff
@@ -0,0 +1,96 @@
...
+    {% include '_snippets/conf-file-location.md' %}<!--_ -->
+
+2. Test running the [can_delete method][] to prompt the server to run online deletion.
+
+    You can use the [`rippled` commandline interface](get-started-with-the-rippled-api.html#commandline) to run this command. For example:
+
+        $ rippled --conf=/etc/opt/ripple/rippled.cfg can_delete now
+
+    The response indicates the maximum ledger index that may the server may delete from its ledger store. For example, the following message indicates that ledger versions up to and including ledger index 43633667 are to be deleted:
+
```
mtrippled, line 52: are to be deleted once the latest ledger acquired is >= that value plus the value of online_delete.

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
```diff
@@ -0,0 +1,96 @@
...
+    Edit your `cron` configuration:
+
+        $ crontab -e
+
+    The following example sets the server to run deletion at 1:05 AM server time daily:
+
+        5 1 * * * rippled --conf /etc/opt/ripple/rippled.cfg can_delete now
+
+    Be sure that you schedule the command to run based on the time zone your server's clock is configured to use. Ripple recommends running all production servers with the UTC time zone.
```
mtrippled: Why tell people what time zone to use?

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
```diff
@@ -0,0 +1,96 @@
...
+
+    **Tip:** You do not need to schedule a `cron` job to run online deletion if you have `advisory_delete` disabled. In that case, `rippled` runs online deletion automatically when the server's oldest available ledger is twice as old as the number of ledgers to keep after deletion.
```
mtrippled: Actually, it's triggered once the difference between the earliest ledger acquired and the latest level acquired are at least the value of online_delete. This also means that, eventually, the oldest available ledger will be roughly twice that value.

content/tutorials/manage-the-rippled-server/configuration/configure-full-history.md
```diff
@@ -0,0 +1,98 @@
...
+      	open_files=2000
+      	filter_bits=12
+      	cache_mb=256
+      	file_size_mb=8
+      	file_size_mult=2
+        compression=1
+      	#online_delete=2000
+      	#advisory_delete=0
+
+    You can store full history with either RocksDB or NuDB as the key-value store behind the ledger store. NuDB requires less RAM but RocksDB requires less disk space. For more information, see [Capacity Planning](capacity-planning.html).
```
mtrippled: Don't store full history with rocksdb. Just nudb. Based on RAM utilization of each. RAM requirements of rocksdb for full history are unknown.

content/tutorials/manage-the-rippled-server/configuration/configure-full-history.md
```diff
@@ -0,0 +1,98 @@
...
+
+    If you are using RocksDB, add `compression=1` to enable Snappy compression, which reduces the disk space necessary to store full history (at a cost of greater CPU usage when reading or writing history).
```
mtrippled: It may also reduce CPU utilization because less time spent in I/O. Compression is enabled by default. There's no reason to not compress rocksdb unless somebody wants to test this. Recommend leaving it alone, also recommend not even mentioning it.

mtrippled (dismissed):

content/concepts/the-rippled-server/online-deletion.md
ximinez: Does "by default" mean "in the default / sample config file"?

mDuo13: That was the intention. Rewording to make that clearer.

content/concepts/the-rippled-server/online-deletion.md
mDuo13: The documentation calls it the "ledger store" because the previous term "node store" (still common in the code, etc.) has shown a history of being confusing.

content/concepts/the-rippled-server/online-deletion.md
mDuo13: You mean, the server backfills gaps in the ledger history up to the configured online_delete setting even if it's higher than the ledger_history setting? (I'm making a distinction between "gaps" and ledgers missing before the earliest ledger locally available, here.)

mDuo13: Clarification (from offline convo): this is only true if advisory_delete is on (as mentioned in the other comment below)

content/concepts/the-rippled-server/online-deletion.md
mDuo13: Gotcha, that's an outdated thing from the old wiki page.

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
mDuo13: Added a note following the code example to clarify this.

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -116,32 +117,49 @@ For best performance in enterprise production environments, Ripple recommends ru
...
-<!-- {# ***TODO: Update the dated storage consideration above, as needed. ***#} -->
-<!-- {# ***TODO: DOC-1331 tracks: Create historic metrics that a user can use to derive what will be required. For ex, a chart with 1TB in 2014, 3TB in 2015, 7TB in 2018 ***#} -->
+You can control how much data you keep with [online deletion](online-deletion.html); the default config file has the server keep the latest 2000 ledger versions. Without online deletion, the server's disk requirements grow without bounds.
 
-SSD storage should support several thousand of both read and write IOPS. Ripple engineers observed the following maximum reads and writes per second:
+The following table approximates the requirements for different amounts of history, at the time of writing (2018-12-13):
+
+| Real Time Amount | Number of Ledger Versions | Disk Space Required (RocksDB) | Disk Space Required (NuDB) |
+|:-----------------|:--------------------------|:------------------------------|:--|
+| 2 hours          | 2,000                     | ***(TBD)*** | ***(TBD)*** |
```
mDuo13: I'm going to test these and update later.

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
```diff
@@ -0,0 +1,98 @@
+# Configure Advisory Deletion
+
+The default config file sets `rippled` to automatically delete outdated history of XRP Ledger state and transactions as new ledger versions become available. If your server uses most of its hardware resources during peak hours, you can configure the server to delete ledgers only when prompted by a command scheduled to run during off-peak hours to reduce the chances of online deletion.
```
ximinez: "to run during off-peak hours to reduce the chances of online deletion"... doing what? I assume it's something like "running during peak hours", but the sentence reads awkwardly to me.

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
ximinez: This sentence reads awkwardly to me. Instead of "are to be deleted", I think it should say "are available to be deleted", "may be deleted", or something like that.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Interrupting Online Deletion
+
+Online deletion automatically stops if the [server state](rippled-server-states.html) becomes less than `full`. If this happens, the server writes a log message with the prefix `SHAMapStore::WRN`. The server attempts to start online deletion again after the next validated ledger version after becoming fully synced.
+
+If you stop the server or it crashes while online deletion is running, online deletion resumes after the server is restarted and the server becomes fully synced.
+
+To temporarily disable online deletion, you can use the [can_delete method][] with an argument of `never`. This change persists until you re-enable online deletion by calling [can_delete][can_delete method] again, even if you change the . For more information on controlling when online deletion happens, see [Advisory Deletion](#advisory-deletion).
```
ximinez: "even if you change the " what?

content/concepts/the-rippled-server/ledger-history/online-deletion.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Configuration
+
+The following settings relate to online deletion:
+
+- **`online_delete`** - Specify a number of validated ledger versions to keep. The server periodically deletes any ledger versions that are older than this number. If not specified, no ledgers are deleted.
+
+    The default config file specifies 2000 for this value. This cannot be less than 256, because some events like [Fee Voting](fee-voting.html) and the [Amendment Process](amendments.html#amendment-process) update only every 256 ledgers.
+
+    **Caution:** If you run `rippled` with `online_delete` disabled, then later enable `online_delete` and restart the server, the server disregards but does not delete existing ledger history that your server already downloaded while `online_delete` was disabled. To save disk space, delete your existing history before re-starting the server with `online_delete` back on.
```
ximinez: The converse is also true. Should that be mentioned here?

content/concepts/the-rippled-server/ledger-history/online-deletion.md
```diff
@@ -0,0 +1,107 @@
...
+
+For estimates of how much disk space is required to store different amounts of history, see [Capacity Planning](capacity-planning.html#disk-storage).
+
+### Advisory Deletion
+
+By default, online deletion happens automatically and periodically. If the `advisory_delete` setting is enabled in the config file, online deletion only happens when an administrator triggers it using the [can_delete method][].
+
+You can use advisory deletion with a scheduled job to trigger automatic deletion based on clock time instead of the number of ledger versions closed. If your server is heavily used, the extra load from online deletion can cause your server to fall behind and temporarily de-sync from the consensus network. If this is the case, you can use advisory deletion and schedule online deletion to happen only during off-peak times.
+
+Alternatively, you can use advisory deletion for other reasons, such as if you want to manually confirm that transaction data is backed up to a separate server before deleting it.
```
ximinez: Another reason is wanting to manually confirm that data is processed by another task before deleting it. Is that worth mentioning?

ximinez (changes requested):
A few suggestions and questions.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
```diff
@@ -0,0 +1,107 @@
...
+- The server may store up to twice the configured number of ledger versions if online deletion is set to run automatically. (Each time it runs, it reduces the number of stored ledger versions to approximately the configured number.)
+- If advisory deletion is enabled, the server stores all the ledger versions that it has acquired and built until its administrator calls the [can_delete method][].
+
+    If you call [can_delete][can_delete method] on a regular basis, the maximum amount of data the server stores is either of the following, whichever is larger:
+
+    - A number of ledger versions approximately equal to twice the `online_delete` value. (After deletion, this is reduced to approximately the `online_delete` value.)
+    - Ledger versions spanning an amount of time that is approximately twice the interval between `can_delete` calls. (After deletion, this is reduced to approximately one interval's worth of data.)
```
thejohnfreeman: If I understand this section correctly, l'd like to add a hint before each case, like this:

- If you call `can_delete` too frequently, a number of ledger versions approximately equal to twice the `online_delete` value. (After deletion, this is reduced to approximately the `online_delete` value.)
- If you call `can_delete` too infrequently, ledger versions spanning an amount of time that is approximately twice the interval between `can_delete` calls. (After deletion, this is reduced to approximately one interval's worth of data.)

content/concepts/the-rippled-server/ledger-history/online-deletion.md
```diff
@@ -0,0 +1,107 @@
...
+
+    For example, if you call `can_delete` with a value of `now` once per day and an `online_delete` value of 2000, the server typically stores up two full day's worth of ledger versions before running deletion. After running deletion, the server keeps approximately 1 day's worth, but never less than 2000 ledger versions.
```
thejohnfreeman: One correction: "never less than 2000" -> "never fewer than 2000"

One suggested addition: "the server keeps approximately 1 day's worth (around 20,000 ledger versions)"

content/concepts/the-rippled-server/ledger-history/online-deletion.md
```diff
@@ -0,0 +1,107 @@
...
+
+The following settings relate to online deletion:
+
+- **`online_delete`** - Specify a number of validated ledger versions to keep. The server periodically deletes any ledger versions that are older than this number. If not specified, no ledgers are deleted.
+
+    The default config file specifies 2000 for this value. This cannot be less than 256, because some events like [Fee Voting](fee-voting.html) and the [Amendment Process](amendments.html#amendment-process) update only every 256 ledgers.
+
+    **Caution:** If you run `rippled` with `online_delete` disabled, then later enable `online_delete` and restart the server, the server disregards but does not delete existing ledger history that your server already downloaded while `online_delete` was disabled. To save disk space, delete your existing history before re-starting the server with `online_delete` back on.
+
+- **`[ledger_history]`** - Specify a number of validated ledgers, equal to or less than `online_delete`. If the server does not have at least this many validated ledger versions, it attempts to backfill them by fetching the data from peers.
```
thejohnfreeman, line 67: 1. Why are `ledger_history` and `fetch_depth` in brackets, but `online_delete` and `advisory_delete` are not?

2. Am I right in thinking of `ledger_history` as a *minimum* number of validated ledgers to keep, which will be backfilled when necessary? If so, can we explicitly say "minimum" in its description?

content/concepts/the-rippled-server/ledger-history/ledger-history.md
```diff
@@ -0,0 +1,56 @@
+# Ledger History
+
+The [consensus process](intro-to-consensus.html) creates a chain of [validated ledger versions](ledgers.html), each derived from the previous one by applying a set of transactions. Every `rippled` server stores ledger versions and transaction history locally. The amount of transaction history a server stores depends on how long that server has been online and how much history it is configured to fetch and keep.
+
+Servers in the peer-to-peer XRP Ledger network share transactions and other data with each other as part of the consensus process. Each server each independently builds each new ledger version and compares results with its trusted validators to ensure consistency. (If a consensus of trusted validators disagrees with a server's results, that server fetches the necessary data from its peers to achieve consistency.) Servers can download older data from their peers to fill gaps in their available history. The structure of the ledger uses cryptographic [hashes](basic-data-types.html#hashes) of the data so that any server can verify the integrity and consistency of the data.
```
jhaaaa: Suggested nit edit - if it doesn't change your intended meaning:

>Each server each independently builds each new ledger version

to

>Each server independently builds each new ledger version

content/concepts/the-rippled-server/ledger-history/ledger-history.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Available History
+
+By design, all data and transactions in the XRP Ledger are public, and anyone can search or query anything. However, your server can only search data that it has available locally. If you try to query for a ledger version or transaction that your server does not have available, your server replies that it cannot find that data. Other servers that have the necessary history can respond successfully to the same query. If you have a business that uses XRP Ledger data, you should be mindful of how much history your server has available.
+
+The [server_info method][] reports how many ledger versions your server has available in the `complete_ledgers` field.
+
+## Fetching History
+
+When it starts, a `rippled` server's first priority is to get a complete copy of the latest validated ledger. From there, it keeps up with advances in the ledger progress. If configured to do so, the server also backfills ledger history up to a configured amount, which must be equal or less than the cutoff beyond which online deletion is configured to delete.
```
jhaaaa: suggested nit edit:

>which must be equal or less than the cutoff

to 

>which must be equal to or less than the cutoff

content/concepts/the-rippled-server/ledger-history/ledger-history.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Full History
+
+Some servers in the XRP Ledger network are configured as "full-history" servers. These servers, which require significantly more disk space than other tracking servers, collect all available XRP Ledger history and **do not use online deletion**.
+
+Ripple provides a set of public full-history servers as a public service at `s2.ripple.com`. This service is provided for the benefit of the larger XRP community. Ripple reserves the right to block those who abuse the servers or use more than their fair share of the servers' resources.
+
+**Tip:** Unlike some cryptocurrency networks, servers in the XRP Ledger do not need full history to know the current state and keep up with current transactions.
+
+For instructions setting up full history, see [Configure Full History](configure-full-history.html).
```
jhaaaa: Suggested edit:

>For instructions setting up full history,

to

>For instructions on setting up full history,

content/concepts/the-rippled-server/ledger-history/online-deletion.md
```diff
@@ -0,0 +1,107 @@
...
+The default config file sets the `rippled` server to keep the most recent 2000 ledger versions and automatically delete older data.
+
+**Tip:** Even with online deletion, the amount of disk space required to store the same time span's worth of ledger data increases over time, because the size of individual ledger versions tends to grow over time. This growth is very slow in comparison to the accumulation of data that occurs without deleting old ledgers. For more information on disk space needs, see [Capacity Planning](capacity-planning.html).
+
+
+## Background
+
+The `rippled` server stores [ledger history](ledger-history.html) in its _ledger store_. This data accumulates over time.
+
+Inside the ledger store, ledger data is "de-duplicated": data that doesn't change from version to version is only stored once. The records themselves in the ledger store do not indicate which ledger version(s) contain them; part of the work of online deletion is identifying which records are only used by outdated ledger versions. This process is time consuming and affects the disk I/O and application cache, so it is not feasible to delete old data on every ledger close.
```
jhaaaa: Nit - stray ":"? 

>ledger data is "de-duplicated": data

to

>ledger data is "de-duplicated" data

jhaaaa: Is this a valid suggested edit? 🤔 

>is "de-duplicated" data that doesn't change from version to version is only stored once.

to

>is "de-duplicated" data that doesn't change from version to version and is only stored once.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
jhaaaa: suggested nit edit:

>the server typically stores up two full day's worth

to

>the server typically stores up to two full day's worth

content/concepts/the-rippled-server/ledger-history/ledger-history.md
mtrippled: Default ledger_history is 256, but can be set as low as 0.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
mtrippled: If it's called too frequently (meaning before online_delete quantity of ledgers have passed), then can_delete has no effect. Rotation won't happen until at least the minimum configured online_delete ledgers have passed.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
mtrippled: Yeah, not clearly expressed here, whatever it is.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
mtrippled: In the config file, sections are enclosed in brackets whereas sub-items within a section are simply k=v. online_delete and advisory_delete are within the section [node_db].

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -55,45 +55,47 @@ The example `rippled-example.cfg` file has the `type` field in the `[node_db]` s
...
 ```
 
+(Adjust the `path` to the directory where you want to keep the ledger store on disk. Adjust the `online_delete` and `advisory_delete` settings as desired for your configuration.)
+
 #### More About Using NuDb
 
 [NuDB](https://github.com/vinniefalco/nudb#introduction) is an append-only key-value store that is optimized for SSD drives.
 
-NuDB has nearly constant performance and memory footprints regardless of the amount of data being [stored](#storage). NuDB _requires_ a solid-state drive, but uses much less RAM than RocksDB to access a large database.
+NuDB has nearly constant performance and memory footprints regardless of the [amount of data being stored](#disk-space). NuDB _requires_ a solid-state drive, but uses much less RAM than RocksDB to access a large database.
```
mtrippled, line 85: SSD's are recommended for all configurations, not just nudb.

content/references/rippled-api/public-rippled-methods/server-info-methods/server_info.md
```diff
@@ -283,7 +283,7 @@ The `info` object may have some arrangement of the following fields:
...
 | `build_version`                     | String                    | The version number of the running `rippled` version. |
 | `closed_ledger`                     | Object                    | (May be omitted) Information on the most recently closed ledger that has not been validated by consensus. If the most recently validated ledger is available, the response omits this field and includes `validated_ledger` instead. The member fields are the same as the `validated_ledger` field. |
-| `complete_ledgers`                  | String                    | Range expression indicating the sequence numbers of the ledger versions the local rippled has in its database. This may be a disjoint sequence, for example `24900901-24900984,24901116-24901158`. |
+| `complete_ledgers`                  | String                    | Range expression indicating the sequence numbers of the ledger versions the local rippled has in its database. This may be a disjoint sequence, for example `24900901-24900984,24901116-24901158`. If the server does not have any complete ledgers (for example, it just started syncing with the network), this is the string `empty`. |
```
jhaaaa: super nit:

>rippled

to

>`rippled`

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
```diff
@@ -0,0 +1,98 @@
...
+
+    Ubuntu Linux runs a `cron` daemon by default.
+
+    On RHEL or CentOS, you can install the `cronie` package:
+
+        $ sudo yum install cronie
+
+- Your server has enough disk space to store your chosen amount of history in its ledger store.
+
+    See [Capacity Planning](capacity-planning.html) for details of how much storage is required for different configuration. With advisory deletion enabled, the maximum history a server may accumulate before deletion is equal to the number of ledger versions configured in the `online_delete` setting **plus** the amount of time between online deletion prompts.
```
jhaaaa: suggested nit edit:

>for different configuration.

to

>for different configurations.

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
jhaaaa: suggested edit:

>maximum ledger index that may the server may delete from its ledger store.

to

>maximum ledger index that the server may delete from its ledger store.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
thejohnfreeman: I understand that the subsequent calls will (likely) not trigger a delete, but I get the impression that this section of documentation isn't about whether an individual call has an effect, it is about the upper bound on data stored as a function of call frequency. I would like a rule of thumb, and the way it's phrased is not very intuitive to me. Feel free to rephrase "too frequently" to "very frequently" or "frequently enough" or similar; it means the same to me. I'm trying to picture two extremes on a spectrum of `can_delete` frequency, and assign storage bounds to those extremes, to get a sense of the relationship between call frequency and storage cost.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
thejohnfreeman: Is `ledger_history` a config section? Why does the documentation describe it as a number? Do you just put the number on its own line under `[ledger_history]`? If so, I cannot wait for the TOML migration...

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
```diff
@@ -0,0 +1,98 @@
...
+4. Start (or restart) the `rippled` service.
+
+        $ sudo systemctl restart rippled
+
+5. Periodically check your server's `complete_ledgers` range using the [server_info method][] to confirm that ledgers are being deleted as scheduled.
+
+    The lowest ledger index in `complete_ledgers` should increase after online deletion.
+
+    Deletion may take several minutes to complete when it runs, depending on how busy your server is and how much history you delete at a time.
+
```
jhaaaa, line 85: I appreciate step 5 and the tips it provides. 🙌

content/concepts/the-rippled-server/ledger-history/online-deletion.md
thejohnfreeman: @jhaaaa Your suggested edit is valid, but I actually prefer the way it was written. I like short sentences, and the colon splits one long sentence into two short ones, but this is just my opinion.

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
```diff
@@ -0,0 +1,98 @@
...
+## Troubleshooting
+
+If online deletion does not seem to be running after configuring it, try the following:
+
+- Check that the user who configured the `cron` job has permissions to run the `rippled` server as a commandline client.
```
jhaaaa, line 90: This troubleshooting section is great!

For this particular troubleshooting tip above - would it be useful to mention this in the step for creating the `cron` job? Or is it obvious to a user who would be performing that task? 

And just to be sure I understand why this needs to be true -- the user that created the `cron` job will be the user used to issue the `can_delete` method at the scheduled time.

content/tutorials/manage-the-rippled-server/configuration/configure-full-history.md
jhaaaa: I know this is just an example and not sure if this matters - but maybe the `type` should be set to `NuDB`, just to reinforce our recommendation that you use NuDB for full history?

jhaaaa: Thank you for writing this tutorial, Rome - great info, so concise, and I think the topic is of great interest.

content/tutorials/manage-the-rippled-server/configuration/configure-online-deletion.md
```diff
@@ -0,0 +1,75 @@
...
+- Your server has [enough disk space](capacity-planning.html) to store your chosen amount of history in its ledger store.
+
+
+## Configuration Steps
+
+To change the amount of history your server stores, perform the following steps:
+
+1. Decide how many ledger versions' worth of history to store.
+
+    New ledger versions are usually validated 3 to 4 seconds apart, so the number of ledger versions corresponds roughly to the amount of time you want to store. See [Capacity Planning](capacity-planning.html) for details of how much storage is required for different configuration.
```
jhaaaa: suggested nit edit:

>for different configuration.

to

>for different configurations.

content/tutorials/manage-the-rippled-server/configuration/configure-online-deletion.md
```diff
@@ -0,0 +1,75 @@
...
+
+0. Wait for your server to sync to the network.
+
+    Depending on your network and system capabilities and how long your server was offline, it may take between 5 and 15 minutes to fully sync.
+
+    When your server is synced with the network, the [server_info method][] reports a `server_state` value of `"full"`, `"proposing"`, or `"validating"`.
+
+0. Periodically check your server's `complete_ledgers` range using the [server_info method][] to confirm that ledgers are being deleted.
+
+    After online deletion runs, the `complete_ledgers` range reflects that older ledgers are no longer available. As your server accumulates history, the total number of ledgers available should slowly increase to twice the `online_delete` value you configured, then decrease when online deletion runs.
```
jhaaaa, line 51: Great explanation of expected behavior.

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
intelliot: Maybe “to run during off-peak hours to mitigate the performance impact of online deletion.”

content/tutorials/manage-the-rippled-server/configuration/configure-full-history.md
mDuo13: Good catch on the RocksDB thing. I changed the `[node_db]` stanza but missed the later `[import_db]` example on my first pass. I think technically they _can_ be different types, but most likely if you're importing from another DB that has full history, it'll be NuDB→NuDB.

content/concepts/the-rippled-server/online-deletion.md
mDuo13: Reworded to make this more precise.

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
mDuo13: Good catch. This was a copy-edit failure. Changed to:

> If your server uses most of its hardware resources during peak hours, you can configure the server to delete ledgers only when prompted by a command scheduled to run during off-peak hours, so that online deletion is less likely to impact server performance.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
mDuo13: Good catch. Changed to:

> This change persists until you re-enable online deletion by calling [can_delete] again.

_Edit: Actually, disabling advisory deletion in the config file is kind of redundant with this anyway, so simplified again._

content/concepts/the-rippled-server/ledger-history/online-deletion.md
mDuo13: Good point. Changed the last sentence to:

> To save disk space, delete your existing history before re-starting the server after changing the `online_delete` setting.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
mDuo13: Good point. Rephrased:

> You can use advisory deletion for other reasons. For example, you may want to manually confirm that transaction data is backed up to a separate server before deleting it. Alternatively, you may want to manually confirm that a separate task has finished processing transaction data before you delete that data.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
mDuo13: Rewrote this section with an example for each case:

> The amount of data the server stores depends on how often you call [can_delete][can_delete method] and the how big an interval of time your `online_delete` setting represents:
>
>    - If you call `can_delete` _more often_ than your `online_delete` interval, the server stores at most a number of ledger versions approximately equal to **twice the `online_delete` value**. (After deletion, this is reduced to approximately the `online_delete` value.)
>
>        For example, if you call `can_delete` with a value of `now` once per day and an `online_delete` value of 50,000, the server typically stores up to 100,000 ledger versions before running deletion. After running deletion, the server keeps at least 50,000 ledger versions (about two days' worth). With this configuration, approximately every other `can_delete` call results in no change because the server does not have enough ledger versions to delete.
>
>    - If you call `can_delete` _less often_ than your `online_delete` interval, the server stores at most ledger versions spanning an amount of time that is approximately **twice the interval between `can_delete` calls**. (After deletion, this is reduced to approximately one interval's worth of data.)
>
>        For example, if you call `can_delete` with a value of `now` once per day and an `online_delete` value of 2000, the server typically stores up to two full days' worth of ledger versions before running deletion. After running deletion, the server keeps approximately one day's worth (about 25,000 ledger versions), but never fewer than 2000 ledger versions.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
mDuo13: Yes, `[ledger_history]` and `[fetch_depth]` are sections, and yes, you put the number on its own line under `[ledger_history]`.

And don't count on the config file changing soon. Backwards-compatibility and all.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
mDuo13: The colon was not stray. It indicates that the remainder is an independent clause that clarifies the preceding clause. I see where you mis-read "de-duplicated" as being an adjective that modifies the following data, but it's actually a predicate adjective that modifies the previous instance of "data".

I don't like connecting the clauses with "that" because it implies that there is another meaning or use of "de-duplicated".  To make it easier for others to avoid the same mistake, I've rephrased it into two separate sentences:

> Inside the ledger store, ledger data is "de-duplicated". In other words, data that doesn't change from version to version is only stored once.

content/concepts/the-rippled-server/ledger-history/ledger-history.md
mDuo13: That's good to know, but I'm going to leave it out here. (It would be more appropriate in a config reference than in this context.)

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
mDuo13: Yes, that is clarified a couple other places in this document, in particular lines 44-50 which lead into this section.

content/tutorials/manage-the-rippled-server/configuration/configure-advisory-deletion.md
mDuo13: Regarding your second point, yes, that's correct.

For the first point, I think that should be fine because the previous step has you run `rippled`, so I would expect that to come up before that point anyway. But it's in the troubleshooting in case users skipped past that.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Online Deletion Behavior
+
+The online deletion settings configure how many ledger versions the `rippled` server should keep available in the ledger store at a time. However, the specified number is a guideline, not a hard rule:
+
+- The server never deletes data more recent than the configured number of ledger versions, but it may have less than that amount available if it has not been running for long enough or if it lost sync with the network at any time. (The server attempts to backfill at least some history; see [fetching history](ledger-history.html#fetching-history) for details.)
+- The server may store up to twice the configured number of ledger versions if online deletion is set to run automatically. (Each time it runs, it reduces the number of stored ledger versions to approximately the configured number.)
+- If advisory deletion is enabled, the server stores all the ledger versions that it has acquired and built until its administrator calls the [can_delete method][].
+
+    The amount of data the server stores depends on how often you call [can_delete][can_delete method] and the how big an interval of time your `online_delete` setting represents:
```
jhaaaa: nit: change

>and the how big an interval

to

>and how big an interval

jhaaaa (approved):
Left just one more nit of an edit. Otherwise, this looks fantastic to me. Such a huge leap forward in providing crucial details about the behavior of online deletion. Thank you!

mDuo13:
Rebased on master and re-staged with the latest changes.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
```diff
@@ -0,0 +1,119 @@
...
+
+
+## Online Deletion Behavior
+
+The online deletion settings configure how many ledger versions the `rippled` server should keep available in the ledger store at a time. However, the specified number is a guideline, not a hard rule:
+
+- The server never deletes data more recent than the configured number of ledger versions, but it may have less than that amount available if it has not been running for long enough or if it lost sync with the network at any time. (The server attempts to backfill at least some history; see [fetching history](ledger-history.html#fetching-history) for details.)
+- The server may store up to twice the configured number of ledger versions if online deletion is set to run automatically. (Each time it runs, it reduces the number of stored ledger versions to approximately the configured number.)
```
mtrippled: Will store up to twice the configured ledger versions if online deletion is not delayed. If delayed, ledgers will continue to accumulate. In fact, online deletion takes awhile to complete, in which time a handful of new ledgers have been created. So in practice, slightly more than double will be held if all is functioning properly.

content/concepts/the-rippled-server/ledger-history/online-deletion.md
```diff
@@ -0,0 +1,119 @@
...
+- If advisory deletion is enabled, the server stores all the ledger versions that it has acquired and built until its administrator calls the [can_delete method][].
+
+    The amount of data the server stores depends on how often you call [can_delete][can_delete method] and how big an interval of time your `online_delete` setting represents:
+
+    - If you call `can_delete` _more often_ than your `online_delete` interval, the server stores at most a number of ledger versions approximately equal to **twice the `online_delete` value**. (After deletion, this is reduced to approximately the `online_delete` value.)
```
mtrippled, line 33: See above, it will be slightly above twice (unless online deletion is delayed).

mtrippled: Though I don't know if you want to be that particular. Maybe more generic language like "in the vicinity of" rather than "up to", because it could actually go a bit higher.

content/references/rippled-api/admin-rippled-methods/logging-and-data-management-methods/can_delete.md
```diff
@@ -47,7 +47,7 @@ The request includes the following optional parameter:
 
 | `Field`      | Type              | Description                               |
 |:-------------|:------------------|:------------------------------------------|
-| `can_delete` | String or Integer | The maximum ledger to allow to be deleted. For `ledger_index` or `ledger_hash`, see [Specifying Ledgers][]. `never` sets the value to 0, and effectively disables online deletion until another `can_delete` is appropriately called.  `always` sets the value to the maximum possible ledger (4294967295), and online deletion occurs as of each configured `online_delete` interval. `now` triggers online deletion at the next validated ledger that meets or exceeds the configured `online_delete` interval, but no further. |
+| `can_delete` | String or Integer | The [Ledger Index][] of the maximum ledger version to allow to be deleted. The special case `never` disables online deletion. The special case `always` enables automatic online deletion as if advisory deletion was disabled. The special case `now` allows online deletion one time at the next validated ledger that meets or exceeds the configured `online_delete` value. |
```
mtrippled: Also maybe note that can_delete returns the ledger sequence when executed. So if somebody wants to know that value without making any changes, they can simply call can_delete with no parameters.

content/tutorials/manage-the-rippled-server/configuration/configure-full-history.md
```diff
@@ -0,0 +1,92 @@
...
+      	path=/tmp/full_history_dump/
+
+0. Remove your server's existing database files, if you have any from previously running `rippled`.
+
+    After disabling online deletion, the server ignores any data that was downloaded while online deletion was enabled, so you may as well clear up the disk space. For example:
+
+        rm -r /var/lib/rippled/db/rocksdb/*
+        rm /var/lib/rippled/db/*.db
+        rm /var/lib/rippled/db/*.sqlite
+
```
mtrippled, line 65: I've seen SQLite sometimes createing files with a slightly different extension. Basically, wipe out the whole directory:
rm -fr /var/lib/rippled/db/*

content/tutorials/manage-the-rippled-server/installation/capacity-planning.md
```diff
@@ -53,47 +53,49 @@ The example `rippled-example.cfg` file has the `type` field in the `[node_db]` s
 
 #### More About Using RocksDB
 
-[RocksDB](https://rocksdb.org/docs/getting-started.html) is an embeddable persistent key-value store that is optimized for rotational disks.
+[RocksDB](https://rocksdb.org/docs/getting-started.html) is an embeddable persistent key-value store that works well with both rotational disks and solid state disks.
```
mtrippled: No rippled works well with rotational disks!

mtrippled:
:+1:

content/concepts/the-rippled-server/ledger-history/online-deletion.md
mDuo13: It already says "approximately equal" here, so I think that's fine as-is. I added a more detailed explanation above (around line 25) where you also pointed it out.

content/references/rippled-api/admin-rippled-methods/logging-and-data-management-methods/can_delete.md
mDuo13: Yes, the lines before and after this table made that clear, but I've changed it to be more consistent with how these tables are styled elsewhere in the docs.

PR #487, opened 2018-12-17
Improve doc for running a rippled validator

jhaaaa:
- Included step to open port 443 on the validator's associated web domain
- Restated task headings to be action-oriented
- Provided a bit more context for each task
- Tried to create a clear distinction between tasks that must be performed to simply run a `rippled` server as a validator and tasks that should be performed if you aspire to have your validator listed on a UNL and participate in consensus.

### DIFF
--- a/content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
+++ b/content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
@@ -1,74 +1,223 @@
 # Run rippled as a Validator
 
-Running a `rippled` validator that participates in the Consensus process is simple:
+A `rippled` server running in validator mode does everything a stock server does:
 
-1. [Enable validation](run-rippled-as-a-validator.html) on your `rippled` server.
-    * At first, your server is an _untrusted validator_. Others can see the validations your server issues, but they disregard them in the consensus process.
-2. Share the public key with the public, especially other `rippled` operators.
-3. When other `rippled` operators add your public key to their list of trusted servers, you have become a _trusted validator_.
-    * Also see [Properties of a Good Validator](rippled-server-modes.html#properties-of-a-good-validator) for best practices.
+- Connects to a network of peers
 
+- Relays cryptographically signed transactions
 
-## Validator Setup
+- Maintains a local copy of the complete shared global ledger
 
-The `validator-keys` tool (included in the `rippled` RPM) is the recommended means to securely generate and manage your validator keys.
+What makes a validator _different_ is that it also issues validation messages, which are sets of candidate transactions for evaluation by the XRP Ledger network during the [consensus process](consensus-principles-and-rules.html#how-consensus-works).
 
-1. [Install a `rippled` server.](install-rippled.html)
+It's important to understand that these validation messages are used only by `rippled` servers that explicitly trust your validator by listing it on their Unique Node Lists (UNLs). Servers that do not include your validator in their UNLs ignore your validator's messages in the consensus process. A validator that is not included in any UNL is called an _untrusted_ validator.
 
-2. Generate a validator key pair:
+For this reason, aside from getting your validator up and running, another key aspect of operating a validator is putting infrastructure in place that builds trust in your validator. Before validation list publishers, such as [https://vl.ripple.com](https://vl.ripple.com), can add your validator to their UNLs, they must first have trust in your validator. Once your validator has been added to a UNL, it is a _trusted_ validator that participates in the consensus process.
 
-        $ /opt/ripple/bin/validator-keys create_keys
+To run a `rippled` server as a validator, complete the following tasks:
 
-    **Warning:** Store the generated `validator-keys.json` key file in a secure but recoverable location, such as an encrypted USB flash drive. Do not modify its contents.
+1. Install a `rippled` server. For more information, see [Install rippled](install-rippled.html).
 
-3. Generate a validator token and edit your `rippled.cfg` file to add the `[validator_token]` value.
+2. [Enable validation on your `rippled` server](#enable-validation-on-your-rippled-server).
 
-        $ /opt/ripple/bin/validator-keys create_token --keyfile /path/to/your/validator-keys.json
+Beyond this, you can optionally complete the following tasks that help build trust in your validator:
 
-    If you had previously configured your validator without using the `validator-keys` tool, you must also delete the `[validation_seed]` from your `rippled.cfg` file. This changes your validator public key.
+1. [Understand the traits of a good validator](#understand-the-traits-of-a-good-validator).
 
-4. Start `rippled`:
+2. [Set up proxies to help protect your validator](#set-up-proxies-to-help-protect-your-validator) from [DDoS](https://en.wikipedia.org/wiki/Denial-of-service_attack) attacks.
 
-        $ sudo service rippled restart
-
-See [the `validator-keys-tool` GitHub repository](https://github.com/ripple/validator-keys-tool/blob/master/doc/validator-keys-tool-guide.md) for more information about managing validator keys.
+3. [Provide domain verification](#provide-domain-verification).
 
 
-## Public-Facing Server
 
-To protect a production validator from [DDoS](https://en.wikipedia.org/wiki/Denial-of-service_attack) attacks, you can use a stock `rippled` server as a proxy between the validator and the outside network.
+## Enable Validation on Your `rippled` Server
 
-1. [Set up the `rippled` validator.](run-rippled-as-a-validator.html)
+Enabling validation on your `rippled` server means providing a validator token in your server's `rippled.cfg` file. Ripple recommends using the `validator-keys` tool (included in `rippled` RPMs) to securely generate and manage your validator keys and tokens.
 
-2. [Set up one or more stock `rippled` servers.](install-rippled.html)
+In a location **not** on your validator:
 
-3. Configure the validator and stock `rippled` servers to be [clustered](cluster-rippled-servers.html) with each other.
+1. Manually build and run the `validator-keys` tool, if you don't already have it installed via a `rippled` RPM.
 
-4. Make the following configuration changes to your validator:
-    * Copy the `[ips_fixed]` list and paste it under `[ips]`. These fields should contain only the IP addresses and ports of the public-facing rippled(s). The validator connects to only these peers.
-    * Change `[peer_private]` to `1` to prevent its IP address from being forwarded.
+    For information about manually building and running the `validator-keys` tool, see [validator-keys-tool](https://github.com/ripple/validator-keys-tool).
 
-5. Configure the validator host machine's firewall to only accept inbound connections from its public-facing rippled(s).
+2. Generate a validator key pair using the `create_keys` command.
 
-Remember to restart `rippled` for config changes to take effect.
+        $ validator-keys create_keys
 
-Take care not to publish the IP address of your validator.
+      Sample output on Ubuntu:
+
+        Validator keys stored in /home/my-user/.ripple/validator-keys.json
+
+        This file should be stored securely and not shared.
+
+      Sample output on macOS:
+
+        Validator keys stored in /Users/my-user/.ripple/validator-keys.json
+
+        This file should be stored securely and not shared.
+
+      **Warning:** Store the generated `validator-keys.json` key file in a secure, offline, and recoverable location, such as an encrypted USB flash drive. Do not modify its contents. In particular, be sure to not store the key file on the validator where you intend to use the keys. If your validator's `secret_key` is compromised, [revoke the key](https://github.com/ripple/validator-keys-tool/blob/master/doc/validator-keys-tool-guide.md#key-revocation) immediately.
+
+      For more information about the `validator-keys` tool and the key pairs it generates, see the [Validator Keys Tool Guide](https://github.com/ripple/validator-keys-tool/blob/master/doc/validator-keys-tool-guide.md).
+
+3. Generate a validator token using the `create_token` command.
+
+        $ validator-keys create_token --keyfile /PATH/TO/YOUR/validator-keys.json
+
+      Sample output:
+
+        Update rippled.cfg file with these values:
+
+        # validator public key: nHUtNnLVx7odrz5dnfb2xpIgbEeJPbzJWfdicSkGyVw1eE5GpjQr
+
+        [validator_token]
+        eyJ2YWxpZGF0aW9uX3NlY3J|dF9rZXkiOiI5ZWQ0NWY4NjYyNDFjYzE4YTI3NDdiNT
+        QzODdjMDYyNTkwNzk3MmY0ZTcxOTAyMzFmYWE5Mzc0NTdmYT|kYWY2IiwibWFuaWZl
+        c3QiOiJKQUFBQUFGeEllMUZ0d21pbXZHdEgyaUNjTUpxQzlnVkZLaWxHZncxL3ZDeE
+        hYWExwbGMyR25NaEFrRTFhZ3FYeEJ3RHdEYklENk9NU1l1TTBGREFscEFnTms4U0tG
+        bjdNTzJmZGtjd1JRSWhBT25ndTlzQUtxWFlvdUorbDJWMFcrc0FPa1ZCK1pSUzZQU2
+        hsSkFmVXNYZkFpQnNWSkdlc2FhZE9KYy9hQVpva1MxdnltR21WcmxIUEtXWDNZeXd1
+        NmluOEhBU1FLUHVnQkQ2N2tNYVJGR3ZtcEFUSGxHS0pkdkRGbFdQWXk1QXFEZWRGdj
+        VUSmEydzBpMjFlcTNNWXl3TFZKWm5GT3I3QzBrdzJBaVR6U0NqSXpkaXRROD0ifQ==
+
+On your validator:
+
+1. Add `[validator_token]` and its value to your validator's `rippled.cfg` file.
+
+    If you previously configured your validator without the `validator-keys` tool, delete `[validation_seed]` and its value from your `rippled.cfg` file. This changes your validator public key.
+
+2. Restart `rippled`.
+
+        $ sudo systemctl restart rippled.service
+
+3. Use the `server_info` command to get information about your validator to verify that it is running as a validator.
+
+        $ rippled server_info
+
+      - The `pubkey_validator` value in the response should match the `public_key` in the `validator-keys.json` file that you generated for use with your validator.
+
+      - The `server_state` value should be _**proposing**_.
 
 
-## Domain Verification
 
-Network participants are unlikely to trust validators without knowing who is operating them. To address this concern, validator operators can associate their validator with a web domain that they control.
+## Understand the Traits of a Good Validator
 
-1. Find your validator public key by running the following on the validator server:
+Not every validator is likely to be widely trusted and validator list publishers may require validators to meet stringent criteria before they list them in their UNLs.
 
-        $ /opt/ripple/bin/rippled server_info -q | grep pubkey_validator
+Strive to have your validator always embody the following properties. The more of these properties your validator embodies, the more reasons validator list publishers have to include it on their UNLs.
 
-2. Sign the validator public key (from step 1) using the SSL private key used for your domain. The SSL private key file does not need to be stored on the validator server.
+- **Available**
 
-        $ openssl dgst -sha256 -hex -sign /path/to/your/ssl.key <(echo <your-validator-public-key>)
+    A good validator is always running and submitting validation votes for every proposed ledger. Strive for 100% uptime.
 
-3. Using `validator-keys` tool (included in the `rippled` RPM), sign the domain name:
+- **In agreement**
 
-        $ /opt/ripple/bin/validator-keys --keyfile /path/to/your/validator-keys.json sign <your-domain-name>
+    A good validator's votes match the outcome of the consensus process as often as possible. To do otherwise could indicate that your validator's software is outdated, buggy, or intentionally biased. Always run the [latest `rippled` release](https://github.com/ripple/rippled/tree/master) without modifications. [Watch `rippled` releases](https://github.com/ripple/rippled/releases) to be notified of new releases.
 
-4. To have the verified validator domain included on [XRP Charts](https://xrpcharts.ripple.com/#/validators), submit this [Google Form](https://docs.google.com/forms/d/e/1FAIpQLScszfq7rRLAfArSZtvitCyl-VFA9cNcdnXLFjURsdCQ3gHW7w/viewform) with your validator's information.
+- **Issuing timely votes**
+
+    A good validator's votes arrive quickly and not after a consensus round has already finished. To keep your votes timely, make sure your validator meets the recommended [system requirements](system-requirements.html), which include a fast internet connection.
+
+- **Identified**
+
+    A good validator has a clearly identified owner. Providing [domain verification](#provide-domain-verification) is a good start. Ideally, XRP Ledger network UNLs include validators operated by different owners in multiple legal jurisdictions and geographic areas. This reduces the chance that any localized events could interfere with the impartial operations of trusted validators.
+
+It's worth noting that running an untrusted validator can also be helpful to the overall health of the network. Untrusted validators help set the standard that trusted validators are measured against. For example, if a trusted validator is disagreeing with a lot of untrusted validators, that might indicate a problem.
+
+Ripple (the company) publishes a [validator list](https://github.com/ripple/rippled/blob/develop/cfg/validators-example.txt) with a set of recommended validators. Ripple strongly recommends using exactly this list for production servers.
+
+
+
+## Set Up Proxies to Help Protect Your Validator
+
+To help protect a validator from [DDoS](https://en.wikipedia.org/wiki/Denial-of-service_attack) attacks, set up one or more stock `rippled` servers as proxies between your validator and inbound and outbound traffic.
+
+**Note:** While these servers are acting as proxies, they are not web proxies for HTTP(S) traffic.
+
+<!-- { TODO: Future: add a recommended network architecture diagram to represent the proxy, clustering, and firewall setup: https://ripplelabs.atlassian.net/browse/DOC-2046 }-->
+
+1. [Enable validation](#enable-validation-on-your-rippled-server) on your `rippled` server.
+
+2. Set up stock `rippled` servers. For more information, see [Install rippled](install-rippled.html).
+
+3. Configure your validator and stock `rippled` servers to run in a [cluster](cluster-rippled-servers.html).
+
+4. In your validator's `rippled.cfg` file:
+
+    1. Copy the `[ips_fixed]` IP address list you defined in the previous cluster configuration step and paste it under `[ips]`.
+
+        For this purpose, the `[ips_fixed]` and `[ips]` values must be identical and contain only the IP addresses and ports of the stock `rippled` servers you clustered with your validator. This ensures that your validator connects to only the stock `rippled` servers that you control.
+
+    2. Set `[peer_private]` to `1` to prevent your validator's IP address from being forwarded. For more information, see [Private Peers](peer-protocol.html#private-peers).
+
+        **Warning:** Be sure that you don't publish your validator's IP address in other ways.
+
+5. Configure your validator host machine's firewall to allow the following traffic only:
+
+    - Inbound traffic: Only from IP addresses of the stock `rippled` servers in the cluster you configured.
+
+    - Outbound traffic: Only to the IP addresses of the stock `rippled` servers in the cluster you configured and to <https://vl.ripple.com> through port 443.
+
+6. Restart `rippled`.
+
+        $ sudo systemctl restart rippled.service
+
+7. Use the [Peer Crawler](peer-protocol.html#peer-crawler) endpoint on one of your stock `rippled` servers. The response should not include your validator. This verifies that your validator's `[peer_private]` configuration is working. One of the effects of enabling `[peer_private]` on your validator is that your validator's peers do not include it in their Peer Crawler results.
+
+
+
+## Provide Domain Verification
+
+To help validation list publishers and other participants in the XRP Ledger network understand who operates your validator, provide domain verification for your validator. At a high level, domain verification is a two-way link:
+
+- Use your domain to claim ownership of a validator key.
+
+- Use your validator key to claim ownership of a domain.
+
+Creating this link establishes strong evidence that you own both the validator key and the domain.
+
+To provide domain verification:
+
+1. Choose a domain name you own that you want to be publicly associated with your validator. You must run a public-facing HTTPS server on port 443 of that domain and you must have access to the private key file associated with that server's TLS certificate. (Note: [TLS was formerly called SSL](https://en.wikipedia.org/wiki/Transport_Layer_Security).)
+
+2. Share your validator's public key with the public, especially other `rippled` operators. For example, you can share your validator's public key on your website, on social media, in the [XRPChat community forum](https://www.xrpchat.com/), or in a press release.
+
+3. Submit a request to have your validator listed in XRP Charts' [Validator Registry](https://xrpcharts.ripple.com/#/validators) using this [Google Form](https://docs.google.com/forms/d/e/1FAIpQLScszfq7rRLAfArSZtvitCyl-VFA9cNcdnXLFjURsdCQ3gHW7w/viewform). Having your validator listed in this registry provides another form of public verification that your validator and domain are owned by you. To complete the form, you'll need the following information:
+
+      1. Find your validator public key by running the following on the validator server.
+
+            $ /opt/ripple/bin/rippled server_info | grep pubkey_validator
+
+        Provide the value returned in the **Validator Public Key** field of the Google Form.
+
+      2. Sign the validator public key using your web domain's TLS private key. The TLS private key file does not need to be stored on your validator's server.
+
+            $ openssl dgst -sha256 -hex -sign /PATH/TO/YOUR/TLS.key <(echo YOUR_VALIDATOR_PUBLIC_KEY_HERE)
+
+        Sample output:
+
+            4a8b84ac264d18d116856efd2761a76f3f4544a1fbd82b9835bcd0aa67db91c53342a1ab197ab1ec4ae763d8476dd92fb9c24e6d9de37e3594c0af05d0f14fd2a00a7a5369723c019f122956bf3fc6c6b176ed0469c70c864aa07b4bf73042b1c7cf0b2c656aaf20ece5745f54ab0f78fab50ebd599e62401f4b57a4cccdf8b76d26f4490a1c51367e4a36faf860d48dd2f98a6134ebec1a6d92fadf9f89aae67e854f33e1acdcde12cfaf5f5dbf1b6a33833e768edbb9ff374cf4ae2be21dbc73186a5b54cc518f63d6081919e6125f7daf9a1d8e96e3fdbf3b94b089438221f8cfd78fd4fc85c646b288eb6d22771a3ee47fb597d28091e7aff38a1e636b4f
+
+        Provide the value returned in the **SSL Signature** field of the Google Form.
+
+      3. Using the [`validator-keys` tool](https://github.com/ripple/validator-keys-tool/blob/master/doc/validator-keys-tool-guide.md) (included in `rippled` RPMs), sign the domain name.
+
+            $ validator-keys --keyfile /PATH/TO/YOUR/validator-keys.json sign YOUR_DOMAIN_NAME
+
+        Sample output:
+
+            E852C2FE725B64F353E19DB463C40B1ABB85959A63B8D09F72C6B6C27F80B6C72ED9D5ED6DC4B8690D1F195E28FF1B00FB7119C3F9831459F3C3DE263B73AC04
+
+        Provide the value returned in the **Domain Signature** field of the Google Form.
+
+4. After submitting the completed Google Form, you'll receive an email from XRP Charts that tells you whether your domain verification succeeded or failed. If your domain verification succeeds, XRP Charts lists your validator and domain in its [Validator Registry](https://xrpcharts.ripple.com/#/validators).
+
+<!--{ ***TODO: For the future - add a new section or separate document: "Operating a Trusted Validator" -- things that you need to be aware of once your validator has been added to a UNL and is participating in consensus. We should tell the user what to expect once they are listed in a UNL. How to tell if your validator is participating in the consensus process? How to tell if something isn't right with your validator - warning signs that they should look out for? How to tell if your validator has fallen out of agreement - what is the acceptable vs unacceptable threshold? Maybe provide a script that will alert them when something is going wrong.*** }-->
+
+
+
+## Revoke Validator Keys
+
+If your validator's master private key is compromised, you must revoke it immediately and permanently.
+
+For information about how to revoke a master key pair you generated for your validator using the `validator-keys` tool, see [Key Revocation](https://github.com/ripple/validator-keys-tool/blob/master/doc/validator-keys-tool-guide.md#key-revocation).
--- a/content/use-cases/run-a-rippled-validator.md
+++ b/content/use-cases/run-a-rippled-validator.md
@@ -37,14 +37,14 @@ To configure your `rippled` server to run in validator mode, generate a validato
 
 <span class="use-case-step-num">{{n.next()}}</span>
 <!-- <span class="use-case-step-length">(1 hour)</span> -->
-## [Set up a stock rippled server as a proxy](run-rippled-as-a-validator.html#public-facing-server)
+## [Set up a stock rippled server as a proxy](run-rippled-as-a-validator.html#set-up-proxies-to-help-protect-your-validator)
 
 To protect a production validator from DDoS attacks, you can use a stock `rippled` server as a proxy between the validator and the outside network.
 
 
 <span class="use-case-step-num">{{n.next()}}</span>
 <!-- <span class="use-case-step-length">(1 hour)</span> -->
-## [Associate your validator with a web domain you control](run-rippled-as-a-validator.html#domain-verification)
+## [Associate your validator with a web domain you control](run-rippled-as-a-validator.html#provide-domain-verification)
 
 Network participants are unlikely to trust a validator without knowing who is operating it. To address this concern, associate your validator with a web domain you control.
 You may also wish to have your validator listed with one or more validator tracking services, such as the <a href="https://xrpcharts.ripple.com/#/validators" target="_blank">XRP Charts Validator Registry <i class="fa fa-external-link" aria-hidden="true"></i></a>.<!--#{ fix md highlighting_ #}-->
--- a/content/concepts/the-rippled-server/rippled-server-modes.md
+++ b/content/concepts/the-rippled-server/rippled-server-modes.md
@@ -34,21 +34,8 @@ Not all `rippled` servers need to be validators: trusting more servers from the 
 
 If your organization runs a validating server, you may also run one or more stock servers, to balance the computing load of API access, or as a proxy between your validation server and the outside network.
 
+For more information about running a validator, see [Run `rippled` as a Validator](run-rippled-as-a-validator.html).
 
-### Properties of a Good Validator
-
-There are several properties that define a good validator. The more of these properties your server embodies, the more reason others have to include your server in their list of trusted validators:
-
-* **Availability**. An ideal validator should always be running, submitting validation votes for every proposed ledger.
-    * Strive for 100% uptime.
-* **Agreement**. A validator's votes should match the outcome of the consensus process as often as possible. To do otherwise could indicate that the validator's software is outdated, buggy, or intentionally biased.
-    * Always run the latest `rippled` release without modifications.
-* **Timeliness**. A validator's votes should arrive quickly, and not after a consensus round has already finished.
-    * A fast internet connection helps with this.
-* **Identified**. It should be clear who runs the validator. Ideally, a list of trusted validators should include validators operated by different owners in multiple legal jurisdictions and geographic areas, to reduce the chance that any localized events could interfere with the validator's impartial operation.
-    * Setting up [Domain Verification](run-rippled-as-a-validator.html#domain-verification) is a good start.
-
-At present, Ripple (the company) cannot recommend any validators aside from those in the default validator list. However, we are collecting data on other validators and building tools to report on their performance. For metrics on validators, see the [XRPCharts Validator Registry](https://xrpcharts.ripple.com/#/validators).
 
 
 ## Reasons to Run a `rippled` Server in Stand-Alone Mode
--- a/content/references/rippled-api/admin-rippled-methods/key-generation-methods/validation_create.md
+++ b/content/references/rippled-api/admin-rippled-methods/key-generation-methods/validation_create.md
@@ -7,7 +7,7 @@ _The `validation_create` method is an [admin method](admin-rippled-methods.html)
 
 You can configure your server to use the generated key pair to sign validations (validation key pair) or regular peer-to-peer communications (node key pair).
 
-**Tip:** For configuring a robust validator, you should use the `validator-keys` tool (included in the `rippled` RPM) to generate validator tokens (which can be rotated) with an offline master key. For more information, see [Validator Setup](run-rippled-as-a-validator.html#validator-setup).
+**Tip:** For configuring a robust validator, you should use the `validator-keys` tool (included in the `rippled` RPM) to generate validator tokens (which can be rotated) with an offline master key. For more information, see [Validator Setup](run-rippled-as-a-validator.html#enable-validation-on-your-rippled-server).
 
 
 ### Request Format


content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
 # Run rippled as a Validator
 
-Running a `rippled` validator that participates in the Consensus process is simple:
+A `rippled` server running in validator mode does everything a stock server does: ***TODO: I see stock servers referred to as "tracking servers" too - should we always call them one or the other? "Tracking servers" is a little more descriptive and it used in the Consensus article: https://developers.ripple.com/consensus.html. Or would it be confusing because "tracking" is also a server state: 	The server is in agreement with the network?***
```
garay-daniel: Lets avoid tracking for now (I also like it more), this is something we need to align on with marketing.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
 
-1. [Install a `rippled` server.](install-rippled.html)
+It's important to understand that these validation messages are used only by `rippled` servers that explicitly trust your validator by listing it on their Unique Node Lists (UNLs). Until your validator is listed on a UNL, it is considered an _untrusted_ validator and its validation messages are disregarded in the consensus process.
 
-2. Generate a validator key pair:
+For this reason, aside from getting your validator up and running, another key aspect of operating a validator is putting infrastructure in place that engenders trust in your validator. Before validation list publishers can add your validator to their UNLs, they must first have trust in your validator. Once your validator has been added to a UNL, it is a _trusted_ validator that participates in the consensus process. ***TODO: what do we mean by "validator list publishers"? Is this specifically about getting a validator listed in this blob? https://vl.ripple.com/ Why is it in a blob and can we provide guidance about how to decode it? Whatever is in this list would be reflected on XRPCharts' list of UNL validators, correct?***
```
garay-daniel: Anyone can publish a list to follow. Ours is the default list and the one that is trusted by most of the network. Nodes should follow lists that are very similar to avoid forking. There are various websites that decode Ripple's UNL, not sure how to do it directly.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+
+For more information about using the `validator-keys` tool, see the [Validator Keys Tool Guide](https://github.com/ripple/validator-keys-tool/blob/master/doc/validator-keys-tool-guide.md).
+
+1. Generate a validator key pair.
 
         $ /opt/ripple/bin/validator-keys create_keys
 
-    **Warning:** Store the generated `validator-keys.json` key file in a secure but recoverable location, such as an encrypted USB flash drive. Do not modify its contents.
+    **Warning:** Store the generated `validator-keys.json` key file in a secure but recoverable location, such as an encrypted USB flash drive. Do not modify its contents. ***TODO: Is the main point here that this is an offline location? Specifically, is it not secure enough to keep it in the default location it is generated in? Like if I'm running my server in AWS EC2 - should I move the json file to my flash drive and delete it from EC2?***
```
garay-daniel: Yes, you want a secure offline location. If someone finds your private validator key, they can impersonate your validator and sign validations on your behalf.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+
+3. Add `[validator_token]` and its value to your `rippled.cfg` file. ***TODO: Is it okay to also put the "# validator public key: XXXXX" text in the cfg file?***
+
+    If you previously configured your validator without the `validator-keys` tool, delete `[validation_seed]` and its value from your `rippled.cfg` file. This changes your validator public key. ***TODO: validation_seed has been disabled since 0.29.1 -- can we safely delete this text?***
+
+5. Restart `rippled`.
+
+        $ sudo systemctl restart rippled.service
+
+***TODO: how can I tell that my server is behaving as a validator? what command can tell me that my server is issuing validation messages? How can I look at a list of validations on the network and see if my validator is associated with some of them?***
```
garay-daniel: server_info has a pubkey_validator field, which should have your public key in it if you are validating. Your server state should also be "proposing".

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
 
-4. Make the following configuration changes to your validator:
-    * Copy the `[ips_fixed]` list and paste it under `[ips]`. These fields should contain only the IP addresses and ports of the public-facing rippled(s). The validator connects to only these peers.
-    * Change `[peer_private]` to `1` to prevent its IP address from being forwarded.
+To help protect a validator from [DDoS](https://en.wikipedia.org/wiki/Denial-of-service_attack) attacks, set up one or more stock `rippled` servers as proxies between the validator and inbound traffic. ***TODO: Stated correctly? I think inbound traffic is clearer than "outside network" - but is it accurate? Do we have any recommendations around the minimum number of proxies to set up? Is the primary purpose of providing more than one proxy that if one goes down, you don't lose your connection to the network? Also curious about how the validator's "view" of the network changes when proxies are in place -- what does inbound traffic to a proxy-less validator look like vs inbound traffic to a validator with two proxies, for example? I don't quite understand how the views could differ, if at all.***
```
garay-daniel: This is correct. You want to set peer_private to avoid incoming connections and then setup your own outgoing connections to a proxy (rippled server). We could also instruct validators they can connect to our public hubs here as well, these will ensure more peers and higher connectivity. @mDuo13 @jhaaaa 

In general a "connect to the network" section would be nice for validators.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+
+        $ sudo systemctl restart rippled.service
+
+***TODO: should we provide some tips on how to test that the firewall has been configured correctly? maybe ping the validator IP address from a machine outside the cluster? The test that the cluster is set up correctly (peers command) is covered in the clustering doc.***
+
+
+
+## Request Domain Verification
+
+***TODO: I'm unclear on what constitutes domain verification. Is it simply associating the validator with a web domain - like putting your public key on a website and claiming ownership? Or is it additionally getting your validator listed on XRP Charts or some other site like https://bithomp.com/validators or https://peerisland.com/validators?***
```
garay-daniel: Domain verification is associating your validator with a domain name. Right now, that can only happen by going through us and having us "check your work" essentially.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+
+To help validation list publishers and other participants in the XRP Ledger network understand who operates your validator, associate your validator with a web domain that you control and then apply to have your validator listed on XRP Charts' [Validator Registry](https://xrpcharts.ripple.com/#/validators). This provides a way for the public to verify your ownership of your validator.
+
+1. Set up the web domain that you want to use to publicly share your ownership of your validator. Secure the domain with SSL and import the SSL certificate into the domain. Ensure that the web domain's port 443 is open. You'll need the SSL private key file to complete a subsequent step.
+
+2. Share your validator's public key with the public, especially other `rippled` operators. ***TODO: what are recommended ways of doing this? put the public key in an obvious location on the web domain you are associating with your validator? request to have your validator listed on XRPCharts -- which will list its public key?***
```
garay-daniel: You could put it on your website, make a blog post, tweet, etc.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+
+3. Submit a request to have your validator listed in XRP Charts' [Validator Registry](https://xrpcharts.ripple.com/#/validators) using this [Google Form](https://docs.google.com/forms/d/e/1FAIpQLScszfq7rRLAfArSZtvitCyl-VFA9cNcdnXLFjURsdCQ3gHW7w/viewform). ***TODO: is this step of listing it on XRP Charts the thing that will make it a "verified validator domain"? Before this step, all I've done is sign values locally. The link to the doc in this form needs to be updated to point to this section. On XRP Charts, why are there listings with no Domain values? What do the chainXX values mean?*** To complete the form, you'll need the following information:
```
garay-daniel: Right now the only way to be "verified" is for us to check your work through the google form. The end result is being listed on XRP charts. In the near future, we will change this system so anyone can check your work and verify you, this will hopefully mean more dashboards where validators are listed.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
         $ openssl dgst -sha256 -hex -sign /path/to/your/ssl.key <(echo <your-validator-public-key>)
+        ```
+
+        Sample output:
+
+        ```
+        4a8b84ac264d18d116856efd2761a76f3f4544a1fbd82b9835bcd0aa67db91c53342a1ab197ab1ec4ae763d8476dd92fb9c24e6d9de37e3594c0af05d0f14fd2a00a7a5369723c019f122956bf3fc6c6b176ed0469c70c864aa07b4bf73042b1c7cf0b2c656aaf20ece5745f54ab0f78fab50ebd599e62401f4b57a4cccdf8b76d26f4490a1c51367e4a36faf860d48dd2f98a6134ebec1a6d92fadf9f89aae67e854f33e1acdcde12cfaf5f5dbf1b6a33833e768edbb9ff374cf4ae2be21dbc73186a5b54cc518f63d6081919e6125f7daf9a1d8e96e3fdbf3b94b089438221f8cfd78fd4fc85c646b288eb6d22771a3ee47fb597d28091e7aff38a1e636b4f
+        ```
+
+        Provide the value returned in the **SSL Signature** field of the Google Form. ***TODO: correct? Is this step solely meant to provide the value that you need to put in the SSL Signature field of the XRP Charts Google Form? Or is there another purpose for generating the value?***
```
garay-daniel: Used to respond to the form.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
         $ /opt/ripple/bin/validator-keys --keyfile /path/to/your/validator-keys.json sign <your-domain-name>
+        ```
+
+        Sample output:
+
+        ```
+        E852C2FE725B64F353E19DB463C40B1ABB85959A63B8D09F72C6B6C27F80B6C72ED9D5ED6DC4B8690D1F195E28FF1B00FB7119C3F9831459F3C3DE263B73AC04
+        ```
+
+        Provide the value returned in the **Domain Signature** field of the Google Form. ***TODO: same question as above - is this only for filling out the google form? Or am I supposed to do something else with this value?***
```
garay-daniel: For the form

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+
+  ***TODO: we should tell the user what to expect after submitting the form. Who reviews the request and based on what criteria? How long is the wait time for request approval or denial? There are validators listed on XRPCharts that don't have domain names listed - based on our recommendations in this doc, how did those validators get approved for listing?***
```
garay-daniel: Reviews automatically. We simply make sure the signatures match the certificate we can reach via their domain (via a script that runs every so often). We have automated emails that we send stating if you were successful or failed to verify.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+
+  ***TODO: we should tell the user how to tell that their validator has been added to a UNL - whether it is the default UNL or another UNL.***
+
+<!--{ ***TODO: For the future - add a new section or separate document: "Operating a Trusted Validator" -- things that you need to be aware of once your validator has been added to a UNL and is participating in consensus. We should tell the user what to expect once they are listed in a UNL. How to tell if your validator is participating in the consensus process? How to tell if something isn't right with your validator - warning signs that they should look out for? How to tell if your validator has fallen out of agreement - what is the acceptable vs unacceptable threshold? Maybe provide a script that will alert them when something is going wrong.*** }-->
```
garay-daniel, line 215: Great idea.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+
+        ```
+        E852C2FE725B64F353E19DB463C40B1ABB85959A63B8D09F72C6B6C27F80B6C72ED9D5ED6DC4B8690D1F195E28FF1B00FB7119C3F9831459F3C3DE263B73AC04
+        ```
+
+        Provide the value returned in the **Domain Signature** field of the Google Form. ***TODO: same question as above - is this only for filling out the google form? Or am I supposed to do something else with this value?***
+
+  ***TODO: we should tell the user what to expect after submitting the form. Who reviews the request and based on what criteria? How long is the wait time for request approval or denial? There are validators listed on XRPCharts that don't have domain names listed - based on our recommendations in this doc, how did those validators get approved for listing?***
+
+  ***TODO: we should tell the user how to tell that their validator has been added to a UNL - whether it is the default UNL or another UNL.***
```
garay-daniel: Right now being added to a UNL is a fairly direct process where we talk extensively before it happens, so they will know :)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+- Relays cryptographically signed transactions
 
-## Validator Setup
+- Maintains a local copy of the complete shared global ledger
 
-The `validator-keys` tool (included in the `rippled` RPM) is the recommended means to securely generate and manage your validator keys.
+What makes a validator _different_ is that it also issues validation messages, which are sets of candidate transactions for evaluation by the XRP Ledger network during the [consensus process](consensus-principles-and-rules.html#how-consensus-works).
 
-1. [Install a `rippled` server.](install-rippled.html)
+It's important to understand that these validation messages are used only by `rippled` servers that explicitly trust your validator by listing it on their Unique Node Lists (UNLs). Until your validator is listed on a UNL, it is considered an _untrusted_ validator and its validation messages are disregarded in the consensus process.
```
mDuo13: > Until your validator is listed on a UNL, it is considered an _untrusted_ validator

Passive-voice "it is considered" leaves the question hanging, "by whom?" Also, "until" implies an eventuality. Suggested rephrase:

> Servers that do not include your validator in their UNLs ignore your validator's messages in the consensus process. A validator that is not included in any UNL is called an _untrusted_ validator.

(Yes, I know "is called" is also passive, but at least IMO that phrasing is a more idiomatic way of introducing a term's definition.)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
mDuo13: Plain language suggestion: instead of "engenders" (a great, but somewhat advanced word), I suggest "encourages" or even "builds."

As for your other questions:

> What do we mean "validator list publishers"?

Those who publish lists of recommended validators. As Daniel said, anyone could publish a list, but servers whose lists aren't sufficiently similar run the risk of forking away from each other (declaring different ledger data as "validated"). In practice, the risk of accidentally forking may still be pretty small in a lot of configurations because it requires network delays and pretty crazy byzantine behavior (single validators sending different validations to different parties) to achieve a fork unless the two lists are _really_ different. But we're following the research and taking no risks on that.

> Is this specifically about getting a validator listed in this blob? https://vl.ripple.com/

Sort of, but not exactly. You could, in theory, list validators in plain text and rely on people to manually configure their servers' UNLs to match your list. The problem is, that step of people manually editing the list leaves more room for lists to diverge (because, even if you and I both follow the recommendations from some publisher, what if I haven't updated my server's list as recently as you have?) So we have a specific list format that a list publisher can publish, and people can configure their servers with, "just trust exactly what's on this publisher's list".

> Why is it in a blob and can we provide guidance about how to decode it?

I think it has to do with using digital signatures to sign the blob (so you know no one in between muddled with the list on its way from the publisher to you). We have, so far, not prioritized making official documentation on how to create or decode such lists because we don't really want to encourage the proliferation of these lists at the moment. (That's because, as mentioned above, two people trusting fairly different lists can make it possible for those two to fork from each other.) We may reconsider, especially if we have better handling of what to do when lists diverge, but basically, we want people to trust the same list so their chance of forking is zero.

That said, the list isn't that complicated. The "blob" is just standard [base64-encoded](https://tools.ietf.org/html/rfc4648) JSON, with a list of validators, their latest subkey authorization (aka "manifest") and a sequence number and expiration for the list... Anyway, it's not that we're against documenting it, it's just we see more downside than upside in explaining how it works right now since the safest case is everyone using one list anyway.

> Whatever is in this list would be reflected on XRPCharts' list of UNL validators, correct?

Yes, XRPCharts' list is based on decoding the official list from the vl site, so it should match (possibly with some slight delays if XRPCharts doesn't process it immediately?)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+
+2. [Set up a proxy to help protect your validator](#set-up-a-proxy-to-help-protect-your-validator) from [DDoS](https://en.wikipedia.org/wiki/Denial-of-service_attack) attacks.
+
+3. [Request domain verification](#request-domain-verification).
+
+
+
+## Enable Validation on Your `rippled` Server
+
+Enabling validation on your `rippled` server means providing a validator token in your server's `rippled.cfg` file. Ripple recommends using the `validator-keys` tool (included in `rippled` yum and Alien RPMs) to securely generate and manage your validator keys and tokens. ***TODO: correct reference to yum and Alien RPMs? These are the only RPMs, correct? Should we also point to https://github.com/ripple/validator-keys-tool for folks who want to manually compile and use it for a validator running on Windows?***
```
mDuo13: "Alien RPMs" is not really a proper way of describing them. The RPMs are the same; Alien is just a program for installing RPMs on Ubuntu since Ubuntu doesn't use yum.

And no, we don't need to include instructions for running validators on platforms that aren't supported in production.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
mDuo13: > If someone finds your private validator key, they can impersonate your validator and sign validations on your behalf.

Technically, it's the validator (master) key that you want to store online, which if someone finds it, they wouldn't use it to sign validations directly, but they'd use it to authorize a new subkey they controlled, which they could use to sign validations on your behalf. Doing that would also invalidate validations you were trying to publish with your old subkey. (Only one validator subkey can be active at once.)

mDuo13: Also, ideally, you shouldn't create your validator keys in an EC2 instance at all... Maybe not even on the production validator where you intend to use the subkeys.

I guess, contrary to my comment above, that may be a good case for including instructions on how to run the validator keys tool on other platforms, so people can generate their master keys on an offline machine.) But yeah, I think the point here is, generating and leaving the master keys on the validator where you plan to use the (ephemeral) subkeys undermines the advantage of using master/subkeys in the first place, which is that you can rotate your subkeys if, say, a system admin who had root permission to your production servers leaves your company. (Potentially lots of sysadmins need to be able to log in with root permission on servers that are running nonstop, because how else are you going to keep things running 24/7/365? The problem is, any of those sysadmins—or malicious actors who gained such access through social engineering, phishing, or exploits—could theoretically access and write down any keys stored on those machines, and use them later. But since the validator master keys are only needed when you rotate your ephemeral subkeys, those can be kept offline in a machine that far fewer people have access to.)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+        eyJ2YWxpZGF0aW9uX3NlY3J|dF9rZXkiOiI5ZWQ0NWY4NjYyNDFjYzE4YTI3NDdiNT
+        QzODdjMDYyNTkwNzk3MmY0ZTcxOTAyMzFmYWE5Mzc0NTdmYT|kYWY2IiwibWFuaWZl
+        c3QiOiJKQUFBQUFGeEllMUZ0d21pbXZHdEgyaUNjTUpxQzlnVkZLaWxHZncxL3ZDeE
+        hYWExwbGMyR25NaEFrRTFhZ3FYeEJ3RHdEYklENk9NU1l1TTBGREFscEFnTms4U0tG
+        bjdNTzJmZGtjd1JRSWhBT25ndTlzQUtxWFlvdUorbDJWMFcrc0FPa1ZCK1pSUzZQU2
+        hsSkFmVXNYZkFpQnNWSkdlc2FhZE9KYy9hQVpva1MxdnltR21WcmxIUEtXWDNZeXd1
+        NmluOEhBU1FLUHVnQkQ2N2tNYVJGR3ZtcEFUSGxHS0pkdkRGbFdQWXk1QXFEZWRGdj
+        VUSmEydzBpMjFlcTNNWXl3TFZKWm5GT3I3QzBrdzJBaVR6U0NqSXpkaXRROD0ifQ==
+
+3. Add `[validator_token]` and its value to your `rippled.cfg` file. ***TODO: Is it okay to also put the "# validator public key: XXXXX" text in the cfg file?***
```
mDuo13: It's not really necessary, but it's fine. It _is_ a public key after all.

I'm curious if you mean the master pubkey or the sub-pubkey.

Technically both are encoded in a reversible format in the token, so that's why I say it's not really necessary.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+
+    If you previously configured your validator without the `validator-keys` tool, delete `[validation_seed]` and its value from your `rippled.cfg` file. This changes your validator public key. ***TODO: validation_seed has been disabled since 0.29.1 -- can we safely delete this text?***
```
mDuo13: No, we should leave it. To my knowledge, you can still use `[validation_seed]` if you want (not recommended for production, but maybe you're just testing or you know what you're doing and you have another plan) and it was never configured by default to begin with (that would be like giving everyone a default validator identity rather than having them not be a validator, which would be, uh, bad)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+
+***TODO: how can I tell that my server is behaving as a validator? what command can tell me that my server is issuing validation messages? How can I look at a list of validations on the network and see if my validator is associated with some of them?***
+
+
+
+## Understand the Traits of an Ideal Validator
+
+As discussed earlier, not every validator is likely to be widely trusted and validator list publishers may require validators to meet stringent criteria before they list them in their UNLs.
+
+Strive to have your validator always embody the following properties. The more of these properties your validator embodies, the more reasons validator list publishers have to include it on their UNLs.
```
mDuo13, line 107: Is this meant to replace the [Properties of a Good Validator](http://localhost/ripple-dev-portal/rippled-server-modes.html#properties-of-a-good-validator) section? You should remove it from that page and maybe even spin it off to a separate page.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
mDuo13: I previously drafted up a diagram of the recommended network architecture. (It needs review and probably some revisions, plus maybe it should be redone with more "stylish" visuals.)

rippled-network-diagram-rough

> I think inbound traffic is clearer than "outside network" - but is it accurate?

The proxy connects both inbound and outbound traffic. Also, be careful with the word "proxy" here because the typical use of the word "proxy" here because [(as stated on Wikipedia) most proxies today are web proxies](https://en.wikipedia.org/wiki/Proxy_server) specifically for HTTP(S) traffic, but that is not what a rippled server is proxying. They're still proxies, but they're unusual proxies so be aware of the common usage of "proxy" to mean "HTTP proxy".

> Do we have any recommendations around the minimum number of proxies to set up?

I'm not aware of any specific recommendations.

> Is the primary purpose of providing more than one proxy that if one goes down, you don't lose your connection to the network?

Yes.

> Also curious about how the validator's "view" of the network changes when proxies are in place -- what does inbound traffic to a proxy-less validator look like vs inbound traffic to a validator with two proxies, for example? I don't quite understand how the views could differ, if at all.

The view of traffic should not differ much if at all, especially if the proxies are clustered. If they are not clustered, it's possible that an abusive peer in the outside network may be blocked by one proxy but not (yet) by another, so the validator behind the proxies would only see message from the abusive peer if they went through certain proxies.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
-    * Change `[peer_private]` to `1` to prevent its IP address from being forwarded.
+To help protect a validator from [DDoS](https://en.wikipedia.org/wiki/Denial-of-service_attack) attacks, set up one or more stock `rippled` servers as proxies between the validator and inbound traffic. ***TODO: Stated correctly? I think inbound traffic is clearer than "outside network" - but is it accurate? Do we have any recommendations around the minimum number of proxies to set up? Is the primary purpose of providing more than one proxy that if one goes down, you don't lose your connection to the network? Also curious about how the validator's "view" of the network changes when proxies are in place -- what does inbound traffic to a proxy-less validator look like vs inbound traffic to a validator with two proxies, for example? I don't quite understand how the views could differ, if at all.***
 
-5. Configure the validator host machine's firewall to only accept inbound connections from its public-facing rippled(s).
+1. [Enable validation](#enable-validation-on-your-rippled-server) on your `rippled` server.
```
mDuo13, line 139: Nit: since the link goes to a header whose text is almost exactly the full sentence, expand the link text to match, e.g. 

```
[Enable validation on your `rippled` server.](#enable-validation-on-your-rippled-server)
```

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
 
-Remember to restart `rippled` for config changes to take effect.
+2. Set up stock `rippled` servers. For more information, see [Install rippled](install-rippled.html).
 
-Take care not to publish the IP address of your validator.
+3. Configure your validator and stock `rippled` servers to run in a [cluster](cluster-rippled-servers.html).
```
mDuo13, line 143: I am not sure if the recommended setup is for the validator to be part of the cluster with the proxies or not. Please double-check with one of our integration engineers or another expert.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
 
+4. In your validator's `rippled.cfg` file:
 
-## Domain Verification
+    1. Copy the `[ips_fixed]` IP address list you defined in the previous cluster configuration step and paste it under `[ips]`.
 
-Network participants are unlikely to trust validators without knowing who is operating them. To address this concern, validator operators can associate their validator with a web domain that they control.
+        The `[ips_fixed]` and `[ips]` values must be identical and contain only the IP addresses and ports of the stock `rippled` servers you clustered with your validator. This ensures that your validator connects to only the stock `rippled` servers that you control.
```
mDuo13: Saying that the two "must" be identical implies that they _always_ need to be identical, but what we mean is that _for this purpose_ they must be identical.

Also, maybe emphasize that you should _not_ change your proxy rippleds' `[ips]` config in this regard (otherwise they won't attempt to connect to the outside network)—only the validator's. It might be easy to miss the fact that this step is specifically for the validator.

mDuo13: Oh, and you're missing a step where you configure your proxy `rippled` servers' `[ips_fixed]` to include the validator, too. It would _probably_ work without that, but it's better to be explicit, I think.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
 
-1. Find your validator public key by running the following on the validator server:
+    2. Set `[peer_private]` to `1` to prevent your validator's IP address from being forwarded. For more information, see [Private Peers](peer-protocol.html#private-peers).
 
+        **Warning:** Be sure that you don't publish your validator's IP address via any other channels as well.
+
+5. Configure your validator host machine's firewall to accept inbound connections from only the stock `rippled` servers you clustered it with.
```
mDuo13: Here's an important omission: the `rippled` needs to be able to make _outbound connections_ on port 443 to retrieve the recomended validator list from https://vl.ripple.com

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+
+6. Restart `rippled`.
+
+        $ sudo systemctl restart rippled.service
+
+***TODO: should we provide some tips on how to test that the firewall has been configured correctly? maybe ping the validator IP address from a machine outside the cluster? The test that the cluster is set up correctly (peers command) is covered in the clustering doc.***
```
mDuo13: Rules for testing the firewall itself are out of scope for this procedure (we don't know what firewall they're using, after all) but we could include instructions for how to test that peer_private is working correctly.

For that, you'd want to use the [peers method](https://developers.ripple.com/peers.html) on one of your proxies to confirm that your proxies are connected to your validator, then use the special [Peer Crawler Endpoint](https://developers.ripple.com/peer-protocol.html#peer-crawler) on the same proxy to confirm that your validator isn't listed.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
mDuo13: Domain verification is setting up a two-way link:

- You use your domain to claim that you have ownership over a validator key
- You use your validator key to claim ownership of a domain

Doing so establishes pretty strong evidence that one entity in particular owns both, so you're more likely to know who they are. (Domains sometimes come with contact information, via [WHOIS](https://whois.easycounter.com/ripplelabs.com), although a lot of them are registered with privacy-protecting "generic" contact information provided by the name registrar.)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+
+
+
+## Request Domain Verification
+
+***TODO: I'm unclear on what constitutes domain verification. Is it simply associating the validator with a web domain - like putting your public key on a website and claiming ownership? Or is it additionally getting your validator listed on XRP Charts or some other site like https://bithomp.com/validators or https://peerisland.com/validators?***
+
+To help validation list publishers and other participants in the XRP Ledger network understand who operates your validator, associate your validator with a web domain that you control and then apply to have your validator listed on XRP Charts' [Validator Registry](https://xrpcharts.ripple.com/#/validators). This provides a way for the public to verify your ownership of your validator.
+
+1. Set up the web domain that you want to use to publicly share your ownership of your validator. Secure the domain with SSL and import the SSL certificate into the domain. Ensure that the web domain's port 443 is open. You'll need the SSL private key file to complete a subsequent step.
```
mDuo13: > web domain

Technically domains are not limited to the "web" (HTTP) traffic. But you do need to run an HTTPS server at the domain that 

> Secure the domain with SSL

Technically, SSL is the old name for the protocol, and all the versions that were called "SSL" are now considered insecure. The new name is TLS. Also, it's slightly inaccurate to talk about securing "the domain" because you're really securing _HTTP content served from that domain_ (and thereby converting it to HTTPS traffic).

> Ensure that the web domain's port 443 is open.

This is a requirement for serving HTTPS (unless the user explicitly specifies a different port) so this is not quite right either.

In total, I'd recommend rephrasing this to something like:

> 1. Choose a domain name you own that you want to be publicly associated with your validator. You must run a public-facing HTTPS server on port 443 of that domain and you must have access to the private key file associated with that server's TLS certificate. (Note: TLS was formerly called SSL.)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
mDuo13: Post on the XRPChat forum, share on social media or your website, put out a press release, whatever. I don't think we have specific recommendations.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
mDuo13: Nit: we're not adding them to "a UNL". We're adding them to "Ripple's recommended UNL."

Technically there is no way to know whether someone else has listed your validator in their UNL. It's like asking whether someone is paying attention to your webinar.

But in this case, when Ripple adds them to our list of recommendations (arguably not technically a UNL??), they'll know as Daniel mentioned.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,191 @@
...
+
+  ***TODO: we should tell the user how to tell that their validator has been added to a UNL - whether it is the default UNL or another UNL.***
+
+<!--{ ***TODO: For the future - add a new section or separate document: "Operating a Trusted Validator" -- things that you need to be aware of once your validator has been added to a UNL and is participating in consensus. We should tell the user what to expect once they are listed in a UNL. How to tell if your validator is participating in the consensus process? How to tell if something isn't right with your validator - warning signs that they should look out for? How to tell if your validator has fallen out of agreement - what is the acceptable vs unacceptable threshold? Maybe provide a script that will alert them when something is going wrong.*** }-->
+
+
+
+## Revoke Validator Keys
+
+If your validator's private key is compromised, you must revoke it immediately and permanently.
```
mDuo13: I think this is referring specifically to the master key?

mDuo13 (changes requested):
Great start. Left a lot of info. I hope I addressed your questions.

Sadly, I think we'll have to update these docs again in the next few months as we improve the way validator verification works (on both ends of the two-way link)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
garay-daniel: Yes, Ripple's UNL is the exact words we should use. I am lazy.

mDuo13: Even just "Ripple's UNL" is, arguably, a slightly different concept:

- The list of validators we have our servers configured to trust ("Ripple's UNL")
- The list of validators we tell other people to trust ("Ripple's recommended UNL")

We keep them in sync but there's no technical reason they have to be.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Fantastic call on replacing "engenders" with "builds" 🙌 ... and thank you for taking the time to answer my questions!

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Thank you, Rome! I ask b/c the response from the `./validator-keys create_token` command instructs the user to update `rippled.cfg` with the validator public key and the validator_token. See the code block above. Just wanted to be sure that is okay. Absolutely - I know it is a public key -- but I just get a little hypervigilant whenever keys are involved! I need to stop being _unnecessarily_ hypervigilant when it is clearly a public key. 😹 Thank you for humoring me.

I believe this is the master pubkey because it is the `public_key` in the `validator-keys.json` file. Just to be sure I'm keeping this straight - the sub-pubkey would be the `validation_public_key` from `./rippled validation_create`, for example, correct? 🤔

mDuo13: 1. after looking at the validator-keys-tool docs again, it's definitely the master pubkey
2. The sub-pubkey is encoded into the token. Well, actually, I think it's derived from the sub-private key that's encoded into the token. But the `validation_create` output is just a key pair, not intrinsically "master" or "sub". It depends on how you use it, which if you're using the validator keys tool, you won't use `validation_create` for either one since it kind of does the same thing. (Only the keys tool knows how to relate to keypairs to make one authorize the other.) You might use `validation_create` to make a node keypair since that's in the same format but doesn't support the master key/subkey thing.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Thank you for this! I think the [ips_fixed] step you call out is covered in the Clustering doc we link folks to in step 3 above. 

And just to talk it through - in this step here, we are taking the [ips_fixed] value we set for the validator in step 3 and we're copying it to the validator's [ips] stanza, which isolates the validator's connection to only the stock servers in the cluster.

The result is, for example, when I use the `peers` command on the validator - the response includes only the two stock servers I set up in the cluster with the validator. But when I use the `peers` command on each of the two stock servers I set up in the cluster, the response includes 11-12 peers, which include the validator and other stock server.

Please let me know if I'm missing the point. 😅  🙌

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Oh, great catch! 🙏 

Just to talk this through - is this an accurate picture of the firewall settings for the validator:

- Inbound traffic: Only from IP addresses of the stock servers in the cluster. 
- Outbound traffic: Only to the IP addresses of the stock servers in the cluster and to https://vl.ripple.com through port 443. Because even if the validator is getting all other info it needs to participate in consensus from the stock servers in the cluster, the one other thing the validator needs that the stock servers can't provide is the UNL from https://vl.ripple.com? And this is a silly question - but does inbound traffic also need to be open to https://vl.ripple.com for the list of validators to be returned to the validator?

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Thank you for this. Yes, the clustering doc provides the check using the `peers` method. I've added the test using the Peer Crawler Endpoint - 👌 !

Can I just verify the following with you?

The validator (and its IP address) is not included in the Peer Crawler endpoint response from the stock server in the cluster. Is this because [peer_private] is set to `1` in the validator's `rippled.cfg`? 

Though, when I run `peers` on the same stock server in the cluster, the validator does appear with its IP address in the results. This is okay because I can run `peers` only on servers I have admin control over and because of the way my validator's [ips] and [ips_fixed] values are set - I know that my validator will not connect to any other peers than the ones I control in the cluster.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
mDuo13: Inbound traffic: yes, correct.
Outbund traffic: yes, to the stock servers and to vl.

No need for inbound traffic from VL. vl.ripple.com won't be reaching out and connecting to the validator. Only the reverse.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
mDuo13: Correct.

One thing that may be confusing is that enabling `[peer_private]` does several things:

- Prevents the server from opening peer connections to unknown servers. (You have to explicitly configure which servers you want to peer with.)
- Makes the server reject peer connections from unknown servers. (This is an extra defense alongside the firewall you're expected to use.)
- Requests that the server's peers don't mention it in their peer crawler results.

The `peers` command is an admin command, so it provides an accurate picture of which servers are peered with the server you query. That's important for administering the server!

The peer crawler's results are treated as public. You can get peer crawler results without being an administrator. Honest servers (that is, ones that haven't been modified to break this rule) won't report their private peers in the peer crawler results. You could modify your server so that it _would_ report its private peers, but they can only spill their own secrets (that is, private peers they're _directly_ connected with) since they don't know anyone else's secrets.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Yes, it is! Sorry I forgot to handle that piece in this commit. Handling in next push!

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Hi @garay-daniel:

> We could also instruct validators they can connect to our **public hubs** here as well, these will ensure more peers and higher connectivity.

Just making sure I understand - what do you mean by "public hubs"?

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Thank you so much for the explanation and for being very exact about this, Rome. Also, much gratitude for taking the time to craft the excellent rephrasing.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
mDuo13: Ripple runs a few `rippled` servers for the specific purpose of creating reliable connection points in the XRP Ledger peer-to-peer network. Those are our "public hub servers" (public hubs for short)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Daniel and Rome - thank you for these clarifications. Very help for my understanding of the process and of what we mean by a "UNL."

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Yes - this is referring to the `secret_key` value in the `validator-keys.json` generated for the validator. I'm updating the text to be more specific.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Thank you, Rome! 

Sorry I'm being a little thick here:

- Where can validator operators find a list of public hub servers? This is not about ripple.com validators in the Validator Registry, right?
- Now this feels like a really silly question - but what kind of connection are we talking about? How should they connect to these public hubs? This is not about clustering or proxying, correct?

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Great - thank you for raising this. Working on verifying.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+**Note:** While these servers are acting as proxies, they are not web proxies for HTTP(S) traffic.
+
+<!-- { TODO: Future: add a recommended network architecture diagram to represent the proxy, clustering, and firewall setup: https://ripplelabs.atlassian.net/browse/DOC-2046 }-->
+
+1. [Enable validation](#enable-validation-on-your-rippled-server) on your `rippled` server.
+
+2. Set up stock `rippled` servers. For more information, see [Install rippled](install-rippled.html).
+
+3. Configure your validator and stock `rippled` servers to run in a [cluster](cluster-rippled-servers.html). ***TODO: JHA still working on verifying that the validator should be in the cluster.***
```
jhaaaa: Hi @nbougalis! I've checked with a few folks on this - but we'd like to get your expert call on this, please. 

When setting up proxy stock servers to help protect a validator from DDoS attacks, we tell the user to cluster the `rippled` servers - the proxy stock servers and the validator (https://developers.ripple.com/cluster-rippled-servers.html). 

Can you confirm that the validator should be included in the cluster? Thank you for any guidance you can provide!

jhaaaa: Confirmed with Nik that it is ok to include the validator in the cluster with the proxy stock servers.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+## Enable Validation on Your `rippled` Server
 
-To protect a production validator from [DDoS](https://en.wikipedia.org/wiki/Denial-of-service_attack) attacks, you can use a stock `rippled` server as a proxy between the validator and the outside network.
+Enabling validation on your `rippled` server means providing a validator token in your server's `rippled.cfg` file. Ripple recommends using the `validator-keys` tool (included in `rippled` RPMs) to securely generate and manage your validator keys and tokens.
 
-1. [Set up the `rippled` validator.](run-rippled-as-a-validator.html)
+In a location **not** on your validator:
 
-2. [Set up one or more stock `rippled` servers.](install-rippled.html)
+1. Manually build and run the `validator-keys` tool, if you don't already have it installed via a `rippled` RPM. ***TODO: added this step for users who need to use the validator-keys tool delivered outside the rippled RPM. For example, they don't want to install rippled via RPM on their non-validator machine and just need to be able to generate the required keys and token. Okay?***
```
mDuo13: :+1:

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
-Remember to restart `rippled` for config changes to take effect.
+        ```
+        $ /opt/rippled/bin/validator-keys create_keys
+        ```
 
-Take care not to publish the IP address of your validator.
+      - For a manually built `validator-keys` tool install, run:
 
+        ```
+        $ /build/gcc.debug/validator-keys create_keys
```
mDuo13: This is almost certainly the wrong path. You probably don't have permission to create a top-level `/build/` folder when you create the tool.

Also, if I'm not mistaken, [the RPM installation adds the validator-keys tool to a binary path](https://github.com/ripple/rippled-package-builder/blob/99596cbcd739899b08c6e2c34ae022b40e5d93d1/rpm-builder/rippled.spec#L62) so you wouldn't need to specify the full `/opt/rippled/bin/` path when running it from the RPM installation.

In other words, I think you can combine the two examples (here and elsewhere in the guide) into simply:

```
$ validator-keys create_keys
```

On a manually-built system, the user will have to either navigate to the folder where they built the output, or manually add the binary to their PATH, but those are standard provisions when manually installing software, so I don't think we need to document that specifically. Obviously, the folder where they built the output varies user by user.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+        ```
 
-## Domain Verification
+      Sample output:
 
-Network participants are unlikely to trust validators without knowing who is operating them. To address this concern, validator operators can associate their validator with a web domain that they control.
+        Validator keys stored in /Users/local/.ripple/validator-keys.json
```
mDuo13: This looks like a macOS-specific path (`/Users` is a giveaway). I'm not sure where it would end up on different OSes. Possibly `/home/ripple/.local/validator-keys.json` or something? (Just guessing)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+    If you previously configured your validator without the `validator-keys` tool, delete `[validation_seed]` and its value from your `rippled.cfg` file. This changes your validator public key.
+
+2. Restart `rippled`.
+
+        $ sudo systemctl restart rippled.service
+
+3. Use the `server_info` command to get information about your validator to verify that it is running as a validator.
+
+        $ ./rippled server_info
```
mDuo13: Remove the `./` from here. That's incorrect on a system where you installed with the RPM.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+      - The `pubkey_validator` value in the response should match the `public_key` in the `validator-keys.json` file that you generated for use with your validator.
+
+      - The `server_state` value should be _**proposing**_.
+
+
+
+## Understand the Traits of an Ideal Validator
```
mDuo13: Maybe let's tone "Ideal" back down to "Good"?

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+As discussed earlier, not every validator is likely to be widely trusted and validator list publishers may require validators to meet stringent criteria before they list them in their UNLs.
```
mDuo13: "As discussed earlier" implies that other reading is necessary (meaning, this is not [page one](https://www.oreilly.com/library/view/every-page-is/9781492001959/)). I think you can safely strike it from this sentence.

mDuo13: Also, it may be worth noting that even if you run an untrusted validator, that can be helpful for the overall health of the network because it sets the standard that trusted validators are measured against. If a trusted validator is disagreeing with a lot of untrusted validators, that might indicate a problem.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+Strive to have your validator always embody the following properties. The more of these properties your validator embodies, the more reasons validator list publishers have to include it on their UNLs.
+
+- Available
```
mDuo13: I suggest bolding these points

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
-1. Find your validator public key by running the following on the validator server:
+        This file should be stored securely and not shared.
 
+      **Warning:** Store the generated `validator-keys.json` key file in a secure, offline, and recoverable location, such as an encrypted USB flash drive. Do not modify its contents. In particular, be sure to not store the key file on the validator where you intend to use the keys. If your validator's `secret_key` is compromised, [revoke the key](https://github.com/ripple/validator-keys-tool/blob/master/doc/validator-keys-tool-guide.md#key-revocation) immediately.
+
+      For more information about the `validator-keys` tool and the key pairs it generates, see the [Validator Keys Tool Guide](https://github.com/ripple/validator-keys-tool/blob/master/doc/validator-keys-tool-guide.md).
+
+3. Generate a validator token using the `create_token` command.
+
+      - For an RPM-based `validator-keys` tool install, run:
```
mDuo13: As suggested above, I think you can combine these two examples.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+        ```
```
mDuo13: Indented code fences won't work well with Dactyl's Markdown parser. Indent the code blocks an extra four spaces (compared to the surrounding text) instead.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+    An ideal validator is always running and submitting validation votes for every proposed ledger. Strive for 100% uptime.
+
+- In agreement
+
+    An ideal validator's votes match the outcome of the consensus process as often as possible. To do otherwise could indicate that your validator's software is outdated, buggy, or intentionally biased. Always run the [latest `rippled` release](https://github.com/ripple/rippled/tree/master) without modifications. [Watch `rippled` releases](https://github.com/ripple/rippled/releases) to be notified of new releases.
+
+- Issuing timely votes
+
+    An ideal validator's votes arrive quickly and not after a consensus round has already finished. Meeting recommended `rippled` server [system requirements](system-requirements.html), which include having a fast internet connection, helps with this.
```
mDuo13: It might be easier to follow the last sentence if it didn't start with a gerund ("Meeting"). Suggested plainer-language rephrase (And before you ask, yes, the bad sentence was my fault in the first place):

> To keep your votes timely, make sure your validator meets the recommended [system requirements](system-requirements.html), which include a fast internet connection.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+- Identified
+
+    An ideal validator has a clearly identified owner. Providing [domain verification](#provide-domain-verification) is a good start. Ideally, XRP Ledger network UNLs include validators operated by different owners in multiple legal jurisdictions and geographic areas. This reduces the chance that any localized events could interfere with the impartial operations of trusted validators.
+
+At present, Ripple (the company) cannot recommend any validators aside from those in the [default validator list](https://github.com/ripple/rippled/blob/develop/cfg/validators-example.txt). However, we are collecting data on other validators and building tools to report on their performance. For metrics on validators, see the XRP Charts [Validator Registry](https://xrpcharts.ripple.com/#/validators).
```
mDuo13: Suggest changing the first sentence from negative to positive:

> Ripple (the company) publishes a [validator list] with a set of recommended validators. Ripple strongly recommends using exactly this list for production servers.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+4. In your validator's `rippled.cfg` file:
+
+    1. Copy the `[ips_fixed]` IP address list you defined in the previous cluster configuration step and paste it under `[ips]`.
+
+        For this purpose, the `[ips_fixed]` and `[ips]` values must be identical and contain only the IP addresses and ports of the stock `rippled` servers you clustered with your validator. This ensures that your validator connects to only the stock `rippled` servers that you control.
+
+    2. Set `[peer_private]` to `1` to prevent your validator's IP address from being forwarded. For more information, see [Private Peers](peer-protocol.html#private-peers).
+
+        **Warning:** Be sure that you don't publish your validator's IP address via any other channels as well.
```
mDuo13: Nit: "via any other channels as well"→"in other ways"

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+5. Configure your validator host machine's firewall to allow the following traffic only:
+
+    - Inbound traffic: Only from IP addresses of the stock `rippled` servers in the cluster you configured.
+
+    - Outbound traffic: Only to the IP addresses of the stock `rippled` servers in the cluster you configured and to [https://vl.ripple.com](https://vl.ripple.com) through port 443.
```
mDuo13: Tip: If you want to make a link whose label is the same as its target, you can use angle brackets. In other words, the following two lines are equivalent. I prefer the second one since it's easier to read & maintain:

```
[https://vl.ripple.com](https://vl.ripple.com)
<https://vl.ripple.com>
```

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+6. Restart `rippled`.
+
+        $ sudo systemctl restart rippled.service
+
+7. Use the [Peer Crawler](peer-protocol.html#peer-crawler) endpoint on one of your stock `rippled` servers. The response should not include your validator. This verifies that your validator's `[peer_private]` configuration is working. One of the effects of enabling `[peer_private]` on your validator is that your validator's peers do not include it in their Peer Crawler results.
```
mDuo13, line 165: :+1:

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+1. Choose a domain name you own that you want to be publicly associated with your validator. You must run a public-facing HTTPS server on port 443 of that domain and you must have access to the private key file associated with that server's TLS certificate. (Note: TLS was formerly called SSL.)
+
+2. Share your validator's public key with the public, especially other `rippled` operators. For example, you can share your validator's public key on your website, on social media, in the [XRPChat forum](https://www.xrpchat.com/), or in a press release.
+
+3. Submit a request to have your validator listed in XRP Charts' [Validator Registry](https://xrpcharts.ripple.com/#/validators) using this [Google Form](https://docs.google.com/forms/d/e/1FAIpQLScszfq7rRLAfArSZtvitCyl-VFA9cNcdnXLFjURsdCQ3gHW7w/viewform). Having your validator listed in this registry provides another form of public verification that your validator and domain are owned by you. To complete the form, you'll need the following information:
+
+      1. Find your validator public key by running the following on the validator server.
+
+        ```
```
mDuo13: Avoid indented code fences

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+## Provide Domain Verification
+
+To help validation list publishers and other participants in the XRP Ledger network understand who operates your validator, provide domain verification for your validator. At a high level, domain verification is a two-way link:
+
+- Use your domain to claim ownership of a validator key.
+
+- Use your validator key to claim ownership of a domain.
+
+Creating this link establishes strong evidence that you own the validator key and the domain.
```
mDuo13: maybe add "both" for emphasis here?

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+To provide domain verification:
+
+1. Choose a domain name you own that you want to be publicly associated with your validator. You must run a public-facing HTTPS server on port 443 of that domain and you must have access to the private key file associated with that server's TLS certificate. (Note: TLS was formerly called SSL.)
```
mDuo13: Maybe the note can link the Wikipedia article or RFC for Transport Layer Security?

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+2. Share your validator's public key with the public, especially other `rippled` operators. For example, you can share your validator's public key on your website, on social media, in the [XRPChat forum](https://www.xrpchat.com/), or in a press release.
```
mDuo13: :1st_place_medal: 

Totally optional suggestion: maybe say "XRPChat community forum" to emphasize that it's not a first-party site?

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+      1. Find your validator public key by running the following on the validator server.
+
+        ```
         $ /opt/ripple/bin/rippled server_info -q | grep pubkey_validator
+        ```
+
+        Provide the value returned in the **Validator Public Key** field of the Google Form.
+
+      2. Sign the validator public key using your web domain's SSL private key. The SSL private key file does not need to be stored on your validator's server.
```
mDuo13: Nit: use "TLS" instead of "SSL" here

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+        $ openssl dgst -sha256 -hex -sign /PATH/TO/YOUR/SSL.key <(echo YOUR_VALIDATOR_PUBLIC_KEY_HERE)
+        ```
+
+        Sample output:
+
+        ```
+        4a8b84ac264d18d116856efd2761a76f3f4544a1fbd82b9835bcd0aa67db91c53342a1ab197ab1ec4ae763d8476dd92fb9c24e6d9de37e3594c0af05d0f14fd2a00a7a5369723c019f122956bf3fc6c6b176ed0469c70c864aa07b4bf73042b1c7cf0b2c656aaf20ece5745f54ab0f78fab50ebd599e62401f4b57a4cccdf8b76d26f4490a1c51367e4a36faf860d48dd2f98a6134ebec1a6d92fadf9f89aae67e854f33e1acdcde12cfaf5f5dbf1b6a33833e768edbb9ff374cf4ae2be21dbc73186a5b54cc518f63d6081919e6125f7daf9a1d8e96e3fdbf3b94b089438221f8cfd78fd4fc85c646b288eb6d22771a3ee47fb597d28091e7aff38a1e636b4f
+        ```
+
+        Provide the value returned in the **SSL Signature** field of the Google Form.
```
mDuo13, line 201: I guess we're stuck with "SSL" here to match the form wording. But whoever created the form should've known better. :weary:

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+      3. Using the [`validator-keys` tool](https://github.com/ripple/validator-keys-tool/blob/master/doc/validator-keys-tool-guide.md) (included in `rippled` RPMs), sign the domain name.
+
+        - If running a `validator-keys` tool installed via 'rippled' RPM:
```
mDuo13: As above, you can combine these (and ditch the indented code fences)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -1,74 +1,253 @@
...
+
+        - If running a manually built `validator-keys` tool:
+
+            ```
+            $ ./validator-keys  --keyfile /PATH/TO/YOUR/validator-keys.json sign YOUR_DOMAIN_NAME
+            ```  
+
+        Sample output:
+
+        ```
```
mDuo13: More nested code fences to avoid

content/use-cases/run-a-rippled-validator.md
```diff
@@ -37,14 +37,14 @@ To configure your `rippled` server to run in validator mode, generate a validato
 
 <span class="use-case-step-num">{{n.next()}}</span>
 <!-- <span class="use-case-step-length">(1 hour)</span> -->
-## [Set up a stock rippled server as a proxy](run-rippled-as-a-validator.html#public-facing-server)
+## [Set up a stock rippled server as a proxy](run-rippled-as-a-validator.html#set-up-proxies-to-help-protect-your-validator)
```
mDuo13, line 40: Great catch to make sure the use case is updated!

mDuo13 (changes requested):
Found a couple little things that should be cleaned up, but by and large I'm happy with where this is at.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Ah, great catch about the `/` in front of the `build` 🤦‍♀️ and great guidance about being able to just combine the examples - thank you!

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Great point about "As discussed earlier." It is discussed earlier on this same page - so I thought I'd just tie it back. But sure thing, it works just as well without it! 👍 I am a big fan of Every Page is Page One - great to know that others look to it as a reference as well!

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: @mDuo13 - Ah, sorry - I thought I provided an answer to this open question. Nik confirmed that it doesn't hurt to have the validator in the cluster.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: @mDuo13 - I addressed this comment by adding sample outputs for both macOS and Ubuntu. Let me know if this is not what you had in mind -- happy to update! 😄

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -36,25 +36,21 @@ Enabling validation on your `rippled` server means providing a validator token i
...
-      - For an RPM-based `validator-keys` tool install, run:
+        $ validator-keys create_keys
 
-        ```
-        $ /opt/rippled/bin/validator-keys create_keys
-        ```
+      Sample output on Ubuntu:
 
-      - For a manually built `validator-keys` tool install, run:
+        Validator keys stored in /home/local/.ripple/validator-keys.json
```
mDuo13: This path is probably still not correct. The [validator keys tool docs](https://github.com/ripple/validator-keys-tool/blob/master/doc/validator-keys-tool-guide.md#validator-keys) suggest it should be:

```
/home/ubuntu/.ripple/validator-keys.json
```

(In this case, `ubuntu` is presumably the username. Linux convention is `/home/{USER}/`.)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -186,57 +174,39 @@ To help validation list publishers and other participants in the XRP Ledger netw
...
+2. Share your validator's public key with the public, especially other `rippled` operators. For example, you can share your validator's public key on your website, on social media, in the [XRPChat community forum](https://www.xrpchat.com/), or in a press release.
 
 3. Submit a request to have your validator listed in XRP Charts' [Validator Registry](https://xrpcharts.ripple.com/#/validators) using this [Google Form](https://docs.google.com/forms/d/e/1FAIpQLScszfq7rRLAfArSZtvitCyl-VFA9cNcdnXLFjURsdCQ3gHW7w/viewform). Having your validator listed in this registry provides another form of public verification that your validator and domain are owned by you. To complete the form, you'll need the following information:
 
       1. Find your validator public key by running the following on the validator server.
 
-        ```
-        $ /opt/ripple/bin/rippled server_info -q | grep pubkey_validator
-        ```
+            $ /opt/ripple/bin/rippled server_info -q | grep pubkey_validator
```
mDuo13: Optional nit: I think the `-q` here is [unnecessary](https://developers.ripple.com/commandline-usage.html#unit-tests) and probably ignored by the server anyway.

mDuo13 (approved):
Found one typo you should still fix (and another you can optionally change) but otherwise the changes look good. Thanks!

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: 👋 @mDuo13! Sorry for the misunderstanding here. Yes, I changed my username, which comes back in the actual output. to `local` - just to remove my name from the example. 😂  

In the validator keys tool doc - yes, they changed the username to `ubuntu`. I can change my example here to `ubuntu` -- would that work for you? 🙌 

We should probably come up with some style guide instructions for how to genericize these sample outputs/responses - what do you think?

mDuo13: `.local` is actually a common file to be in one's home directory on Linux, so we should choose something different as a generic name. Preferably something no one could mistake as literal.

In some other places, we assume that there is a `rippled` user and the `rippled` service runs as that user after being installed by RPM (I think this is correct) but in this case, where we're instructing people to do something _not_ on their server, using a more generic name is appropriate.

Perhaps `my-user`?

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jhaaaa: Sounds great - thank you for the link to more info on this.

PR #490, opened 2019-01-04
Add use case: Payment Channels for Digital Exchanges

jhaaaa:
- Adds a use case that describes how a digital exchange can use a Payment Channel to send XRP to another digital exchange


content/use-cases/use-a-payment-channel-to-move-xrp-between-exchanges.md
```diff
@@ -0,0 +1,65 @@
+# Use a Payment Channel to Move XRP Between Exchanges
+
+***TODO: what do we think of this use case title? I like using the name of the feature and also describing the task to be performed. Is the language in the title accurate?***
```
mDuo13: Yes, it's a good title.

jhaaaa:
@mDuo13, @warpaul, and @garay-daniel - I humbly submit for your review, a first draft of the use case for payment channels for exchanges. Still rough -- but wanted to get it out to you sooner than later. To find my questions/comments, search for "TODO". Thank you for any guidance you have the time to provide!

content/use-cases/use-cases.md
```diff
@@ -12,6 +12,10 @@ Here are some of our users' most popular use cases:
 
       Integrate with the XRP Ledger to be able to list XRP in your exchange, enabling your users to deposit and withdraw XRP.
 
+* **[Use a Payment Channel to Move XRP Between Exchanges](use-a-payment-channel-to-move-xrp-between-exchanges.html)**
```
jhaaaa: J-Ha will update this once we approve a title for this use case.

jhaaaa:
@mDuo13 - reviewed with PM to ensure that I'm capturing the features/tasks they'd like to see in the use case. When you have a moment -- I think it is a point where it could really benefit from a review from you. Thank you! I was hoping to be able to actually set up a payment channel before sending this your way, but I went down a terraform rabbit hole and lost time. I'm going to set one up using local builds - but wanted to get this out to you sooner than later!

TheMainlander (commented):
Nice!  🙏

content/use-cases/use-cases.md
```diff
@@ -12,6 +12,10 @@ Here are some of our users' most popular use cases:
 
       Integrate with the XRP Ledger to be able to list XRP in your exchange, enabling your users to deposit and withdraw XRP.
 
+* **[Open a Payment Channel to Enable an Inter-Exchange Network](open-a-payment-channel-to-enable-an-inter-exchange-network.html)**
+
+      Use a payment channel to move XRP from your digital exchange to another digital exchange, while also enabling an inter-exchange network and connecting to the Internet of Value.
```
mDuo13: This feels like it's juggling a lot. I'd scratch the "and connecting to the Internet of Value" part from the short text. Maybe rephrasing from "while also enabling an inter-exchange network" to "creating an inter-exchange network". (I mean, I want it to say, "thereby" but that's not plain language enough.)

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,84 @@
+# Open a Payment Channel to Enable an Inter-Exchange Network
+
+As a digital asset exchange, do you make multiple XRP payments to another exchange with some degree of frequency? If this usage pattern sounds familiar, consider opening an XRP Ledger [payment channel](payment-channels.html) between your exchange (the _payer_ exchange) and the other exchange (the _payee_ exchange). A payment channel enables you to send one-way, "asynchronous" XRP payments that can be divided into very small increments and settled later. Here are some of the benefits of using a payment channel to send XRP instead of using individual payment transactions:
```
mDuo13: Do you feel like you're hearing a sales pitch when an article starts with a question? I dunno about you, but my brain instantly goes on the defensive when I encounter this sentence structure. Here's a rephrase that doesn't come off so strong:

> As a digital asset exchange, if you frequently send XRP to another exchange, you can <s>turbo-charge</s> improve the efficiency of these payments with an XRP Ledger [payment channel].

(I also changed "multiple... with some degree of frequency" to just "frequent")

I also think it may be useful to mention that, if you have a two-way flow with the other exchange, you can simply open two payment channels (one for each direction).

mDuo13: Another thing: I wouldn't be surprised if some exchanges see the word "payments" and think that you're not talking about them. They may not even realize at first how many payments they're making to other exchanges! I'd try working the word "withdrawals" into the intro paragraph to help spark the recognition of what circumstances they're probably already doing this in.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,84 @@
...
+
+- **Achieve faster processing:** A standard payment transaction involves submitting an XRP Ledger transaction and waiting for a new ledger version that includes the transaction to be approved by [consensus](https://developers.ripple.com/consensus.html). When you use a payment channel to send XRP, creation and verification of a claim, which guarantees the payment of XRP, all happen outside of the consensus process. This means that the payer exchange can guarantee XRP payments to the payee exchange at a rate limited only by the participants' ability to create and verify the digital signatures of the claims.
```
mDuo13: I feel the word "achieve" bolded here is extraneous. Since just removing it would destroy the parallel structure, maybe **Process payments faster:** (Or "Process withdrawals faster"?) conveys the same idea while bolding only the parts that _really_ matter.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,84 @@
...
+
+- **Lower transaction costs:** The payee exchange can choose to take the claims it has received from the payer exchange and redeem them in batches to receive the guaranteed XRP amounts. Redeeming a batch of claims in a single transaction incurs a _single_ set of transaction costs (any applicable reference transaction, load, and open ledger costs), while processing multiple XRP payment transactions (one per claim without a payment channel in place, for example) to receive the same amount of XRP incurs _multiple_ sets of transaction costs. ***TODO: accurately stated?***
```
mDuo13: I feel like calling it a "set of transaction costs" implies more than one pay-point, when really, it's one cost. (The different "costs" you mention are really different potential minimums. But you're only paying one cost. Even if the net is busy you can choose to let your transaction wait in the queue rather than paying the—temporarily higher—open ledger cost.)

mDuo13: Also, it may be confusing to say "XRP payment transactions" because while you a claim-redeem can process multiple _conceptual payments_, it only occupies one _on-ledger transaction_, which means it's not the same as multiple "XRP Ledger Payment transactions" (with a capital P).

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,84 @@
...
+
+- **Connect to the Internet of Value:** One of the key requirements of the [Internet of Value](https://ripple.com/insights/the-internet-of-value-what-it-means-and-how-it-benefits-everyone/) is interoperability. The [Interledger Protocol](https://interledger.org/) (ILP), which plays a large role in driving this interoperability, [uses payment channels](https://interledger.org/rfcs/0027-interledger-protocol-4) as its method for rebalancing accounts, whether it uses the XRP Ledger's payment channels, Bitcoin's micropayment channels, or Ethereum's state channels, for example. In effect, when you open a payment channel from your exchange to another, you are connecting to the Internet of Value and helping to create the inter-exchange network that is fundamental to the success of the Internet of Value and the apps that are built on it.
```
mDuo13: - Interledger doesn't require payment channels. It just works best with payment channels.
- I'd end the "uses payment channels" sentence after "... for rebalancing accounts." You can modify or remove the "whether it uses..." portion of the sentence after that.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,84 @@
...
+
+<ul>Here’s a roadmap to the high-level tasks you’ll need to perform to implement this payment channel use case. To go directly to a full payment channels tutorial, see <a href="use-payment-channels.html">Use Payment Channels</a>.</ul> ***TODO: I put this paragraph above in an HTML ul tag to keep the CSS formatting on this page somewhat sane. I know that it is unnecessarily indented, but if I make it a regular paragraph, the formatting will break and look even more conspicuous. We can fix the CSS in a future release.***
```
mDuo13: :nauseated_face: 
I think technically this is illegal HTML. `<div>` would be more appropriate, but we should just fix the formatting that screwed this up in the first place.

I think this can't be done strictly in CSS since there's no widely-supported selector that can know whether a paragraph is before or after a sibling header. Instead, we could either use a simple Dactyl filter that looks for a comment such as `<!-- BEGIN USE CASE STEPS -->` or just define some standard HTML to use like `<div class="use-case-steps" markdown=1>` ... `</div>`

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Understand payment channels](payment-channels.html)
+
+Learn more about payment channels and whether they provide the features you need for your specific implementation.
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Set up and run `rippled` servers](manage-the-rippled-server.html) (payer and payee)
```
mDuo13: For multi-role use cases / tutorials like this, I think I like putting the role at the beginning, rather than the end. So, for example:

> "Payer and payee: Set up and run `rippled` servers"

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,84 @@
...
+
+To use a payment channel to send and receive payments, both the payer and payee exchanges must each have access to a `rippled` server that they can use to send transactions. A great way for an exchange to get access to a `rippled` server is to set up and run one.
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Fund XRP Ledger accounts](accounts.html) (payer and payee)
+
+The payer exchange must have a funded XRP Ledger account to be used to send XRP to the payee exchange.
```
mDuo13: We should provide guidance for how much XRP they'll need, at minimum.

Aside from the account reserve (20 XRP) and the owner reserve of a payment channel (5 XRP), they'll also need enough XRP to set aside in the payment channel to cover some number of transactions. I'm not sure how many; you can top off the channel if you run out, but that requires an actual on-ledger transaction and confirmation (so like 4-5 seconds and a 10ish drops of XRP to send). Obviously the more XRP you pre-fund the less often you have to top off, so you can save some money, but if you put more than you need in the channel you'll need to close it (waiting out two transactions and the `SettleDelay`) to get it back.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,84 @@
...
+
+The payee exchange must have a funded XRP Ledger account to be used to redeem (receive) XRP sent by the payer exchange.
```
mDuo13: For how much XRP they need, the payee only needs the 20 XRP reserve + enough to pay the transaction costs of redeeming (which are trivial; you could redeem thousands of claims for less than 1 XRP in total). So maybe we can just say "21 XRP or more"?

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Open a payment channel](use-payment-channels.html#1-the-payer-creates-a-payment-channel-to-a-particular-recipient) (payer) and [verify it](use-payment-channels.html#2-the-payee-checks-specifics-of-the-payment-channel) (payee)
```
mDuo13: I would rather see these as totally separate steps so each can be given adequate attention.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Create claims](use-payment-channels.html#3-the-payer-creates-one-or-more-signed-claims-for-the-xrp-in-the-channel) and [send them](use-payment-channels.html#4-the-payer-sends-a-claim-to-the-payee-as-payment-for-goods-or-services) (payer)
+
+The payer exchange creates one or more claims for amounts of XRP that it wants to guarantee to the payee exchange and then sends the claim details to the payee exchange off-ledger.
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Verify claims](use-payment-channels.html#5-the-payee-verifies-the-claims) and [redeem them in batches](use-payment-channels.html#8-when-ready-the-payee-redeems-a-claim-for-the-authorized-amount) (payee)
```
mDuo13: I'd like to see these split up into separate steps.

For redeeming them in batches, we should provide guidance for how often they should redeem them or how to decide when to redeem them. The short version is:

- Don't redeem every claim. That's not gaining any benefit from the payment channels.
- Don't wait until you have more than you're scared to lose. If something goes wrong and you miss your chance to redeem a claim, you don't get paid. If that happens and you already credited someone in another system, you could be in trouble. (For example, if you credited a customer with the XRP and they've already traded it for some other cryptocurrency which they withdrew.)
- If the payer requests to close the channel, you won't get paid if you don't redeem your claims before it finishes closing. The amount of time you have is based on the `SettleDelay`.
- If the channel was created with an immutable `CancelAfter` time, be sure to redeem all outstanding claims before that time.
- You can decide to redeem, for example, after a certain amount of time or after accumulating a certain amount of credit, or even other things if you really care. The safest strategy is probably both time- and amount-based.

I'd want to talk to some active PayChan users like Coil people to get some more specific recommendations. But I imagine that settling every few minutes or even hours could make sense depending on how much the payee exchange trusts the payer exchange, how many transactions they do how rapidly, etc.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,84 @@
...
+
+The payee exchange verifies claims sent by the payer exchange.
+
+This is usually when the payee provides goods or services to the payer in exchange for the amount of XRP listed in the claims. However, in this use case, there is likely no exchange of goods or services. The payee exchange can just redeem batches of claims to receive the XRP guaranteed by the payer exchange.
```
mDuo13: The terminology in this tutorial masks the real "services" that are being provided, because the "payer exchange" is not the one _really_ initiating the payments.

The customer—who probably has accounts at both exchanges and is doing some kind of arbitrage (or xRapid stuff)—is the one triggering the payment, and the "services" they're receiving are "being credited in the payee exchange's systems" so they can start trading there.

So after verifying the claims is when the payee exchange should credit the destination for the XRP deposit.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,84 @@
...
+
+The payee exchange then reviews the details of the payment channel.
+
+If the payer exchange has multiple accounts and wants to use them with payment channels, they must open one payment channel per pair of payer and payee accounts. Also keep in mind that payment channels are unidirectional. So, let's say that exchange A wants to move XRP in both directions between an account it owns on exchange A and an account it owns on exchange B. Exchange A must open one payment channel from its exchange A account to its exchange B account, as well as a second payment channel from its exchange B account to its exchange A account. ***TODO: stated correctly? Relevant to surface this info here?***
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Create claims](use-payment-channels.html#3-the-payer-creates-one-or-more-signed-claims-for-the-xrp-in-the-channel) and [send them](use-payment-channels.html#4-the-payer-sends-a-claim-to-the-payee-as-payment-for-goods-or-services) (payer)
+
+The payer exchange creates one or more claims for amounts of XRP that it wants to guarantee to the payee exchange and then sends the claim details to the payee exchange off-ledger.
```
mDuo13: An uncomfortable wrinkle here: the way that Destination Tags work with payment channels is not really compatible with how we recommend exchanges use destination tags (that is, to identify which customer should be credited).

A given payment channel only has one DestinationTag (for the whole channel) but the two exchanges probably don't want to use a separate payment channel for each receiving customer (at the payee exchange). They should probably use one payment channel for everything, but that means that they'll have to communicate the equivalent information—that is, which customer to credit at the payee exchange—some other way.

One way would be to attach a destination tag with the signed claims when sending them. That way the payee exchange at least knows which customer to credit when it verifies a claim. The payer exchange still won't necessarily know which customers' claims have been fully redeemed and which have been only "verified" and credited, but that can be worked out. (I'll talk about this some more in my comments below about verifying and redeeming claims.)

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
mDuo13: If a claim comes with a Destination Tag or destination-tag-like bit of information, you know which customer to credit. (As I mentioned above, the payer exchange has to send this along with the claim, since you can't rely on the payment channel's built-in `DestinationTag` unless you're only ever using the channel for one receiving customer.)

As for how much to credit the customer, you have to be careful. You should only be crediting them for _the difference from the previous claim_. So you need to know how much the last claim on this channel was for! To be extra sure that you're not getting this wrong, it might even behoove the two exchanges to work out a "sequence number"-like system for assuring that claims don't arrive at the payee exchange out of order. Otherwise it's possible the payee exchange might credit a customer with another customer's funds because they didn't realize they were missing a claim!

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,84 @@
...
+
+Regarding the amount of XRP the payment channel should hold, consider the typical best practice of holding the vast majority of XRP across all of your user accounts in a cold wallet, with a small amount of XRP in a hot wallet. Think about your daily traded volume of XRP to help determine how much XRP in your hot wallet to make available in the payment channel. You have the option to increase the amount allocated to the payment channel after you create it.
+
+The payee exchange then reviews the details of the payment channel.
+
+If the payer exchange has multiple accounts and wants to use them with payment channels, they must open one payment channel per pair of payer and payee accounts. Also keep in mind that payment channels are unidirectional. So, let's say that exchange A wants to move XRP in both directions between an account it owns on exchange A and an account it owns on exchange B. Exchange A must open one payment channel from its exchange A account to its exchange B account, as well as a second payment channel from its exchange B account to its exchange A account. ***TODO: stated correctly? Relevant to surface this info here?***
```
mDuo13: >  they must open one payment channel per pair of payer and payee accounts.

It should be clear that you mean exchanges' XRP Ledger accounts, not _customer accounts at the exchanges_ here. Some of my other comments address the challenges of using one channel for every _customer account_.

But yeah, if the exchanges have multiple XRP Ledger accounts (like hot wallets) then they would probably need one payment channel for each.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,84 @@
...
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## When it's time, make a request to [close the payment channel](use-payment-channels.html#9-when-the-payer-and-payee-are-done-doing-business-the-payer-requests-for-the-channel-to-be-closed) (payer)
+
+When the payer exchange and payee exchange are done using the payment channel, the payer exchange can make a request to close the payment channel.
+
+
+### Related Tasks
+
+- [List XRP in Your Exchange](list-xrp-in-your-exchange.html)
```
mDuo13: Maybe list this in the background reading rather than in "Related Tasks"

mDuo13 (changes requested):
This is a good start, but it's going to need a lot more specific information to become a truly useful guide for exchanges.

I feel like we may need to really dig deeper into some of the tutorial steps this links and make more concrete examples targeted at this use case, with examples of the kinds of data structures and math that the target user should be using.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
mDuo13: The exchanges must agree on a system to indicate which customers have been credited (in the asynchronous messaging when they send, receive, verify the claim, and reply to the payer that they've verified and accepted it).

Optionally they could also use something like `Memos` to indicate which specific customer claim(s) a particular redeem transaction covers. (In other words, use a memo to indicate what's in the batch.) That's not strictly necessary (and for greater privacy, they may prefer not to go into the maximum level of detail) but if there's any plans for clawbacks or re-crediting customers at the sending exchange if particular claims _don't_ get redeemed, they could use such a system for both exchanges to communicate which claims have been not only verified and credited but actually redeemed.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
jhaaaa: Thank you for these comments! 🙏  Just to hep me understand the misuse here 😂 -- I called them reference transaction, load, and open ledger costs bc of the way transaction costs are documented here: https://developers.ripple.com/transaction-cost.html. Should I call them "reference transaction, load, and open ledger potential minimums"? Or is there no need to call them out? Or are we talking about different things? I want to use the correct language -- but I think I'm missing something.

For "XRP payment transactions" -- thank you for this clarification. Just to be sure I understand - the comparison I intended to make is this: Instead of making multiple Payments, for which you'd be charged costs for _multiple_ transactions -- you can redeem a batch of claims in a single transaction (that ends up having the same effect as the multiple Payments, for example), incurring the cost of a _single_ transaction. Is this an incorrect description of why/how the ability to redeem payment channel claims in batch can lower costs? Missing something again - thank you for your patience!

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
jhaaaa: Thank you for this deep insight -- I need to do a bit of spike to truly understand the difference you are talking about. I understand that the difference exists - I think I just need to understand how and why the exchange is actually moving XRP between exchanges. The lightbulb WILL go on!

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
jhaaaa: Sorry to make you 🤢 LOL - yah, I know it is illegal. 👮 I was just trying to come up with a quick workaround to be able to publish this. But now I think this is the least of its troubles. 😂 I'd love to be able to update how this template works to be able to accommodate the evolution of the content we're using it to provide. This is a part of the effort to try to provide more compelling Use Case content. For example, see the Escrow-as-Smart Contract PR as well: https://github.com/ripple/ripple-dev-portal/pull/512. Thank you for being open to updating the template!

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
mDuo13: The reason an exchange would send XRP to another exchange is this:

A customer, who has accounts at both exchanges, asked the "payer exchange" to withdraw their XRP (which the payer exchange has been custodying as they make trades within that exchange). And the address the customer provided to "withdraw" their XRP is the customer's deposit address at another exchange (the "payee exchange").

Two reasons a customer might do that:
1. xRapid—the customer wants to trade the XRP for a local currency that the payee exchange lists (but the payer exchange, presumably, doesn't)
2. Arbitrage—the customer can sell the XRP at a higher rate at the payee exchange. And then, potentially, withdraw the currency they bought, deposit it back at the payer exchange, buy more XRP, and repeat as long as a market imbalance lasts.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
mDuo13: There's only one transaction cost. You don't incur a "set of transaction costs".

It's just that how much you have to pay for the transaction cost depends on several different factors. The transaction cost page is a little confusing, and for that I'm sorry. It's not an easy concept to explain, and it's evolved to become more confusing since that page was originally written. 

Basically, you're going to pay one transaction cost, probably whichever is highest among the costs mentioned. Except not always, because you don't have to pay the open ledger cost if you don't mind waiting in the queue. And the some transactions have to pay more than the "reference transaction cost" because that's just the cost of a so-called "reference transaction" that doesn't have an alternate minimum cost because of things like multi-sigs and crypto-condition fulfillments.

mDuo13: Also, after re-reading, I think the second part makes sense. Maybe it's a little densely worded. But yeah, you seem to have the correct understanding of the idea of individual payments vs. batch claims.

jhaaaa:
@mDuo13, @warpaul, & @garay-daniel - Received some fantastic feedback and education from Rome. I hope that this next draft clarifies the use case (and some of its underlying triggers) and provides details missing from the first draft. I have a feeling that I may have provided too many details for a Use Case in some areas. Perhaps some of the details should be at the tutorial level? Do we need a tutorial specific to exchanges? I don't know -- just putting it out there for discussion.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
+# Open a Payment Channel to Enable an Inter-Exchange Network
+
+As a digital asset exchange, if you frequently send XRP to another exchange, you can improve the efficiency of these payments by opening an XRP Ledger [payment channel](payment-channels.html) between your exchange (the _payer_ exchange) and the other exchange (the _payee_ exchange). The need to send XRP from your exchange to another exchange may originate with your customers withdrawing XRP from your exchange and depositing it to the other exchange, for example.
+
+A payment channel enables you to send one-way, "asynchronous" XRP payments that can be divided into very small increments and settled later. If you have a two-way flow with the other exchange, you can simply open two payment channels (one for each direction).
+
+<div>Here are some of the benefits of using a payment channel to send XRP instead of using individual payment transactions:</div>***TODO: the div workaround isn't working for all of the paras above***
```
mDuo13: This needs to be rebased on master and updated to fit the new use case template.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
+# Open a Payment Channel to Enable an Inter-Exchange Network
+
+As a digital asset exchange, if you frequently send XRP to another exchange, you can improve the efficiency of these payments by opening an XRP Ledger [payment channel](payment-channels.html) between your exchange (the _payer_ exchange) and the other exchange (the _payee_ exchange). The need to send XRP from your exchange to another exchange may originate with your customers withdrawing XRP from your exchange and depositing it to the other exchange, for example.
```
mDuo13: > The need to send XRP from your exchange to another exchange may originate with your customers withdrawing XRP from your exchange and depositing it to the other exchange, for example.

I'd like to see this expanded into a brief subsection, maybe even with a diagram. "Why send XRP to other exchanges?"

Also, this is a total nitpick, but I don't like ending sentences with ", for example." I'd either remove it (it's not a _concrete_ example anyway) or move it to the start of the sentence.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
...
+
+A payment channel enables you to send one-way, "asynchronous" XRP payments that can be divided into very small increments and settled later. If you have a two-way flow with the other exchange, you can simply open two payment channels (one for each direction).
+
+<div>Here are some of the benefits of using a payment channel to send XRP instead of using individual payment transactions:</div>***TODO: the div workaround isn't working for all of the paras above***
+
+- **Process withdrawals faster:** A standard payment transaction involves submitting an XRP Ledger transaction and waiting for a new ledger version that includes the transaction to be approved by [consensus](https://developers.ripple.com/consensus.html). When you use a payment channel to send XRP, creation and verification of a claim, which guarantees the payment of XRP, all happen outside of the consensus process. This means that the payer exchange can guarantee XRP payments to the payee exchange at a rate limited only by the participants' ability to create and verify the digital signatures of the claims.
```
mDuo13: consensus link should not have the full URL

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
...
+## [Understand payment channels](payment-channels.html)
+
+Learn more about payment channels and whether they provide the features you need for your specific implementation.
+
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## Payer and payee: [Set up and run `rippled` servers](manage-the-rippled-server.html)
+
+To use a payment channel to send and receive XRP, both the payer and payee exchanges must each have access to a `rippled` server that they can use to send transactions. A great way for an exchange to get access to a `rippled` server is to set up and run one.
```
mDuo13: If an exchange processes XRP withdrawals directly, they already know how to do this. (If they go through some middleware, I'm not as sure.)

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## Payer and payee: [Fund XRP Ledger accounts](accounts.html)
+
+- The payer exchange must have a funded XRP Ledger account to be used to send XRP to the payee exchange. Aside from the [base reserve](reserves.html) (20 XRP) and the [owner reserve](reserves.html#owner-reserves) of a payment channel (5 XRP), the account must also be able to set aside enough XRP in the payment channel to cover the intended number of transactions.
```
mDuo13: As with the rippled server, if the exchange processes XRP deposits and withdrawals directly, they already have this.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
...
+
+  The payer exchange can always top-off the channel using the [PaymentChannelFund](paymentchannelfund.html) transaction if it runs out of XRP. However, topping-off requires an actual on-ledger transaction and confirmation, so it could take 4-5 seconds of processing time and ~10 drops of XRP to complete the top-off transaction. The more XRP the payer exchange pre-funds, the less often they need to top-off, so they can save some money by pre-funding more XRP.
```
mDuo13: Need a full 4 spaces to make this part of the bullet point.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
...
+
+  However, if the payer exchange puts in more XRP than they need, they need to [close the payment channel](use-payment-channels.html#9-when-the-payer-and-payee-are-done-doing-business-the-payer-requests-for-the-channel-to-be-closed) to get the XRP back, which means waiting out two transactions (a request to close the channel and another to actually close the channel) and the `SettleDelay` time.
```
mDuo13: 4 spaces indent also required here.

The way the part about waiting out the two transactions and the `SettleDelay` time is phrased makes it seem like the settlement delay is the least important, but that's probably backwards. I'm not sure what settlement delay amounts are "industry standard" these days but I would imagine something on the scale of hours to days.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
...
+
+- The payee exchange must have a funded XRP Ledger account to be used to redeem (receive) XRP sent by the payer exchange. The account needs at least 21 XRP, which provides the 20 XRP [base reserve](reserves.html), plus enough to pay the transaction costs of redeeming claims, which are trivial. For example, you could redeem thousands of claims for less than 1 XRP in total.
```
mDuo13: Maybe also worth a similar note about how if they process withdrawals and deposits of XRP, they already have an XRP Ledger account

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
mDuo13: In fact, there's a good chance they may follow industry best practices and have a cold account plus one or more hot accounts. If that's the case, which one should they use for their payment channels? Probably the hot one? Probably?

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## Payer: [Open a payment channel](use-payment-channels.html#1-the-payer-creates-a-payment-channel-to-a-particular-recipient)
+
+The payer exchange opens a payment channel from their XRP Ledger account to the payee exchange's XRP Ledger account. Opening a payment channel includes setting certain specifics of the channel, such as its expiration date and the amount it can hold.
+
+For this exchange use case, there is no real need to ever close the channel, so the payer exchange may not want to define a `CancelAfter` (expiration) value. If they ever need to close the channel, they can still do so.
+
+Regarding the amount of XRP the payment channel should hold, the payer exchange should consider the typical best practice of holding the vast majority of XRP across all of their user accounts in a cold wallet, with a small amount of XRP in a hot wallet. The payer exchange should also think about their daily traded volume of XRP to help determine how much XRP in their hot wallet to make available in the payment channel. They can always add more XRP to the payment channel after they create it.
```
mDuo13: > The payer exchange should also think about their daily traded volume of XRP to help determine how much XRP in their hot wallet to make available in the payment channel.

I think this can be more specific:

> As the payer exchange, you should decide approximately how often you want to add more XRP to the channel—for example, daily, every 4 hours, or every 15 minutes—and estimate the volume of XRP that you send to the payee exchange during that interval. You should fund the payment channel with enough to cover at least that much volume _or_ the largest withdrawal that you want to process without delay, whichever is larger. For example, if you plan to refill the channel every 15 minutes, have an average volume of 50 XRP every 15 minutes, but occasionally send transfers of 10,000 XRP, you should supply the channel with at least 10,000 XRP. For withdrawals that are larger than the amount of XRP you have in the payment channel, you must process them specially. Either you can send large withdrawals as normal XRP payments, skipping the payment channel entirely, or you can first send a transaction to add the full withdrawal amount to the payment channel before creating claims for those. (See below for details on creating claims.)

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
...
+
+If the payer exchange has multiple XRP Ledger accounts (one for each hot wallet, for example) and wants to use them with payment channels, they must open one payment channel per pair of payer exchange and payee exchange XRP Ledger accounts. To be clear, this use case is about creating a payment channel between the exchanges' XRP Ledger accounts, not between customer accounts.
+
+Also keep in mind that payment channels are unidirectional. So, let's say that exchange A and exchange B want to move XRP in both directions between their XRP Ledger accounts. Exchange A must open a payment channel from their XRP Ledger account to exchange B's XRP Ledger account, and exchange B must open a payment channel from their XRP Ledger account to exchange A's XRP Ledger account. ***TODO: I reworded this a bit to portray the accounts at each end of the payment channel as being owned by separate exchanges. Are both scenarios possible - both accounts owned by the same exchange, each account owned by a different exchange? Which one do we want to surface in this use case?***
```
mDuo13: I don't think using payment channels between two of your own accounts makes any sense. 

The point of payment channels is to allow fast, asynchronous payments as the payer adds to their tab, by providing the payee the freedom to batch-redeem at will without needing specific intervention on the part of the payer. If you control both accounts, you don't need a channel to keep tabs of how much one account owes the other and initialize a transfer when the amount is sufficient.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
...
+
+For this exchange use case, there is no real need to ever close the channel, so the payer exchange may not want to define a `CancelAfter` (expiration) value. If they ever need to close the channel, they can still do so.
+
+Regarding the amount of XRP the payment channel should hold, the payer exchange should consider the typical best practice of holding the vast majority of XRP across all of their user accounts in a cold wallet, with a small amount of XRP in a hot wallet. The payer exchange should also think about their daily traded volume of XRP to help determine how much XRP in their hot wallet to make available in the payment channel. They can always add more XRP to the payment channel after they create it.
+
+If the payer exchange has multiple XRP Ledger accounts (one for each hot wallet, for example) and wants to use them with payment channels, they must open one payment channel per pair of payer exchange and payee exchange XRP Ledger accounts. To be clear, this use case is about creating a payment channel between the exchanges' XRP Ledger accounts, not between customer accounts.
```
mDuo13: > they must open one payment channel per pair of payer exchange and payee exchange XRP Ledger accounts

I don't think this is what we want. More likely, they want to pick a pair of accounts to have a channel between them, then use only that pair of accounts for anything that needs the channel.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
...
+| **Sequence**        | A value that indicates the sequence in which the payer exchange created the claims. The payee exchange needs this value to keep track of and redeem claims in the correct order. For example, if the payer exchange did not provide the sequence value and the payee exchange lost track of the second claim above, the payee exchange might incorrectly credit 2500 XRP to destination tag 18273645. If the payer exchange did provide the sequence value, the payee exchange would know that it needs to account for a claim between claim 1 and claim 3. With claim 2 accounted for, the payee exchange could correctly credit 1500 XRP to destination tag 87654321 and 1000 XRP to destination tag 18273645. |
+| **Signature**       | Signature of the claim. The payee exchange needs this value to verify and redeem the claim. |
+| **Amount**          | Amount of the claim created by the payer exchange. The payee exchange needs this value to verify and redeem the claim. For information about how to calculate the actual amount the payee exchange needs to credit the customer, see [Verify claims](#payee-verify-claims). |
```
mDuo13: Probably best to mention that this is a _cumulative_ amount of XRP withdrawn from the channel.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
...
+| **Destination Tag** | Destination tag of the customer account on the payee exchange that needs to be credited based on the claim. The payer exchange can get this value from their customer's withdrawal transaction, which should require a destination tag for the deposit to the payee exchange. When the payee exchange redeems claims, the XRP is deposited into the payee exchange's XRP Ledger account. The payee exchange can then deposit the XRP from the claim to the appropriate customer account based on the destination tag provided. |
```
mDuo13: > The payer exchange can get this value from their customer's withdrawal transaction, which should require a destination tag for the deposit to the payee exchange.

The word "transaction" is confusing here. I suggest "withdrawal request" and I'd also change "require" to "provide" here.

> The payee exchange can then deposit the XRP from the claim to the appropriate customer account based on the destination tag provided.

Instead of "deposit" I'd say "credit" here. The XRP is probably not moving to an on-ledger customer account; it's about crediting the user in the exchange's internal ledger.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## Payee: [Verify claims](use-payment-channels.html#5-the-payee-verifies-the-claims)
+
+The payee exchange verifies claims sent by the payer exchange.
+
+After verifying claims, the payee exchange should credit the claimed XRP to the customer accounts indicated by the destination tags sent by the payer exchange. Because claim amounts are cumulative, the payee exchange needs to be careful to credit the customer for only the _the difference from the previous claim_.
+
+For example, to know how much to credit a customer for a claim amount of 3000, the payee exchange needs to know that the previous claim amount was 1500. The difference between the claim amount and the previous claim amount (3000 - 1500 = 1500) is the amount the payee exchange must credit to the customer account.
```
mDuo13: Just a thought, but maybe use an example where the previous claim amount and the new net claim amount aren't both 1500. For example, 3000 - 2000 = 1000.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## Payee: [Redeem them in batches](use-payment-channels.html#8-when-ready-the-payee-redeems-a-claim-for-the-authorized-amount)
+
+This is usually when the payee provides goods or services to the payer in exchange for the amount of XRP listed in the claims. However, in this use case, there is likely no exchange of goods or services. The payee exchange can just redeem batches of claims after verifying them to receive the XRP guaranteed by the payer exchange.
```
mDuo13: > However, in this use case, there is likely no exchange of goods or services.

As mentioned in my previous review, this isn't _quite_ true. I suggest rephrasing:

> In this use case, the service provided is for the customer to be credited in the payee exchange's own systems, so the customer can start trading.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,149 @@
...
+
+Here are some guidelines the payee exchange can use to decide how often to redeem claims:
+
+- Don't redeem every claim. That's not gaining any benefit from the payment channels.
+
+- Don't wait until you have more in claims than you're scared to lose. If something goes wrong and you miss your chance to redeem a claim, you don't get paid. If that happens and you already credited someone in another system, you could be in trouble. (For example, if you credited a customer with the XRP and they've already traded it for some other cryptocurrency that they withdrew.)
```
mDuo13: > If that happens and you already credited someone in another system, you could be in trouble. (For example, if you credited a customer with the XRP and they've already traded it for some other cryptocurrency that they withdrew.)

Let's make this more explicit and also fix the fact that the parenthetical is a fragment. Suggestion:

> If that happens and you have already credited one or more customers in your system, you could be in trouble. Those customers may have already traded the XRP for other cryptocurrencies and withdrawn them. That leaves you with more XRP owed in your system than you were holding for your customers, and it's too late to correct the balances of the customers whose deposits you didn't receive.

content/use-cases/use-cases.md
```diff
@@ -12,6 +12,10 @@ Here are some of our users' most popular use cases:
 
       Integrate with the XRP Ledger to be able to list XRP in your exchange, enabling your users to deposit and withdraw XRP.
 
+* **[Open a Payment Channel to Enable an Inter-Exchange Network](open-a-payment-channel-to-enable-an-inter-exchange-network.html)**
+
+      Use a payment channel to move XRP from your digital exchange to another digital exchange, while also creating an inter-exchange network.
```
mDuo13: > while also creating an inter-exchange network.

This makes it sound like more work. We want it to sound like an outcome. Or we could just trim it out for brevity's sake.

mDuo13 (changes requested):
This is shaping up well. I left a good number of suggestions, but at this point they're more about the language used than the process described.

Obviously this needs to get put in the new template but it's definitely getting close.

jhaaaa:
@mDuo13 - thank you for your latest review! I've updated the content - when you have a moment, please feel free to take a look and let me know if you have any other feedback. 🙌  I've also created a diagram as discussed - I'll send a link b/c it is definitely not ready for even a PR. 😹

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,177 @@
...
+
+      For your customers who are moving XRP to take advantage of arbitrage opportunities or to do algorithmic trading, speed matters. Enabling a customer to move XRP and start trading in an instant is a compelling differentiator for your exchange.
+
+- **Lower transaction costs:** The payee exchange can choose to take the claims it has received from the payer exchange and redeem them in batches to receive the guaranteed XRP amounts. Redeeming a batch of claims in a single transaction incurs a _single_ transaction cost, while processing multiple XRP payment transactions (one per claim without a payment channel in place, for example) to receive the same amount of XRP incurs _multiple_ transaction costs.
```
mDuo13: One wrinkle here.

_Receiving_ a simple Payment transaction doesn't incur a transaction cost. Only sending it. But sending a PaymentChannelClaim _does_ incur a transaction cost. So, sadly, by using a payment channel, the receiving side goes from zero to nonzero (but very small) transaction costs. 

The _sending_ side is where the savings are, since you're only paying the costs of the initial channel creation and the "top off" transactions to refill the channel with XRP. Sadly again, keeping the channel open involves holding the 5 XRP owner reserve, which is locked up until you close the channel. Assuming you keep the channel open indefinitely and you would've paid 10-12 drops of XRP per transaction, you're going to need to send on the order of 40,000-50,000 claims before you actually break even compared to just sending Payment transactions every time. And that's not even discussing the opportunity costs of locking up XRP in the channel (like maybe needing to have more XRP "hot" for withdrawals that can't use the channel).

So, the receiving side doesn't see lower transaction costs, although the fees they do pay are so negligible that it's probably fine. (The amount they spend depends on how often they redeem, probably 10-12 drops of XRP each time.) And the sending side sees lower transaction costs only when they've processed a very high volume of transactions to the specific destination, or if they close the channel and get their XRP back (getting the leftover channel balance & freeing up the owner reserve).

I don't know what scale of withdrawals particular exchanges are sending to other specific exchanges, but it would have to be sky high for the savings to pay the salary of whoever would be implementing this project. That makes me think that the potential savings should be viewed as more of a footnote in case the Internet of Value really explodes transaction volumes rather than a direct benefit. :disappointed:

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,177 @@
...
+If not, a great way for an exchange to get access to a `rippled` server is to set up and run one.
+
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## Payer and payee: [Fund XRP Ledger accounts with enough XRP](accounts.html)
+
+If your exchange processes XRP deposits and withdrawals directly, you probably have existing funded XRP Ledger accounts that you can use for this purpose. Just ensure that they are funded with enough XRP as described here.
+
+Along these lines, there's a good chance that you are following industry best practices and have a cold account plus one or more hot accounts. Also according to best practices, you probably want to use hot accounts for your payment channels.
```
mDuo13: >  Also according to best practices, you probably want to use hot accounts for your payment channels.

Suggest simpilfying this to:

> Use the hot accounts for your payment channels.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,177 @@
...
+
+- The payer exchange must have a funded XRP Ledger account to be used to send XRP to the payee exchange.
+
+      Aside from the [base reserve](reserves.html) (20 XRP) and the [owner reserve](reserves.html#owner-reserves) of a payment channel (5 XRP), the account must also be able to set aside enough XRP in the payment channel to cover the intended number of transactions.
+
+      The payer exchange can always top-off the channel using the [PaymentChannelFund](paymentchannelfund.html) transaction if it runs out of XRP. However, topping-off requires an actual on-ledger transaction and confirmation, so it could take 4-5 seconds of processing time and ~10 drops of XRP to complete the top-off transaction. The more XRP the payer exchange pre-funds, the less often they need to top-off, so they can save some money by pre-funding more XRP.
```
mDuo13: > so they can save some money by pre-funding more XRP

Saving the confirmation time is more relevant than saving the (miniscule) transaction cost. I suggest rephrasing to "time and money"

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,177 @@
...
+
+      However, if the payer exchange puts in more XRP than they need, they need to [close the payment channel](use-payment-channels.html#9-when-the-payer-and-payee-are-done-doing-business-the-payer-requests-for-the-channel-to-be-closed) to get the XRP back, which means waiting out the `SettleDelay` time they set for the payment channel and two transactions (a request to close the channel and another request to actually close the channel).
```
mDuo13: > a request to close the channel and another request to actually close the channel

I suggest changing this to:

> a request to start closing the channel and a later request to finish closing it after the `SettleDelay` has passed.

Maybe that can be merged with the non-parenthetical part of the sentence for more clarity.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,177 @@
...
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## Payer: [Open a payment channel](use-payment-channels.html#1-the-payer-creates-a-payment-channel-to-a-particular-recipient)
+
+The payer exchange opens a payment channel from their XRP Ledger account to the payee exchange's XRP Ledger account. Opening a payment channel includes setting certain specifics of the channel, such as its expiration date and the amount it can hold.
+
+For this exchange use case, there is no real need to ever close the channel, so the payer exchange may not want to define a `CancelAfter` (expiration) value. If they ever need to close the channel, they can still do so.
+
+Regarding the amount of XRP the payment channel should hold, the payer exchange should consider the typical best practice of holding the vast majority of XRP across all of their user accounts in a cold wallet, with a small amount of XRP in a hot wallet.
```
mDuo13: I wonder if it would be helpful to suggest that the exchange think of the payment channel as a special sub-wallet exclusively for a particular destination, and estimate how much XRP it requires similar to how you would estimate a hot wallet's needs.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,177 @@
...
+
+For withdrawals that are larger than the amount of XRP you have in the payment channel, you must process them specially. Either you can send large withdrawals as normal XRP payments, skipping the payment channel entirely, or you can first send a transaction to add the full withdrawal amount to the payment channel before creating claims for those. (See below for details on creating claims.)
+
+If the payer exchange has multiple XRP Ledger accounts (one for each hot wallet, for example), they should pick a pair of accounts to have a channel between them, then use only that pair of accounts for anything that needs the channel. To be clear, this use case is about creating a payment channel between the exchanges' XRP Ledger accounts, not between customer accounts.
```
mDuo13: > they should pick a pair of accounts to have a channel between them

The antecedents of "they" and "them" are different in this sentence. This whole paragraph sounds a little confused. Here's a suggested rephrase for this and the following paragraph:

> If either exchange has multiple hot accounts in the XRP Ledger, the two exchanges should each designate a specific hot account to use with the payment channel between them. Although there are other potentially viable configurations, this use case assumes a configuration with one payment channel connecting two exchanges. This channel can serve _all_ customers sending XRP from the payer exchange to the payee exchange.
>
> Since payment channels are unidirectional, you need a second channel in the opposite direction to send XRP from the "payee" exchange to the "payer" exchange. This second channel does not need to connect the exact same pair of hot accounts, but it is most convenient to do so. With two unidirectional channels, each exchange can use the XRP it redeems from its incoming channel to refill its outgoing channel.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -0,0 +1,177 @@
...
+After verifying claims, the payee exchange should credit the claimed XRP to the customer accounts indicated by the destination tags sent by the payer exchange. Because claim amounts are cumulative, the payee exchange needs to be careful to credit the customer for only the _the difference from the previous claim_.
+
+For example, to know how much to credit a customer for a claim amount of 3000, the payee exchange needs to know that the previous claim amount was 2000. The difference between the claim amount and the previous claim amount (3000 - 2000 = 1000) is the amount the payee exchange must credit to the customer account.
+
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## Payee: [Redeem them in batches](use-payment-channels.html#8-when-ready-the-payee-redeems-a-claim-for-the-authorized-amount)
+
+This is usually when the payee provides goods or services to the payer in exchange for the amount of XRP listed in the claims. In this use case, the service provided is for the customer to be credited in the payee exchange's own systems, so the customer can start trading.
```
mDuo13: This step (redeeming) is _not_ when the payee provides goods or services. Redeeming the claims is a separate process that is not intrinsically tied to providing the service of crediting customers (which is already described, appropriately, in the previous step).

mDuo13 (changes requested):
It's coming along, but I caught a few things that felt a little out of place still.

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
jhaaaa: Sure thing, Rome -- thank you so much for the thoughtful and thorough explanation. 🙇‍♀️ I'm going to remove this benefit point for now. Let's find out the scale of withdrawals for typical sending exchanges and see if there is a noticeable cost benefit to using a payment channel. If yes, we can add the point back. I've created a research ticket to handle this task.

jhaaaa:
@mDuo13 - Thank you for yet another thoughtful and thorough review! I've addressed your comments. When you have a moment, please take a look. 🙏

content/use-cases/open-a-payment-channel-to-enable-an-inter-exchange-network.md
```diff
@@ -55,15 +51,19 @@ If not, a great way for an exchange to get access to a `rippled` server is to se
...
 
-      However, if the payer exchange puts in more XRP than they need, they need to [close the payment channel](use-payment-channels.html#9-when-the-payer-and-payee-are-done-doing-business-the-payer-requests-for-the-channel-to-be-closed) to get the XRP back, which means waiting out the `SettleDelay` time they set for the payment channel and two transactions (a request to close the channel and another request to actually close the channel).
+      However, if the payer exchange puts in more XRP than they need, they need to [close the payment channel](use-payment-channels.html#9-when-the-payer-and-payee-are-done-doing-business-the-payer-requests-for-the-channel-to-be-closed) to get the XRP back. This means waiting out the following events:
+
+      1. Completion of the payer's request to start closing the payment channel.
+      2. Passage of the `SettleDelay` time set for the payment channel.
+      3. Completion of a request to finish closing the payment channel after the `SettleDelay` has passed.
```
mDuo13, line 66: :100:

mDuo13 (approved):
I'm happy with this at this point. We can, naturally, still improve it, but I think it's good enough to publish as-is!

PR #491, opened 2019-01-09
Replace ECDSA link

mDuo13:
Since the NIST article it previously linked is down for an unclear
amount of time as a result of the government shutdown, I replaced it
with a link to the Wikipedia ECDSA article (which is fairly complete).

Assuming this passes, then after merging this, we can rebase the other open PRs onto master to get those much-coveted :heavy_check_mark: s.


jhaaaa (approved):
Works for me! Thank you for handling this, Rome!

PR #492, opened 2019-01-10
Transaction Malleability Documentation

mDuo13:
Migrates the [old wiki page](https://web.archive.org/web/20170608153028/https://wiki.ripple.com/Transaction_Malleability) on transaction malleability and adds information about a new form of malleability relating to multi-signing.

### DIFF
--- a/content/concepts/consensus-network/transaction-malleability.md
+++ b/content/concepts/consensus-network/transaction-malleability.md
@@ -45,6 +45,11 @@ To calculate a fully-canonical ECDSA signature, one must compare S and N-S to de
 
 All XRP Ledger software that Ripple publishes (including `rippled`, and ripple-lib/RippleAPI) generates only fully-canonical signatures. To further protect users, Ripple has configured its code to enable the **tfFullyCanonicalSig** flag by default where possible. Ripple strongly encourages third-party implementations of XRP Ledger software to generate only fully-canonical signatures, and enable tfFullyCanonicalSig on transactions by default.
 
+There are two cases where Ripple's signing implementations for the XRP Ledger do not automatically enable the tfFullyCanonicalSig flag. Users should take care to set the flag in these situations:
+
+- When the user explicitly specifies the `Flags` field of the transaction. Use bitwise OR to apply tfFullyCanonicalSig _and_ any other desired flags.
+- When the user provides a multi-signature for a transaction. Since different participants in a multi-signature must sign _exactly_ the same data, the signing code does not pre-process the transaction instructions to add the tfFullyCanonicalSig flag. For multi-signed transactions, always enable the tfFullyCanonicalSig flag explicitly.
+
 
 ### Malleability with Multi-Signatures
 
@@ -53,10 +58,21 @@ An important, explicit feature of multi-signing is that multiple different possi
 All of the following cases can lead to transaction malleability:
 
 - If a transaction still has enough signatures to meet its quorum after removing one or more. Any third party could remove a signature and calculate the hash of the modified transaction.
-- If one can add a valid signature to a transaction that already has a quorum. Only one of the sending account's authorized signers could add such a signature.
-- If one can replace one signature from a transaction with another valid signature while maintaining a quorum. Only one of the sending account's authorized signers could add such a signature.
+- If one can add a valid signature to a transaction that already has a quorum. Only an authorized signer of the sending account could create such a signature.
+- If one can replace one signature from a transaction with another valid signature while maintaining a quorum. Only an authorized signer of the sending account could create such a signature.
 
-Generally, you can avoid problems when multi-signing if you do not collect more signatures than necessary and your authorized signers are not trying to exploit your software. Even if your authorized signers are not intentionally malicious, confusion or poor coordination could cause several signers to submit different valid versions of the same transaction. When using multi-signatures, you should be prepared for the possibility that transaction executes with a different hash and set of signatures than you expected. Carefully monitor your account's sent transactions (for example, using the [account_tx method][]) and do not automatically re-create multi-signed transactions.
+Even if your authorized signers are not intentionally malicious, confusion or poor coordination could cause several signers to submit different valid versions of the same transaction.
+
+**Good operational security can protect against these problems.** Generally, you can avoid transaction malleability problems when multi-signing if you follow good operational security practices, including the following:
+
+- Do not collect more signatures than necessary.
+- Either designate one party to assemble a transaction after collecting the necessary number of signatures, or designate a "chain" where signers pass the transaction instructions forward to be signed in a predefined order.
+- Do not add unnecessary or untrusted signers to your multi-signing lists, even if the `weight` values associated with their keys are insufficient to authorize a transaction.
+- Be prepared for the possibility that a transaction executes with a different hash and set of signatures than you expected. Carefully monitor your account's sent transactions (for example, using the [account_tx method][]).
+- If you re-create transactions to be multi-signed, _do not_ change the `Sequence` number unless you have manually confirmed that the intended actions have not already executed.
+- Confirm that the tfFullyCanonicalSig flag is enabled before signing.
+
+For greater security, these guidelines provide multiple layers of protection.
 
 
 ## Exploit With Malleable Transactions
@@ -117,7 +133,8 @@ The process to exploit a vulnerable system follows a series of steps similar to 
 
     Worse, the vulnerable system might construct a new transaction to replace the transaction, picking new `Sequence`, `LastLedgerSequence`, and `Fee` parameters based on the current state of the network, but keeping the rest of the transaction the same as the original. If this new transaction is also malleable, the system could be exploited in the same way an indefinite number of times.
 
+
 <!--{# common link defs #}-->
-{% include '_snippets/rippled-api-links.md' %}			
-{% include '_snippets/tx-type-links.md' %}			
+{% include '_snippets/rippled-api-links.md' %}
+{% include '_snippets/tx-type-links.md' %}
 {% include '_snippets/rippled_versions.md' %}
--- a/content/concepts/consensus-network/transaction-malleability.md
+++ b/content/concepts/consensus-network/transaction-malleability.md
@@ -12,7 +12,7 @@ There are two circumstances that could lead to transaction malleability:
 
 2. The transaction is [multi-signed](multi-signing.html) and has more signatures than necessary. Even if the transaction originally did not have more signatures than necessary, it could be malleable if one of the authorized signers provides an additional signature.
 
-    There is no solution that prevents all variations of this problem. When using multi-signatures, monitor your account for transactions sent even if they have different hashes, and try not to collect more signatures than necessary.
+    Good operational security can protect against these problems. See [Mitigations for Multi-Signature Malleability](#mitigations-for-multi-signature-malleability) for guidelines.
 
 
 ## Background
@@ -62,6 +62,8 @@ All of the following cases can lead to transaction malleability:
 - If one can replace one signature from a transaction with another valid signature while maintaining a quorum. Only an authorized signer of the sending account could create such a signature.
 
 Even if your authorized signers are not intentionally malicious, confusion or poor coordination could cause several signers to submit different valid versions of the same transaction.
+
+#### Mitigations for Multi-Signature Malleability
 
 **Good operational security can protect against these problems.** Generally, you can avoid transaction malleability problems when multi-signing if you follow good operational security practices, including the following:
 
--- a/content/concepts/consensus-network/transaction-malleability.md
+++ b/content/concepts/consensus-network/transaction-malleability.md
@@ -25,6 +25,8 @@ In the XRP Ledger, a transaction cannot execute unless:
 
 Any change to the signed fields, no matter how small, would invalidate the signature, so no part of the transaction can be malleable except for the signature itself. In most cases, any change to a signature itself also invalidates the signature, but there are some specific exceptions, described below.
 
+Since the signature is among the data used to compute a transaction's identifying hash, any changes to a malleable transaction result in a different hash.
+
 ### Alternate secp256k1 Signatures
 
 To be "canonical", signatures created with the ECDSA algorithm and secp256k1 curve (the default) must meet the following requirements:
@@ -41,7 +43,7 @@ Thus, to have _fully_ canonical signatures, one must choose which of the two pos
 
 To maintain compatibility with older software that did not always generate fully canonical signatures, the XRP Ledger accepts transactions that are not fully canonical. To protect new users from exploits, the XRP Ledger has a flag on transactions called [**tfFullyCanonicalSig**](transaction-common-fields.html#global-flags), which requires that the transaction use a _fully-canonical_ signature to be valid.
 
-To calculate a fully-canonical ECDSA signature, one must compare S and N-S to determine which is smaller, then use that value in the `Signature` field of the transaction, and compute the identifying `hash` of the transaction using that version of the signature.
+To calculate a fully-canonical ECDSA signature, one must compare S and N-S to determine which is smaller, then use that value in the `Signature` field of the transaction.
 
 All XRP Ledger software that Ripple publishes (including `rippled`, and ripple-lib/RippleAPI) generates only fully-canonical signatures. To further protect users, Ripple has configured its code to enable the **tfFullyCanonicalSig** flag by default where possible. Ripple strongly encourages third-party implementations of XRP Ledger software to generate only fully-canonical signatures, and enable tfFullyCanonicalSig on transactions by default.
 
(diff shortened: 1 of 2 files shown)


content/concepts/consensus-network/transaction-malleability.md
```diff
@@ -0,0 +1,142 @@
+# Transaction Malleability
+
+A transaction is "malleable" if it can be changed in any way after being signed, without the keys to sign it. In the XRP Ledger, the **functionality** of a signed transaction cannot change, but in some circumstances a third party _could_ change the signature and identifying hash of a transaction.
+
+If vulnerable software submits malleable transactions and assumes they can only execute under the original hash, it may lose track of transactions. In the worst case, malicious actors could take advantage of this to steal money from the vulnerable system.
+
+There are two circumstances that could lead to transaction malleability:
+
+1. The transaction does not specify the tfFullyCanonicalSig flag on a transaction signed using the default signing algorithm (ECDSA with the secp256k1 curve).
```
scottschurr, line 9: I have a question aimed at @nbougalis.

What are the reasons that we don't just publicly declare that with some specific version of the server we will no longer accept non-canonical signatures?  Say we make a public announcement and give the development community a minimum of a year to make the adjustment.  It seems like that would be a responsible way to close the single-signing hole that Rome is describing here.

There must be a reason we haven't done that yet.  What is that reason?  Thanks.

content/concepts/consensus-network/transaction-malleability.md
```diff
@@ -0,0 +1,142 @@
...
+
+All XRP Ledger software that Ripple publishes (including `rippled`, and ripple-lib/RippleAPI) generates only fully-canonical signatures. To further protect users, Ripple has configured its code to enable the **tfFullyCanonicalSig** flag by default where possible. Ripple strongly encourages third-party implementations of XRP Ledger software to generate only fully-canonical signatures, and enable tfFullyCanonicalSig on transactions by default.
+
+There are two cases where Ripple's signing implementations for the XRP Ledger do not automatically enable the tfFullyCanonicalSig flag. Users should take care to set the flag in these situations:
+
+- When the user explicitly specifies the `Flags` field of the transaction. Use bitwise OR to apply tfFullyCanonicalSig _and_ any other desired flags.
+- When the user provides a multi-signature for a transaction. Since different participants in a multi-signature must sign _exactly_ the same data, the signing code does not pre-process the transaction instructions to add the tfFullyCanonicalSig flag. For multi-signed transactions, always enable the tfFullyCanonicalSig flag explicitly.
```
scottschurr, line 53: FWIW, it would be possible to adjust the `sign_for` RPC command to insert a `Flags` field in the transaction (with the `tfFullyCanonicalSig` bit set) if a `Flags` field is not specified when the transaction is submitted.  The `sign_for` already inserts an empty `SigningPubKey` in the transaction if one is not already present (https://github.com/ripple/rippled/blob/develop/src/ripple/rpc/impl/TransactionSign.cpp#L942-L945).

content/concepts/consensus-network/transaction-malleability.md
```diff
@@ -0,0 +1,142 @@
...
+
+
+### Malleability with Multi-Signatures
+
+An important, explicit feature of multi-signing is that multiple different possible configurations can render a transaction valid. For example, an account can be configured so that signatures from any three of five signers could authorize a transaction. However, this inherently means that there can be several different variations of a valid transaction, each with a different identifying hash.
+
+All of the following cases can lead to transaction malleability:
+
+- If a transaction still has enough signatures to meet its quorum after removing one or more. Any third party could remove a signature and calculate the hash of the modified transaction.
```
scottschurr: FWIW, I don't think they need to compute the hash.  Just remove one or more spare signatures and submit the modified transaction to the network.

content/concepts/consensus-network/transaction-malleability.md
```diff
@@ -0,0 +1,142 @@
...
+
+    Three ways that a transaction may not enable the tfFullyCanonicalSig flag are:
+
+    - The system explicitly specifies a `Flags` field that does not have the tfFullyCanonicalSig bit enabled.
+    - The transaction is multi-signed and does not explicitly enable the tfFullyCanonicalSig flag.
+    - The system omits the `Flags` field from the transaction fields, but uses a non-standard implementation that does not automatically enable tfFullyCanonicalSig when signing.
+
+    To be vulnerable, the transaction must be signed with an ECDSA key pair. If multi-signed, the transaction must be signed by at least one ECDSA key pair.
+
+    Most likely, the vulnerable transaction uses a fully-validated signature, but the flags indicate that the transaction would also be valid with a non-fully-canonical one. The transaction may also use `LastLedgerSequence` so that its final outcome is clear in a finite amount of time.
```
scottschurr: "fully-validated" -> "fully-canonical"?

content/concepts/consensus-network/transaction-malleability.md
```diff
@@ -0,0 +1,142 @@
...
+
+2. The system notes the identifying hash of the vulnerable transaction, submits it to the XRP Ledger network, then begins monitoring for that hash to be included in a validated ledger version.
+
+3. A malicious actor sees the transaction propagating through the network before it becomes confirmed.
+
+4. The malicious actor calculates the alternate signature for the vulnerable transaction, and derives the hash of the resulting transaction data.
```
scottschurr: I think they don't need to compute the hash.  As I understand it, it is sufficient to compute the alternative signature and submit the transaction with the alternate signature to the network.

content/concepts/consensus-network/transaction-malleability.md
```diff
@@ -0,0 +1,142 @@
...
+
+    Unlike creating a signature for a different transaction data, this does not require a large amount of computational work. It can be done in much less time than it takes to generate a signature in the first place.
```
scottschurr: I'm not a cryptographer, so I'm assuming you are correct -- that the alternative signature is relatively easy to compute given only the original transaction with the original signature.  @nbougalis, could you confirm?  Thanks.

content/concepts/consensus-network/transaction-malleability.md
```diff
@@ -0,0 +1,142 @@
...
+
+    Servers in the peer-to-peer network have no way of knowing which one "came first" or was intended by its original sender. Delays or other coincidences in network connectivity could result in validators seeing only one or the other by the time they finalize their consensus proposals, so either one could "win the race".
+
+    A malicious actor could increase the chances of getting non-canonical transactions confirmed if they controlled some number of well-connected servers in the peer-to-peer network, even if those servers are not trusted as validators.
+
+    If the malicious actor controls the only server to which the vulnerable system submitted the transaction, the malicious actor can easily control which version is distributed to the rest of the network.
+
+6. The malicious actor's version of the transaction achieves a consensus and becomes included in a validated ledger.
```
scottschurr: Do you want the "a" between "achieves" and "consensus"?

content/concepts/consensus-network/transaction-malleability.md
```diff
@@ -0,0 +1,142 @@
...
+
+    At this point, the transaction has executed and cannot be reversed. Its effects (such as sending XRP) are final. The original version of the transaction is no longer valid because its `Sequence` number has been used.
+
+    The effects of the transaction in the XRP Ledger are exactly the same as if the original version had executed.
+
+7. The vulnerable system does not see the transaction hash it is expecting, and erroneously concludes that the transaction did not execute.
+
+    If the transaction included the `LastLedgerSequence` field, this would occur after the specified ledger index has passed.
+
+    If the transaction omitted the `LastLedgerSequence` field, this could be wrong in another way: there is no amount of time after which you can definitively conclude that such a transaction could not later succeed. (See [Reliable Transaction Submission](reliable-transaction-submission.html) for details.)
```
scottschurr: I think this statement is not quite true?  If the user can confirm that the Sequence of the transaction has been consumed by the account (by, e.g., looking at the `Sequence` in `account_info`) then the user can conclude the original transaction can never succeed.  Or did I miss something?

scottschurr (approved):
:+1: Looks good to me.  I learned some useful stuff.  I left a few questions and suggestions, but you should deal with those as you see fit.  Thanks for writing this up!

content/concepts/consensus-network/transaction-malleability.md
```diff
@@ -0,0 +1,142 @@
...
+
+An ECDSA signature consists of two integers, called R and S. The secp256k1 modulus, called N, is a constant value for all secp256k1 signatures. Specifically, N is the value `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141`. For any given signature `(R,S)`, the signature `(R, N-S)` (that is, using N minus S in place of S) is also valid.
+
+Thus, to have _fully_ canonical signatures, one must choose which of the two possibilities is preferred and declare the other to be invalid. The creators of the XRP Ledger decided arbitrarily to prefer the _smaller_ of the two possible values, `S` or `N-S`. A transaction is considered _fully canonical_ if it uses the preferred (smaller) value of `S`, and follows all the normal rules for being canonical.
+
+To maintain compatibility with older software that did not always generate fully canonical signatures, the XRP Ledger accepts transactions that are not fully canonical. To protect new users from exploits, the XRP Ledger has a flag on transactions called [**tfFullyCanonicalSig**](transaction-common-fields.html#global-flags), which requires that the transaction use a _fully-canonical_ signature to be valid.
```
ximinez, line 44: Do we have any data on how common non-canonical signatures are? Maybe it's time to start deprecating the backward compatibility with not fully canonical sigs.

content/concepts/consensus-network/transaction-malleability.md
```diff
@@ -0,0 +1,142 @@
...
+
+#### Mitigations for Multi-Signature Malleability
+
+**Good operational security can protect against these problems.** Generally, you can avoid transaction malleability problems when multi-signing if you follow good operational security practices, including the following:
+
+- Do not collect more signatures than necessary.
+- Either designate one party to assemble a transaction after collecting the necessary number of signatures, or designate a "chain" where signers pass the transaction instructions forward to be signed in a predefined order.
+- Do not add unnecessary or untrusted signers to your multi-signing lists, even if the `weight` values associated with their keys are insufficient to authorize a transaction.
+- Be prepared for the possibility that a transaction executes with a different hash and set of signatures than you expected. Carefully monitor your account's sent transactions (for example, using the [account_tx method][]).
+- If you re-create transactions to be multi-signed, _do not_ change the `Sequence` number unless you have manually confirmed that the intended actions have not already executed.
```
ximinez, line 77: Can we explicitly suggest carefully monitoring the account's sequence number? This item kind of implies it (because you won't need to change the sequence number if no transactions processed), but an explicit check can also indicate that there needs to be some manual intervention.

ximinez (changes requested):
Looks good overall. I left a couple of comments, and I'll sign off once those are addressed (either way).

content/concepts/consensus-network/transaction-malleability.md
intelliot: I haven't compiled long-term data but I grabbed some current data because I'm curious. Unless there's a bug in my script, it looks like there are quite a few transactions that aren't setting the flag.

For each ledger version, this shows the # of transactions with tfFullyCanonicalSig set, the total # of transactions in the ledger, and % calculated from these two numbers.
```
Ledger version: 44337368
tfFullyCanonicalSig count: 20
Total tx count: 33
tfFullyCanonicalSig percent: 61%
------------------------
Ledger version: 44337369
tfFullyCanonicalSig count: 27
Total tx count: 29
tfFullyCanonicalSig percent: 93%
------------------------
Ledger version: 44337370
tfFullyCanonicalSig count: 24
Total tx count: 25
tfFullyCanonicalSig percent: 96%
------------------------
Ledger version: 44337371
tfFullyCanonicalSig count: 22
Total tx count: 38
tfFullyCanonicalSig percent: 58%
------------------------
Ledger version: 44337372
tfFullyCanonicalSig count: 24
Total tx count: 25
tfFullyCanonicalSig percent: 96%
------------------------
Ledger version: 44337373
tfFullyCanonicalSig count: 23
Total tx count: 36
tfFullyCanonicalSig percent: 64%
```

ximinez: Oh well.

content/concepts/consensus-network/transaction-malleability.md
intelliot: I agree - I think every app/user should carefully monitor their sequence number. It's a great way to, e.g., ensure they aren't accidentally double-sending a tx that already succeeded.

content/concepts/consensus-network/transaction-malleability.md
mDuo13: I [made the same suggestion](https://ripplelabs.atlassian.net/wiki/spaces/~Rome/pages/376045583/Proposal+Amendment+to+Always+Require+Fully-Canonical+Signatures) and would also like to know.

content/concepts/consensus-network/transaction-malleability.md
mDuo13: I think this would be a good change as long as it's synced with ripple-lib and there's an announcement that explains it. (Otherwise you could get people "multi-signing" different data with different software. Obviously the workaround is to explicitly specify the `Flags` field in all cases.)

But also, if we're going to move on making non-fully-canonical signatures irrelevant (as mentioned above) this would be unnecessary.

content/concepts/consensus-network/transaction-malleability.md
mDuo13: My understanding is that the "work required" is literally subtracting an integer mod another integer (so the description here may be an understatement!) but I didn't test it myself so I can only go on what I read elsewhere.

content/concepts/consensus-network/transaction-malleability.md
mDuo13: You're not really wrong, but the statement is not wrong either; it's supposed to poke holes on the logic of, "Well, it's been a long time and the transaction hasn't been included in a ledger, so it must have failed." Like, even if you assume the transaction isn't malleable, if you left off `LastLedgerSequence`, there are other unintuitive cases you should also watch out for.

I'll work on rewording this.

content/concepts/consensus-network/transaction-malleability.md
mDuo13: Ah, but the _not setting the flag_ and _actually using a non-fully-canonical signature_ are quite different.

It's my understanding that almost all software (if not entirely all software) signing transactions exclusively creates fully-canonical signatures at this point. The problem is people don't know to set the flag, so their transactions are vulnerable anyway—the fully canonical signature could be replaced with a non-fully-canonical signature and still be valid.

content/concepts/consensus-network/transaction-malleability.md
ximinez: Rough suggestion:
> If the transaction omitted the `LastLedgerSequence` field, this could be wrong in another way: in the absence of a different validated transaction consuming the same sequence number, there is no amount of time after which you can definitively conclude that such a transaction could not later succeed.

content/concepts/consensus-network/transaction-malleability.md
ximinez: @intelliot Since you've already got something to look at the flag, how hard would it be to add a check to look at the canonicality of the signature?

content/concepts/consensus-network/transaction-malleability.md
jhaaaa: nit: possible edit?

>Unlike creating a signature for a different transaction data, this 

to

>Unlike creating a signature for a different transaction, this

dactyl-config.yml
```diff
@@ -470,6 +470,15 @@ pages:
         targets:
             - local
 
+    -   md: concepts/consensus-network/transaction-malleability.md
+        html: transaction-malleability.html
+        funnel: Docs
+        doc_type: Concepts
+        category: Consensus Network
+        blurb: Be aware of ways transactions could be changed have a different hash than expected.
```
jhaaaa: nit: possible edit?

>Be aware of ways transactions could be changed have a different hash than expected.

to

>Be aware of ways transactions could be changed to have a different hash than expected.

content/concepts/consensus-network/transaction-malleability.md
mDuo13: Went slightly the other direction with the edit:

> Unlike creating a signature for different transaction instructions, this

content/concepts/consensus-network/transaction-malleability.md
mDuo13: _shrug_

I guess it works just as well without, so may as well remove it.

content/concepts/consensus-network/transaction-malleability.md
mDuo13: I went with this:

> If the transaction omitted the `LastLedgerSequence` field, this could be wrong in another way: if no other transaction from the same sender uses the same `Sequence` number, then the transaction could theoretically succeed later regardless of how much time has passed.

mDuo13:
!rebuild

PR #494, opened 2019-01-22
Add Check & Preauth txn types to common fields

professorhantzen:


mDuo13:
Thanks for catching this.

I'm hesitant to merge this because Checks aren't yet enabled on the production XRP Ledger so people could be tricked/confused by their presence in this list.

We could have a link like "Some transaction types depend on amendments being enabled" but I'm worried that for a relative beginner that could take them down a rathole or spook them into worrying that a bunch of (long-since enabled) transaction types might not be available. What do you think?

professorhantzen:
Good point. Specifically in this case because I was myself fooled by the Checks' presence in the sidebar TransactionTypes list - I'd assumed from that they were already enabled.

I now notice some of the "requires xxx amendment" notes might need to be updated to "This transaction type was introduced by the xxx amendment", as per SignerListSet.

I'd be happy to go through the TransactionType's Common Fields and match it as per https://developers.ripple.com/known-amendments.html if it's useful.  

I also wonder if unimplemented TransactionTypes should have something next to their name in the sidebar list to differentiate unimplemented from implemented types?  (Maybe have their BG-coloured differently, and match that colour with the BG of a "requires xxx amendment" note in the main doc?)

mDuo13:
Yes, please change the list in the Common Types to match what's currently enabled in production.

And you're right that we should be a little more consistent with the way we call out which changes depend on which amendments, both in the text and in the web design. I'll look into a way to mark specific pages as depending on an amendment that's not enabled in production yet.

PR #500, opened 2019-01-28
Dependency updates for rippled 1.2

jhaaaa:
- Updates "Build and Run `rippled` on macOS" tutorial to provide instructions on how to manually compile and install Boost 1.67. Previously, this was accomplished using the boost@1.67 brew formula, but the boost@1.67 formula is no longer available. (Note: These instructions will likely have to change soon once Boost 1.68 is tested and used in CI. The boost@1.68 formula does exist.) I used these instructions to get the `rippled` `develop` branch running on my Mac using Boost 1.67 and CMake 3.13.3.

- Updates "Build and Run `rippled` on Ubuntu" tutorial to provide instructions on how to manually compile and install CMake 3.9.0 or higher. `rippled` 1.2 requires CMake 3.9.0 or higher and the version provided by Ubuntu is CMake 3.5.0, thus the need for manual installation. I used these instructions to get the `rippled` `develop` branch running on my AWS EC2 Ubuntu 16.04 instance using Boost 1.67 and CMake 3.13.3.

### DIFF
--- a/content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
+++ b/content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
@@ -29,43 +29,18 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
 
         sudo apt-get -y install git pkg-config protobuf-compiler libprotobuf-dev libssl-dev wget
 
-4. Compile CMake.
+4. Install CMake.
 
-    Version 1.2.0 of `rippled` requires CMake version 3.9.0 or higher. Because CMake version 3.9.0 or higher isn't available in the Ubuntu 16.04 software repositories, you must compile it yourself.
+    Version 1.2.0 of `rippled` requires CMake 3.9.0 or higher. For the purposes of this tutorial, we used CMake 3.13.3, which was the latest version available at the time of writing.
 
-    If you have previously built CMake 3.9.0 or higher for `rippled` and configured the `CMAKE_ROOT` environment variable, you can skip these steps.
+    If you have previously installed the latest version of CMake, you can skip these steps.
 
-      1. Download CMake 3.13.3. For the purposes of this tutorial, we used CMake 3.13.3, which was the latest version available at the time of writing.
+    To install CMake 3.13.3:
 
-              wget https://cmake.org/files/v3.13/cmake-3.13.3.tar.gz
+        wget https://github.com/Kitware/CMake/releases/download/v3.13.3/cmake-3.13.3-Linux-x86_64.sh
+        sudo sh cmake-3.13.3-Linux-x86_64.sh --prefix=/usr/local --exclude-subdir
 
-      2. Extract `cmake-3.13.3.tar.gz`.
-
-              tar xvzf cmake-3.13.3.tar.gz
-
-      3. Change to the new `cmake-3.13.3` directory.
-
-              cd cmake-3.13.3
-
-      4. Prepare to build the CMake installer.  ***TODO: stated correctly?***
-
-              ./bootstrap
-
-      5. Build the CMake installer. ***TODO: stated correctly?***
-
-              make
-
-      6. Install CMake.
-
-              sudo make install
-
-      7. Set the environment variable `CMAKE_ROOT` to point to the new `cmake-3.13.3` directory. It's best to put this environment variable in your `.profile`, or equivalent, file for your shell so it's automatically set when you log in. Add the following line to the file:
-
-              export CMAKE_ROOT=/home/my_user/cmake-3.13.3
-
-      8. Source your updated `.profile` file. For example:
-
-              source ~/.profile
+    Use `cmake --version` to verify that the installation worked.
 
 5. Compile Boost.
 
@@ -95,7 +70,7 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
 
           **Tip:** This example uses 4 processes to build in parallel. The best number of processes to use depends on how many CPU cores your hardware has available. You can use `cat /proc/cpuinfo` to get information about your hardware's processor.
 
-      6. Set the environment variable `BOOST_ROOT` to point to the new `boost_1_68_0` directory. It's best to put this environment variable in your `.profile`, or equivalent, file for your shell so it's automatically set when you log in. Add the following line to the file:
+      6. Set the environment variable `BOOST_ROOT` to point to the new `boost_1_67_0` directory. It's best to put this environment variable in your `.profile`, or equivalent, file for your shell so it's automatically set when you log in. Add the following line to the file:
 
               export BOOST_ROOT=/home/my_user/boost_1_67_0
 
@@ -133,7 +108,7 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
 
           **Tip:** The default build includes debugging symbols, which can be useful for development but are inefficient in production. To build `rippled` for use on production servers, add the `-DCMAKE_BUILD_TYPE=Release` flag when running the `cmake` command.
 
-      2. Build the `rippled` binary executable.
+      2. Build the `rippled` binary executable. This may take about 30 minutes, depending on your hardware specs.
 
               cmake --build .
 
--- a/content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
+++ b/content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
@@ -33,7 +33,7 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
 
     Version 1.2.0 of `rippled` requires CMake 3.9.0 or higher. For the purposes of this tutorial, we used CMake 3.13.3, which was the latest version available at the time of writing.
 
-    If you have previously installed the latest version of CMake, you can skip these steps.
+    If you have previously installed CMake 3.9.0 or higher, you can skip these steps.
 
     To install CMake 3.13.3:
 
--- a/content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
+++ b/content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
@@ -4,7 +4,7 @@
 
 For an overview of `rippled`, see [Operating rippled Servers](install-rippled.html).
 
-Use these instructions to build a `rippled` executable from source version 1.1.0 or higher on Ubuntu Linux 16.04 or higher. These instructions were tested on Ubuntu 16.04 LTS.
+Use these instructions to build a `rippled` executable from source version 1.2.0 or higher on Ubuntu Linux 16.04 or higher. These instructions were tested on Ubuntu 16.04 LTS.
 
 For information about building `rippled` for other platforms, see [Builds](https://github.com/ripple/rippled/tree/develop/Builds) in the `rippled` GitHub repository.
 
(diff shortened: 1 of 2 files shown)


content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
```diff
@@ -97,13 +133,11 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
 
           **Tip:** The default build includes debugging symbols, which can be useful for development but are inefficient in production. To build `rippled` for use on production servers, add the `-DCMAKE_BUILD_TYPE=Release` flag when running the `cmake` command.
 
-      2. Build the `rippled` binary executable. Replace `<number of parallel jobs>` with the number of jobs to run in parallel. Choose this value based on the number of CPU cores you want to use for building.
+      2. Build the `rippled` binary executable.
 
-              cmake --build . -- -j 4
```
jhaaaa, line 102: I removed the `-- -j 4` instructions b/c the build hung at 83% repeatedly when I included them. In the past, these instructions worked fine.

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
```diff
@@ -27,11 +27,49 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
...
+4. Compile CMake.
 
-    Version 1.1.0 of `rippled` requires Boost version 1.67.0 exactly. Because Boost version 1.67.0 isn't available in the Ubuntu 16.04 software repositories, you must compile it yourself.
+    Version 1.2.0 of `rippled` requires CMake version 3.9.0 or higher. Because CMake version 3.9.0 or higher isn't available in the Ubuntu 16.04 software repositories, you must compile it yourself.
+
+    If you have previously built CMake 3.9.0 or higher for `rippled` and configured the `CMAKE_ROOT` environment variable, you can skip these steps.
+
+      1. Download CMake 3.13.3. For the purposes of this tutorial, we used CMake 3.13.3, which was the latest version available at the time of writing.
+
+              wget https://cmake.org/files/v3.13/cmake-3.13.3.tar.gz
```
mellery451: hmm - I can't think of any reason to recommend building cmake from source since the binary package works fine for standard distros:  https://github.com/Kitware/CMake/releases/download/v3.13.3/cmake-3.13.3-Linux-x86_64.sh.

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
```diff
@@ -27,11 +27,49 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
...
+
+              make
+
+      6. Install CMake.
+
+              sudo make install
+
+      7. Set the environment variable `CMAKE_ROOT` to point to the new `cmake-3.13.3` directory. It's best to put this environment variable in your `.profile`, or equivalent, file for your shell so it's automatically set when you log in. Add the following line to the file:
+
+              export CMAKE_ROOT=/home/my_user/cmake-3.13.3
```
mellery451: I'm not aware of any reason for setting `CMAKE_ROOT`, unless you use it somewhere else in these instructions. Also, if you did a `make install` like above or use the binary installer, it will install in `/usr/local` by default which should be part of the standard search path.

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
mellery451: as long as we are recommending cmake 3.13, then we can suggest `cmake --build . --parallel` ...the `--parallel` option is new and it will figure out what to do for the underlying generator instead of having to pass the makefile specific `-j`.

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
```diff
@@ -57,37 +95,35 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
 
           **Tip:** This example uses 4 processes to build in parallel. The best number of processes to use depends on how many CPU cores your hardware has available. You can use `cat /proc/cpuinfo` to get information about your hardware's processor.
 
-      6. Set the environment variable `BOOST_ROOT` to point to the new `boost_1_67_0` directory. It's best to put this environment variable in your `.profile`, or equivalent, file for your shell so it's automatically set when you log in. Add the following line to the file:
+      6. Set the environment variable `BOOST_ROOT` to point to the new `boost_1_68_0` directory. It's best to put this environment variable in your `.profile`, or equivalent, file for your shell so it's automatically set when you log in. Add the following line to the file:
```
mellery451: I think this should still be 1_67 ?

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
```diff
@@ -57,37 +95,35 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
...
 
-        cd ~
         git clone https://github.com/ripple/rippled.git
         cd rippled
         git checkout master
 
-6. Check the commit log to be sure you're compiling the version you intend to. The most recent commit should be signed by a well-known Ripple developer and should set the version number to the latest released version. For example:
+7. Check the commit log to be sure you're compiling the version you intend to. The most recent commit should be signed by a well-known Ripple developer and should set the version number to the latest released version. For example:
 
         $ git log -1
```
mellery451, line 89: since we sign our version bumps (head of release), should we suggest adding `--show-signature` to the git log command line? I don't know if it helps instill confidence or not.

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
mellery451: also, I'm not sure why `-j 4` hangs for you..we should try to figure that out, although I'm curious if `--parallel` shows the same behavior.

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
```diff
@@ -27,11 +27,49 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
...
-        sudo apt-get -y install git cmake pkg-config protobuf-compiler libprotobuf-dev libssl-dev wget
+        sudo apt-get -y install git pkg-config protobuf-compiler libprotobuf-dev libssl-dev wget
 
-4. Compile Boost.
+4. Compile CMake.
 
-    Version 1.1.0 of `rippled` requires Boost version 1.67.0 exactly. Because Boost version 1.67.0 isn't available in the Ubuntu 16.04 software repositories, you must compile it yourself.
+    Version 1.2.0 of `rippled` requires CMake version 3.9.0 or higher. Because CMake version 3.9.0 or higher isn't available in the Ubuntu 16.04 software repositories, you must compile it yourself.
+
+    If you have previously built CMake 3.9.0 or higher for `rippled` and configured the `CMAKE_ROOT` environment variable, you can skip these steps.
```
mellery451: I feel like we mention version 3.9 several times here and then jump into instructions for 3.13...that could be confusing ?  I don't have any specific suggestions about how to ameliorate this other than maybe just referring to "latest release" or something like that.

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
jhaaaa: Discussed with Mike. `--parallel` resulted in `virtual memory exhausted: Cannot allocate memory` messages for my build, so leaving all option commands out for now.

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
jhaaaa: Discussed with Mike. Leaving it off for now because the user must have access to the signer's public key and those keys aren't readily accessible right now. I will follow up with Nik to see if we can make those available. We can then add this option to the command line and give users a link to the location of the public keys of folks who can sign `rippled` releases.

jhaaaa:
Thank you for your review, @mellery451! I've updated the PR to address your feedback. 

@mDuo13 - if you have the time and interest, your feedback is more than welcome.

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
```diff
@@ -27,11 +27,24 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
...
-        sudo apt-get -y install git cmake pkg-config protobuf-compiler libprotobuf-dev libssl-dev wget
+        sudo apt-get -y install git pkg-config protobuf-compiler libprotobuf-dev libssl-dev wget
 
-4. Compile Boost.
+4. Install CMake.
 
-    Version 1.1.0 of `rippled` requires Boost version 1.67.0 exactly. Because Boost version 1.67.0 isn't available in the Ubuntu 16.04 software repositories, you must compile it yourself.
+    Version 1.2.0 of `rippled` requires CMake 3.9.0 or higher. For the purposes of this tutorial, we used CMake 3.13.3, which was the latest version available at the time of writing.
+
+    If you have previously installed the latest version of CMake, you can skip these steps.
```
mDuo13: Nit: Instead of "the latest version of CMake" here, let's say, "CMake 3.9.0 or higher"—since, installing a version of CMake that was the latest at the time but is lower than 3.9.0 is not sufficient.

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
```diff
@@ -59,35 +72,33 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install the sof
...
         cd rippled
         git checkout master
 
-6. Check the commit log to be sure you're compiling the version you intend to. The most recent commit should be signed by a well-known Ripple developer and should set the version number to the latest released version. For example:
+7. Check the commit log to be sure you're compiling the version you intend to. The most recent commit should be signed by a well-known Ripple developer and should set the version number to the latest released version. For example:
 
         $ git log -1
 
-        commit 3e22a1e9e8f2de450eded6ca4c2db6411e329b2a
+        commit 4f3a76dec00c0c7ea28e78e625c68499debbbbf3
```
mDuo13, line 91: Aside: We need to remember to update this as soon as the actual 1.2.0 release is live.

mDuo13 (approved):
The updated instructions look good.

As an aside, I was able to use these to build on Arch Linux (latest updates) as well, with the following caveats:

- I had to re-build Boost after upgrading my system. Leaving the old boost 1.67.0 directory there didn't work, and for some reason it picked up my system's boost 1.69.0 (which isn't compatible) even though the environment variable appeared to be set correctly.
- As you might expect, I used Arch's package manager (`pacman`) rather than `apt-get` to install the necessary packages, which are mostly available with similar names. Well, to be precise, I already had all the dependencies from building `rippled` in the past.
    - Arch has CMake 3.13.3 already, so I didn't have to run the install script. (Consequently, I didn't test those instructions.)
    - Arch doesn't have an up-to-date `openssl-static` package, so I used `cmake .. -Dstatic=OFF` instead of just `cmake ..` during the "generate the build system" part (step 9.1).

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
jhaaaa: Ah - great point. Thank you - I'll be sure to take care of this.

PR #501, opened 2019-01-29
Add (DEV) sidebar label for features not enabled

mDuo13:
Inspired by #494, I made a label that can be added to mark some pages as "dev" in the sidebar, in case (like Checks), they're documented but not currently available on the production XRP Ledger.

Preview:

Hovering over the "DEV" text shows the tooltip "This feature is not currently available on the production XRP Ledger."


jhaaaa:
This is great. I like everything about this idea! How wonderful to be able to provide this key info, at-a-glance. Thank you for taking the time to develop this.

Just a couple of thoughts that came to mind (no pressure to do anything with them):

- I wonder if it might be good to use the statuses we use for the Known Amendments as the text for the markers? So Checks would have a "Planned" marker? I know it is 4 more characters, but it may be a good to keep them consistent? This is what it might look like: https://drive.google.com/file/d/1tO6pa-pnzjnVOFh8jmVvkbwlEfO5oWzl/view?usp=sharing Or maybe shorten it to "Plan"? 

- Not sure if this is doable, or worth taking the time to do, but just wondering if it would be useful to provide this marker inline with text, too. So, for example, it can call out these Planned and In Dev features when they are discussed at a more granular level than a page. This is what it might look like: https://drive.google.com/file/d/19qe9LQc4-RRAvoWwmDfXLhwjqVgpNbQP/view?usp=sharing. If we do this, I think we'd have to standardize the way we refer to these Planned or In Dev amendments so that the text and marker look right together. We should also look at the areas that would make heavy use of this to ensure that using the markers inline doesn't make the content LESS readable.

mDuo13:
I think there's room to have more than one status, but I'd like to limit it to **as few as possible** both for readers' sanity and our own maintenance burden. The level of granularity associated with the Known Amendments page is probably overkill. Without condensing them to names brief enough to fit, here are some categories we could consider:

- Features that are available on the production network after being enabled by amendments. I'd like to keep this, like it is now, unlabeled.
- Features that are available on the test net but not production yet. (These could be labeled as "TEST".) The Known Amendments page doesn't have a specific status for this, as its statuses are purely relating to what's up with the amendment in production.
- Features that are not yet in a release, but are expected to be released soon. The Known Amendments page calls these "In Development" for the most part. (Right now we have kind of an exception where some stuff that's confirmed in 1.2.0 is listed even though 1.2.0 isn't officially released yet.)
- Features that are in a release, but not enabled in production nor the test net. These might be "planned", "open for voting", or "expected" in the Known Amendments page depending on how much validator support there is and what Ripple is doing. So far, Ripple has controlled a big enough percentage of the validator list to hold back amendments from receiving 80% support until the company has been ready to throw its weight behind them, but that's likely to change soon, which will make the "Planned" status basically obsolete.

The only very clear-cut distinction I can think of is "enabled in prod" vs. "not enabled in prod". I think the term "dev" reasonably captures in very few letters that the feature is not currently prime-time available.

The number of letters for a status isn't strictly a problem, but things will, in general, be easier to fit in more places if the statuses are smaller. We could also theoretically use icons, which if done well are probably not _much_ less intuitive than abbreviations like "DEV", but I didn't see an obvious choice on, say, [fontawesome's list](https://fontawesome.com/icons?d=gallery&m=free).

If Ripple starts maintaining multiple test nets (as has been discussed—for example, a "mirror production" test net alongside a "future feature" test net) then we may also want to revisit adding statuses that convey features' status with respect to different nets.

ryangyoung:
> The only very clear-cut distinction I can think of is "enabled in prod" vs. "not enabled in prod".

Agreed. 

> I think the term "dev" reasonably captures in very few letters that the feature is not currently prime-time available.

I can see where you're coming from here but my gut instinct is that it's confusing to have two different sets of statuses for the same thing, even if one is "reasonably" intuitive. (Also, I'm not as sure that DEV is immediately intuitive to mean what we mean here: I'd probably assume that the feature was in development, not not available on the production ledger.)

I'd vote for using the same statuses throughout the docs. Or, if we do use different statuses, to create a definitions table that explains what each means in the different contexts they're used. 

> If Ripple starts maintaining multiple test nets (as has been discussed—for example, a "mirror production" test net alongside a "future feature" test net) then we may also want to revisit adding statuses that convey features' status with respect to different nets.

This sounds good. And would be another reason for clearly defining the different statuses.

ryangyoung:
As for whether to add the statuses inline on each mention, I think that depends on whether we end up using the same statuses or not and the level of effort it would take to implement the solution. It seems like a nice-to-have feature at this point, unless user testing reveals more significant confusion.

mDuo13:
Switched to a blue flask icon for not enabled. Tried adding a filter that makes it easy to inline the icons in body text, and I seem to have encountered a Dactyl bug in the process. (It's running certain filters more than they ought to run.)

I'll have to look into that...

mDuo13:
:point_up_2: Preview of what in-body icons would look like

mDuo13:
Sidebar with the flask icons

PR #503, opened 2019-01-30
Transaction Queue Reserve Changes for 1.2.0

mDuo13:
Updates the guidance in the transaction queue page based on the changes from https://github.com/ripple/rippled/pull/2750.

Fixes DOC-2094.

### DIFF
--- a/content/concepts/consensus-network/transaction-queue.md
+++ b/content/concepts/consensus-network/transaction-queue.md
@@ -30,7 +30,10 @@ The `rippled` server uses a variety of heuristics to estimate which transactions
 
 * Transactions must be properly-formed and [authorized](transaction-basics.html#authorizing-transactions) with valid signatures.
 * Transactions with an `AccountTxnID` field cannot be queued.
-* A single sending address can have at most 10 transactions queued at the same time. In order for a transaction to be queued, the sender must have enough XRP to pay all the XRP costs of all the sender's queued transactions including both the `Fee` fields and the sum of the XRP that each transaction could send. [New in: rippled 0.32.0][]
+* A single sending address can have at most 10 transactions queued at the same time. In order for a transaction to be queued, the sender must have enough XRP to pay all the XRP costs of all the sender's queued transactions, including all of the following:
+    - The `Fee` fields of all queued transactions.
+    - The sum amount the XRP that each transaction could send.
+    - The amount of XRP the account must keep in [reserve](reserves.html). [Updated in: rippled 1.2.0][New in: rippled 1.2.0]
 * If a transaction affects how the sending address authorizes transactions, no other transactions from the same address can be queued behind it. [New in: rippled 0.32.0][]
 * If a transaction includes a `LastLedgerSequence` field, the value of that field must be at least **the current ledger index + 2**.
 
--- a/content/concepts/consensus-network/transaction-queue.md
+++ b/content/concepts/consensus-network/transaction-queue.md
@@ -32,7 +32,7 @@ The `rippled` server uses a variety of heuristics to estimate which transactions
 * Transactions with an `AccountTxnID` field cannot be queued.
 * A single sending address can have at most 10 transactions queued at the same time. In order for a transaction to be queued, the sender must have enough XRP to pay all the XRP costs of all the sender's queued transactions, including all of the following:
     - The `Fee` fields of all queued transactions.
-    - The sum amount the XRP that each transaction could send.
+    - The sum amount of XRP that all queued transactions could send.
     - The amount of XRP the account must keep in [reserve](reserves.html). [Updated in: rippled 1.2.0][New in: rippled 1.2.0]
 * If a transaction affects how the sending address authorizes transactions, no other transactions from the same address can be queued behind it. [New in: rippled 0.32.0][]
 * If a transaction includes a `LastLedgerSequence` field, the value of that field must be at least **the current ledger index + 2**.
--- a/content/concepts/consensus-network/transaction-queue.md
+++ b/content/concepts/consensus-network/transaction-queue.md
@@ -32,7 +32,7 @@ The `rippled` server uses a variety of heuristics to estimate which transactions
 - Transactions with an `AccountTxnID` field cannot be queued.
 - A single sending address can have at most 10 transactions queued at the same time.
 - To queue a transaction, the sender must have enough XRP for all of the following: [Updated in: rippled 1.2.0][New in: rippled 1.2.0]
-    - Destroying the XRP [transaction cost](transaction-cost.html) as specified in the `Fee` fields of all the sender's queued transactions. This amount cannot be more than the base account reserve (currently 20 XRP). (Transactions paying significantly more than the minimum transaction cost of 0.00001 XRP typically skip the queue and go straight into the open ledger.)
+    - Destroying the XRP [transaction cost](transaction-cost.html) as specified in the `Fee` fields of all the sender's queued transactions. The total amount among queued transactions cannot be more than the base account reserve (currently 20 XRP). (Transactions paying significantly more than the minimum transaction cost of 0.00001 XRP typically skip the queue and go straight into the open ledger.)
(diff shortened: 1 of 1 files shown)


content/concepts/consensus-network/transaction-queue.md
```diff
@@ -30,7 +30,10 @@ The `rippled` server uses a variety of heuristics to estimate which transactions
 
 * Transactions must be properly-formed and [authorized](transaction-basics.html#authorizing-transactions) with valid signatures.
 * Transactions with an `AccountTxnID` field cannot be queued.
-* A single sending address can have at most 10 transactions queued at the same time. In order for a transaction to be queued, the sender must have enough XRP to pay all the XRP costs of all the sender's queued transactions including both the `Fee` fields and the sum of the XRP that each transaction could send. [New in: rippled 0.32.0][]
+* A single sending address can have at most 10 transactions queued at the same time. In order for a transaction to be queued, the sender must have enough XRP to pay all the XRP costs of all the sender's queued transactions, including all of the following:
+    - The `Fee` fields of all queued transactions.
+    - The sum amount the XRP that each transaction could send.
```
jhaaaa: nit: possibly change

>The sum amount _the_ XRP that each transaction could send.

to

>The sum amount _of_ XRP that each transaction could send.

jhaaaa (approved):
Left one note about a possible nit change... LGTM! 👍

content/concepts/consensus-network/transaction-queue.md
```diff
@@ -30,7 +30,10 @@ The `rippled` server uses a variety of heuristics to estimate which transactions
 
 * Transactions must be properly-formed and [authorized](transaction-basics.html#authorizing-transactions) with valid signatures.
 * Transactions with an `AccountTxnID` field cannot be queued.
-* A single sending address can have at most 10 transactions queued at the same time. In order for a transaction to be queued, the sender must have enough XRP to pay all the XRP costs of all the sender's queued transactions including both the `Fee` fields and the sum of the XRP that each transaction could send. [New in: rippled 0.32.0][]
+* A single sending address can have at most 10 transactions queued at the same time. In order for a transaction to be queued, the sender must have enough XRP to pay all the XRP costs of all the sender's queued transactions, including all of the following:
+    - The `Fee` fields of all queued transactions.
+    - The sum amount of XRP that all queued transactions could send.
+    - The amount of XRP the account must keep in [reserve](reserves.html). [Updated in: rippled 1.2.0][New in: rippled 1.2.0]
```
ximinez: Two things:
1. I'm not really clear about about what this would mean to a typical user. The rule in the code is that an account can not queue up multiple transactions where the total fees are more than the base reserve. I'm not sure how to phrase that in the context of this list. 
2. You have both "Updated in" and "New in" indicating the same version: 1.2.0.

ximinez (changes requested):

content/concepts/consensus-network/transaction-queue.md
intelliot: > an account can not queue up multiple transactions where the total fees are more than the base reserve.

base reserve = 20 XRP (currently)? Why don't we allow an account to queue up multiple transactions where the total fees are more than 20 XRP?

ximinez: @intelliot Because we want all of those transactions to be "likely to claim a fee" with simple heuristics that don't vary as conditions change on the ledger.

Example:
* alice has 10000 XRP.
* alice submits tx 10: an `OfferCreate` that offers 20000XRP for 200USD with a 9XRP fee. This transaction gets queued.
* alice submits tx 11: a `Payment` that sends 5000XRP to bob with a 9XRP fee. This transaction gets queued.
* alice submits tx 12: an `OfferCancel` that cancels the above offer with a 9XRP fee. It also gets queued.
* finally, alice submits tx 13: an `AccountSet` that is a noop, also with a 9XRP fee. It gets queued.
* At this point, all 4 transactions have been broadcast and attempted by all the other nodes on the network. Assume they all make the same decisions. alice has a total of 36XRP in fees in the queued transactions.
* Some time later, tx 10 gets processed to the open ledger, and just by luck, it crosses. Because she offers more XRP than her balance, it only partially crosses, though - her reserve is not processed as part of the offer. So now her XRP balance is 11XRP (the 20XRP reserve minus the 9XRP fee), and some amount of USD that I'm not going to worry about. Also, the remainder of the offer is left in the order book (IIRC).
* Then tx 11 gets processed to the open ledger. The payment can't send, because she has an insufficient balance, but we can still claim a fee. So her XRP balance is 2XRP.
* Now tx 12 tries to process to the open ledger . Because her balance of is less than the fee, it can't get into an open ledger, it fails with a `terINSUF_FEE_B` and stays in the queue for a few more attempts. If alice doesn't get any incoming payments, it's eventually dropped. (If the transaction did somehow make it into an open ledger, it would wipe out alice's balance to get as much fee as possible, with a result of `tecINSUFF_FEE`, but wouldn't make any other changes to the ledger.)
* Once tx 12 is dropped, if another tx with a good fee is not submitted, tx 13 will be attempted (only once, though, IIRC), and will fail with `terPRE_SEQ`, since there was no tx 12 for the account. Then it gets dropped as well.

Why is this bad?
1. Alice should have paid 36XRP in fees, but ended up only paying 18XRP.
2. Two of alice's transactions were broadcast to the entire network, incurring resource consumption (signature checking is the big one here, but also memory, time, CPU, etc.), and yet they did not claim any fee at all.

Because there are so many possible interactions that can reduce an account's balance directly and indirectly (eg. an offer already on the books gets taken while transactions are queued), we decided that the design of the queue would not allow transactions to be queued if, in the worst case, they wouldn't be able to claim a fee. So we prevent txs 12 and 13 from queuing because we do not consider them "likely to claim a fee".

Realistically, though, nobody should be paying nearly that much in fees, and since an account is only allows to queued up 10 txs as once, they're not likely to get near that total. If they do :man_shrugging:

ximinez: Oh, it's also worth noting that a transaction paying 9XRP is really unlikely to be queued, so this is a contrived example in that aspect, too.

mDuo13: > You have both "Updated in" and "New in" indicating the same version: 1.2.0.

That's not a mistake, it's just confusing markup. This link gets turned into a badge link. I'm changing the label to say "Updated in" but reusing the "New in" badge definition for the badge color and link URL.

As for the phrasing, I'm not really sure how I can rephrase it. What I was trying to convey was simply, "You need enough XRP left over to meet the reserve requirement" without dipping into the "fees can spend from the reserve" technicality.

content/concepts/consensus-network/transaction-queue.md
```diff
@@ -28,11 +28,15 @@ The transaction queue plays an important role in selecting the transactions that
...
-* If a transaction affects how the sending address authorizes transactions, no other transactions from the same address can be queued behind it. [New in: rippled 0.32.0][]
-* If a transaction includes a `LastLedgerSequence` field, the value of that field must be at least **the current ledger index + 2**.
+- Transactions must be properly-formed and [authorized](transaction-basics.html#authorizing-transactions) with valid signatures.
+- Transactions with an `AccountTxnID` field cannot be queued.
+- A single sending address can have at most 10 transactions queued at the same time.
+- To queue a transaction, the sender must have enough XRP for all of the following: [Updated in: rippled 1.2.0][New in: rippled 1.2.0]
+    - Destroying the XRP [transaction cost](transaction-cost.html) as specified in the `Fee` fields of all the sender's queued transactions. This amount cannot be more than the base account reserve (currently 20 XRP). (Transactions paying significantly more than the minimum transaction cost of 0.00001 XRP typically skip the queue and go straight into the open ledger.)
```
ximinez: Nit for clarification: "This *total* amount cannot be more than..." or "These amounts cannot total more than..."

ximinez (approved):
Yeah, this seems a lot clearer. I left a small suggestion, but I'm signing off whether you take it or not. :+1:

PR #506, opened 2019-02-01
"Set Up Secure Signing" Tutorial

mDuo13:
Includes diagrams and steps for getting yourself an environment where you can sign transactions securely. I figure this will be a key pre-requisite for a bunch of other tutorials.


content/tutorials/get-started/set-up-secure-signing.md
```diff
@@ -0,0 +1,134 @@
...
+
+3. Configure the `rippled` machine to accept connections from your local private network.
+
+    In the `rippled`'s config file, modify the `[port_rpc_admin_local]` and `[port_ws_admin_local]` stanzas to use your server's private-LAN address:
+
+        [port_rpc_admin_local]
+      	port = 5005
+      	ip = 10.1.0.0 # Changed from 127.0.0.1
+      	admin = 10.1.1.2 # Only the client with this IP is treated as admin
+      	protocol = http
```
mellery451: shouldn't we still recommend `https/wss` for this LAN config?

mDuo13: Good question. I copied the stanza from the example config and wrote text with the assumption that the communications would be unencrypted, but if https/wss is an option, that would be an improvement.

The question, of course, is the certificates. I don't think having `rippled` use self-signed certs here is a good idea. But setting up rippled to use a proper certificate (assuming that's even possible) from either a well-known CA or an internal CA is probably complicated enough to warrant a separate tutorial.

content/tutorials/get-started/set-up-secure-signing.md
```diff
@@ -0,0 +1,134 @@
...
+
+[![Diagram of using a client library with local signing](img/secure-signing-client-library.png)](img/secure-signing-client-library.png)
+
+This configuration uses a client library in the programming language you are using to sign transactions locally. You need a client library for the programming language you use. Ripple publishes the following client libraries that can sign transactions for the XRP Ledger locally:
+
+- **RippleAPI (ripple-lib) for JavaScript**
+    - [Setup](get-started-with-rippleapi-for-javascript.html)
+    - [API Reference](rippleapi-reference.html)
+- **ripple-libpp for C++**
+    - [Source and Documentation](https://github.com/ripple/ripple-libpp/)
```
mellery451: I don't think we should recommend ripple-libpp, instead:  https://github.com/ripple/rippled/tree/develop/Builds/linux#signing-library

mellery451: for that matter, I don't know why we don't just make one or two simple command line utils that sign transactions using this lib and include them with rippled...but we don't for now.

content/tutorials/get-started/set-up-secure-signing.md
```diff
@@ -0,0 +1,134 @@
...
+- [Use a client library](#use-a-client-library-with-local-signing) that can do local signing.
+- [Use a secure VPN to connect to a remote `rippled` machine](#use-a-secure-vpn-with-a-remote-rippled-server) you trust.
+
+<!-- Source for all diagrams in this article: https://docs.google.com/presentation/d/1BfGyWgC0njoPiKUZz3gXHMVSUINE3Q-_lHqY_D0TGwg/ -->
+
+## Insecure Configurations
+
+[![Diagram of insecure configurations](img/insecure-signing-options.png)](img/insecure-signing-options.png)
+
+Any configuration in which outside sources may gain access to your secret key is dangerous, and is likely to result in a malicious user stealing all your XRP. Examples of such configurations include ones where you use the [sign method][] of someone else's `rippled` server over the internet, or you transmit your secret key in plain text over the internet to your own server.
```
nbougalis: A malicious user could do more than than steal all your XRP. If you had IOUs they may steal those, etc. Consider rewording to suggest that the malicious user would gain access to the account and be able to authorize the execution of transactions.

content/tutorials/get-started/set-up-secure-signing.md
nbougalis: We should make such tools, period. Let's do so for **1.3**.

content/tutorials/get-started/set-up-secure-signing.md
```diff
@@ -0,0 +1,134 @@
...
+If you use a client library not published by Ripple, make sure it uses proper, secure implementations of the signing algorithm(s) it implements. (For example, if it uses the default ECDSA algorithm, it should also use deterministic nonces as described in [RFC6979](https://tools.ietf.org/html/rfc6979).) All of Ripple's published libraries listed above follow industry best practices.
+
+For best security, be sure to keep your client library updated to the latest stable version.
+
+### Example Local Signing with RippleAPI
+
+The following code sample shows how to sign transaction instructions locally with RippleAPI for JavaScript:
+
+```js
+{% include '_code-samples/secure-signing/js/signPayment.js' %}
```
nbougalis, line 82: We need to highlight that this snippet is _insecure_ and should not be used as is, in that it includes the key directly in the script file.

content/tutorials/get-started/set-up-secure-signing.md
nbougalis: We should avoid describing this setup _without_ referencing WSS/HTTPS. We should recommend that users install real SSL certificates, issued by a CA and link to Let's Encrypt as one such authority.

nbougalis (changes requested):
Should we consider highlighting hardware-based signing solutions, like the Ledger? The Trezor Model T also is (theoretically) compatible, thought no wallet capable of leveraging it exists, at this time.

content/tutorials/get-started/set-up-secure-signing.md
mellery451: ssl certs on a local network seem meh...I mean they are fine, but sure is a lot of work for not much gain. I see value in enabling ssl, just not convinced that certs on the 10. network buy you all that much unless you think something on your LAN can hijack the 10. . That said, I'm not an expert and will defer to Nik et. al.

content/tutorials/get-started/set-up-secure-signing.md
```diff
@@ -0,0 +1,142 @@
...
+
+1. [Install `rippled`](install-rippled.html).
+
+    Be sure that your local machine meets the minimum [system requirements for `rippled`](system-requirements.html).
+
+2. When you need to sign transactions, connect to your server on `localhost` or `127.0.0.1`. Use the [sign method][] (for single signatures) or [sign_for method][] (for multi-signatures).
+
+    The [example config file](https://github.com/ripple/rippled/blob/8429dd67e60ba360da591bfa905b58a35638fda1/cfg/rippled-example.cfg#L1050-L1073) listens for connections on the local loopback network (127.0.0.1), with JSON-RPC (HTTP) on port 5005 and WebSocket (WS) on port 6006, and treats all connected clients as admin.
+
+    **Caution:** Using the [commandline API](request-formatting.html#commandline-format) for signatures is less secure than [using the Websocket or JSON-RPC APIs](get-started-with-the-rippled-api.html) through non-commandline client. When using the commandline syntax, your secret key may be visible to other users in the system's process listing, and your shell history may save the key in plaintext.
```
jhaaaa, line 41: Great point - thank you for this. 

Does this caution also apply to the "Run rippled on the same LAN" option?

content/tutorials/get-started/set-up-secure-signing.md
```diff
@@ -0,0 +1,142 @@
...
+
+As always, follow industry-standard practices for securing your machines, such as using a firewall, anti-virus, appropriate user permissions, and so on.
+
+**Warning:** This configuration comes with the additional downside that anyone on the LAN can sniff traffic between your machines, potentially gaining access to your secret keys. Do not use this configuration on a network that may have strangers on it. For example, on the LAN at a colocation facility or cloud hosting, other customers may be able to get access to the traffic between your machines. If you employ several developers sending test transactions, you could run one `rippled` machine for your whole office, while the developers use cheaper hardware, but any user on your office network could potentially use a packet sniffer to get access to developers' secret keys.
```
jhaaaa: possible nit edit? change

>cloud hosting

to

>cloud host

content/tutorials/get-started/set-up-secure-signing.md
mDuo13: Let's try to figure out a plausible way of demonstrating securely importing the secret instead

content/tutorials/get-started/set-up-secure-signing.md
mDuo13: No. (Because if you're connecting on LAN, you're almost certainly not using the commandline interface.)

content/tutorials/get-started/set-up-secure-signing.md
mDuo13: Would be nice if they worked in a way that doesn't expose the secret to the process listing and shell history, btw.

content/tutorials/get-started/set-up-secure-signing.md
jhaaaa: Ah, duh. Got it. Thank you!

mDuo13:
- Moved the LAN config tutorial to its own (draft) file to be finished later
- Mentioned dedicated signing devices such as the Ledger Nano S
- Improved the JS example's handling of secrets

content/tutorials/get-started/set-up-secure-signing.md
mDuo13: I'm kicking this can down the road to a future task. I simplified the instructions for now, and left a draft page to be expanded later.

mDuo13:
(Rebased onto the latest master branch.)

mDuo13:
OK, gonna publish the "Set up Secure Signing" doc as-is and follow up with instructions for setting the LAN config w/ actual certificates in a later task.

PR #507, opened 2019-02-05
Add xrp-ledger.toml spec

mDuo13:
This proposed specification provides a way for people and businesses to provide information about themselves in a machine-readable way that client applications could potentially use. This specification is intended as a replacement for the dated and poorly specified "ripple.txt" file.

At the time of this writing, there are no known parsers for this data, but I hope to demonstrate at least a sample script. Similarly, I hope for Ripple to provide an example of such a file on the company website shortly after (or maybe even before?) the publication of this script to the dev portal.

I welcome feedback from the XRP Ledger community on this and future versions of the spec. For the time being, I prefer to keep the scope of this specification small so that we can publish it sooner and extend it later, rather than having a specification whose basics are in flux for a long time.

Here are some meta-notes on the choices that went into the making of this specification:

- I chose the TOML language with the support of the `rippled` engineering team at Ripple, for several reasons, including its similarity to the old `ripple.txt` file format, its support for comments (a requirement which ruled out JSON), and the fact that Stellar already uses TOML for effectively the same purpose.
- I have tried to make the `xrp-ledger.toml` file format similar to the [`stellar.toml` file format](https://www.stellar.org/developers/guides/walkthroughs/how-to-complete-stellar-toml.html) where it makes sense to do so. It just makes sense for the better of the Internet of Value not to reinvent the wheel nor to make things different for the sake of being different. There are some differences even in the features they have in common: for example, this file does not require listed principals to have Twitter accounts, because I do not think using Twitter should be a prerequisite for being a point of contact.
- I have tried to make the specification as simple as possible. For example, the rule that **all fields are optional** has no exceptions, even though obviously there are some cases where providing one field without another would be strange or not very useful. I have intentionally omitted many proposed features from this specification to keep the scope of the first version small. I am open to including more fields and features if support in the community is widespread, there is a clear need or purpose, and the field's format can be specified precisely. I expect more fields to become included in later versions of this spec as people writing client applications find use cases and needs. Some ideas intentionally omitted from this version of the spec:
    - A warrant canary. This was a little too difficult to define and there are other equally good places for an organization to provide a warrant canary if they desire.
    - A status URL for validators. We may provide this in a later version of the spec, but as it stands there isn't enough of a standard in terms of what should be provided, and whether the URL in question should be read by humans or machines.
    - A "domain" field for the document. This is not useful because the document is only useful in the context of having been served by a domain whose identity is verified using a mechanism such as TLS with well-known certificate authorities. Despite the known flaws in the certificate-authority system, there isn't another system as widely supported that is nearly as effective.
    - Specific fields for marking an address as a hot wallet, cold wallet, etc. If you issue currencies in the XRP ledger, you can use a `[[CURRENCIES]]` entry to indicate your issuing address. For other hot/cold accounts, you can mention their purpose in the `desc` field of the `[[ACCOUNTS]]` entry.
    - A field for validators to provide their "manifest" (that is, latest ephemeral public key etc.). Manifests are best provided by `rippled` servers in the XRP Ledger's peer-to-peer communications. We may revisit this point in the future.

### DIFF
--- a/content/references/xrp-ledger-toml.md
+++ b/content/references/xrp-ledger-toml.md
@@ -108,6 +108,25 @@ desc = "Ripple-owned address from old ripple.txt file"
 #   "Domain" field of the account in the XRP Ledger matches the
 #   domain this file was served from.
 
+[[SERVERS]]
+json_rpc = "https://s1.ripple.com:51234/"
+ws = "wss://s1.ripple.com/"
+peer = "https://s1.ripple.com:51235/"
+desc = "General purpose server cluster"
+
+[[SERVERS]]
+json_rpc = "https://s2.ripple.com:51234/"
+ws = "wss://s2.ripple.com/"
+peer = "https://s2.ripple.com:51235/"
+desc = "Full-history server cluster"
+
+[[SERVERS]]
+json_rpc = "https://s.altnet.rippletest.net:51234"
+ws = "wss://s.altnet.rippletest.net:51233"
+peer = "https://s.altnet.rippletest.net:51235/"
+network = "altnet"
+desc = "Test Net public server cluster"
+
 [[PRINCIPALS]]
 name = "Rome Reginelli" # Primary spec author
 email = "rome@example.com" # Not my real email address
@@ -175,6 +194,20 @@ The principals list (`[[PRINCIPALS]]`) provides information about the people (or
 | `email` | String | The email address where this principal can be contacted. |
 
 You may provide other contact information as desired. (See [Custom Fields](#custom-fields) for information about custom fields.)
+
+
+### Servers
+
+The servers list (`[[SERVERS]]`) provides information about XRP Ledger servers (`rippled`) you run with public access. Each entry describes a different server or server cluster. For _each_ `[[SERVERS]]` entry, you MAY provide any of the following fields:
+
+| Field   | Type   | Description                                              |
+|:--------|:-------|:---------------------------------------------------------|
+| `json_rpc` | String (URL) | The URL where you serve a public JSON-RPC API. This MUST begin with either `http://` or `https://`. HTTPS is RECOMMENDED for public APIs. |
+| `ws` | String (URL) | The URL where you serve a public WebSocket API. This MUST begin with either `ws://` or `wss://`. WSS is RECOMMENDED for public APIs. |
+| `peer` | String (URL) | The URL where your server is listening for the XRP Ledger Peer Protocol. Other XRP Ledger servers can connect at this URL. If your server provides a Peer Crawler response, it is served from this URL with `crawl` appended. |
+| `network`  | String | Which network chain this server follows. If omitted, clients SHOULD assume that the server follows the production XRP Ledger. Use `main` to explicitly specify the production XRP Ledger. Use `altnet` for Ripple's XRP Ledger Test Net. You MAY provide other values to describe other test nets or non-standard network chains. |
+
+For all URLs in this section, the trailing slash is RECOMMENDED. If omitted, client applications SHOULD assume that there is a trailing slash implied.
 
 
 ### Currencies
--- a/content/references/xrp-ledger-toml.md
+++ b/content/references/xrp-ledger-toml.md
@@ -1,0 +1,275 @@
+# xrp-ledger.toml File
+
+If you run an XRP Ledger validator or use the XRP Ledger for your business, you can provide information about your usage of the XRP Ledger to the world in a machine-readable **`xrp-ledger.toml`** file. Scripts and applications can use the information contained in your `xrp-ledger.toml` file to better understand and represent you in the XRP Ledger. In some cases, humans may also find it useful to read the same file.
+
+One of the primary use cases for the `xrp-ledger.toml` file is [domain verification](#domain-verification).
+
+#### Notational Conventions
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Serving the File
+
+The `xrp-ledger.toml` file is meant to be served by a web server. The file should be available at the following URL:
+
+```
+https://{DOMAIN}/.well-known/xrp-ledger.toml
+```
+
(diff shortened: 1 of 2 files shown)


content/references/xrp-ledger-toml.md
```diff
@@ -0,0 +1,275 @@
...
+The contents of the `xrp-ledger.toml` file MUST be formatted in [TOML](https://github.com/toml-lang/toml). **All contents are optional.** Comments are optional, but encouraged for readability.
+
+Example contents:
+
+```
+# Example xrp-ledger.toml file. These contents should not be considered
+# authoritative for any real entity or business.
+# Note: all fields and all sections are optional.
+
+[METADATA]
```
nbougalis, line 86: My _preference_ would be to avoid CAPITALIZING things; I have no rational argument for this, it's something purely stylistic.

content/references/xrp-ledger-toml.md
```diff
@@ -0,0 +1,275 @@
...
+modified = 2019-01-22T00:00:00.000Z
+expires = 2019-03-01T00:00:00.000Z
+
+[[VALIDATORS]]
```
nbougalis, line 90: I'd prefer using `[[validator]]` rather than the plural.

content/references/xrp-ledger-toml.md
```diff
@@ -0,0 +1,275 @@
...
+
+Tools that parse the `xrp-ledger.toml` file MUST accept documents that contain any other fields that the application is not familiar with. Those tools MAY make those additional fields available to higher-level applications that call them, or MAY discard those fields. To maintain forward-compatibility with future versions of this specification, tools MAY also discard fields specified in this standard. Tools MUST NOT return an error if an `xrp-ledger.toml` file contains an unrecognized field. To detect typos, tools MAY provide a warning on unrecognized fields, especially if those field names are similar to the names of standard fields.
+
+Tools MAY return an error if a field they recognize is not formatted as expected, even if that field is not defined in this specification.
+
+When creating custom fields, be mindful of the field name you choose. If you use a very generic field name, other users may use the same name to mean something different, or formatted in a conflicting way. If you use a custom field that you think others will find useful, please contribute a specification for your field to the maintainers of this document.
```
MarkusTeufelberger, line 234: Maybe add `via the "Edit on Github" button on top of this page` to `please contribute a specification for your field to the maintainers of this document.`?

nbougalis: We should define a `manifest` entry, likely as a field under a particular validator. The problem is that a manifest can be very long, which is always awkward.

content/references/xrp-ledger-toml.md
MarkusTeufelberger: I hope that eventually https://github.com/ripple/rippled/issues/2601 is implemented, then this section doesn't need to be renamed.

I would also expect people/organizations that run a validator to also run one on testnet at least.

MarkusTeufelberger:
Possible additional fields that could already be specified now:

* A `[[SERVERS]]` or `[[API_ENDPOINTS]]` section that lists RPC and Websocket ports as well as /crawl endpoints of your servers (not validators, these shouldn't be on the public internet anyways).

* A `[[VALIDATOR_LISTS]]` (naming?) section that links to lists in the format of vl.ripple.com and a git repository (including the current commit hash of master) like https://github.com/ripple/vl or something similar that can be used to tell others that you're publishing a recommended UNL and where it is being publicly(?) archived.

content/references/xrp-ledger-toml.md
mDuo13: For some fields, I went with the uppercase to match how the stellar.toml file is formatted. Then, for others, I went with uppercase to match those.

I think maintaining compatibility / consistency with a similar file format outranks personal preference here.

content/references/xrp-ledger-toml.md
mDuo13: There's a good reason to use a plural: when you parse the file, each `[[VALIDATORS]]` entry becomes an object in a list named `VALIDATORS`.

I think we can agree that a list of validators should not be named "validator".

mDuo13:
@nbougalis Is there a definition for a manifest format somewhere public? If not, let's wait until we have one of those and then add them at the same time.

@MarkusTeufelberger I like the idea of a `[[SERVERS]]` list. I'll see if I can spec one up. 

The `[[VALIDATOR_LISTS]]` idea is interesting. The idea would be to advertise if you publish a validator list as Ripple and Coil currently do, right? My only concern would be people shooting themselves in the foot by using a little-known validator list and then diverging from the network because it doesn't overlap enough with the generally-accepted lists. Who would consume this data, though? Web crawlers? I'd like to revisit this in the next version of this spec rather than do it now.

MarkusTeufelberger:
> The idea would be to advertise if you publish a validator list as Ripple and Coil currently do, right?

Exactly. I for example didn't even know that Coil publishes one and I wouldn't know where to look. I'd expect to look in this file in the future. People can shoot themselves into their feet already by not using a list with enough overlap, that's the cost of decentralization.

content/references/xrp-ledger-toml.md
```diff
@@ -0,0 +1,308 @@
...
+The validators list (`[[VALIDATORS]]`) provides information about validating servers you operate. Each entry describes a separate validating server.
+
+The _first_ `[[VALIDATORS]]` entry in the file is treated as your primary validator. If you operate one or more validators for the production XRP Ledger, you should put the one you want others to trust first.
+
+For _each_ `[[VALIDATORS]]` entry, you MAY provide any of the following fields:
+
+| Field        | Type   | Description                                          |
+|:-------------|:-------|:-----------------------------------------------------|
+| `public_key` | String | The master public key of your primary validator, encoded in the XRP Ledger's base58 format (typically, this starts with `n`). |
+| `network`  | String | Which network chain this validator follows. If omitted, clients SHOULD assume that the validator follows the production XRP Ledger. Use `main` to explicitly specify the production XRP Ledger. Use `testnet` for Ripple's XRP Ledger Test Net. You MAY provide other values to describe other test nets or non-standard network chains. |
```
mDuo13, line 168: I expect the Data API to change from using `altnet` to `testnet` in API requests/responses in a forthcoming version, for consistency.

mDuo13:
Sounds like people have nothing but good ideas to add to this, so I'm excited to move forward and declare the first version done so we can start revising it!

content/references/xrp-ledger-toml.md
MarkusTeufelberger: Fyi, there are other mentions of "altnet" in this repo, might be worth a grep.

PR #512, opened 2019-02-07
Provide an Escrow-as-Smart Contract use case

jhaaaa:
- Provides a use case that describes how to use a conditionally-held escrow as a smart contract


jhaaaa:
@mDuo13 , @warpaul , & @garay-daniel - Sorry for the delay in getting this use case out to review with you. When you have a moment, please take a look and feel free to comment. As I mention in a TODO comment in the markdown, I'm using a far-out scenario right now. (What, you don't know party planners who want to generate condition and fulfillment values? 😂) But I needed to use a dead simple scenario to help me focus on and work through the logistics. I'm open to changing up the scenario! Thank you.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+***TODO: Content in this use case breaks the Use Case template. Let's figure out a way to adjust either the template or content to make this work.***
+
+***TODO: Interested in how we talk about escrow vs smart contract here. Is the escrow something that can provide, be used as, or serve as a smart contract? I'm open to the language we use to describe this relationship.***
+
+A smart contract is a blockchain-based program that encodes the conditions and fulfillment of an agreement between two or more parties and automatically fulfills the terms of the agreement once conditions are met.
+
+A smart contract can help you exchange anything of value in a transparent, traceable, tamper-resistant, and irreversible way.
+
+The benefit of encoding a smart contract into a blockchain is that it enables the contract to be securely performed without traditional third-parties, like financial or legal institutions, which can result in lower costs. Instead, the contract is supervised by the distributed, decentralized network of computers that run the blockchain.
```
mDuo13: I think "which can result in lower costs" is a misplaced modifier here. Also, <sup>[citation needed]</sup>, so maybe just leave it off.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
+# Use an Escrow as a Smart Contract
+
+***TODO: Content in this use case breaks the Use Case template. Let's figure out a way to adjust either the template or content to make this work.***
+
+***TODO: Interested in how we talk about escrow vs smart contract here. Is the escrow something that can provide, be used as, or serve as a smart contract? I'm open to the language we use to describe this relationship.***
+
+A smart contract is a blockchain-based program that encodes the conditions and fulfillment of an agreement between two or more parties and automatically fulfills the terms of the agreement once conditions are met.
+
+A smart contract can help you exchange anything of value in a transparent, traceable, tamper-resistant, and irreversible way.
```
mDuo13: Maybe combine with the previous paragraph?

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+Let's use this scenario to help illustrate this use case: A party planner uses smart contracts to manage payments from party hosts to party vendors. Specifically, the party planner wants to use a smart contract to have the party host pay the party band 2000 XRP once they are done with their set. ***TODO: this is a far-out example -- but I had to use something really tangible to be able to make myself understand the logistics of how this might work. I'm open to any other scenario that might be more appropriate.***
+
+In this use case, the party host is the sender of the escrow, the party band is the receiver of the escrow, and the party planner is playing the role of an _oracle_. In the context of smart contracts, an oracle is a neutral third-party agent that can verify real-world events to either fulfill or invalidate a smart contract. This use case uses a human oracle for illustrative purposes, but in real-life, a software application would more likely play the role of the oracle to automate the process.
```
mDuo13: I wish we had an example that used an actual software oracle. I'll see if I can think of a scenario that's a little more "true to life". For now, this is OK.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+Using an XRP Ledger escrow to provide this smart contract is a great arrangement because the party planner, as the third-party oracle, never "holds" the funds as one might in a traditional escrow arrangement, and can't possibly take the funds for themselves.
```
mDuo13, line 13: :+1:

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+{% set n = cycler(* range(1,99)) %}
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## Meet the prerequisites
+
+The party host (sender) must have:
+
+- An XRP Ledger [account](accounts.html#creating-accounts) that holds enough XRP to pay for escrow and any fees incurred.
+
+- Access to a [`rippled` server](install-rippled.html) that they can use to create the escrow.
```
mDuo13: Technically they only need a [secure signing environment](https://github.com/ripple/ripple-dev-portal/pull/506), I think. Which should include having a network connection to a `rippled` server (any server) they can submit their signed transaction to.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+- An XRP Ledger [account](accounts.html#creating-accounts) that can receive the XRP paid by the escrow.
+
+- The ability to look up the details of an XRP Ledger transaction hash. For example, they can use the [RPC Tool](https://developers.ripple.com/xrp-ledger-rpc-tool.html) or access a [`rippled` server](install-rippled.html) that they can use to look up the transaction hash.
+
+The party planner (oracle) must have:
+
+- The ability to generate a condition and a fulfillment.
+
+- Access to a [`rippled` server](install-rippled.html) that they can use to submit the fulfillment value and finish the escrow.
```
mDuo13: Having a rippled server and the ability to finish the escrow should be under the band's requirements. (You need an XRP Ledger account to send any kind of transaction, after all.) The payee (the band) can be the one to actually construct and submit the transaction to get paid, after the oracle tells them what fulfillment to use. 

The party planner (the oracle) needs:

- To be able to keep a secret (the fulfillment) until the time is right.
- A way to communicate the fulfillment publicly or at least to the payee (the band) when the time is right.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## Define the terms of the smart contract
+
+To create the escrow as a smart contract, the participants must first define the terms of the contract. In this scenario, the participants need to agree on the following details.
+
+- Should the escrow be fulfilled after a specific time has passed or after a cryptographic condition has been fulfilled?
```
mDuo13: Feels a little odd to ask this here after this whole use case has been building up using the crypto-condition.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+    - *[Conditionally-held escrow](send-a-conditionally-held-escrow.html)*: In this case, the participants agree to use a conditionally-held escrow. This is a good choice because the smart contract can be fulfilled only after the party planner (oracle) verifies that the party band completed their set and submits the cryptographic condition and fulfillment to finish the escrow and release the XRP.
```
mDuo13: As mentioned above, the oracle doesn't submit the transaction to finish the escrow. The oracle merely provides the information that makes it possible for the payee to finish the escrow.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+    To handle a no-show scenario, the participants configure the escrow to be eligible for cancellation after a specific time. Typically, this cancellation time should be after the time at which the party planner is expected to finish the escrow. The the cancellation time in this way helps prevent the party host from maliciously cancelling the escrow before the party planner can finish the escrow.
```
mDuo13: This isn't quite right.

An XRP Ledger Escrow cannot be "maliciously" canceled. The escrow exists in three states (I just made up the names of these states for explanation. They're not used elsewhere):

- "Held." No one can finish or cancel it.
- "Ripe." Anyone can finish it. No one can cancel it.
- "Expired." Anyone can cancel it. No one can finish it.

A time-based escrow is "ripe" after the `FinishAfter` time has passed.

A conditional escrow is "ripe" after the oracle has published the fulfillment.

A time-and-conditional escrow is "ripe" after both have happened.

An escrow that passes its `CancelAfter` time is expired, regardless of what happened with the fulfillment. An escrow that doesn't have a `CancelAfter` time never expires. If you're using an oracle, you _definitely_ want to use `CancelAfter` so that you can get your money back if the oracle does not publish the fulfillment.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+    The conditionally-held escrow provides the functionality required to handle the terms of this smart contract by putting the ability to finish the escrow solely in the hands of the neutral party planner (oracle).
+
+    - *[Time-held escrow](send-a-time-held-escrow.html)*: A time-held escrow is not the best choice for this scenario because the party band could be a no-show. If the party band is scheduled to finish their set at 11 p.m., the participants could agree to allow the escrow to finish and release the XRP at 11 p.m. At 11 p.m., any participant, including the party band, can finish the escrow and release the XRP, even if the band didn't show up. This does not meet the terms of the smart contract.
+
+    The participants could agree to allow the cancellation of the escrow before 11 p.m. to accommodate the no-show scenario. However, consider that the party band could show up and play their set, but the party host cancels the escrow before a participant can finish the escrow at 11 pm.
```
mDuo13: You can't use the cancellation on XRP Ledger escrows this way.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+- Should the escrow disallow fulfillment until a specific time?
+
+  The participants agree to disallow the escrow from being fulfilled until 11 p.m. the night of the party, which is when the party band is scheduled to complete their set. ***TODO: not necessary, but just an extra precaution? Or just unnecessary?***
```
mDuo13: Unnecessary. I don't think it provides any additional security, since whether the escrow can be finished still depends entirely on whether the oracle publishes the fulfillment before the expiration.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+- Should the participants be able to cancel the escrow after a specific time period if the condition is not fulfilled?
```
mDuo13: > Should the participants be able to cancel the escrow

Suggest changing this to "Should the escow expire" and the answer is "absolutely yes" because otherwise the only outcomes are:

- The party planner releases the condition: the band gets paid
- The party planner doesn't release the condition: the money stays locked in escrow forever

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+  The participants agree that the escrow can be cancelled after 12 noon the day after the party. This gives the party planner enough time to finish the escrow, if the party band fulfills their end of the contract. After cancellation, the locked XRP returns to the party host's account.
+
+  This timing also prevents the party host from maliciously cancelling the escrow before the party planner can finish the fulfilled contract.
```
mDuo13: As mentioned, remove this line; that's not how cancellation works.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+    The participants could agree to allow the cancellation of the escrow before 11 p.m. to accommodate the no-show scenario. However, consider that the party band could show up and play their set, but the party host cancels the escrow before a participant can finish the escrow at 11 pm.
+
+- Should the escrow disallow fulfillment until a specific time?
+
+  The participants agree to disallow the escrow from being fulfilled until 11 p.m. the night of the party, which is when the party band is scheduled to complete their set. ***TODO: not necessary, but just an extra precaution? Or just unnecessary?***
+
+- Should the participants be able to cancel the escrow after a specific time period if the condition is not fulfilled?
+
+  The participants agree that the escrow can be cancelled after 12 noon the day after the party. This gives the party planner enough time to finish the escrow, if the party band fulfills their end of the contract. After cancellation, the locked XRP returns to the party host's account.
```
mDuo13: Suggest changing to:

> This gives the band enough time to finish the escrow, after the party planner verifies that they fulfilled their end of the contract and publishes the cryptographic fulfillment.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+  This timing also prevents the party host from maliciously cancelling the escrow before the party planner can finish the fulfilled contract.
+
+- How much XRP should the escrow to lock up and potentially pay?
+
+  The participants agree that the escrow should lock up and potentially pay 2000 XRP, which is the party band's fee.
+
+- Which XRP Ledger account should the escrow lock up and potentially pay XRP out of?
```
mDuo13: This wording implies that the escrow "locks up" an account. It only locks up the specified amount of XRP, not the whole account.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+  The participants agree that the escrow should lock up and potentially pay XRP out of the party host's XRP Ledger account.
+
+- Which XRP Ledger account should the escrow lock up and potentially pay XRP into?
```
mDuo13: I think this sentence reads more clearly if you remove "lock up and" from it:

> Which XRP Ledger account should the escrow potentially pay XRP to?

(Also, I like "pay XRP to" rather than "pay XRP into" here, but that's not that important.)

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Generate a condition and a fulfillment](send-a-conditionally-held-escrow.html#1-generate-condition-and-fulfillment) (oracle)
+
+Because participants want to create a conditionally-held escrow to provide the smart contract, they need a condition value and a fulfillment value. In this scenario, the participant that creates these values is the neutral party planner (oracle).
+
+The party planner generates the condition and fulfillment values and provides the condition value to the party host, who creates the escrow.
+
+The party planner must keep the fulfillment value a secret. They will use the condition and fulfillment values to finish the escrow.
```
mDuo13: > They will use the condition and fulfillment values to finish the escrow.

Anyone can use these, but most likely the band will since they're the ones who're motivated to get paid.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Calculate time values needed for the escrow](send-a-conditionally-held-escrow.html#2-calculate-release-or-cancel-time) (sender)
+
+Because the participants want the escrow to be eligible for cancellation after 12 noon the day after the party, the party host (sender) must calculate a `CancelAfter` value to include in the escrow definition.
+
+Because the participants want to disallow the escrow from being finished until after 11:00 pm the night of the party (the time when the party band is scheduled to finish its set), the party host must calculate a `FinishAfter` value to include in the escrow definition.
```
mDuo13: Since I don't think specifying a `FinishAfter` alongside a condition is useful, scratch this.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+The party host (sender) creates the escrow that provides the smart contract. The party host must create the escrow because they are the only participant that can authorize the lock up and potential payout of XRP from their XRP Ledger account.
+
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Wait for validation](send-a-conditionally-held-escrow.html#4-wait-for-validation) and [confirm escrow creation](send-a-conditionally-held-escrow.html#5-confirm-that-the-escrow-was-created) (sender and receiver)
+
+The party host (sender) waits for validation of the ledger that contains the escrow creation transaction and then confirms that the escrow was created.
+
+The party host then provides the escrow transaction's `hash` value to the party band (receiver). The party band can use the `hash` value to look up the escrow transaction on the XRP Ledger to ensure that it was created according to the smart contract terms they agreed to.
```
mDuo13: Right. As part of this step, the band should probably confirm that the condition matches the one the oracle provided. If the condition's wrong, the fulfillment the oracle provides won't let the band get paid.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+  The participants agree that the escrow should lock up and potentially pay XRP to the party band's XRP Ledger account.
+
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Generate a condition and a fulfillment](send-a-conditionally-held-escrow.html#1-generate-condition-and-fulfillment) (oracle)
+
+Because participants want to create a conditionally-held escrow to provide the smart contract, they need a condition value and a fulfillment value. In this scenario, the participant that creates these values is the neutral party planner (oracle).
+
+The party planner generates the condition and fulfillment values and provides the condition value to the party host, who creates the escrow.
```
mDuo13: The party planner needs to tell the condition to the band also so that they know that this is the right condition.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+The party host then provides the escrow transaction's `hash` value to the party band (receiver). The party band can use the `hash` value to look up the escrow transaction on the XRP Ledger to ensure that it was created according to the smart contract terms they agreed to.
+
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Finish the escrow](send-a-conditionally-held-escrow.html#6-submit-escrowfinish-transaction) (oracle)
+
+The party band (receiver) shows up and plays their set.
+
+The party planner (oracle) finishes the escrow using the fulfillment value they generated and kept secret.
```
mDuo13: Change this to reflect the band finishing the escrow using the fulfillment the oracle publishes.

mDuo13: Also note that this has to happen before 12 noon or the escrow expires and the band doesn't get paid.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+Alternatively, if the party band is a no-show, the party planner does not finish the escrow. In this case, after 12 noon the next day, any participant can [cancel the escrow](cancel-an-expired-escrow.html) to return the held XRP to the party host's (sender's) account.
```
mDuo13: > the party planner does not finish the escrow

Change to:

> the party planner does not publish the fulfillment.

Also change "any participant can cancel..." to "anyone can cancel" because literally anyone can cancel an expired escrow.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Wait for validation](send-a-conditionally-held-escrow.html#7-wait-for-validation) and [confirm escrow finish](send-a-conditionally-held-escrow.html#8-confirm-final-result) (oracle, sender, and receiver)
+
+The party planner (oracle) waits for validation of the ledger that contains the escrow finish transaction and then confirms that the escrow was finished.
```
mDuo13: The band does this, but otherwise that's correct.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,149 @@
...
+
+At this time, the party planner provides the transaction's `hash` value to the party host (sender) and party band (receiver). The party host and party band can use the `hash` value to look up the escrow transaction on the XRP Ledger to ensure that it is been finished correctly. The party host and party band can also just check their XRP Ledger account balances to ensure that their balances have decreased and increased by 2000 XRP respectively.
```
mDuo13: The host's balance won't change at this step (unless the escrow was canceled) because the escrow creation already debited the locked-up XRP from their account.

mDuo13 (changes requested):
This use case isn't quite ready for primetime. There are some misconceptions about how fulfillments and expirations work, which I've tried to clarify. 

I also would really like to come up with a more plausible scenario than the party planning one, but I'm not sure I'll be able to come up with something. Still, I guess it's better to publish with this fanciful scenario than not to publish.

It might also be worth emphasizing that the oracle doesn't need an XRP Ledger account, or any secret communication channels... just a way to publish a condition and later a fulfillment reliably to the payer/payee, and a way to verify that the payee held up their end of the bargain.

content/use-cases/use-an-escrow-as-a-smart-contract.md
jhaaaa: Thank you for this! Sure thing. I included it because it points out that you have these options to pick from when you are creating an escrow -- but you are right, the scenario has already picked the conditionally-held option!

jhaaaa:
@mDuo13 - wow, wow, wow - thank you so much for such an informative and insightful review. I've addressed your feedback. Please let me know if there is anything else you'd like me to take a look at. 

If the content is good for now - I think we still need to address the use case template to be able to publish this. I think we discussed using some dactyl filters to allow the content to break some template rules? 😂

mDuo13:
Yes, I looked into doing a Dactyl filter thing today, but at minimum it needs the bug fix that's in the forthcoming Dactyl version.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,135 @@
...
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## Define the terms of the smart contract
+
+To create the escrow as a smart contract, the participants must first define the terms of the contract. In this scenario, the participants need to agree on the following details.
+
+- Should the escrow disallow fulfillment until a specific time?
+
+  While this is an option, the participants agree that it is unnecessary for their escrow. For conditionally-held escrows, enabling this option doesn't provide any additional security, since whether the escrow can be finished still depends entirely on whether the party planner (oracle) publishes the fulfillment before the expiration.
```
mDuo13: On this and the other "answers" in this Q&A-style section of this step, you need to indent the answers by 4 spaces instead of 2 so the answers are properly nested in the list items.

Also, you might consider bolding the "Questions" to make it a little easier to read.

jhaaaa:
👋 @warpaul and @mDuo13 - Thanks to Rome, we now have a Use Case template that can handle all kinds of formatting - so this escrow use case now looks presentable! 💅 

Any thoughts on our ability to publish this use case? I believe I've addressed all content feedback. However, I know that we agree that the scenario is a bit fanciful -- but do we think the content is accurate and valuable enough to publish? 

For the future, we can aim to come up with a more realistic use case that perhaps integrates with an actual software oracle?

content/use-cases/use-an-escrow-as-a-smart-contract.md
mDuo13: Suggestion:
```
< a software application would more likely play the role of the oracle to automate the process.
---
> a software application would more likely play the role of the oracle.

File: content/use-cases/use-an-escrow-as-a-smart-contract.md [hunk 26]
```diff
@@ -0,0 +1,137 @@
...
+
+- Access to a [`rippled` server](install-rippled.html) that they can use to look up the details of an XRP Ledger transaction hash and submit the fulfillment value to finish the escrow.
+
+The party planner (oracle) must have:
+
+- The ability to generate a condition and a fulfillment.
+
+- To be able to keep a secret (the fulfillment) until the time is right.
+
+- A way to communicate the fulfillment publicly or at least to the party band when the time is right.
```
mDuo13, line 42: One last thing the oracle must have:

- The ability to recognize whether the receiver (the party band) has fulfilled their end of the contract (played at the party).

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,137 @@
...
+
+- **From which XRP Ledger account should the escrow lock up XRP for potential payment to the party band?**
+
+      The participants agree that the escrow should lock up and potentially pay XRP out of the party host's XRP Ledger account.
+
+- **Which XRP Ledger account should the escrow potentially pay XRP to?**
+
+      The participants agree that the escrow should potentially pay XRP to the party band's XRP Ledger account.
+
+
```
mDuo13, line 75: Another (arguably the most important) part of the puzzle: how shall the oracle know that the receiver did in fact play at the party?

If the party planner trusts the band and the party host to tell the truth, the oracle could hear conflicting reports about whether the band even played. Part of the job of an oracle is to know the truth (that's where the name comes from), so the party planner should be at the party to see for themselves.

You could picture a case where the party planner has some trusted partygoer(s) who report whether the band played, but in that case the partygoers and the party planner are jointly acting the role of oracle.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,137 @@
...
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Generate a condition and a fulfillment](send-a-conditionally-held-escrow.html#1-generate-condition-and-fulfillment) (oracle)
```
mDuo13: Similar to the Payment Channels case, I like to structure these sorts of headings as "Role: task" rather than "task (role)" because I prefer to set the context first.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,137 @@
...
+
+Because participants want to create a conditionally-held escrow to provide the smart contract, they need a condition value and a fulfillment value. In this scenario, the participant that creates these values is the neutral party planner (oracle).
+
+The party planner generates the condition and fulfillment values. The party planner provides the condition value to the party host, who creates the escrow. The part planner also provides the condition to the party band so that they know that this is the right condition.
+
+The party planner must keep the fulfillment value a secret. Anyone can use the condition and fulfillment values to finish the escrow, but in this scenario, it will most likely be the party band who will finish the escrow because they're the ones who are motivated to get paid.
```
mDuo13: > Anyone can use the condition and fulfillment values to finish the escrow, but in this scenario, it will most likely be the party band who will finish the escrow because they're the ones who are motivated to get paid.

I suggest breaking this up into two sentences and rewording to avoid future tense:

> Anyone can use the condition and fulfillment values to finish the escrow. Most often, the receiver finishes the escrow because they're the ones who are motivated to get paid.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,137 @@
...
+
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Wait for validation](send-a-conditionally-held-escrow.html#4-wait-for-validation) and [confirm escrow creation](send-a-conditionally-held-escrow.html#5-confirm-that-the-escrow-was-created) (sender and receiver)
+
+The party host (sender) waits for validation of the ledger that contains the escrow creation transaction and then confirms that the escrow was created.
+
+The party host then provides the escrow transaction's `hash` value to the party band (receiver). The party band can use the `hash` value to look up the escrow transaction on the XRP Ledger to ensure that it was created according to the smart contract terms they agreed to. As part of this step, the party band should also confirm that the condition matches the one the party planner (oracle) provided. If the condition's wrong, the fulfillment the party planner provides won't let the party band finish the escrow and get paid.
```
mDuo13: > If the condition's wrong,

I suggest avoiding the contraction in this case and spelling out "condition is" because it's not a common contraction and it could be confused for a possessive by non-native speakers.

mDuo13: > As part of this step, the party band should also confirm that the condition matches

I suggest removing "also" from this phrase.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,137 @@
...
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Finish the escrow](send-a-conditionally-held-escrow.html#6-submit-escrowfinish-transaction) (receiver)
+
+The party band (receiver) shows up and plays their set.
+
+The party planner (oracle) confirms that the party band has fulfilled their contract and publishes the fulfillment publicly, or at least to the party band.
+
+The party band finishes the escrow using the fulfillment value published by the party planner. The party band must finish the escrow before 12 noon or the escrow expires, the party band doesn't get paid, and the XRP returns to the party host's (sender's) XRP Ledger account.
```
mDuo13: > The party band must finish the escrow before 12 noon or the escrow expires, the party band doesn't get paid, and the XRP returns to the party host's (sender's) XRP Ledger account.

I suggest breaking this up into two sentences so that it doesn't read as a list. Also, the next paragraph explains how the XRP doesn't automatically return, so it's probably less confusing if you just leave that part off.

> The party band must finish the escrow before 12 noon. If they don't, the escrow expires and the party band doesn't get paid.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,137 @@
...
+
+Alternatively, if the party band is a no-show, the party planner does not publish the fulfillment. In this case, after 12 noon the next day, any participant can [cancel the escrow](cancel-an-expired-escrow.html) to return the held XRP to the party host's account.
```
mDuo13: I think you can rephrase this slightly to continue from either the "fulfillment not published" or "fulfillment published, but no one bothered to finish the escrow" cases:

> Alternatively, if the party band is a no-show, the party planner does not publish the fulfillment. In this case, the escrow cannot be finished. After 12 noon the next day, ...

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,137 @@
...
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## [Wait for validation](send-a-conditionally-held-escrow.html#7-wait-for-validation) and [confirm escrow finish](send-a-conditionally-held-escrow.html#8-confirm-final-result) (receiver and sender)
+
+The party band (receiver) waits for validation of the ledger that contains the escrow finish transaction and then confirms that the escrow was finished.
+
+At this time, the party band provides the transaction's `hash` value to the party host (sender). They can use the `hash` value to look up the escrow transaction on the XRP Ledger to ensure that it is been finished correctly.
+
+The party band can also just check their XRP Ledger account balance to ensure that their balance has increased by 2000 XRP. The party host's balance won't change at this step (unless the escrow was cancelled) because the escrow creation already debited the locked-up XRP from their account.
+<!-- USE_CASE_STEPS_START -->
```
mDuo13: This should be `<!-- USE_CASE_STEPS_END -->`

mDuo13 (changes requested):
Looking good. Left a few copy-edits and pointed out a couple small but important changes, but I think this is very close.

content/use-cases/use-an-escrow-as-a-smart-contract.md
jhaaaa: @mDuo13 -Yes, I see! In my mind, the party planner is at the party. Maybe I've seen too many movies with control-freak wedding planners in them so I just assumed the planner would be onsite. LOL. I will update the doc to make it clear that in this scenario, the party planner is at the party to witness the party band playing and therefore knows the truth firsthand.

jhaaaa:
@mDuo13 - Thank you for yet another round of very thoughtful feedback. Please let me know if there is anything else you'd like to see implemented. 🙌

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,139 @@
...
+<span class="use-case-step-num">{{n.next()}}</span>
+## Receiver: [Finish the escrow](send-a-conditionally-held-escrow.html#6-submit-escrowfinish-transaction)
+
+The party band (receiver) shows up and plays their set.
+
+The party planner (oracle) is present at the party to ensure that everything is going smoothly. The party planner confirms first-hand that the party band has fulfilled their contract and publishes the fulfillment publicly, or at least to the party band.
+
+The party band must finish the escrow before 12 noon. If they don't, the escrow expires and the party band doesn't get paid.
+
+If the party planner does not publish the fulfillment (the party band is a no show) or if the party planner publishes the fulfillment, but no one finishes the escrow; after 12 noon the next day, any participant can [cancel the escrow](cancel-an-expired-escrow.html). Cancelling the escrow returns the held XRP to the party host's account.
```
mDuo13: Suggested change:

```
- any participant can
---
+ anyone can
```

because, _literally anyone_ can cancel an expired escrow.

content/use-cases/use-an-escrow-as-a-smart-contract.md
```diff
@@ -0,0 +1,139 @@
...
+
+
+<span class="use-case-step-num">{{n.next()}}</span>
+## Receiver and Sender: [Wait for validation](send-a-conditionally-held-escrow.html#7-wait-for-validation) and [confirm final result](send-a-conditionally-held-escrow.html#8-confirm-final-result)
+
+The party band (receiver) waits for validation of the ledger that contains the escrow finish transaction and then confirms that the escrow was finished.
+
+At this time, the party band provides the transaction's `hash` value to the party host (sender). They can use the `hash` value to look up the escrow transaction on the XRP Ledger to ensure that it is been finished correctly.
+
+The party band can also just check their XRP Ledger account balance to ensure that their balance has increased by 2000 XRP. The party host's balance won't change at this step (unless the escrow was cancelled) because the escrow creation already debited the locked-up XRP from their account.
```
mDuo13: > The party band can also just check

Suggest simplifying this to:

> The party band can check

mDuo13: Also, _super-duper nit:_ prefer [the US spelling "canceled"](https://www.dictionary.com/browse/canceled) over the British spelling that has two L's.

mDuo13 (approved):
The revised version looks good. I left a couple copy-edits but I have no substantive improvements to suggest to the content.

PR #514, opened 2019-02-13
Simple XRP Payments

mDuo13:
Creates an interactive tutorial walking you through the steps of sending an XRP payment using RippleAPI for JavaScript. Hopefully this fills in some of the gaps in our documentation for advanced beginners.

You can view a rendered version of the tutorial [on staging (Ripple VPN required)](http://developers-staging.ripple.com/send-xrp.html) or you can build it yourself with Dactyl.


content/tutorials/use-simple-xrp-payments/send-xrp.md
```diff
@@ -0,0 +1,593 @@
...
+
+**Caution:** Ripple operates the XRP Test Net for testing purposes only, and regularly resets the state of the test net along with all balances. As a precaution, Ripple recommends **not** using the same addresses on the Test Net and production.
+
+
+## Send a Payment on the Test Net
+{% set n = cycler(* range(1,99)) %}
+
+### {{n.next()}}. Connect to a Test Net Server
+
+To provide the necessary auto-fillable fields, ripple-lib must be connected to a server where it can get the current status of your account and the shared ledger itself. (You _can_ sign transactions while being offline, but you need to provide more fields yourself.) You also, obviously, need to be connected to the network to submit transactions to it.
```
ximinez: > You also, obviously, need to be connected to the network to submit transactions to it.

I never cease to be surprised at the seemingly obvious things that perfectly reasonable people find not to be obvious...

content/tutorials/use-simple-xrp-payments/send-xrp.md
```diff
@@ -0,0 +1,593 @@
...
+```js
+// Continues from previous examples.
+// earliestLedgerVersion was noted when the transaction was submitted.
+// txID was noted when the transaction was signed.
+tx = await api.getTransaction(txID, {minLedgerVersion: earliestLedgerVersion})
+console.log("Transaction result:", tx.outcome.result)
+console.log("Balance changes:", JSON.stringify(tx.outcome.balanceChanges))
+```
+
+The RippleAPI `getTransaction()` method only returns success if the transaction is in a validated ledger version.
```
ximinez: Is this code valid if the transaction is _not_ in a validated ledger version?

content/tutorials/use-simple-xrp-payments/send-xrp.md
```diff
@@ -0,0 +1,593 @@
...
+- [You must connect to a server that's synced with the production XRP Ledger network.](#connecting-to-the-production-xrp-ledger)
+
+### Getting a Real XRP Account
+
+This tutorial uses a button to get an address that's already funded with Test Net XRP, which only works because Test Net XRP is not worth anything. For actual XRP, you need to get XRP from someone who already has some. (For example, you might buy it on an exchange.) You can generate an address and secret that'll work on either production or the test net using RippleAPI's [generateAddress() method](rippleapi-reference.html#generateaddress):
+
+```js
+const generated = api.generateAddress()
+console.log(generated.address) // Example: rGCkuB7PBr5tNy68tPEABEtcdno4hE6Y7f
+console.log(generated.secret) // Example: sp6JS7f14BuwFY8Mw6bTtLKWauoUs
```
ximinez, line 575: Maybe mask this secret?

content/tutorials/use-simple-xrp-payments/send-xrp.md
```diff
@@ -0,0 +1,593 @@
+# Send XRP
+
+This tutorial walks through send a simple XRP Payment using JavaScript, by first stepping through the process using the XRP Test Net, then comparing that to the additional requirements for doing the equivalent in production.
```
gbarr01: * Missing "how to" or something like it
* Is the comma right? Seems wrong. Not sure.
* Also, not to edit for style but: "through ... through" and "using ... using"

Maybe:
> This tutorial explains how to send a simple XRP Payment using JavaScript by stepping through the process with the XRP Test Net, then comparing that to the additional requirements for doing the equivalent in production.

content/tutorials/use-simple-xrp-payments/send-xrp.md
```diff
@@ -0,0 +1,593 @@
...
+  const EXAMPLE_ADDR = "rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe"
+  const EXAMPLE_SECRET = "s████████████████████████████"
+  $("#populate-creds-button").click( () => {
+
+  })
+
+})
+</script>
+
+**Caution:** Ripple operates the XRP Test Net for testing purposes only, and regularly resets the state of the test net along with all balances. As a precaution, Ripple recommends **not** using the same addresses on the Test Net and production.
```
gbarr01:
```
< resets the state of the test net
---
> resets the state of the Test Net
```

content/tutorials/use-simple-xrp-payments/send-xrp.md
```diff
@@ -0,0 +1,593 @@
...
+```
+
+The bare minimum set of instructions you must provide for an XRP Payment is:
+
+- An indicator that this is a payment. (`"TransactionType": "Payment"`)
+- The sending address. (`"Account"`)
+- The address that should receive the XRP (`"Destination"`). This can't be the same as the sending address.
+- The amount of XRP to send (`"Amount"`). Typically, this is specified as an integer in "drops" of XRP, where 1,000,000 drops equals 1 XRP.
+
+Technically, a viable transaction must contain some additional fields, and certain optional fields like `LastLedgerSequence` are strongly recommended. The [`prepareTransaction()` method](rippleapi-reference.html#preparetransaction) automatically fills in good defaults for the remaining fields of a transaction. Here's an example of preparing the above payment:
```
gbarr01:
```
< certain optional fields like `LastLedgerSequence`
---
> certain optional fields such as `LastLedgerSequence`
```

content/tutorials/use-simple-xrp-payments/send-xrp.md
```diff
@@ -0,0 +1,593 @@
...
+```
+
+This method returns the **tentative** result of trying to apply the transaction locally. This result _can_ change when the transaction is included in a validated ledger: transactions that succeed initially might ultimately fail, and transactions that fail initially might ultimately succeed. Still, the tentative result often matches the final result, so it's OK to get excited if you see `tesSUCCESS` here. 😁
+
+If you see any other result, you should check the following:
+
+- Are you using the correct addresses for the sender and destination?
+- Did you forget any other fields of the transaction, skip any steps, or make any other typos?
+- Do you have enough Test Net XRP to send the transaction? The amount of XRP you can send is limited by the [reserve requirement](reserves.html), which is currently 20 XRP with an additional 5 XRP for each "object" you own in the ledger. (If you generated a new address with the Test Net Faucet, you don't own any objects.)
+- That you are connected to a server on the test network.
```
gbarr01: This is a fragment and doesn't match your 3 previous questions:

```
< That you are connected to a server on the test network.
---
> Are you connected to a server on the test network?
```

content/tutorials/use-simple-xrp-payments/send-xrp.md
```diff
@@ -0,0 +1,593 @@
...
+      $("#loader-{{n.current}}").hide()
+      $("#submit-output").text("Error: "+error)
+    }
+
+  })
+</script>
+
+### {{n.next()}}. Wait for Validation
+
+Most transactions are accepted into the next ledger version after they're submitted, which means it may take 4-7 seconds for a transaction's outcome to be final. If the XRP Ledger is busy or poor network connectivity delays a transaction from being relayed throughout the network, a transaction may take longer to be confirmed. (For information on how to set an expiration for transactions, see )
```
gbarr01: See what?
```
< (For information on how to set an expiration for transactions, see )
---
> (For information on how to set an expiration for transactions, see )
```

content/tutorials/use-simple-xrp-payments/send-xrp.md
```diff
@@ -0,0 +1,593 @@
...
+
+```js
+const generated = api.generateAddress()
+console.log(generated.address) // Example: rGCkuB7PBr5tNy68tPEABEtcdno4hE6Y7f
+console.log(generated.secret) // Example: sp6JS7f14BuwFY8Mw6bTtLKWauoUs
+```
+
+**Warning:** You should only use an address and secret that you generated securely, on your local machine. If another computer generated the address and secret and sent it to you over a network, it's possible that someone else on the network may see that information. If they do, they'll have as much control over your XRP as you do. It's also recommended not to use the same address for the test net and for production, because transactions that you created for use on one network could potentially also be viable on the other network, depending on the parameters you provided.
```
gbarr01, line 578:
```
< the same address for the test net
---
> the same address for the Test Net
```

content/tutorials/use-simple-xrp-payments/send-xrp.md
```diff
@@ -0,0 +1,593 @@
...
+
+If you [install `rippled`](install-rippled.html) yourself, it connects to the production network by default. (You can also [configure it to connect to the Test Net](connect-your-rippled-to-the-xrp-test-net.html) instead.) After the server has synced (typically within about 15 minutes of starting it up), you can connect RippleAPI to it locally, which has [various benefits](rippled-server-modes.html#reasons-to-run-a-stock-server). The following example shows how to connect RippleAPI to a server running the default configuration:
+
+```js
+ripple = require('ripple-lib')
+api = new ripple.RippleAPI({server: 'ws://localhost:6006'})
+api.connect()
+```
+
+**Tip:** The local connection uses the Websocket protocol (`ws`) unencrypted rather than the TLS-encrypted version (`wss`). This is secure only because the communications never leave the same machine, and is easier to set up because it does not require a TLS certificate. For connections on an outside network, always use `wss`.
```
gbarr01:
```
< Websocket protocol
---
> WebSocket protocol
```

content/tutorials/use-simple-xrp-payments/send-xrp.md
ximinez: Maybe include a note around here explaining that you need to generate a new account to use the demos, because we don't want to risk anybody accidentally including their mainnet secret key, and because you should never give any secret key to a third party.

content/tutorials/use-simple-xrp-payments/send-xrp.md
ximinez: Also, we seem to be moving in a direction where users _should_ sign transactions offline. 

> (You _should_ sign transactions while being offline, but you need to provide more fields yourself, which are out of scope for this demo.)

content/tutorials/use-simple-xrp-payments/send-xrp.md
```diff
@@ -0,0 +1,593 @@
...
+    const tx = await api.getTransaction(txID, {minLedgerVersion: earliestLedgerVersion})
+
+    $("#get-tx-output").html(
+      "<div><strong>Transaction result:</strong> " +
+      tx.outcome.result + "</div>" +
+      "<div><strong>Balance changes:</strong> <pre><code>" +
+      JSON.stringify(tx.outcome.balanceChanges, null, 2) +
+      "</pre></code></div>"
+    )
+  })
```
ximinez, line 558: This function doesn't put the little check mark in the breadcrumb list after "Check".

ximinez (changes requested):

content/tutorials/use-simple-xrp-payments/send-xrp.md
mDuo13: Went with this:

> (For more security, you should sign transactions while being offline, but you must provide the auto-fillable fields manually if you do so.) You must be connected to the network to submit transactions to it.

content/tutorials/use-simple-xrp-payments/send-xrp.md
```diff
@@ -0,0 +1,593 @@
...
+    }
+
+  })
+</script>
+
+### {{n.next()}}. Wait for Validation
+
+Most transactions are accepted into the next ledger version after they're submitted, which means it may take 4-7 seconds for a transaction's outcome to be final. If the XRP Ledger is busy or poor network connectivity delays a transaction from being relayed throughout the network, a transaction may take longer to be confirmed. (For information on how to set an expiration for transactions, see )
+
+You can trigger code to run using the `ledger` event type in RippleAPI. For example:
```
jhaaaa: A little bit confused by this sentence. Maybe adding what running the code will do -- will help? Humbly suggested edit:

> You can run code that uses the RippleAPI [`ledger` event type] to signal whenever a new ledger version is validated on the connected server.

content/tutorials/use-simple-xrp-payments/send-xrp.md
mDuo13: It will raise an exception, I think. Added a try/catch to the code sample and the actual code.

But there's also a question for @intelliot about whether providing a minLedgerVersion may cause it to search non-validated ledgers. (I think there may be a case where it returns an error because it fails to find the `ledger_index` field in `rippled`'s response while doing so.)

content/tutorials/use-simple-xrp-payments/send-xrp.md
mDuo13: I wanted at least one example of an un-masked secret so that users could see what one _actually_ looks like. This is the same secret that's used in the RippleAPI reference docs for `generateAddress()`.

content/tutorials/use-simple-xrp-payments/send-xrp.md
mDuo13: Was wondering if someone else would catch that. ;) I noticed after deploying.

content/tutorials/use-simple-xrp-payments/send-xrp.md
```diff
@@ -0,0 +1,593 @@
...
+      "</pre></code></div>"
+    )
+  })
+</script>
+
+## Differences for Production
+
+To send an XRP payment on the production XRP Ledger, the steps you take are largely the same. However, there are some key differences in the necessary setup:
+
+- [Getting real XRP isn't free.](#getting-a-real-xrp-account)
```
jhaaaa, line 565: 😂 - nice.

content/tutorials/use-simple-xrp-payments/send-xrp.md
mDuo13: Rephrased to:

> You use the `ledger` event type in RippleAPI to trigger your code to run whenever there is a new validated ledger version.

jhaaaa:
Wow - talk about USEFUL documentation. Thank you for this *next-level* doc, @mDuo13.

intelliot (approved):
This is awesome! It's very cool for the tutorial to be interactive in the browser. Thanks for writing this.

A few ideas for the future:
1. A warning about transaction malleability.
2. A way to address (1) is to take note of the sequence number [*] that was auto-filled by ripple-lib. If the account ends up with a transaction using that sequence number, see if that tx is what you expected. It might be helpful to note that the sequence number can be used as a nonce/idempotency key in some respects.
3. Show the reader how to specify the sequence number manually, which is necessary if you are quickly submitting multiple transactions (without waiting for the prior tx to be fully validated). ripple-lib currently fails to advance the auto-filled sequence number until the tx is fully validated, so without specifying it manually, you can't execute more than 1 tx per ledger.

[*] In production, I would recommend auditing the sequence number to make sure it's the expected one. For example, if your last tx had sequence 5 and the next tx has sequence 7, you should check what happened in between to ensure that there's no bug in your code and that your account hasn't been compromised.

Similarly, I would recommend auditing your account's balances on the XRP Ledger and ensuring that the number matches what you expect (in your internal ledger).

(Some of these ideas may be better suited for other pages of the docs.)

Overall, this is a great tutorial. When we release new APIs in the future, it will be very useful to have a similar tutorial for them.

mDuo13:
Those are all good ideas, @intelliot! I've added a [task](https://ripplelabs.atlassian.net/browse/DOC-2172) to our backlog to collect and publish a tutorial on using Sequence numbers and sending multiple transactions

PR #524, opened 2019-02-26
Update text on some old diagrams to change 'Ripple'→'XRP Ledger'

mDuo13:
Some old diagrams still said "Ripple" where they mean "XRP Ledger" so I changed them. I also changed the title of the Funds Flow diagram to make it clearer that it's not how XRP itself works.

I don't like using the XRP symbol in this context, where we're talking about using issued currencies in the XRP Ledger, but I don't think we have anything better at the moment. A previous PR already did a one-for-one swap of the triskelion for the "X" mark, but neglected to change the text where it said "Ripple" to say "XRP Ledger" instead.

I broke the "e2g" diagram source files off from the all-in-one "Gateway Guide Graphics v6.svg" file, which was a little hard to work with.


ryangyoung (approved):
👍

PR #525, opened 2019-02-27
Updated use case template

mDuo13:
The revised template is capable of fitting all kinds of things in and after the steps without falling apart. To use it:

- Add the `use_case` filter to the page in the Dactyl config. Example:

            -   md: use-cases/contribute-code-to-ripple-lib.md
                 html: contribute-code-to-ripple-lib.html
                 funnel: Use Cases
                 template: template-use-case.html
                 useful_background:
                     - rippleapi-reference.html
                     - get-started-with-rippleapi-for-javascript.html
                 filters:
                     - use_case
                 targets:
                     - local

- Mark the start of the numbered steps with `<!-- USE_CASE_STEPS_START -->`
- Optionally, mark the end of the numbered steps with `<!-- USE_CASE_STEPS_END -->`. If you skip this, it'll automatically go to the end of the page.

Otherwise, works the same as the current template.

### DIFF
--- a/assets/css/devportal.css
+++ b/assets/css/devportal.css
@@ -778,12 +778,19 @@ a.current {
 }
 
 /* Use Cases ---------------------------------------------------------------- */
+.use-case-steps {
+  padding-left: 1.5rem;
+  margin-left: 1.5rem;
+  border-left: 1px dashed #3B4147;
+  position: relative;
+}
+
 .use-case-step-num {
   border: 1px solid #3B4147;
   border-radius: 50%;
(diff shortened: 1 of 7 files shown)


jhaaaa:
This is a 🎁! So helpful to be able to use more formatting options within steps. Thank you for evolving the template as the definition of a use case evolved. I also appreciate the updated paragraph spacing -- so easy on the eyes. 🙇‍♀️

jhaaaa (approved):
Looks great 💯

PR #528, opened 2019-03-04
Update assign-a-regular-key-pair.md

intelliot:
I think `comprised` was a typo. Also, I don't see how it's less severe since someone with the regular private key could steal the account from its current owner.


mDuo13:
Good catch. Thank you.

PR #529, opened 2019-03-05
Delete downloaded RPM after installation on Ubuntu

alloynetworks:
The downloaded RPM should be deleted before the next update. To ensure that `sudo alien -i --scripts rippled*.rpm` doesn't accidentally install an earlier version of `rippled` that has already been downloaded.


mellery451:
@mDuo13 we have some upcoming packaging changes for 1.3. This doc change probably makes sense in the broader context of reviewing our installation instructions WRT 1.3. Thanks.

mDuo13 (approved):
After looking a little more closely, I think it's best to add this step now (I edited it to provide a _little_ more context) and revisit when we know more about the packaging changes for 1.3.

PR #531, opened 2019-03-08
removed verb "use" when talking about open source licenses limitation

federicoculloca:
No proper open source license can limit how you can *use* software. See point 6 of the [Open Source Definition](https://opensource.org/osd)


mDuo13 (approved):
That's a fair point, and I agree with the revised language (after removing the comma that's now unnecessary).

PR #532, opened 2019-03-10
Remove duplicated section

keefertaylor:
Public servers are already enumerated and describe on line 23. This PR removes the duplicated section below.


mDuo13:
That's not a duplicated section. One table describes the servers' WebSocket ports, the other describes their JSON-RPC ports.

We could make that clearer by changing the headers or listing them all together in one table, but just removing the JSON-RPC information is not the way to go.

PR #535, opened 2019-03-12
Escrow state diagram

mDuo13:
Adds a new [Escrow state diagram](https://github.com/ripple/ripple-dev-portal/blob/3363e92faadb7414f357ec2e365b90266c8380b3/img/escrow-states.png), and a section describing it, to the Escrow concept page. This diagram shows the states an Escrow progresses through and visualizes the relationship between FinishAfter and CancelAfter.

Also removes an outdated mention of Interledger Protocol requiring escrow (with ILPv4, any settlement method is possible and payment channels are preferred) and updates the requirements for an escrow following the fix1571 amendment.

(Reviewers can skip over the `escrow-states.uxf` file, which is the [umlet](https://www.umlet.com/) source file for the new diagram.)


thejohnfreeman:
If `CancelAfterTime` is unspecified, can the Escrow ever be canceled?

mDuo13:
@thejohnfreeman Nope. If there is no `CancelAfter` time, the escrow never becomes expired and can never be canceled.

(You're required to specify at least one of `CancelAfter` and `FinishAfter` though.)

content/concepts/complex-payment-types/escrow.md
```diff
@@ -33,12 +33,32 @@ Escrow is designed as a feature to enable the XRP Ledger to be used in the [Inte
 - Escrow only works with XRP, not issued currencies.
 - Escrow requires sending at least two transactions: one to create the escrow, and one to finish or cancel it. Thus, it may not be financially sensible to escrow payments for very small amounts, because the participants must destroy the [transaction cost](transaction-cost.html) of the two transactions.
     - When using Crypto-Conditions, the [cost of the transaction to finish the escrow](#escrowfinish-transaction-cost) is higher than usual.
-- All escrows must have a "finish-after" time, an expiration time, or both. Neither time can be in the past when the transaction to create the escrow executes.
+- All escrows must be created with a "finish-after" time or a [crypto-condition][]. If the escrow does not have a finish-after time, it must have an expiration time.
```
seelabs: Right now, we have one condition phrased with a logical or (finish after or crypto-condition), and anther phrased with an if statement(if not finish-after time then expiration time). These statements are true; however I'd phrase both conditions using logical or, rather than mix in the `if`. I.e `All escrows must be created with a finish-after time or an expiration time.`. In both cases we have a logical or, but in one case you phrased it with an `if` statement.

content/concepts/complex-payment-types/escrow.md
```diff
@@ -33,12 +33,32 @@ Escrow is designed as a feature to enable the XRP Ledger to be used in the [Inte
...
+
+- **Combination Escrow (center):** If the escrow specifies both a crypto-condition (`Condition` field) _and_ a "finish-after" time (`FinishAfter` field), the escrow is **Held** until its finish-after time has passed. Then it becomes **Conditionally Ready**, and can finish it if they supply the correct fulfillment to the crypto-condition. If the escrow has an expiration time (`CancelAfter` field), and no one finishes it before that time passes, then the escrow becomes **Expired**. In the expired state, an escrow cannot be finished, and anyone can cancel it.
+
+- **Conditional Escrow (right):** If the escrow specifies a crypto-condition (`Condition` field) and not a finish-after time, the escrow becomes **Conditionally Ready** immediately when it is created. During this time, anyone can finish the escrow, but only if they supply the correct fulfillment to the crypto-condition. If no one finishes the escrow before its expiration time (`CancelAfter` field), the escrow becomes **Expired**. (An escrow without a finish-after time _must_ have an expiration time.) In the expired state, the escrow can no longer be finished, and anyone can cancel it.
+
+
```
seelabs, line 60: Nice diagram. One nit: in the `condition only` scenario, the cancel-after time is required, but the diagram has it as `if specified`. I don't have a good solution for you (and I don't want to clutter up your nice, clean diagram). It's probably fine, as the text does state these conditions, but something to think about if you can cleanly sneak in that info.

seelabs (approved):
Left two nits, but LGTM. :+1:

content/concepts/complex-payment-types/escrow.md
mDuo13: It's complicated, because the following combos are all valid:
- FinishAfter plus any combination of CancelAfter and Condition
- Both Condition & CancelAfter

The following combinations are not valid:
- CancelAfter only (post-fix1571)
- Condition only

I couldn't find a more concise way of expressing the full set of cases than the "if" statement here.

seelabs: Here are four equavalent ways of saying this (if we consider english to be boolean logic, which I know it isn't):

1) What's there now:

All escrows must be created with a "finish-after" time or a crypto-condition. If the escrow does not have a "finish-after" time, it must have an expiration time.


2) Rephrase so both conditions use "or":

All escrows must be created with a "finish-after" time or a crypto-condition. All escrows must also be created with a "finish-after" time or an expiration time.


3) Rephrase so we group the "finish-after" time as the condition to an "if" statement.

If the escrow does not have a "finish-after" time, it must have *both* a cancel after time and a cryptocondition.


4) Rephrase so we group "finish-after" time as an "or" statement

All escrows must be created with a "finish-after" time or *both* a cancel after time and a cryptocondition.


None of these are prefect; I would probably vote for (4). But what's there is now is fine with me if you like that best.

Edit: The "or" here is inclusive or. Clarifying this fits best with (2). So maybe:

All escrows must be created with a "finish-after" time or a crypto-condition (or both). All escrows must also be created with a "finish-after" time or an expiration time (or both).

mDuo13:
Rebased onto master to get the link fix.

content/concepts/complex-payment-types/escrow.md
mDuo13: I think this is a case where treating English like boolean logic doesn't work very well. Making it two "or" statements both including the "finish-after" time in the two successive statements raises questions, for me, like, "Is that a typo? Are there two separate parts that have different requirements?"

I think the "or" followed by "if" works best in English. You're right that it's better to be clear that it's an inclusive or. (In general, I treat "or" in English as inclusive unless the text says "either ... or ..." but there's no harm in adding "or both" here.)

seelabs: Cool. FYI, still :+1: after latest changes.

PR #543, opened 2019-04-01
Remove instructions on how to run a historical database

jhaaaa:
- Removes the "Running a Historical Database" section from https://developers.ripple.com/data-api.html. The section was removed due to complexity of the task.
- Adds a note stating that users can run a historical database on their own, while also acknowledging that the task is complex. The note asks users to contact `docs@ripple.com` if they want to do it anyway and need the instructions.
- Is docs@ripple.com the best email to provide? Or should they open an issue in the `rippled` repo?

### DIFF
--- a/content/references/data-api.md
+++ b/content/references/data-api.md
@@ -6,13 +6,15 @@ Ripple provides a live instance of the Data API with as complete a transaction r
 
 [**https://data.ripple.com**](https://data.ripple.com)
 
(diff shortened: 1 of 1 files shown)


mDuo13:
Probably best to send them to Ripple Technical Services: <support@ripple.com>, which is the general-purpose email we also send people to from dev blog announcements.

PR #548, opened 2019-04-06
Refer to Ripple's recommended list

MarkusTeufelberger:
For quite a while now best practice and the default setting in rippled has been to trust anything that's published at https://vl.ripple.com instead of only Ripple's validators.

There might also be some further opportunity to explain what validator registries or lists are, what data they contain and how to operate one.


mDuo13:
Thank you for catching this. You are indeed correct.

I changed the text from "validator registry" to "validator list site" because, to my understanding, the idea for a validator registry involves more than just providing a list, but also publishing criteria to be listed, providing metrics of validator performance, and vouching for the identity of the listed validators, at least.

This isn't the time & place to editorialize, but I hope to see more of those things in the future, whether from Ripple, the community, or a consortium/foundation of some sort. Until then, I'm going to stick to calling vl.ripple.com a "validator list site" and save the term registry for something better. 🤞

mDuo13 (approved):

MarkusTeufelberger:
Yeah, fair enough, I wasn't sure about the current nomenclature. https://xrpcharts.ripple.com/#/validators would be probably the closest thing to a registry these days (though Ripple doesn't publish criteria for inclusion and exclusion/removal from their list).

https://developers.ripple.com/xrp-ledger-toml.html also calls this a "validator list site" for reference.

PR #550, opened 2019-04-11
Update devblog links

mDuo13:
Replaces all old dev portal links in the dev portal to go to their new locations at developers.ripple.com/blog

### DIFF
--- a/content/concepts/decentralized-exchange/autobridging.md
+++ b/content/concepts/decentralized-exchange/autobridging.md
@@ -8,6 +8,6 @@ Auto-bridging happens automatically on any OfferCreate transaction. [Payment tra
 
 ## See Also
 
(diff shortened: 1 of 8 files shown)


jhaaaa:
Thank you for these updates, @mDuo13!

Here are a couple of details I'm hoping we can address -- possible? Perhaps via another PR?

1. In Dev Portal docs, links to blog posts are displaying with the `fa-external-link` icon. Now that the blog posts are in the Dev Portal, this icon should not display.

2. In Dev Portal blog posts, external links do not display with the`fa-external-link` icon. Can we fix this?

jhaaaa (approved):
LGTM!

Added a couple of comments about fixing the display of the `fa-external-link` icon, but I think that's probably for a separate PR.

mDuo13:
I agree we should make some changes to when the external link icon appears, in a different task.

PR #552, opened 2019-04-16
Clarify tx finality and open vs. closed ledgers

mDuo13:
This is the first of several pull requests containing the doc changes I've been making for understanding metadata. I divided them up to make it easier to review them individually.

This section doesn't have to do with metadata specifically, but provides some clarifying information around how the process of closing ledgers works. This is useful background, because it helps to explain why the outcome (and, consequently, the metadata) of a transaction may be different in the tentative result than in the final result, even for cases where it succeeded both initially and finally.

### DIFF
--- a/content/concepts/payment-system-basics/ledgers.md
+++ b/content/concepts/payment-system-basics/ledgers.md
@@ -23,10 +23,28 @@ As its name might suggest, a ledger's state tree is a tree data structure. Each 
 
 In the case of transactions, the identifying hash is based on the signed transaction instructions, but the contents of the transaction object when you look it up also contain the results and metadata of the transaction, which are not taken into account when generating the hash.
 
+## Open, Closed, and Validated Ledgers
+
+The `rippled` server makes a distinction between ledger versions that are _open_, _closed_, and _validated_. A server has one open ledger, a history of immutable, validated ledgers, and any number of closed but unvalidated ledgers. The following table summarizes the difference:
+
+| Ledger Type                      | Open                        | Closed                                     | Validated |
+|:---------------------------------|:----------------------------|:-------------------------------------------|:--|
+| **Purpose:**                     | Temporary workspace         | Proposed next state                        | Confirmed previous state |
+| **Number used:**                 | 1                           | Any number, but usually 0 or 1             | One per ledger index, growing over time |
+| **Can contents change?**         | Yes                         | No, but the whole ledger could be replaced | Never |
+| **Transactions are applied in:** | The order they are received | Canonical order                            | Canonical order |
+
+Unintuitively, the XRP Ledger never "closes" an open ledger to convert it into a closed ledger. Instead, the server throws away the open ledger, creates a new closed ledger by applying transactions on top of a previous closed ledger, then creates a new open ledger using the latest closed ledger as a base. This is a consequence of [how consensus solves the double-spend problem](consensus-principles-and-rules.html#simplifying-the-problem).
+
+For an open ledger, servers apply transactions in the order those transactions appear, but different servers may see transactions in different orders. Since there is no central timekeeper to decide which transaction was actually first, servers may disagree on the exact order of transactions that were sent around the same time. Thus, the process for calculating a closed ledger version that is eligible for [validation](consensus.html#validation) is different than the process of building an open ledger from proposed transactions in the order they arrive. To create a "closed" ledger, each XRP Ledger server starts with a set of transactions and a previous, or "parent", ledger version. The server puts the transactions in a canonical order, then applies them to the previous ledger in that order. The canonical order is designed to be deterministic and efficient, but hard to game, to increase the difficulty of front-running Offers in the [decentralized exchange](decentralized-exchange.html).
+
+Thus, an open ledger is only ever used as a temporary workspace, which is a major reason why transactions' [tentative results may vary from their final results](finality-of-results.html).
+
 
 ## See Also
 
-For more information about ledger headers, ledger object IDs, and ledger object types, see [Ledger Data Formats](ledger-data-formats.html)
+- For more information about ledger headers, ledger object IDs, and ledger object types, see [Ledger Data Formats](ledger-data-formats.html)
+- For information on how servers track the history of changes to ledger state, see [Ledger History](ledger-history.html)
 
 
 <!--{# common link defs #}-->
--- a/content/concepts/payment-system-basics/ledgers.md
+++ b/content/concepts/payment-system-basics/ledgers.md
@@ -25,9 +25,9 @@ In the case of transactions, the identifying hash is based on the signed transac
 
 ## Open, Closed, and Validated Ledgers
 
-The `rippled` server makes a distinction between ledger versions that are _open_, _closed_, and _validated_. A server has one open ledger, a history of immutable, validated ledgers, and any number of closed but unvalidated ledgers. The following table summarizes the difference:
+The `rippled` server makes a distinction between ledger versions that are _open_, _closed_, and _validated_. A server has one open ledger, any number of closed but unvalidated ledgers, and an immutable history of validated ledgers. The following table summarizes the difference:
 
-| Ledger Type                      | Open                        | Closed                                     | Validated |
+| Ledger Type:                      | Open                        | Closed                                     | Validated |
 |:---------------------------------|:----------------------------|:-------------------------------------------|:--|
 | **Purpose:**                     | Temporary workspace         | Proposed next state                        | Confirmed previous state |
 | **Number used:**                 | 1                           | Any number, but usually 0 or 1             | One per ledger index, growing over time |
--- a/content/concepts/payment-system-basics/transaction-basics/finality-of-results.md
+++ b/content/concepts/payment-system-basics/transaction-basics/finality-of-results.md
@@ -22,7 +22,7 @@ When you initially submit a transaction, the `rippled` server tentatively applie
 
 - The transaction may be delayed until a later ledger version, or may never be included in a validated ledger. For the most part, the XRP Ledger follows a principle that all valid transactions should be processed as soon as possible. However, there are exceptions, including:
 
-    - If a proposed transaction has not been relayed a majority of validators by the time a [consensus round](consensus.html) begins, it may be postponed until the next ledger version, to give the remaining validators time to fetch the transaction and confirm that it is valid.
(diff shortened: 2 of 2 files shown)


content/concepts/payment-system-basics/ledgers.md
```diff
@@ -23,10 +23,28 @@ As its name might suggest, a ledger's state tree is a tree data structure. Each
 
 In the case of transactions, the identifying hash is based on the signed transaction instructions, but the contents of the transaction object when you look it up also contain the results and metadata of the transaction, which are not taken into account when generating the hash.
 
+## Open, Closed, and Validated Ledgers
+
+The `rippled` server makes a distinction between ledger versions that are _open_, _closed_, and _validated_. A server has one open ledger, a history of immutable, validated ledgers, and any number of closed but unvalidated ledgers. The following table summarizes the difference:
```
nbougalis: So, closed ledgers are immutable too but I don't know if that's a distinction you care to make.

content/concepts/payment-system-basics/transaction-basics/finality-of-results.md
```diff
@@ -15,3 +15,41 @@ For any other result code, it can be difficult to determine if the result is fin
...
+
+    - If a [cross-currency payment](cross-currency-payments.html) succeeds by consuming an [Offer](offers.html) in the [decentralized exchange](decentralized-exchange.html), but a different transaction consumes or creates offers in the same order book, the cross-currency payment may succeed with a different exchange rate than it had when it executed tentatively. If it was a [partial payment](partial-payments.html), it could also deliver a different amount.
+
+    - A [Payment transaction][] that tentatively failed because the sender did not have enough funds may later succeed because another transaction delivering the necessary funds came first in the canonical order. The reverse is also possible: a transaction that tentatively succeeded may fail because a transaction delivering the necessary funds did not come first after being put into canonical order.
```
nbougalis, line 41: Is the empty `[]` intentional here?

mDuo13: Yes, that syntax tells it to reuse a reference link (it's [standard markdown syntax](https://daringfireball.net/projects/markdown/syntax#link); but, in this case, the definitions are pulled in from a separate file using non-standard `{% include 'FILE' %}` syntax.)

content/concepts/payment-system-basics/ledgers.md
mDuo13: Yeah, not sure if it's helpful or confusing to clarify that validated ledgers are a subset of closed ledgers... Since even though validated ledgers are immutable, they can be thrown out and replaced with alternate versions that have the same sequence number, which even though it fits the programming definition of immutability, may not meet some people's expectations of what "immutable" means...

intelliot: I would consider a ledger "mutated" if it is not included in the history of the latest fully validated ledger. I think it's OK to use this alternate meaning of "immutable" because we are saying that the ledger history is immutable. On the other hand, if we're concerned that "immutable" has multiple meanings, we could use one of these terms used by other public blockchains: `confirmed` or `finalized`.

content/concepts/payment-system-basics/ledgers.md
```diff
@@ -23,10 +23,28 @@ As its name might suggest, a ledger's state tree is a tree data structure. Each
...
+
+| Ledger Type                      | Open                        | Closed                                     | Validated |
```
ximinez: Looking at the rendered version of this table, the "Ledger Type" cell is a little confusing, and runs into a "is this a header for the row of the column?" problem. I think it can be improved by some combination of:
1. Put a colon at the end `Ledger Type:`/
2. Right justify the text so it visually lines up with the type names.
3. Remove it entirely.

ximinez (changes requested):

content/concepts/payment-system-basics/ledgers.md
jhaaaa: Suggested nit edit: Would it be helpful to change the order in which we describe the ledger types above to match the order in which the table below presents the ledger types? So instead of open, validated, and closed - open, closed, and validated?

content/concepts/payment-system-basics/transaction-basics/finality-of-results.md
```diff
@@ -15,3 +15,41 @@ For any other result code, it can be difficult to determine if the result is fin
...
+
+## How can non-final results change?
+
+When you initially submit a transaction, the `rippled` server tentatively applies that transaction to its current open ledger, then returns the tentative [transaction results](transaction-results.html) from doing so. However, the transaction's final result may be very different than its tentative results, for several reasons:
+
+- The transaction may be delayed until a later ledger version, or may never be included in a validated ledger. For the most part, the XRP Ledger follows a principle that all valid transactions should be processed as soon as possible. However, there are exceptions, including:
+
+    - If a proposed transaction has not been relayed a majority of validators by the time a [consensus round](consensus.html) begins, it may be postponed until the next ledger version, to give the remaining validators time to fetch the transaction and confirm that it is valid.
```
jhaaaa: nit edit:

> has not been relayed a majority

to

>has not been relayed to a majority

content/concepts/payment-system-basics/transaction-basics/finality-of-results.md
```diff
@@ -15,3 +15,41 @@ For any other result code, it can be difficult to determine if the result is fin
...
+
+    - If an address sends two different transactions using the same sequence number, at most one of those transactions can become validated. If those transactions are relayed through the network in different paths, a tentatively-successful transaction that some servers saw first may end up failing because the other, conflicting transaction reached a majority of servers first.
+
+    - To protect the network from spam, all transactions must destroy a [transaction cost](transaction-cost.html) in XRP to be relayed throughout the XRP Ledger peer-to-peer network. If heavy load on the peer-to-peer network causes the transaction cost to increase, a transaction that tentatively succeeded may not get relayed to enough servers to achieve a consensus, or may be [queued](transaction-queue.html) for later.
```
jhaaaa, line 29: Just curious about causes the transaction to not be relayed to enough servers or be queued for later -- is it the heavy load on the network? Or is it that the transaction cost increased and the account wasn't able to cover the cost? Or both?

content/concepts/payment-system-basics/transaction-basics/finality-of-results.md
```diff
@@ -15,3 +15,41 @@ For any other result code, it can be difficult to determine if the result is fin
...
+
+    - Combinations of two or more of these factors can also occur.
+
+- The [order transactions apply in a closed ledger](ledgers.html#open-closed-and-validated-ledgers) is usually different than the order those transactions were tentatively applied to a current open ledger; depending on the transactions involved, this can cause a tentatively-successful transaction to fail or a tentatively-failed transaction to succeed. Some examples include:
+
+    - If two transactions would each fully consume the same [Offer](offers.html) in the [decentralized exchange](decentralized-exchange.html), whichever one comes first succeeds, and the other fails. Since the order in which those transactions apply may change, the one that succeeded can fail and the one that failed can succeed. Since offers can be partially executed, they could also still succeed, but to a greater or lesser extent.
```
jhaaaa, line 37: >they could also still succeed, but to a greater or lesser extent.

Would this suggested edit be helpful and accurate? 

>they could both also still succeed, but to a greater or lesser extent.

content/concepts/payment-system-basics/ledgers.md
mDuo13: Changed the sentence in question to:

> A server has one open ledger, any number of closed but unvalidated ledgers, and an immutable history of validated ledgers.

content/concepts/payment-system-basics/ledgers.md
mDuo13: Went with adding a colon.

Re: option 3, [accessibility guidelines for tables](https://webaim.org/techniques/tables/data) say:

> Table headers should never be empty. This is particularly of concern for the top-left cell of some tables

content/concepts/payment-system-basics/transaction-basics/finality-of-results.md
mDuo13: Both, because they're related. The load on the network dictates—through a somewhat complex decentralized mechanism—how big of a transaction cost is necessary for a transaction to be relayed throughout the network and proposed for inclusion in a validated ledger.

Every transaction has a fixed cost hardcoded in the `Fee` field when it's signed. Whenever a server sees a new proposed transaction, if the transaction's signed cost isn't high enough for the current load level on that server, the server "drops the transaction on the floor", which is to say it doesn't relay the transaction to its peers nor does it apply the transaction to its open ledger or include it in any proposals for validation. So if enough servers _aren't_ under heavy load that the transaction gets passed around to a majority of validators who aren't busy, then there's a good chance the transaction may become validated anyway. Load among servers in the network tends to be correlated (they're processing the same transactions, after all) so the transaction might not make it to enough servers. Also, the topology of the peer-to-peer network can affect this, since even if most of the servers aren't busy, if you submit to a server that has to relay things through busy peers to get to the rest of the network, then the necessary cost is going to vary a bit there.

There's also the open ledger cost, which works a little differently since transactions that don't make the cut are queued & still relayed to peers if they're "likely to succeed".

By the way, it matters less how much the account _could_ pay, and more how much the transaction "promises" to pay with that hardcoded, signed transaction cost. If a transaction is signed with a cost of, say, 12 drops, then if it gets relayed through the network and included in a validated ledger, _it will definitely destroy those 12 drops of XRP_ if at all possible. The only exceptions are if a transaction promises to pay more than its sender has. If the servers notice that while looking at the transaction provisionally, they'll throw it out with a `terINSUF_FEE_B` code. If for some reason the transaction makes it to enough servers to reach a consensus and _then_ it turns out it doesn't have enough XRP (probably something else changed how much XRP the account had at the last minute) then it consumes as much as possible and fails with a tec code. (Not sure which one, maybe `tecUNFUNDED`?)

content/concepts/payment-system-basics/transaction-basics/finality-of-results.md
mDuo13: After thinking carefully about it, it's accurate, but I'm not sure it's helpful. The original statement is true about all offers in general, even outside the context of this example.

content/concepts/payment-system-basics/ledgers.md
ximinez: It's a small thing, but I think it helps. Thanks.

ximinez (approved):

PR #553, opened 2019-04-16
Clarify some RippleState, AccountTxnID stuff

mDuo13:
This is the second PR breaking up some of the work I've done regarding transaction metadata, after #552. This one provides some clarifications and links to the RippleState ledger object description and the AccountTxnID field of accounts.


content/references/rippled-api/ledger-data-formats/ledger-object-types/ripplestate.md
```diff
@@ -1,9 +1,14 @@
...
+
+## High vs. Low Account
+
+There can only be one `RippleState` object per currency for any given pair of accounts. Since no account is privileged in the XRP Ledger, a `RippleState` object sorts their account addresses numerically, to ensure a canonical form. Whichever address is numerically lower when [decoded](accounts.html#address-encoding) is deemed the "low account" and the other is the "high account". The net balance of the trust line is stored from the low account's perspective.
+
+The "issuer" for the balance in a trust line is whichever account has a negative balance. If a `RippleState` object shows a negative balance, the low account is the issuer. If the balance is positive, the high account is the issuer. Frequently, but not always, you can also recognize the issuer as the account whose limit is set to 0.
```
nbougalis: I don't know if I like the "_Frequently, but not always, you can also recognize the issuer as the account whose limit is set to 0._" bit. It's confusing, because it's not accurate (a fact that you allude to by the way you start that sentence). I'd simply not include it.

mDuo13: I think it's a really handy shortcut to have available, even if it's not reliable, so I'm hesitant to remove it entirely. How about this rephrasing?

> Frequently, the issuer has its limit set to 0 and the other account has a positive limit, but this is not reliable because limits can change without affecting an existing balance.

intelliot: Is it useful to consider what the limits mean? The reason why the issuer has its limit set to 0 is that it is the one being trusted with off-ledger value. The account receiving currency from the issuer is trusting that the issuer will honor its obligation to redeem that currency. But only up to a certain (positive) limit.

mDuo13: I thought about describing that, but this isn't the right place to discuss that sort of thing. The "trust lines and issuing" page linked earlier in the diff covers that; this part only covers how to go from the low-level technical implementation of that to the conceptual intent.

Oh, and there are some weird caveats because you can, say, use an Offer to acquire issued currency in excess of your stated limit. (The thinking behind that decision is: if you're trading for that currency, then clearly you are OK with holding more of it.)

mDuo13: Maybe it's wise to link that article again here.

content/references/rippled-api/ledger-data-formats/ledger-object-types/ripplestate.md
```diff
@@ -1,9 +1,14 @@
...
+The `RippleState` object type connects two accounts in a single currency. Conceptually, a `RippleState` object represents two [trust lines](trust-lines-and-issuing.html) between the accounts, one from each side. Each account can change the settings for its side of the `RippleState` object, but the balance is a single shared value. A trust line that is entirely in its default state is considered the same as a trust line that does not exist, so `rippled` deletes `RippleState` objects when their properties are entirely default.
+
+## High vs. Low Account
+
+There can only be one `RippleState` object per currency for any given pair of accounts. Since no account is privileged in the XRP Ledger, a `RippleState` object sorts their account addresses numerically, to ensure a canonical form. Whichever address is numerically lower when [decoded](accounts.html#address-encoding) is deemed the "low account" and the other is the "high account". The net balance of the trust line is stored from the low account's perspective.
```
gbarr01: Perhaps: 
```
< a `RippleState` object sorts their account addresses numerically
---
> a `RippleState` object sorts account addresses numerically
```

content/references/rippled-api/ledger-data-formats/ledger-object-types/ripplestate.md
```diff
@@ -1,9 +1,14 @@
...
+
+The ["issuer"](trust-lines-and-issuing.html) for the balance in a trust line is whichever account has a negative balance. If a `RippleState` object shows a negative balance, the low account is the issuer. If the balance is positive, the high account is the issuer. Frequently, the issuer has its limit set to 0 and the other account has a positive limit, but this is not reliable because limits can change without affecting an existing balance.
```
gbarr01: Perhaps: 
```
< The issuer for the balance in a trust line is whichever account has a negative balance. 
---
> The issuer for the balance in a trust line depends upon whether the balance is negative or positive. 
```

gbarr01 (approved):
LGTM

PR #555, opened 2019-04-16
WIP: Provide Helpfulness Survey on Dev Portal

jhaaaa:
This PR is a proof-of-concept for using a TypeForm survey to provide a "Was this content helpful?" mini-survey on every appropriate page on the Dev Portal.

### DIFF
--- a/tool/template-doc.html
+++ b/tool/template-doc.html
@@ -25,6 +25,12 @@
     <div class="content">
         {{ content }}
     </div>
+    <!--{# start embed of the typeform survey widget at the bottom of the page. #}-->
+    <hr>
+    <div id="typeform-survey" class="typeform-widget" data-url="" style="width: 100%; height: 300px"></div>
+    <script> $('#typeform-survey').attr('data-url', 'https://ryan1106.typeform.com/to/CROjck?source=' + window.location.pathname); </script>
+    <script> (function() { var qs,js,q,s,d=document, gi=d.getElementById, ce=d.createElement, gt=d.getElementsByTagName, id="typef_orm", b="https://embed.typeform.com/"; if(!gi.call(d,id)) { js=ce.call(d,"script"); js.id=id; js.src=b+"embed.js"; q=gt.call(d,"script")[0]; q.parentNode.insertBefore(js,q) } })() </script>
+    <!--{# end embed of the typeform survey widget at the bottom of the page. #}-->
   </article>
 {% endblock %}
 
@@ -34,11 +40,5 @@
     <ul class="dev_nav_sidebar card-body" id="dactyl_toc_sidebar">
       {{ sidebar_content }}
     </ul>
-    <hr>
-    <!--{# embeds the typeform survey widget. #}-->
-    <!--{# TODO: On smaller screens, the survey displays before the content - turn off the survey for mobile screens? #}-->
-    <div id="typeform-survey" class="typeform-widget" data-url="" style="width: 100%; height: 300px;"></div>
-    <script> $('#typeform-survey').attr('data-url', 'https://ryan1106.typeform.com/to/CROjck?source=' + window.location.pathname); </script>
-    <script> (function() { var qs,js,q,s,d=document, gi=d.getElementById, ce=d.createElement, gt=d.getElementsByTagName, id="typef_orm", b="https://embed.typeform.com/"; if(!gi.call(d,id)) { js=ce.call(d,"script"); js.id=id; js.src=b+"embed.js"; q=gt.call(d,"script")[0]; q.parentNode.insertBefore(js,q) } })() </script>
   </div>
 {% endblock %}
--- a/tool/template-doc.html
+++ b/tool/template-doc.html
@@ -28,7 +28,7 @@
     <!--{# start embed of the typeform survey widget at the bottom of the page. #}-->
     <hr>
     <div id="typeform-survey" class="typeform-widget" data-url="" style="width: 100%; height: 300px"></div>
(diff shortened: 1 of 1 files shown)


MarkusTeufelberger:
Do people really click on these and are the answers of people that do representative of the general user base or the helpfulness of the page?

mDuo13:
> Do people really click on these and are the answers of people that do representative of the general user base or the helpfulness of the page?

Yes, some people do. It's an open question whether they're representative, but collecting this data provides a dimension to our analytics that's otherwise completely absent: specifically, recognizing when someone leaves a page if that means they got their answer or gave up. Even if most people don't answer, and nobody answers "yes", we can recognize a page that needs help if it gets more than a random noise-level amount of "No" answers.

mDuo13 (commented):
As a proof of concept, it works pretty well (assuming we're seeing the data on the backend and we can recognize which page it comes from). There are some things I'd like to look into tweaking before we push it live:

- The form takes a second or two to load, and the empty space in the sidebar looks a little weird while it does. Plus having it slide in just after the page finishes loading is a little distracting. If it's not going to load as fast as the full site (ideally it would just _be there_ rather than popping up after the page loads) then maybe we should come up with a guideline of when it should appear (a while later, after the user has had some time to actually pay attention to the content) and see if we can animate it expanding into place during that time. 

- I don't know if we can control the size of the fonts in the embedded TypeForm, but making them just a _little_ smaller would probably help, since it doesn't currently fit much text. (Even the word "helpful" gets hyphenated on my screen!) It would be nice if they looked about the same size as some of the other body text on the page.

- Similarly, I'd like to make the up & down buttons in the form smaller and/or less brightly colored. They demand a _lot_ of attention for being a very unimportant part of the page. (How relevant is "go to the next/previous question of this form" in the context of most doc pages?)

- Agreed that the form is in a terrible place when you drop to the mobile layout breakpoint. We could easily use a CSS breakpoint to hide it at that breakpoint, but it would be pretty hard to do much else with it using that technique. And it would kind of suck to lose the form entirely on mobile, since our existing metrics show a (surprisingly) non-trivial proportion of views come from mobile devices.

One approach that may solve most or all of those problems—what if we moved the form to immediately follow the page content, rather than being in the sidebar? Users probably won't ever see it on super-long pages, but those pages are and should continue to be the exception. Having it at the bottom seems to be a common pattern on other similar sites, and it makes sense—users are asked to evaluate the content when they reach the end, that is, after they've had a chance to consume it and decide the answer to the question. 

That can be done by moving the code up a few lines to between lines 27 and 28 of this same template file:

```
{% block main %}
  <article class="p-3">
    <div class="content">
        {{ content }}
    </div>

    <hr />
    <!--{# embedded typeform survey widget. #}-->
    <div id="typeform-survey" class="typeform-widget" data-url="" style="width: 100%; height: 300px;"></div>
    <script> $('#typeform-survey').attr('data-url', 'https://ryan1106.typeform.com/to/CROjck?source=' + window.location.pathname); </script>
    <script> (function() { var qs,js,q,s,d=document, gi=d.getElementById, ce=d.createElement, gt=d.getElementsByTagName, id="typef_orm", b="https://embed.typeform.com/"; if(!gi.call(d,id)) { js=ce.call(d,"script"); js.id=id; js.src=b+"embed.js"; q=gt.call(d,"script")[0]; q.parentNode.insertBefore(js,q) } })() </script>
  </article>
{% endblock %}
```

That works pretty well on mobile especially, although it really makes it obvious how much wasted space the embedded TypeForm UI has, and the spacing can look a little ugly alongside the other UI elements:



Compared with having a fully native interface, it certainly doesn't blend in as well.

jhaaaa:
Thank you for the spot-on review, @mDuo13! Agreed on all points.

- I moved the survey to the bottom of the page.
- Putting the survey at the bottom of the page gives the survey a bit more time to load before the user actually sees it. 😅 
- I changed the color of the survey buttons to blend in a bit more. Better? 
- I tried to override the survey's CSS to reduce the size of at least the "question" and "thank you" font. I was unsuccessful. I contacted TypeForm for help and they said it is not currently possible to do this. They've added it as an enhancement request. 
   - We have a few enhancement requests open with them. The TypeForm person said that whether they actually develop the enhancements depends on how many other customers want the enhancements. I'll receive an email if any of our requests actually gets approved for development.
- I'll provide you with a copy of the TypeForm response report. When a user submits either a Yes or No answer, the report shows the Yes or No and the URL of the page the user was on when they submitted the answer. It would be great to get a second set of eyes on it to validate that it is capturing what we need.

mDuo13:
Yep, survey at the bottom of the page definitely works better, and the new colors are good. I'd like some time to look at the response report but otherwise, I'm happy with where this is.

intelliot:
After someone says yes or no, is it possible to let them submit a free form text response to explain why? Just an idea.

jhaaaa:
Hi @intelliot - thank you for asking! 

If the user answers No, we prompt them to optionally tell us how we can improve the content. Thinking that possibly getting this key feedback is worth risking survey abandonment due to the friction added by prompting for a "why."

If the user answers Yes, we just accept the Yes and don't prompt for any further info. Thinking that this quick Yes approach will reduce the chances of folks abandoning the survey before they are done. 

Please let me know if you feel strongly about enabling a user to provide a free-form response for a Yes answer. 🙌

mDuo13:
Going to merge this out and try it for a while. We can remove it or evolve it to a new thing if it doesn't work out, but we may as well see if we can start collecting responses.

PR #561, opened 2019-04-24
Transaction sender tool

mDuo13:
Adds a new dev tool that sends various transactions to the XRP Test Net address of your choice at the push of a button. The motivation for this is, if you're writing some kind of integration software that listens for transactions in the XRP Ledger, this tool provides a convenient interface to send stuff on demand so you can see how your code handles it.

My next goal is to make an interactive tutorial for receiving transactions that uses this tool to demonstrate/be the sending side.

You can try out the tool live [on staging](https://developers-staging.ripple.com/tx-sender.html) (Ripple VPN required).

### DIFF
--- a/assets/js/tx-sender.js
+++ b/assets/js/tx-sender.js
@@ -1,4 +1,8 @@
 const set_up_tx_sender = async function() {
+  //////////////////////////////////////////////////////////////////////////////
+  // Connection / Setup
+  //////////////////////////////////////////////////////////////////////////////
+
   FAUCET_URL = "https://faucet.altnet.rippletest.net/accounts"
   TESTNET_URL = "wss://s.altnet.rippletest.net:51233"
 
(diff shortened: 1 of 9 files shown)


jhaaaa:
What a grand vision for a tool! 🎉  Thank you for having the expertise to develop this, @mDuo13. Just a few comments, do with them what you will!

- For Create Escrow, Create Payment Channel, Send Issued Currency, and Trust for, you might want to link to relevant docs, like you do for Send XRP Payment and Send Partial Payment.

- I think the Connected status in the Status sidebar should use a different style and color than the buttons in the main tool area. The current treatment makes it look clickable.

- I know I can copy and paste the hash value that displays in the ephemeral tx succeeded message, but it might be useful to display a history of transaction hashes for the current session. Maybe under the Status sidebar section?

- It might be useful to provide access to the raw request and response messages for a transaction. The tool can be used to teach newbies about message structure as well as enable advanced users to test their apps.

assets/js/tx-sender.js
```diff
@@ -0,0 +1,465 @@
...
+  console.debug("Getting a sending address from the faucet...")
+
+  faucet_response = function(data) {
+    sending_address = data.account.address
+    sending_secret = data.account.secret
+    xrp_balance = Number(data.balance) // Faucet only delivers ~10,000 XRP,
+                                       // so this won't go over JavaScript's
+                                       // 64-bit double precision
+
+    $("#balance-item").text(xrp_balance) + " drops"
```
intelliot: I think the ` + " drops"` part doesn't do anything and can be deleted. The `balance` returned by the faucet is in XRP.

assets/js/tx-sender.js
```diff
@@ -0,0 +1,465 @@
...
+    let resp = await submit_and_verify({
+      TransactionType: "AccountSet",
+      Account: pp_issuer_address,
+      SetFlag: 8
+    }, pp_issuer_secret, true)
+    if (resp === undefined) {
+      console.log("Couldn't set DefaultRipple for partial payment issuer")
+      return
+    }
+    $("#pp_progress .progress-bar").width("40%")
```
intelliot, line 224: Idea: Since this is a dev tool, I wonder if we could show more detail in the UI about what's happening. For example, it could be a checklist of steps that gets checked off as the steps are completed:

✅Create test account to use as issuer
✅Set DefaultRipple on issuer
➡️Create trust line from sending address to issuer
⏹Issue fake currency to sending address
⏹Place offer to buy fake currency for XRP

assets/js/tx-sender.js
```diff
@@ -0,0 +1,465 @@
...
+  }
+  $("#send_xrp_payment button").click(on_click_send_xrp_payment)
+
+  // 2. Send Partial Payment Handler ---------------------------------------
+  async function on_click_send_partial_payment(event) {
+    const destination_address = $("#destination_address").val()
+    $("#send_partial_payment .loader").show()
+    $("#send_partial_payment button").attr("disabled","disabled")
+
+    // const path_find_result = await api.request("ripple_path_find", {
```
intelliot, line 314: It would be interesting to have a working example of using `ripple_path_find`. Pathfinding is still a mystery to me.

assets/js/tx-sender.js
```diff
@@ -0,0 +1,465 @@
...
+      TransactionType: "Payment",
+      Account: sending_address,
+      Destination: destination_address,
+      Amount: "1000000000000000", // 1 billion XRP
+      SendMax: {
+        value: (Math.random()*.01).toPrecision(15), // random very small amount
+        currency: pp_sending_currency,
+        issuer: pp_issuer_address
+      },
+      Flags: 0x80020000 // tfPartialPayment | tfFullyCanonicalSig
```
intelliot: I think you can actually do:
```js
      Flags: api.txFlags.Payment.PartialPayment | api.txFlags.Universal.FullyCanonicalSig
```

Actually, prepareTransaction always sets FullyCanonicalSig for you, so this should work as well:
```js
      Flags: api.txFlags.Payment.PartialPayment
```

assets/js/tx-sender.js
mDuo13: Yeah, it turns out this particular setup uses the [default cross-currency path](https://developers.ripple.com/paths.html#default-paths) so even if you go pathfinding you'll probably get back an empty PathSet.

For demonstrating paths, I think we'd need an even more complex setup, with a second order book or rippling to different issuers.

assets/js/tx-sender.js
mDuo13: That's good to know, but I think it's still a good practice to set FullyCanonicalSig explicitly since there are APIs and situations, esp. with multisigning, where it won't be enabled automatically. So the first example seems better.

assets/js/tx-sender.js
mDuo13: It's a nice idea, but I'd rather keep the thing visually compact rather than adding a checklist. I could potentially have it update the progress bar text as each step is in progress, though.

My goal is to make this compact enough to fit comfortably on a typical smartphone screen so you can use a totally separate device to send the transactions. (There are some other things I need to tweak for that.)

mDuo13:
Updated the tool on staging. (Careful: I had to hit refresh to see the new version. Just clicking the link again loaded the old one from cache.)

- Fixed the incorrect "drops" label on the balance
- Use RippleAPI flags values
- Added a history of sent transactions
- Made a bunch of things fit better on mobile screens
- Added links for relevant docs on other tx types

> It might be useful to provide access to the raw request and response messages for a transaction. The tool can be used to teach newbies about message structure as well as enable advanced users to test their apps.

You're pretty much describing the existing WebSocket Tool. I do have plans to make a bunch of improvements to the WebSocket tool, including making it easier to [connect it to the TestNet](https://developers.ripple.com/websocket-api-tool.html?base_url=s.altnet.rippletest.net:51233) but that's a separate set of tasks. One of the improvements I have planned is a way to permalink to a specific request & run it automatically. When that's done, then I could make it so the hashes we have displayed in this tool link to that so users can see the full version. (Alternatively, we could link to them in a transaction explainer like XRP Charts has, but for the Test Net.)

PR #563, opened 2019-05-07
Correction: Ethereum does not use a UTXO model

mDuo13:
Corrects a mistake in the consensus intro footnotes, where it wrongly said Ethereum uses a UTXO model. (It uses an account state model [inspired by the XRP Ledger's](https://medium.com/coil/codius-smart-contracts-made-from-containers-b3b16c3e3890).) You do have to [download full block history and state trie](https://medium.freecodecamp.org/how-to-sync-an-ethereum-node-using-geth-and-ethereum-wallet-81423d42a583) unlike the XRP Ledger where you only need the latest state trie, which is smaller than Ethereum's.


jhaaaa (approved):
LGTM!

intelliot (approved):
I would not call myself an Ethereum expert but this looks good to me. I like your point that Ethereum's account state model was inspired by the XRP Ledger.

PR #564, opened 2019-05-10
Added transaction_count to Transaction Metrics

xrpscanner:
Added information about transaction_count that was missing from the Transaction Metrics table.


mDuo13 (approved):
Thanks for catching this! :+1:

PR #572, opened 2019-06-08
xrp-ledger.toml Checker

mDuo13:
- Add xrp-ledger.toml Checker tool. This reads the (defined) contents of an [xrp-ledger.toml file](https://developers.ripple.com/xrp-ledger-toml.html), confirming that CORS is set up properly.
    - For accounts specified in the file, the tool checks their `Domain` fields against the domain being queried.
    - We can potentially add validator domain verification to this tool following the new stuff coming in v1.3.0.
- Remove ripple.txt validator
- Update ripple-lib to 1.2.4

For an example of a working site, enter `mduo13.com` to check my personal XRPL TOML file.


MarkusTeufelberger:
It would be nice to also have a way to put in an address on MainNet or TestNet and have the checker read the `domain` field, go to the correct location and verify the toml file there.

mDuo13:
@MarkusTeufelberger That's a great idea. I've opened #576 to remember to do that later without expanding the scope of this task right now.

assets/js/xrp-ledger-toml-checker.js
```diff
@@ -0,0 +1,214 @@
...
+    let list_wrap = $("<p>"+name+"</p>")
+    let list_ol = $("<ol>").appendTo(list_wrap)
+    for (i=0; i<list.length; i++) {
+      let entry_wrap = $("<li>").appendTo(list_ol)
+      let entry_def = $("<ul class='mb-3'>").appendTo(entry_wrap)
+      let entry = list[i]
+      for (j=0; j<fields.length; j++) {
+        let fieldname = fields[j]
+        if (entry[fieldname] !== undefined) {
+          let field_def = $("<li><strong>"+fieldname+":</strong> ").appendTo(entry_def)
```
intelliot: In `":</strong> "`, the space is ignored/removed by jQuery. Simple fix: Move it to before `</strong>` like so:

`": </strong>"`

intelliot (approved):
One small fix. Overall, this looks great!

intelliot (approved):
LGTM

MarkusTeufelberger:
https://ripple.com/.well-known/xrp-ledger.toml
https://xrpl.org/.well-known/xrp-ledger.toml

:-(

Would be nice to have one of these pre-filled by default instead of "example.com".

alloynetworks:
WIP - TOML reader (returns JSON)
https://xrptools.alloy.ee/checktoml?domain=alloy.ee
1.Will ignore toml files served without https
2. Will strip paths from the domain provided
3. Will return toml for domain with invalid certificate, but with error status.
4. Returns additional server_headers to enable comparing of Last-Modified as seen by server, and as stated in the TOML file.
5. Python source will be published after cleaning up :)

alloynetworks:
Address to TOML finder also active now. Again WIP
https://xrptools.alloy.ee/addresstoml?address=ADDRESS

mDuo13:
@alloyxrp I was wondering why my checker wasn't displaying the information for your file; turns out the checker is right and you have a recurring syntax error in your `toml` file. The `[[VALIDATORS]`, `[[ACCOUNTS]]`, `[[PRINCIPALS]]`, and `[[SERVERS]]` headings need to use _double_ square brackets. That's TOML syntax for "array entry" whereas single square brackets indicates a fixed object. This is a case where TOML doesn't really live up to the "Obvious" in its name.

This might be something I could add to the checker to look for.

_(Apologies if you get a duplicate notification: posted this in the wrong issue a moment ago.)_

alloynetworks:
Not really. I would use the [[ only if I had multiple validators or accounts, like you have in your file.

I am actually using a toml python module to read the file. 

I will try the various possible errors, in case the module isn't doing all the checks.

Will also check today by putting multiple validators with a single brace. I need to be more verbose in my error message.

The docs seem to be pretty clear on this. Per my understanding 😬

mDuo13:
An [array of tables](https://github.com/toml-lang/toml#array-of-tables) with length 1 is not the same as a table on its own.

It's a bit more obvious if I translate it to JSON...

Single-brackets (wrong):

```json
{
    "VALIDATORS": {
        "public_key": "n9KM73uq5BM3Fc6cxG3k5TruvbLc8Ffq17JZBmWC4uP4csL4rFST",
        "desc": "My dogfood rippled machine. Not a reliable validator."
    }
}
```

Double-brackets (correct):

```json
{
    "VALIDATORS": [
        {
            "public_key": "n9KM73uq5BM3Fc6cxG3k5TruvbLc8Ffq17JZBmWC4uP4csL4rFST",
            "desc": "My dogfood rippled machine. Not a reliable validator."
        }
    ]
}
```

I don't think we should mix and match datatypes for this. It should always be an array, even if there's only one entry.

alloynetworks:
It's not wrong. They are both valid json. Maybe in the context of the tool, you require a list.
But fair enough. Then we must document why we insist on double braces. And may as well make the METADATA an array too.
Did you check with and without an array in my TOML reader?

alloynetworks:
> I don't think we should mix and match datatypes for this. It should always be an array, even if there's only one entry.

@mDuo13 : Can you check https://xrptools.alloy.ee/checktoml?domain=alloy.ee now. I've changed it to double quotes (except in METADATA). I'll also add a check to warn if the rest of the sections don't have a `[[....]]`

mDuo13:
Note for later: 

If Firefox reports an XML parsing error in the toml checker, it means your web server is serving the wrong content type. Optionally add the following line to your `.htaccess` to make the browser use the correct content type.

```apache
AddType application/toml .toml
```

alloynetworks:
Yes, I noticed that too. I've modified my tool to check for that, and also give suggestions depending on the webserver.
Taking your observations into account about sections other than METADATA to be arrays, I'm reporting those warnings too.
https://xrptools.alloy.ee/checktoml?domain=mduo13.com

content/tutorials/xrp-ledger-businesses/become-an-xrp-ledger-gateway.md
```diff
@@ -246,7 +246,7 @@ There are several prerequisites that ACME must meet for this to happen:
 - ACME must control an address in the XRP Ledger. Ripple's best practices recommend using a separate issuing address and operational address. See [Issuing and Operational Addresses](issuing-and-operational-addresses.html) for details.
     - ACME must enable the [DefaultRipple Flag](#defaultripple) on its issuing address for customers to send and receive its issuances.
 - Alice must create an accounting relationship (trust line) from her XRP Ledger address to ACME's issuing address. She can do this from any XRP Ledger client application as long as she knows ACME's issuing address.
-    - ACME should publicize its issuing address on its website where customers can find it. It can also use [ripple.txt](#rippletxt) to publish the issuing address to automated systems.
+    - ACME should publicize its issuing address on its website where customers can find it. It can also use an [`xrp-ledger.toml` file](xrp-ledger-toml.html) to publish the issuing address to automated systems.
```
MarkusTeufelberger, line 249: "can also" --> "should also"?

PR #578, opened 2019-06-14
Update link w/ in-house WebSocket client tutorial

mDuo13:
The MDN documentation is still good, but we have some more relevant/specific documentation on this now, so we can replace this link with the (recently added!) tutorial for monitoring on WebSocket.


mDuo13:
!rebuild

PR #598, opened 2019-06-17
Changed Link of Frontend in readme

jetwes:


mDuo13 (approved):
Good catch, thanks. I updated the README a bit more to match the current naming & purpose.

PR #601, opened 2019-06-19
xrp-ledger.toml file - proper syntax for enabling CORS w/ .htaccess 

mDuo13:
also disables a ligature in the code font that's causing `</` to look like a diagonal arrow in this example and the Apache settings sample

### DIFF
--- a/content/references/xrp-ledger-toml.md
+++ b/content/references/xrp-ledger-toml.md
@@ -238,13 +238,22 @@ When creating custom fields, be mindful of the field name you choose. If you use
 
 You MUST configure your web server to allow Cross-Origin Resource Sharing ([CORS][]) for the `xrp-ledger.toml` file. This configuration depends on your web server.
 
-If you use Apache HTTP Server, add the following to your config file or a `.htaccess` file:
+If you run an Apache HTTP Server, add the following to your config file:
 
 ```
 <Location "/.well-known/xrp-ledger.toml">
     Header set Access-Control-Allow-Origin "*"
 </Location>
 ```
+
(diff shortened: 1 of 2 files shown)


mDuo13:
On the ligature thing. Before:
Screenshot from 2019-06-19 16-34-14

After:
Screenshot from 2019-06-19 16-44-15

Might not work on Microsoft Edge (which [doesn't support](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-ligatures#Browser_Compatibility) `font-variant-ligatures`), but I wasn't able to figure out the proper [feature name](https://docs.microsoft.com/en-us/typography/opentype/spec/featurelist) to disable this ligature using `font-feature-settings`.

content/references/xrp-ledger-toml.md
```diff
@@ -238,14 +238,23 @@ When creating custom fields, be mindful of the field name you choose. If you use
...
+Alternatively, you can add the following to a `.htaccess` file in the `/.well-known/` directory of your server:
+
+```
+<Files "xrp-ledger.toml">
+    Header set Access-Control-Allow-Origin "*"
+</Files>
+```
+
+
 If you use nginx, add the following to your config file:
```
gbarr01, line 258: I read `a .htaccess file` as "_an_ H-T access file" ... not sure what is correct.

gbarr01 (approved):
LGTM

content/references/xrp-ledger-toml.md
mDuo13: There isn't a strictly correct answer, but I pronounce it "dot H-T access" file, thus "a" instead of "an".

Issue #419, opened 2018-06-13
XRP and drop naming is not explained


MarkusTeufelberger:
XRP as a name is often used, but on https://developers.ripple.com/xrp.html the explanation what XRP actually stands for (an ISO 4217 compatible code for "Ripple credits/ripples", "1 XRP is one ripple") as well as the inventor of the "drop" name (ThePiachu in https://forum.ripple.com/viewtopic.php?f=1&t=40&p=228 - they might have even been called "Jeds" otherwise) is missing.

yxxyun:
I think this is for reasons,  name XRP "ripples" maybe confused people and made them think it was a securities.

MarkusTeufelberger:
I wouldn't think that I'm buying Amazon shares with Amazon Coins or Google shares if I'm buying something called "googles". I also don't see any reason why naming something like the concept it enables makes it a security. After all, Ripple Labs was called "Open Coin" back when XRP/ripples/Ripple credits were named/invented.

PR #602, opened 2019-06-20
Migrate some content from old wiki "ripple credits" article

mDuo13:
- XRP naming vs. Ripple company naming, plus logo comparison
- origin of the term "drops of XRP"


mDuo13:
Fixes #419

mDuo13:
!rebuild

PR #611, opened 2019-07-03
v1.3.0 install updates

### LINKED ISSUE #544: Updates to Installation Instructions for 1.3 release
We are making some non-trivial changes to our package repository build and infrastructure for the upcoming 1.3 release, so the docs need to change. I'll try to summarize here, but we'll probably have an ongoing discussion as we try to flesh-out the details.

For both package types, I think we need to advise that users do a *manual* upgrade from 1.2, which means:

* stop rippled
* Backup **DATA AND CONFIG FILES** - the location of the config is probably `/opt/ripple/etc/rippled.cfg`, but might be somewhere else if they moved it. The `validators.txt` file in the same location should be saved too. The data directory depends on their config/setup. All data directory files should be saved, including the node-db files.
* uninstall the previous package. For rpm packages, this probably involves `rpm -e rippled`. For things installed via alien on ubuntu, maybe there is an alien uninstall, or just use rpm? (TBD)
* install 1.3 as described below
* restore data and config files to desired location.
* NOTE: If you are just running a *stock server* with no config tweaks and don't care about the data you had before upgrade, then the data/config backup step is not needed.

These steps are specific to the 1.2-->1.3 upgrade. Future versions past 1.3 should not require these manual backup/restore steps.

## rpm installation ##

The repository and instructions for using it have changed. The steps are now something like:

```bash
<with sudo:>
yum -y update
yum -y install yum-utils
cat << REPOFILE > /etc/yum.repos.d/ripple.repo
[ripple-stable]
name=XRP Ledger Packages
baseurl=https://repos.ripple.com/repos/rippled-rpm/stable/
enabled=1
gpgcheck=0
gpgkey=https://repos.ripple.com/repos/rippled-rpm/stable/repodata/repomd.xml.key
repo_gpgcheck=1
REPOFILE
yum -y update
yum -y install rippled
```

(important note: the name used above `[ripple-stable]` is important in order for the update script we provide to work properly - it relies on this naming scheme specifically)

The system enable/start steps in the existing instructions can remain.

## deb package installation ##

We now have a native debian package - no more alien involved. The steps are:

```bash
<with sudo:>
apt -y update
apt -y install apt-transport-https ca-certificates wget gnupg
wget -q -O - "https://repos.ripple.com/repos/api/gpg/key/public" | apt-key add -
# Fingerprint: C001 0EC2 05B3 5A33 10DC 90DE 395F 97FF CCAF D9A2
echo "deb https://repos.ripple.com/repos/rippled-deb <DISTRO> stable" > /etc/apt/sources.list.d/ripple.list
apt -y update
apt-get -y install rippled
```

*Note:* `<DISTRO>` above is one of:

* xenial  (for ub 16.04)
* bionic  (for ub 18.04)
* stretch  (for debian stable/9)

These are currently the only distros we support, but that list will change in the future as we choose to add more debian distros.

I *believe* the daemon will start automatically after installation with deb - but this needs to be confirmed. If it does not, then we can duplicate the systemctl steps from the rpm instructions.

## automatic or manual updates ##

An update script is installed at `/opt/ripple/bin/update-rippled.sh`. This script can be invoked manually, or via cron. We provide a crontab file in `/opt/ripple/etc/update-rippled-cron` which can be enabled by creating a symlink to this file in `/etc/cron.d`

For manual updates on both distros, I think we can just recommend that users run the update script manually, as needed.

mDuo13:
Fixes #544 .

- Changes the Ubuntu install to use the new native package. It also supports Debian 9 now.
    - Changes the filename to not to mention Alien anymore. I considered having it also mention Debian, but decide not to. I left a redirect page for the old URL. (A server-side redirect would be more efficient, but the redirect page doesn't require server admin, so it's more expedient for us.)
        - Updates the sidebar template not to list "redirect" pages if they're subpages. (They were already not listed in some other circumstances.)
- Changes the RHEL install to use the new repository.
- Changes manual upgrade instructions to match. The upgrade instructions now also mention that you can run the upgrade script manually (that is, not using cron) to do the same thing.
- Changes automatic upgrade instructions to use the `cron.d` config file, and to support Ubuntu.
- Adds specific migration instructions for upgrading from 1.2.x to 1.3.0 on either Ubuntu or RHEL. These cover cleaning up stuff needed for the old install styles.
- Changes the build instructions for Ubuntu and macOS to reflect how rippled 1.3.0 supports Boost 1.67 through at least 1.70.


alloynetworks:
@mDuo13 : Maybe we should explicitly say that we recommend installation only on LTS versions of Ubuntu? Sorry, if I missed seeing that in the document.

content/tutorials/manage-the-rippled-server/installation/install-rippled-on-ubuntu.md
```diff
@@ -0,0 +1,88 @@
...
+        $ wget -q -O - "https://repos.ripple.com/repos/api/gpg/key/public" | \
+          sudo apt-key add -
+
+4. Check the fingerprint of the newly-added key:
+
+        $ apt-key finger
+
+    The output should include an entry for Ripple such as the following:
+
+        pub   rsa3072 2019-02-14 [SC] [expires: 2012-02-13]
```
mellery451: transposition of 2012 here (should be 2021 ?)

content/tutorials/manage-the-rippled-server/installation/install-rippled-on-ubuntu.md
```diff
@@ -0,0 +1,88 @@
...
+            sudo tee -a /etc/apt/sources.list.d/ripple.list
+
+    The above example is appropriate for **Ubuntu 18.04 Bionic Beaver**. For other operating systems, replace the word `bionic` with one of the following:
+
+    - `xenial` for **Ubuntu 16.04 Xenial Xerus**
+    - `stretch` for **Debian 9 Stretch**
+
+    If you want access to development or pre-release versions of `rippled`, use one of the following instead of `stable`:
+
+    - `unstable` - Pre-release builds ([`release` branch](https://github.com/ripple/rippled/tree/release))
```
mellery451, line 53: we don't mention the `unstable` and `nightly` repos in the rpm instructions, but they are similarly available there - should we include that info in rpm step 1 also ?

content/tutorials/manage-the-rippled-server/installation/rippled-1-3-0-migration-instructions.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Migration on CentOS or Red Hat Enterprise Linux (RHEL)
+
+Ripple's official RPM repository and instructions for using it have changed. To migrate from the old repository to the new one, complete the following steps:
+
+1. Stop the `rippled` server.
+
+        $ sudo systemctl stop rippled.service
+
+2. Replace Ripple's old yum repo file with the new one:
```
mellery451: what do you think about an additional step here that is `rpm --erase=ripple-repo`? The old step 1 of rpm was to install that repo package which we are no longer supporting and it will probably conflict with step 2 here. This cleanup step is only relevant in the 1.2 to 1.3 upgrade and probably could be removed in future revisions.

content/tutorials/manage-the-rippled-server/installation/update-rippled-automatically-on-linux.md
```diff
@@ -0,0 +1,17 @@
+# Update Automatically on Linux
+
+On Linux, you can set up `rippled` to automatically upgrade to the latest version with a one-time `cron` configuration. Ripple recommends enabling automatic updates if possible.
+
+These instructions assume you have already installed `rippled` [from the `yum` repository (CentOS/RedHat)](install-rippled-on-centos-rhel-with-yum.html) or [using `apt` (Ubuntu/Debian)](install-rippled-on-ubuntu.html).
+
+**Warning:** As of version 1.3.0, the install and upgrade process has changed. Automatic updates **will not** upgrade from `rippled` 1.2.4 or lower to `rippled` 1.3.0 or higher. If you have `rippled` v1.2.4 or lower installed, follow the [rippled 1.3.0 migration instructions](rippled-1-3-0-migration-instructions.html) first. After performing the migration, automatic updates should work again.
```
mellery451: I think this is not strictly true since we are planning on pushing a one-time (and final) update to the ripple-repo package that switches existing auto-update users over to the new repositories and thus they will get the 1.3 update. That said, the manual upgrade steps you've outlined makes things cleaner and allows the users to remove the ripple-repo package.

mellery451 (commented):
left a few small suggestions.

content/tutorials/manage-the-rippled-server/installation/install-rippled-on-ubuntu.md
mDuo13: I would like to, but I couldn't come up with as elegant a way to summarize them

content/tutorials/manage-the-rippled-server/installation/rippled-1-3-0-migration-instructions.md
mDuo13: I'd have to test that it works as intended but I think that's a good idea. The step here overwrites the repo file that was there before (note the lack of `-a` on the `tee` command) but if you were to later remove the (old) repo package it might erase the (updated) file, so probably best to do it as you mentioned and remove the file first.

Will test this and see what happens

content/tutorials/manage-the-rippled-server/installation/update-rippled-automatically-on-linux.md
mDuo13: I'll have to dig more into how that will work for users using the existing auto-update script...

ddibble-rip (commented):
Might want to take the word "latest" out of this sentence:

The latest version of Boost available in the Homebrew repositories isn't new enough, so you must install Boost manually:

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-macos.md
```diff
@@ -24,7 +24,7 @@ For development purposes Ripple recommends running `rippled` as your own user, n
 
         $ brew install git cmake pkg-config protobuf openssl ninja
 
-0. Install Boost 1.67.0. `rippled` 1.2.0 is compatible with Boost 1.67. <!--#{ no boost@1.67 formula, so must manually compile and install. will eventually upgrade to boost 1.68, but needs testing first }# -->
+0. Install Boost 1.67.0 or higher. `rippled` 1.3.0 is compatible with Boost 1.67.0 and higher. The latest version of Boost available in the Homebrew repositories isn't new enough, so you must install Boost manually:
```
ddibble-rip, line 27: Might want to take the word "latest" out of this sentence:

The latest version of Boost available in the Homebrew repositories isn't new enough, so you must install Boost manually:

content/tutorials/manage-the-rippled-server/installation/install-rippled-on-ubuntu.md
```diff
@@ -0,0 +1,88 @@
...
+        $ apt-key finger
+
+    The output should include an entry for Ripple such as the following:
+
+        pub   rsa3072 2019-02-14 [SC] [expires: 2012-02-13]
+              C001 0EC2 05B3 5A33 10DC 90DE 395F 97FF CCAF D9A2
+        uid           [ unknown] TechOps Team at Ripple <techops+rippled@ripple.com>
+        sub   rsa3072 2019-02-14 [E] [expires: 2021-02-13]
+
+    In particular, make sure that the fingerprint  matches. (In the above example, the fingerprint is on the second line, starting with `C001`.)
```
ddibble-rip: Extra space between "fingerprint" and "matches"

content/tutorials/manage-the-rippled-server/installation/update-rippled-automatically-on-linux.md
```diff
@@ -0,0 +1,17 @@
...
+
+To set up automatic updates, complete the following steps:
+
+1. Check that `/opt/ripple/etc/update-rippled-cron` exists. If it does not, update manually ([CentOS/Red Hat](update-rippled-manually-on-centos-rhel.html) or [Ubuntu/Debian](update-rippled-manually-on-ubuntu.html)).
+
+2. Create a symlink in your `cron.d` folder to the `/opt/ripple/etc/update-rippled-cron` config file:
+
+        $ sudo ln -s /opt/ripple/etc/update-rippled-cron /etc/cron.d/
+
+    This cron configuration runs a script to update installed `rippled` package within an hour of each new release. To reduce the chance of outages from all servers updating simultaneously, the script delays the update for a random number of minutes, up to 59.
```
ddibble-rip: Should you add the word "any" here?

This cron configuration runs a script to update **any** installed `rippled` package within an hour of each new release.

content/tutorials/manage-the-rippled-server/installation/rippled-1-3-0-migration-instructions.md
```diff
@@ -0,0 +1,74 @@
+# rippled v1.3.0 Migration Instructions
+
+This document describes the migration process for upgrading from `rippled` 1.2.4 or earlier to `rippled` v1.3.0 or later. This is necessary because the `rippled` install process has changed as of version 1.3.0.
```
ddibble-rip: What is "this"? 

This migration process

content/tutorials/manage-the-rippled-server/installation/rippled-1-3-0-migration-instructions.md
```diff
@@ -0,0 +1,74 @@
...
+
+3. Fetch the latest repo updates:
+
+        $ sudo yum -y update
+
+4. Install the new `rippled` package:
+
+        $ sudo yum install rippled
+
+    Version 1.3.0 does not require any changes to your configuration files (`rippled.cfg` and `validators.txt`). This update procedure leaves your existing config files in place.
```
ddibble-rip: Small nit, but you've been referring to configuration files everywhere else as "config files"

ddibble-rip (approved):
I've reviewed the changes. None of my comments need to be fixed necessarily for the pull request.

content/tutorials/manage-the-rippled-server/installation/install-rippled-on-ubuntu.md
```diff
@@ -0,0 +1,88 @@
...
+
+
+## Prerequisites
+
+Before you install `rippled`, you must meet the [System Requirements](system-requirements.html).
+
+
+## Installation Steps
+
+1. Update apt repositories:
```
ddibble-rip: Should it be 
1. Update APT repositories
?

content/tutorials/manage-the-rippled-server/installation/update-rippled-manually-on-ubuntu.md
```diff
@@ -1,35 +1,25 @@
...
+**Caution:** To upgrade from `rippled` 1.2.x to 1.3.0 or higher on Ubuntu Linux, you should follow the [1.3.0 migration instructions](rippled-1-3-0-migration-instructions.html). The following instructions assume you have already installed the native APT package provided with versions 1.3.0 and up.
 
-2. Download the latest `rippled` package:
+**Tip:** To perform these steps all at once, you can run the `/opt/ripple/bin/update-rippled.sh` script, which is included with the `rippled` package and is compatible with Ubuntu and Debian starting with `rippled` version 1.3.0.
 
-        $ yumdownloader --enablerepo=ripple-stable --releasever=el7 rippled
+To update manually, complete the following steps:
 
-3. Verify the signatures on the downloaded packages:
+1. Update apt repositories:
```
ddibble-rip: Should it be 

1. Update APT repositories
?

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-macos.md
mDuo13: I see where you're coming from, but Homebrew actually contains several versions of Boost, none of which are new enough.

content/tutorials/manage-the-rippled-server/installation/update-rippled-automatically-on-linux.md
mDuo13: Good catch, but I used "the" instead of "any".

content/tutorials/manage-the-rippled-server/installation/rippled-1-3-0-migration-instructions.md
mDuo13: Good point. Turns out we had ~14 instances of "configuration file" and ~74 instances of "config file". The latter is shorter and, although it's less formal, is pretty ubiquitous overall (outside our docs) so I've gone and replaced all instances of "configuration file" with "config file" for consistency.

content/tutorials/manage-the-rippled-server/installation/install-rippled-on-ubuntu.md
mDuo13: Decided less is more here and went with just "Update repositories" instead.

mDuo13:
Oh, by the way, @alloyxrp, to my knowledge we don't discourage installing `rippled` on any version of Ubuntu that's currently supported by Canonical. I'm happy to leave the decision of which OS version to use up to server admins as much as possible.

alloynetworks:
Agreed. I can't remember if we did the tests on 19.04 though. @mellery?

mellery:
> 
> 
> Agreed. I can't remember if we did the tests on 19.04 though. @mellery?

I am not @mellery451

mellery451:
yes - we do a basic smoketest on 19.04:  https://github.com/mellery451/rippled/blob/develop/Builds/containers/gitlab-ci/pkgbuild.yml#L192 -- and we can add next LTS when it's available.

mDuo13:
Only broken link is for the 1.3.0 release tag.

PR #616, opened 2019-07-04
Consistent use of terminology in admin method blurbs

mDuo13:
The inconsistency was pointed out by translators. This takes the more common phrasing and applies it to the other methods.


mDuo13:
The broken links are benign.

(Also, I really should figure out why the Data API build sometimes fails on the type error, because it should not be getting non-array values there. I wonder—could some of the build machines be using an outdated version of BeautifulSoup?)

PR #634, opened 2019-07-25
Missing install keyword

alloynetworks:
`apt -y apt-transport-https ca-certificates wget gnupg` changed to `apt -y install apt-transport-https ca-certificates wget gnupg`


mDuo13:
Good catch, thanks.

PR #621, opened 2019-07-09
rippled v1.3.1 Docs

mDuo13:
- Updated install/upgrade instructions, including migration information for 1.2.x to 1.3.x
- fixMasterKeyAsRegularKey amendment description
- log rotation stuff

This stuff has all been reviewed individually. Link checker will fail until the 1.3.1 tag exists in the `rippled` repository.

_Edit: Changed to 1.3.1 instead of 1.3.0._

PR #617, opened 2019-07-04
Add fixMasterKeyAsRegularKey amendment details

mDuo13:
Adds the fixMasterKeyAsRegularKey amendment, as described in https://github.com/ripple/rippled/pull/2873

PR #635, opened 2019-07-25
Known Amendments: fixMasterKeyAsRegularKey

intelliot:
I tried to model this based on the other amendments on the page.


mDuo13:
This work is already done (in #617) and just awaiting deploy (in #621)

PR #636, opened 2019-07-26
Add service restart instruction for alien-to-apt migration

mellery451:
turns out we missed this in the alien to apt instructions as well (thanks @alloyxrp)


mDuo13 (approved):
:+1:

alloynetworks:
We do not support reload, do we? The doc says reload/restart.

PR #637, opened 2019-07-26
Fix residual ripple.com/build/ links

mDuo13:
Several of these links were still around, but were redirecting correctly until recently (due to a config change on ripple.com)

This fixes them to point directly to the correct source. I also updated a couple lines of text that I spotted near the problematic links.


ryangyoung (commented):
👍

PR #638, opened 2019-07-26
Add custom 404 page

mDuo13:
Builds a file we can then configure the server to use for serving 404s so they're more informative than the current (very empty) default nginx 404 page on xrpl.org.

### DIFF
--- a/tool/template-404.html
+++ b/tool/template-404.html
@@ -1,0 +1,23 @@
+{% extends "template-base.html" %}
+
+{% block left_sidebar %}
+{% set use_page = pages|selectattr("html", 'defined_and_equalto', "docs.html")|first %}
+{% include 'template-sidebar_nav.html' %}
+{% endblock %}
+
+{% block main %}
+  <article class="pt-3 p-md-3">
+    <h1>Not Found</h1>
+    <div class="content">
+        <p>Sorry, this page does not exist. Try looking in the <a href="/docs.html#full-doc-index">Full Documentation Index</a>, or you can search the site with Google:</p>
+        <form role="search" method="get" action="https://www.google.com/search">
+          <div class="form-inline">
+            <label class="sr-only" for="centersearch">Search site with Google...</label>
+            <input name="q" value="site:xrpl.org" type="hidden">
+            <input id="centersearch" name="q" type="text" class="form-control" placeholder="Search site with Google...">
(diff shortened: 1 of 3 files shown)


tool/template-sidebar_nav.html
```diff
@@ -1,107 +1,110 @@
+{% if use_page is undefined %}
+  {% set use_page = currentpage %}
```
mDuo13, line 2: The `use_page` thing provides a way to show the sidebar from a perspective other than the current page's position in the hierarchy.

tool/template-404.html
```diff
@@ -0,0 +1,23 @@
+{% extends "template-base.html" %}
+
+{% block left_sidebar %}
+{% set use_page = pages|selectattr("html", 'defined_and_equalto', "docs.html")|first %}
+{% include 'template-sidebar_nav.html' %}
```
mDuo13, line 5: This awkwardly doesn't actually show the headings for "Concept" / "Tutorial" / "Reference" above their component categories, but it does provide a reasonably useful & thorough list of pages. Actually fixing the thing to show those would be a fair bit harder (probably involving dozens of new lines in the sidebar_nav template) so I decided to go with something faster for the first iteration of this.

ryangyoung: That seems fine to me for the first go. We can always circle back later to polish it. (Famous last words, but I'm sure we'll encounter more broken links in the future.)

ryangyoung (approved):
👍 LGTM. Likely some things to polish here and there but this will definitely be better than the default nginx 404 page.

cancelself:
Who can merge this? and then there is a push job from ops?

cancelself:
i agree that something like this needs a pr (although i don't know why this didn't didn't get merged).

dpurdy:
https://ripplelabs.atlassian.net/wiki/spaces/DOCS/pages/78643238/So+You+re+Going+to+Contribute+to+the+Docs is pretty good.

mDuo13:
The broken link report is interesting. The new link _should_ work but it's reporting as broken. I don't know if that's a problem in the checker or the link itself. Will investigate a little further before merging this.

mDuo13:
Confirmed that the link works for real. If the link checker continues flagging it we'll have to fix the checker.

PR #643, opened 2019-07-30
Solving the confusion for pointing Boost environment variable in .bash_profile

yusufsahinhamza:
Solving the confusion and helping when adding Boost environment variable to .bash_profile

PR #645, opened 2019-07-30
Fix indentation on boost root command sample

mDuo13:
Fixes indentation from #643


mDuo13:
Blegh. The link checker is failing on the 404 page's "full-doc-index" link, which I've confirmed actually works in production. So I'll have to figure out why that is. But of course, that's off the topic of this PR.

PR #646, opened 2019-07-31
Update tx-request-escrowfinish-condition.json

ghost:
The transaction "EscrowFinish" identifying hash value from step 6 was different than the hash value used in tx method in step 8.


mDuo13 (approved):
Took me a bit to confirm this, but it looks good. Thanks!

PR #647, opened 2019-07-31
Update tx-response-escrowfinish-condition.json

ghost:
The hash value has been updated to reflect the hash value from step 6.


mDuo13 (approved):
This is basically a part of #646 and looks good

PR #648, opened 2019-07-31
Update use-payment-channels.md

ghost:
In steps 9 and 10 a PaymentChannelClaim Flag "tfClose" is mentioned (https://xrpl.org/paymentchannelclaim.html#paymentchannelclaim-flags) and yet a global flag (https://xrpl.org/transaction-common-fields.html#global-flags) is given when submitting a transaction in steps 9 and 10.


mDuo13 (changes requested):
The description and example in this PR are mistaken; the original is correct.

The original shows a `Flags` value of 2147614720, which is the bitwise-OR combination of tfFullyCanonicalSig (2147483648) and tfClose (131072). You should always use tfFullyCanonicalSig, so the original examples properly use _both_ flags. See the [flags field](https://xrpl.org/transaction-common-fields.html#flags-field) description for a detailed description of how the flags field is defined.

PR #649, opened 2019-07-31
Update getCreateTx.js

ghost:
The hash value from step 3 should be used. The hash is correct for the "websocket" example but different for RippleAPI and commandline. Both RippleAPI and commandline have a same hash though (but different from step 3).

### DIFF
--- a/content/_code-samples/checks/js/getCreateTx.js
+++ b/content/_code-samples/checks/js/getCreateTx.js
@@ -8,7 +8,7 @@ const api = new RippleAPI({server: 'wss://s.altnet.rippletest.net:51233'})
 api.connect().then(() => {
   console.log('Connected')
 
-  const tx_hash = "C0B27D20669BAB837B3CDF4B8148B988F17CE1EF8EDF48C806AE9BF69E16F441"
+  const tx_hash = "09D992D4C89E2A24D4BA9BB57ED81C7003815940F39B7C87ADBF2E49034380BB"
 
   return api.getTransaction(tx_hash)
 }).then(response => {


mDuo13:
I'm not sure this is correct. The three examples have different blob and hash values that are carried through all the steps. (Their exact details, including the account `Sequence`, are probably different.) So the RippleAPI examples consistently use the hash `C0B27...` while WebSocket examples consistently use the hash `09D99...` and commandline examples _should_ consistently use `07C3B2...`. (The response from step 3 for commandline is wrong, though. I think one of those was copy-pasted from the other.)

ghost:
Hi Rome,
Step 2/Example Response section: the  "tx_blob" : "120010...1F9C39" is identical in all 3 cases (RippleAPI, WebSocket, Commandline). However, the hash value is different in all 3 cases.
RippleAPI example has as a tx_hash "C0B27D...16F441".
WebSocket example has as a tx_hash "09D992...4380BB".
Finally, Commanline example has as a tx_hash "07C3B2...DDF408".
3 different hash values, see https://xrpl.org/send-a-check.html#example-request. 

Then is step3/example response: both websocket (not rippleAPI like I wrote above) and commandline have the same tx_hash "09D992...4380BB". For the Ripple API it's not even mentioned...

Step5/example request and thus example response: both websocket and commandline have the same tx_hash "09D992...4380BB". For the Ripple API it's "C0B27D...16F441".

For consistency I think that you should just pick one hash value and use it throughout entire doc for all 3 cases (RippleAPI, WebSocket, Commandline). The devs won't care, they'll get it but the non-technical folks from the #XRPCommunity will have some long pauses trying to figure it out.
Cheers,
C.

mDuo13:
The blobs are _not_ identical. They just start and end the same way.

RippleAPI blob (188 bytes):
```
12001022800000002400000002201B0077911368400000000000000C694000000005F5E100732103B6FCD7FAC4F665FE92415DD6E8450AD90F7D6B3D45A6CFCF2E359045FF4BB400744630440220181FE2F945EBEE632966D5FB03114611E3047ACD155AA1BDB9DF8545C7A2431502201E873A4B0D177AB250AF790CE80621E16F141506CF507586038FC4A8E95887358114735FF88E5269C80CD7F7AF10530DAB840BBF6FDF8314A8B6B9FF3246856CADC4A0106198C066EA1F9C39
```

WebSocket blob (221 bytes):

```
120010228000000024000000042E00000001501146060241FABCF692D4D934BA2A6C4427CD4279083E38C77CBE642243E43BE29168400000000000000C694000000005F5E100732103B6FCD7FAC4F665FE92415DD6E8450AD90F7D6B3D45A6CFCF2E359045FF4BB40074463044022071A341F911A8EF3B68399487CAF5BA3B59C6FE476B626698AEF044B8183721BC0220166053A859BD907251DFCCF34DD71202180EBABAE7098BB5903D16EBFC993C408114735FF88E5269C80CD7F7AF10530DAB840BBF6FDF8314A8B6B9FF3246856CADC4A0106198C066EA1F9C39
```

Commandline blob (227 bytes):

```
120010228000000024000000012A21FB3DF12E00000001501146060241FABCF692D4D934BA2A6C4427CD4279083E38C77CBE642243E43BE29168400000000000000C694000000005F5E100732103B6FCD7FAC4F665FE92415DD6E8450AD90F7D6B3D45A6CFCF2E359045FF4BB40074473045022100EB5A9001E14FC7304C4C2DF66507F9FC59D17FDCF98B43A4E30356658AB2A7CF02207127187EE0F287665D9552D15BEE6B00D3C6691C6773CE416E8A714B853F44FC8114735FF88E5269C80CD7F7AF10530DAB840BBF6FDF8314A8B6B9FF3246856CADC4A0106198C066EA1F9C39
```

ghost:
Unlike hash values that can be distinguished with the naked eye one should not try to do the same with the tx-blob which is what I did when scrolling through them while focusing only at the 3-4 first and last bytes. I should've placed them on top of each other to quickly see that they have different length (like you did :)). Apologies for the tx-blob confusion!

mDuo13:
OK, so I think regarding this specific tutorial, the correct change is to carry the unique hashes from the step 2 responses through the steps separately for each type. So all the WS examples should use the `09D...` hash, RippleAPI examples should use the `C0B..` one, and CLI examples should use the `07C...` one.

ghost:
Correct!

PR #663, opened 2019-07-31
Update tx-cancel-resp.txt

ghost:
hash value changed to be in line with step 2.


mDuo13:
For reference, the reason I closed this and the other related PRs is because I'd rather address the problem with a different approach as described in #699. I appreciate your contributions and will try to incorporate them as appropriate, but just changing the hashes in some examples introduces some problems because the transaction blobs will then be inaccurate.

PR #666, opened 2019-08-01
Update "Understanding Log Messages"

mDuo13:
- Add new sections for shard "file not found" error ( https://github.com/ripple/rippled/pull/3014 ) and "We are not running on the consensus ledger" warning
- Alphabetize error message list (excluding the "crashes" section which is a special case)
- Make a snippet for log messages that are benign during syncing, so the language can be updated in one place (esp. if/when I add a doc specifically on troubleshooting syncing issues)

Tip: It's probably easier to review the substantive changes in this PR by looking at 176d1f721894f5e0c6c63e70a5345e541233b025 since the diff for the full PR is confusing as a result of the alphabetization thing.


mDuo13:
!rebuild

ddibble-rip (approved):
LGTM, Rome.

miguelportilla (approved):
👍 LGTM @mDuo13 !

PR #667, opened 2019-08-01
Fix payment types table

### LINKED ISSUE #665: Payments page, Payment Types table broken
The new Types of Payments table at https://xrpl.org/payment.html is broken. The cause seems to be that the description column is missing a '|:------'. I considered fixing, but unfamiliar with dactyl and the markup sections like `Address` = `Destination`?.

mDuo13:
Fixes #665


mDuo13:
Only broken link is the `/docs.html#full-doc-index` on the 404 page (again) so this is good to merge

PR #668, opened 2019-08-05
Make 404 page path-independent

mDuo13:
When the 404 page is served from a path that includes slashes for
subfolders (for example, /transactions/ or /some/404/page), any relative
links and content references it has won't work. This change converts all
relative references to absolute references instead.

### DIFF
--- a/tool/template-404.html
+++ b/tool/template-404.html
@@ -1,23 +1,125 @@
-{% extends "template-base.html" %}
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width">
 
-{% block left_sidebar %}
-{% set use_page = pages|selectattr("html", 'defined_and_equalto', "docs.html")|first %}
-{% include 'template-sidebar_nav.html' %}
-{% endblock %}
+    <title>404 Not Found - XRP Ledger Dev Portal</title>
 
-{% block main %}
-  <article class="pt-3 p-md-3">
-    <h1>Not Found</h1>
-    <div class="content">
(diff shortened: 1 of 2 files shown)


intelliot (approved):
I wonder if there's a reason why we couldn't use absolute references in "template-base.html" as well?

mDuo13:
@intelliot The main reason I don't want to use absolute references across the board is that it makes it harder to preview the built pages locally just by opening the HTML files in a web browser (i.e. double clicking the html file in the finder, or using the browser's "File > Open" menu. Anything that results in loading via the `file://` protocol). As it stands, the 404 page won't work if you open it this way, but the other ones all will. But using absolute references across the board will completely break that use case.

mDuo13:
The build failure is a false positive (See: https://github.com/ripple/dactyl/issues/44 )

tool/template-404.html
```diff
@@ -1,23 +1,125 @@
...
+    <link href="/assets/vendor/bootstrap.css" rel="stylesheet" />
+    <link href="/assets/css/devportal.css" rel="stylesheet" />
+
+    <!-- Bootstrap JS -->
+    <script src="/assets/vendor/bootstrap.min.js"></script>
+
+    <!-- fontawesome icons -->
+    <link rel="stylesheet" href="/assets/vendor/fontawesome/css/font-awesome.min.css" />
+
+    <!-- Google Analytics -->
```
ryangyoung, line 33: Good idea to add an event for this page!

ryangyoung (approved):
This looks good, as far as I can tell. I tested it locally and, as @mDuo13 mentioned, it doesn't look right:

Screen Shot 2019-08-05 at 7 14 20 PM

Otherwise, the site built as normal and it looks like the changes should properly address the problem of redirecting links that include trailing slashes (`{anothersite}/foo` or `{anothersite}/foo/bar`).

Issue #670, opened 2019-08-08
Inconsistent documentation for validations stream subscription


jscottbranson:
The documentation for the `validations` stream subscription is inconsistent with regard to which validation messages are returned to the subscriber. 

https://xrpl.org/subscribe.html#subscribe:
> validations - Sends a message whenever the server receives a validation message, **regardless of if the server trusts the validator**.

https://xrpl.org/subscribe.html#validations-stream:
> The validations stream sends messages whenever it receives validation messages, also called validation votes, **from validators it trusts**.

The former message is correct, and all validations (not just trusted) are returned.

PR #671, opened 2019-08-11
Validation stream messages are from all validators

jscottbranson:
Edited the description of the validation stream. This resolves issue [https://github.com/ripple/ripple-dev-portal/issues/670](https://github.com/ripple/ripple-dev-portal/issues/670).


mDuo13:
Thanks for looking into this and taking care of it.

PR #678, opened 2019-08-28
Fix invalid message key

RareData:
The current message key example uses an account address, which is incorrect and can never be used to prepare a valid transaction.


mDuo13:
You are correct that an account address should not be used as the `MessageKey`.

Your updated example works but I wonder if perhaps it would be better for the example _not_ to use the (correct) master public key for the address in question, since you could use any public key as your message key (though of course you'd need the private key to decode any messages). And I think you probably don't want to use an ECDSA key (as this one is) since ECDSA is just for signing, not for encrypting messages.

Ah well. This is still an improvement, and I can revisit later.

PR #679, opened 2019-08-28
Tx Sender updates

mDuo13:
- Add 'initialize' button to reduce wasted testnet accounts
- Change default destination to be the correct faucet account

### DIFF
--- a/assets/js/tx-sender.js
+++ b/assets/js/tx-sender.js
@@ -87,7 +87,7 @@ const set_up_tx_sender = async function() {
       dataType: 'json',
       success: faucet_response,
       error: function() {
-        errorNotif("There was an error with the XRP Ledger Test Net Faucet. Reload this page to try again.")
+        errorNotif("There was an error with the XRP Ledger Testnet Faucet. Reload this page to try again.")
       }
     })
   })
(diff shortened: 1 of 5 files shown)


assets/js/tx-sender.js
```diff
@@ -57,23 +72,32 @@ const set_up_tx_sender = async function() {
...
-    }
+  $("#init_button").click((evt) => {
+    console.debug("Getting a sending address from the faucet...")
+    $.ajax({
+      url: FAUCET_URL,
+      type: 'POST',
+      dataType: 'json',
+      success: faucet_response,
+      error: function() {
+        errorNotif("There was an error with the XRP Ledger Test Net Faucet. Reload this page to try again.")
```
intelliot: Seems like we're changing "Test Net" to "Testnet"?

intelliot (approved):
LGTM.

Outside of this PR, if we're changing "Test Net" to "Testnet", then it may make sense to rename "xrp-test-net-faucet.html" to remove the `-` and create a redirect from the current URL.

mDuo13:
Yes, we are changing from "Test Net" to "Testnet".

I wanted to stick with two words because I think it's easier for non-native English speakers and better for translation than inventing a new compound word, but the usage within the industry is clear.

[Blockchains have "testnets"](https://duckduckgo.com/?q=%22testnet%22&t=h_&ia=web) (one word). [Not "test nets"](https://duckduckgo.com/?q=%22test+net%22&t=h_&ia=web) which are... mostly internet speed tests.

(Had a conversation with @warpaul where he recalled the days when advocated for calling it "block chain" tech. So it goes.)

I plan to update the faucet page text and change the faucet URL soon, but <s>not in this PR</s>. _Edit: On second thought, I can do it in this PR._

PR #682, opened 2019-08-29
Update info about building packages from source tree

mellery451:


mDuo13 (approved):
Slightly cleaned up the wording and formatting. Looks good now.

PR #686, opened 2019-09-06
Combine & minify CSS

mDuo13:
- Combine the CSS (`devportal.css`) and SCSS (`xrpl.scss`) to one file (`xrpl.scss`)
- Minify the output of the SCSS (`devportal.css` now includes the minified version of `bootstrap.css` along with the minified version of its previous contents)
- Add `font-display` values for improving usability when webfonts take a while to load.
- Move some non-blocking assets to the bottom of the file so the browser can start rendering sooner (don't want to move everything though because having jQuery first lets us more easily embed JavaScript stuff in the md, as in the "Send XRP" tutorial.)

Overall these changes should improve the speed of page loading throughout the portal.

**Tip:** Don't forget to refresh pages if they look weird or broken after loading them with this change. Breakages are probably caused by a mix of cached and non-cached content.


intelliot (approved):
LGTM. I built the docs and looked through several pages and did not see any visual difference in the new vs. the old. 👍

PR #688, opened 2019-09-10
Update configure-history-sharding.md

jamesneimanconsulting:
@ryangyoung This is the supplemental writing example you requested.
~James

Notes: @ryangyoung 
- Since the scope of the content changed from just configuring to configuring or programming, I removed the term 'Configuring' from the title.
- I was on the fence about using command line shell syntax, but observed that the multi-tabbed code examples using Websocket and JSON-RPC seem to be prevalent in your documentation.

### DIFF
--- a/content/tutorials/manage-the-rippled-server/configuration/configure-history-sharding.md
+++ b/content/tutorials/manage-the-rippled-server/configuration/configure-history-sharding.md
@@ -1,12 +1,22 @@
-# Configure History Sharding
+# History Sharding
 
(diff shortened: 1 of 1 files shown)


mDuo13:
There are some fundamental misconceptions behind the steps of this tutorial. (For example, you can't use the output of `crawl_shards` to figure out where to point `download_shards` at—much as that would _seem_ to make sense, it's just not true.)

Going to close this one since I think it'd be best to tackle this issue a different way.

PR #689, opened 2019-09-11
Update for ripple-lib 1.3.3

intelliot:
Also updated `github_forkurl`

### DIFF
--- a/dactyl-config.yml
+++ b/dactyl-config.yml
@@ -48,11 +48,11 @@ targets:
         display_name: XRP Ledger Dev Portal
         # These github_ fields are used by the template's "Edit on GitHub" link.
         #  Override them with --vars to change which fork/branch to edit.
-        github_forkurl: https://github.com/ripple/ripple-dev-portal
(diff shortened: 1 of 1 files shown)


dactyl-config.yml
```diff
@@ -48,11 +48,11 @@ targets:
         display_name: XRP Ledger Dev Portal
         # These github_ fields are used by the template's "Edit on GitHub" link.
         #  Override them with --vars to change which fork/branch to edit.
-        github_forkurl: https://github.com/ripple/ripple-dev-portal
+        github_forkurl: https://github.com/ripple/xrpl-dev-portal
         github_branch: master
         recently_updated:
             - html: rippleapi-reference.html
-              date: 2019-06-06
+              date: 2019-09-10
```
mDuo13, line 55: Nice of you to update this, but we actually removed the part of the templates that used this (it was too complicated to maintain) so it wasn't necessary. Actually, I should remove the whole `recently_updated` list for that reason.

mDuo13 (approved):
:+1:

PR #694, opened 2019-10-02
Add XRP-API Reference (Auto-generated)

mDuo13:
Requires [Dactyl v0.9.0](https://github.com/ripple/dactyl/releases/tag/v0.9.0) **(New!)**

- Adds the XRP-API reference README and XRP-API definition, both remotely sourced from the [XRP-API repo](https://github.com/ripple/xrp-api/)
- Adds "custom" templates for OpenAPI spec parsing.
    These templates are the same as the ones built into Dactyl v0.9.0, except they have a warning plastered in that XRP-API is early alpha software. We'll be able to expand and improve on these templates without always needing to update Dactyl, though.
- Fixes PDF styling slightly so you can make a decent-looking standalone XRP-API PDF spec.
- Adds a target for building just the XRP-API reference (the `index.html` is a little wonky. This is mainly useful for just building the XRP-API reference only if you want to update it.)

### DIFF
--- a/dactyl-config.yml
+++ b/dactyl-config.yml
@@ -55,6 +55,7 @@ targets:
         display_name: XRP API Reference
         github_forkurl: https://github.com/ripple/xrpl-dev-portal
         github_branch: master
+        no_cover: true
 
     # Data API target for porting changes to the README in the upstream repo
     # Intended for use in markdown (--md) mode.
--- a/dactyl-config.yml
+++ b/dactyl-config.yml
@@ -2833,7 +2833,7 @@ pages:
     # XRP-API docs (auto-generated)---------------------------------------------
 
     -   name: XRP-API
-        md: https://raw.githubusercontent.com/intelliot/xrp-api/master/README.md
+        md: https://raw.githubusercontent.com/ripple/xrp-api/master/README.md
         html: xrp-api.html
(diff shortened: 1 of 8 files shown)


intelliot (approved):
This is great! Thank you so much for putting this together.

One potential downside is that seeing the whole API requires clicking around quite a bit. I have a soft spot in my heart for the one-long-page approach of ripple-lib/RippleAPI. But there are advantages/disadvantages to both.

intelliot:
By the way, what's the command to generate the PDF and where is the PDF made available?

mDuo13:
The PDF isn't currently published, but you can make a PDF yourself like this:

```
dactyl_build -t xrp-api-only --pdf
```

It's been on my backlog for a while to set up automatic PDF builds for specific subsets of the site. We have the server license for Prince to do it, but I just haven't had time to put the infrastructure in place.

And yeah, I know what you mean about the one-long-page approach. It definitely has its charm, but the drawbacks are big (long page load, poor performance on weaker devices, easy to lose your place, things jump around as stuff loads in, etc.) so I didn't want to commit to it always being a one-pager.

In the future, I could look into extending the generator to support a single-page mode. There's certainly plenty of work to be done to improve it (both in the templates and on the Dactyl side).

PR #701, opened 2019-10-07
update tutorial to add devnet

ryangyoung:


content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
```diff
@@ -1,14 +1,14 @@
...
-**Note:** The XRP Test Net ledger and balances are reset on a regular basis.
+**Note:** The XRP Testnet and Devnet ledger and balances are reset on a regular basis.
 
-To connect your `rippled` server to the XRP Test Net, set the following configurations:
+To connect your `rippled` server to the XRP Testnet or Devnet, set the following configurations:
 
 1. In your `rippled.cfg` file:
 
-    a. Uncomment the following section, as follows:
+    a. Uncomment the following section and add either the [Testnet or Devnet address](xref: xrp-testnet-faucet.md) (this example uses the Testnet address):
```
mDuo13: I think it would be better to provide separate samples for each here. Easy to miss "this example uses the Testnet address" if you're skimming.

mDuo13 (changes requested):
Would like to see one thing changed before merging this. Otherwise looks good.

content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
ryangyoung: Separated the example commands there as well in the verification section (step 4).

content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
```diff
@@ -1,23 +1,30 @@
...
+To connect your `rippled` server to the XRP Testnet or Devnet, set the following configurations:
 
 1. In your `rippled.cfg` file:
 
-    a. Uncomment the following section, as follows:
+    a. To connect to the [Testnet](xref: xrp-testnet-faucet.md), uncomment the following section and add:
 
         [ips]
-        r.altnet.rippletest.net 51235
+        s.altnet.rippletest.net 51233
```
ryangyoung: Providing the `wss` port here but to [verify the connection, we use an rpc method](https://github.com/ripple/xrpl-dev-portal/pull/701/files#diff-de1e359e21d07b051bde0c873c1874b5R56). Should we use `51234` here as well?

mDuo13: The port **should remain `51235`** because the `[ips]` block wants the **Peer protocol** port. And using the JSON-RPC port later is correct as well.

Actually, now that I see it, we should update [Get Started with the `rippled` API](https://xrpl.org/get-started-with-the-rippled-api.html) to use the default port for the Testnet public server (it works; I tested it), which would make it consistent with the other servers.

For reference, there are several ports Ripple tends to use for XRPL servers, some of which are configured by default, including:

- 51233 - WebSocket over TLS, public (WSS)
    Not enabled in the default config, but configured on Ripple's public servers
- 51234 - JSON-RPC public (HTTP)
- 51235 - peer protocol (RTXP)
    The [Peer Crawler](https://xrpl.org/peer-crawler.html) also uses this port. It uses HTTPS with a self-signed certificate.
- 5005 - JSON-RPC admin; not exposed for public servers. (HTTP)
- 6006 - unencrypted WebSocket admin; not exposed for public servers. (WS)
- 2459 - Newly [registered with the IANA as an XRPL port](https://github.com/ripple/rippled/issues/3037) for use with the peer protocol, but not yet in use.
- 443 - **Default TLS port** used by most applications when you don't specify what port to connect on; on Ripple's public XRPL servers this accepts encrypted WebSocket traffic. (WSS) 
    Basically Ripple's public servers forward this to 51233.
    In other contexts, this is the default port for HTTPS traffic.
    Not enabled in the default config (certificates are necessary).
- 80 - Incidentally, this is the default port for unencrypted HTTP traffic. Ripple's public XRPL servers don't use it.

I believe it is also possible to configure `rippled` to speak multiple protocols over the same port, although I haven't experimented with that much myself.

mDuo13 (changes requested):

content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
ryangyoung: Reverted to `51235`. I didn't make the change in the other doc wasn't immediately clear to me where to make the change (plus that should probably be its own PR anyway).

content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
```diff
@@ -1,23 +1,30 @@
...
-**Note:** The XRP Test Net ledger and balances are reset on a regular basis.
+**Note:** The XRP Testnet and Devnet ledger and balances are reset on a regular basis.
 
-To connect your `rippled` server to the XRP Test Net, set the following configurations:
+To connect your `rippled` server to the XRP Testnet or Devnet, set the following configurations:
 
 1. In your `rippled.cfg` file:
 
-    a. Uncomment the following section, as follows:
+    a. To connect to the [Testnet](xref: xrp-testnet-faucet.md), uncomment the following section and add:
```
mDuo13: We don't have the xrefs filter enabled on XRPL.org so this needs to link `xrp-testnet-faucet.html` instead. (That's why Jenkins is unhappy.)

mDuo13 (changes requested):

mDuo13 (approved):

PR #704, opened 2019-10-11
Key Derivation Info

### LINKED ISSUE #624: Address encoding hash order
The figure in [Address encoding](https://xrpl.org/accounts.html#address-encoding) section says: "SHA256 of RIPEMD160". So RIPEMD160 first and SHA256 on the result.

The code below it says the opposite: SHA256 first and RIPEMD160 on the result.

As the address should be 160 bit long I suppose the RIPEMD should be applied secondly.

mDuo13:
Adds a whole bunch more information on how cryptographic keys are derived in the XRPL. Thanks especially to @seelabs, who wrote a [rough draft gist](https://gist.github.com/seelabs/22d3bcf852440c794d210d2eef576cea#generate-the-account-keypair-from-the-root-keypair).

- Adds step-by-step descriptions of how to derive private and public keys for both secp256k1 and Ed25519 types.
- Adds Python sample code showing key derivation for both types from various seed formats.
- Fixes #624 and related issues (DOC-2359, as reported by @nbougalis in the Ripple internal JIRA) about how the diagram incorrectly describes the process of deriving addresses.

**Note for reviewers:** The `.uxf` files are the source files for the diagrams. You do not need to review their internal XML formatting.


ryangyoung (approved):
LGTM, as far as I can understand it.

intelliot (approved):
LGTM

content/concepts/payment-system-basics/accounts/cryptographic-keys.md
```diff
@@ -116,6 +116,106 @@ The supported types of key pairs can be used interchangeably throughout the XRP
...
+
+    1. Concatenate the following in order, for a total of 20 bytes:
+        - The seed value (16 bytes)
+        - A "root sequence" value (4 bytes), as a big-endian unsigned integer. Use 0 as a starting value for the root sequence.
+
+    2. Calculate the [SHA-512Half][] of the concatenated (seed+root sequence) value.
+
+    3. If the result is not a valid secp265k1 private key, increment the root sequence by 1 and start over. [[Source]](https://github.com/ripple/rippled/blob/fc7ecd672a3b9748bfea52ce65996e324553c05f/src/ripple/crypto/impl/GenerateDeterministicKey.cpp#L103 "Source")
+
+        A valid secp256k1 key must not be zero, and it must be numerically less than the _secp256k1 modulus_ (also called the "group order"). The secp256k1 modulus is the constant value `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141`.
```
seelabs: "modulus" and "group order" are two different things. We want "group order", not "modulus". The number listed here is the "group order", so that's correct. However, I'd lose the references to "modulus".

content/concepts/payment-system-basics/accounts/cryptographic-keys.md
```diff
@@ -116,6 +116,106 @@ The supported types of key pairs can be used interchangeably throughout the XRP
...
+
+    2. Calculate the [SHA-512Half][] of the concatenated value.
+
+    3. If the result is not a valid secp265k1 private key, increment the key sequence by 1 and restart deriving the account's intermediate key pair.
+
+    4. With a valid secp256k1 private key, use the standard ECDSA public key derivation with the secp256k1 curve to derive the intermediate public key. (As always with cryptographic algorithms, use a standard, well-known, publicly-audited implementation whenever possible. For example, [OpenSSL](https://www.openssl.org/) has implementations of core Ed25519 and secp256k1 functions.)
+
+4. Derive the master public key pair by adding the intermediate public key to the root public key. Similarly, derive the private key by adding the intermediate private key to the root private key.
+
+    - An ECDSA private key is just a very large integer, so you can calculate the sum of two private keys by summing them modulo the secp256k1 modulus.
```
seelabs: Sum them modulo the secp256k1 "group order", not "modulus"

content/concepts/payment-system-basics/accounts/cryptographic-keys.md
```diff
@@ -116,6 +116,106 @@ The supported types of key pairs can be used interchangeably throughout the XRP
...
+    - [Seed definition](https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/Seed.h)
+    - [General & Ed25519 key derivation](https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/impl/SecretKey.cpp)
+    - [secp256k1 key derivation](https://github.com/ripple/rippled/blob/develop/src/ripple/crypto/impl/GenerateDeterministicKey.cpp)
+- In Python 3 in [this repository's code samples section]({{target.github_forkurl}}/blob/{{target.github_branch}}/content/_code-samples/key-derivation/key-derivation.py).
+- In JavaScript in the [`ripple-keypairs`](https://github.com/ripple/ripple-keypairs/) package.
+
+### Ed25519 Key Derivation
+[[Source]](https://github.com/ripple/rippled/blob/fc7ecd672a3b9748bfea52ce65996e324553c05f/src/ripple/protocol/impl/SecretKey.cpp#L203 "Source")
+
+[![Passphrase → Seed → Private Key → Prefix + Public Key](img/key-derivation-ed25519.png)](img/key-derivation-ed25519.png)
```
seelabs: In the code, we usually refer to "Secret Keys" instead of "Private Keys" (so the variable `pk` is a public key). Unless there's a reason to keep "Private Key" I'd match the terminology from the codebase.

content/concepts/payment-system-basics/accounts/cryptographic-keys.md
```diff
@@ -116,6 +116,106 @@ The supported types of key pairs can be used interchangeably throughout the XRP
...
+
+    4. With a valid secp256k1 private key, use the standard ECDSA public key derivation with the secp256k1 curve to derive the intermediate public key. (As always with cryptographic algorithms, use a standard, well-known, publicly-audited implementation whenever possible. For example, [OpenSSL](https://www.openssl.org/) has implementations of core Ed25519 and secp256k1 functions.)
+
+4. Derive the master public key pair by adding the intermediate public key to the root public key. Similarly, derive the private key by adding the intermediate private key to the root private key.
+
+    - An ECDSA private key is just a very large integer, so you can calculate the sum of two private keys by summing them modulo the secp256k1 modulus.
+
+    - An ECDSA public key is a point on the elliptic curve, so you should use elliptic curve math to sum the points.
+
+    **Tip:** You don't need any private keys to derive the master public key. You can do so using only the root public key.
```
seelabs: I'd probably leave this tip off. That's a useful fact when we're deriving a family of keys, and we have a machine that needs to calculate public keys from the family, but we don't want to put secret keys on that machine. However, we don't use that functionality at all.

seelabs (commented):
Left a couple comments. Note: I have not reviewed the sample python code, just the prose for the key derivation.

content/concepts/payment-system-basics/accounts/cryptographic-keys.md
mDuo13: I've seen the two terms used interchangeably before, but I'll try use "group order" consistently here. (That means I'm also changing the transaction malleability docs.)

content/_code-samples/key-derivation/key-derivation.py
```diff
@@ -0,0 +1,327 @@
...
+
+import ed25519
+import RFC1751
+import base58.base58 as base58
+
+XRPL_SEED_PREFIX = b'\x21'
+XRPL_ACCT_PUBKEY_PREFIX = b'\x23'
+XRPL_VALIDATOR_PUBKEY_PREFIX = b'\x1c'
+ED_PREFIX = b'\xed'
+SECP_MODULUS = 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141
```
seelabs: Similar to my comment in the prose, this is the group order, not the modulus.

seelabs: You can also get the order from `fastecdsa.curve.secp256k1.q`

content/_code-samples/key-derivation/key-derivation.py
```diff
@@ -0,0 +1,327 @@
...
+        Uses the XRPL's convoluted key derivation process to get the
+        secp256k1 master keypair for this seed value.
+        Saves the values to the object for later reference.
+        """
+
+        root_sec_i = secp256k1_secret_key_from(self.bytes)
+        root_pub_point = keys.get_public_key(root_sec_i, curve.secp256k1)
+        root_pub_b = compress_secp256k1_public(root_pub_point)
+        fam_b = bytes(4) # Account families are unused; just 4 bytes of zeroes
+        inter_pk_i = secp256k1_secret_key_from(root_pub_b+fam_b)
```
seelabs: I don't love that we're using `+` for string concatenation in a math-heavy function like this. I'd either:
1) Add a comment that we are concatenating bytes here, not adding points or
2) Use something like. `b''.join([root_pub_b, fam_b])` instead of `root_pub_b + fam_b`

Edit: I do get that you added a `_b` suffix to show type here, and maybe that's good enough. Your call really.

content/_code-samples/key-derivation/key-derivation.py
```diff
@@ -0,0 +1,327 @@
...
+    again.
+
+    Returns a secret key as a 32-byte integer.
+    """
+    seq = 0
+    while True:
+        buf = seed + seq.to_bytes(4, byteorder="big", signed=False)
+        h = sha512half(buf)
+        h_i = int.from_bytes(h, byteorder="big", signed=False)
+        if h_i > SECP_MODULUS or h_i == 0:
```
seelabs: We should check for greater than or equal to, not just greater than. I.e `h_i >= SECP_MODULUS` (although there is no conceivable way we'll find a hash that equals SECP_MODULUS)

content/_code-samples/key-derivation/key-derivation.py
```diff
@@ -0,0 +1,327 @@
...
+            # Not a valid secp256k1 key
+            seq += 1
+            continue
+        break
```
seelabs: Just return `h_i` here instead of breaking?

The `continue` and `break` control flow is a little unusual. A couple alternatives to consider for this loop:
1) Use `itertools.count()`. We'd have to import `itertools` at the top. So something like:
```
    for seq in itertools.count():
        buf = seed + seq.to_bytes(4, byteorder="big", signed=False)
        h = sha512half(buf)
        h_i = int.from_bytes(h, byteorder="big", signed=False)
        if h_i < SECP_MODULUS and h_i != 0:
            return h_i
```

Many people are going to look up what `itertools.count()` does, so while it's terse, it might not be best for educational code like this. We could also do something like:
```
    seq = 0
    while True:
        buf = seed + seq.to_bytes(4, byteorder="big", signed=False)
        h = sha512half(buf)
        h_i = int.from_bytes(h, byteorder="big", signed=False)
        if h_i < SECP_MODULUS and h_i != 0:
            return h_i
        seq += 1
```

content/_code-samples/key-derivation/key-derivation.py
```diff
@@ -0,0 +1,327 @@
+#!/usr/bin/env python3
```
seelabs: Putting a hyphen in the file name makes it more difficult to import (I had to look up how do it with `importlib`)

content/_code-samples/key-derivation/key-derivation.py
```diff
@@ -0,0 +1,327 @@
...
+    from random import SystemRandom
+    randbits = SystemRandom().getrandbits
+else:
+    from secrets import randbits
+
+from fastecdsa import keys, curve
+
+import ed25519
+import RFC1751
+import base58.base58 as base58
```
seelabs: I get an error err that `base58.base58` is not a package (if I change this to just `base58` it works OK).

content/_code-samples/key-derivation/key-derivation.py
```diff
@@ -0,0 +1,327 @@
...
+            self.derive_secp256k1_master_keys()
+        return self._secp256k1_pub
+
+    @property
+    def secp256k1_root_public_key(self):
+        """
+        33-byte secp256k1 root public key (bytes)
+        This is the public key used for validators.
+        """
+        if self._secp256k1_root_pub is None:
```
seelabs: `_secp256k1_root_pub` needs to be set to None in the constructor

seelabs (approved):
:+1:

content/_code-samples/key-derivation/base58/base58.py
```diff
@@ -0,0 +1,177 @@
+'''Base58 encoding
+
+Implementations of Base58 and Base58Check endcodings that are compatible
```
intelliot: `endcodings` -> `encodings` (or `encoders`)

content/_code-samples/key-derivation/base58/base58.py
```diff
@@ -0,0 +1,177 @@
...
+with the XRP Ledger.
+'''
+
+# This This code is adapted from the module by David Keijser at
```
intelliot: remove extra `This`

intelliot (changes requested):
spotted a couple minor typos

PR #710, opened 2019-10-28
Tip to change permissions on rippled.cfg

alloynetworks:
Make the rippled.cfg file readable only by the rippled user and root, with 0600

### DIFF
--- a/content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
+++ b/content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
@@ -114,7 +114,7 @@ On your validator:
 
       - The `server_state` value should be _**proposing**_.
 
-
+**Security Tip:** Change the permissions on your `rippled.cfg` file to be more restrictive. On Linux it is recommended to be `0600`. You can do this with `chmod 0600 rippled.cfg`
 
 ## 4. Connect to the network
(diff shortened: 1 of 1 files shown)


mDuo13:
Should probably mention that you want to do this as the `rippled` user. 

In general it feels kind of incomplete just add a mention about user permissions without actually clarifying the recommended model.

alloynetworks:
You cannot do it as the user rippled. That user doesn't have a $HOME or a `passwd` entry, so I don't quite understand what you're getting at.

alloynetworks:
Well you can su to that user from root, but that will land you in / . The packaging creates the files and directories in /opt/ripple as `rippled:rippled` so doing a chmod will leave it as that user.

mellery451:
@mDuo13 I don't think `chmod` cares who the invoking user is except that it must either owner or root to change perms - so this is just setting the mask to be strictly owner read/write. root seems like the right way to do that. The owner should remain `rippled` and we're not suggesting changing that.

mDuo13:
Yeah, I was thinking about cases other than the supported install, but I was mixed up on how chmod would work. Now that I understand better, I realize you're right that this is a good suggestion regardless.

I still think it would be a good doc to add to describe the recommended permissions setup as implemented by the supported packages. That would make it easier to build a similar setup on non-supported distros.

PR #715, opened 2019-11-01
Domain verification should not expose IP of validator

mayurbhandary:
The verified domain name should not resolve to the ip address of the validator.


mDuo13:
That's a good point, and worth the reminding. :+1:

PR #717, opened 2019-11-08
Update ripple-lib dependency to 1.4.1

ryangyoung:


ryangyoung:
Build failed due to this broken link:

> 07:08:32 Broken remote link in out/peer-protocol.html to https://tools.ietf.org/html/rfc7230#section-6.7

PR #719, opened 2019-11-13
xpring.co -> xpring.io

craustin:
Fixing dead link.


ryangyoung:
LGTM. Thanks for the fix!

mDuo13:
Yeah, Xpring still needs to fix their redirects so that xpring.co goes to the right place, but this is a good thing to do regardless because we want to link directly to the preferred URL!

PR #726, opened 2019-11-15
Update for ripple-lib 1.4.2

intelliot:
Question: I wonder if there's a reason we couldn't have this just use what we have in ripple-lib's `master` branch. I push stable releases to `master` at the time of release. That would be https://raw.githubusercontent.com/ripple/ripple-lib/master/docs/index.md -- maybe using that URL would allow this to auto-update?

### DIFF
--- a/dactyl-config.yml
+++ b/dactyl-config.yml
@@ -5480,7 +5480,7 @@ pages:
     # --------------- end "XRP-API" section --------------------------------
 
     -   name: RippleAPI Reference # name is required for remote-sourced files
-        md: https://raw.githubusercontent.com/ripple/ripple-lib/1.4.1/docs/index.md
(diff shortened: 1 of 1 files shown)


mDuo13:
There's no strong reason we couldn't do it that way, but:

1. We'd still have to manually push a deploy to xrpl.org after any update to ripple-lib.
2. Dactyl uses the URL to know what version to put in the "Updated for version \[whatever\]" line at the top of the page. So we'd have to drop that.

So as it stands, this step is a convenient reminder to do the manual deploy. It might be possible to set up an automatic push to xrpl.org whenever there's an update to ripple-lib master, but that's a level of automation I don't think Ripple's ops team has time to set up yet.

mDuo13 (approved):

PR #729, opened 2019-11-23
Update trust.md

christopheroye:
Please review.


mDuo13 (changes requested):
Adding an extra column for the "Optional" value is not consistent with the style used throughout the documentation, so I don't think it's helpful here.

Also, IMO, "is noted treating" doesn't read like good English to me. I like the original better.

I appreciate you taking the time to open this, but unfortunately I don't think this changeset provides an improvement over the existing doc.

christopheroye:
Thank you for your feedback. This was submitted as a sample assessment performing a PR.

Suggesting the "Optional" tab to document format for better end-user readability. I understand it's not a part of the existing style guide.

fix: "_is noted **for** treating_", not "noted for treating"... any particular reason for using shorthand instead of a term similar such as noted in this case?

PR #730, opened 2019-11-27
DeletableAccounts concept updates & style updates

mDuo13:
- Update descriptions of account sequence numbers to clarify they don't start at 1 w/ deletable accounts
- Standardize on use of "ledger index" instead of "sequence number" for ledgers, where possible
- Update some old docs to add links and rephrase to match current style guidelines in a few nearby places.


content/_snippets/rippled-api-links.md
```diff
@@ -18,6 +18,7 @@
 [Internal Type]: serialization.html
 [内部の型]: serialization.html
 [Ledger Index]: basic-data-types.html#ledger-index
+[ledger index]: basic-data-types.html#ledger-index
```
scottschurr, line 21: Just flagging this so you can decide if it looks right.  The two adjacent entries are _almost_ identical except for the initial capitalization.

content/concepts/payment-system-basics/accounts/accounts.md
```diff
@@ -7,7 +7,7 @@ An "Account" in the XRP Ledger represents a holder of XRP and a sender of [trans
     **Note:** The XRP community has [proposed](https://github.com/xrp-community/standards-drafts/issues/6) (and developed a [codec](https://github.com/xrp-community/xrpl-tagged-address-codec) to support) a new **X**-address format that exchanges and wallets could use instead of [destination tags](https://xrpl.org/source-and-destination-tags.html). These "packed" addresses start with an `X` instead of an `r`. For more information, see the [XRPL 𝗫-address format](https://xrpaddress.info/) site.
 
 - An **XRP balance**. Some of this XRP is set aside for the [Reserve](reserves.html).
-- A **sequence number**, starting at 1 and increasing with each transaction sent from this account. No transaction can be included in a ledger unless the transaction's sequence number matches its sender's next sequence number.
+- A **sequence number**, which increases by 1 with each transaction sent from this account. No transaction can be included in a ledger unless the transaction's sequence number matches its sender's next sequence number. (See also: [Basic Data Types: Account Sequence](basic-data-types.html#account-sequence).)
```
scottschurr: Just for thought, I wonder whether the "next" in "sender's next sequence number" reduces clarity.  Is there maybe a way to say something like, "For a transaction to be applied to the ledger, the sequence in the transaction must match the sequence in the account root.  If the transaction is applied to the ledger, then the sequence in the account root increases by 1."  But better than that, cuz you're a professional...

scottschurr (approved):
I feel like I must not have done a very good review.  I only spotted a couple of things to comment on.  Your call on both of them.  :+1:

content/_snippets/rippled-api-links.md
mDuo13: Yep, that's (unfortunately) necessary. This page defines a bunch of reference links that can be used in any page that includes it, but by the nature of Markdown, the link text is case-sensitive. In some contexts we want a title case version and in some contexts we don't, so they're both here.

content/concepts/payment-system-basics/accounts/accounts.md
mDuo13: On reading this, I realized that this description was putting the mechanism before the rationale. Changed it to this:

> A **sequence number**, which helps make sure any transactions this account sends are applied in the correct order and only once each. To execute a transaction, the transaction's sequence number and its sender's sequence number must match. Then, as part of applying the transaction, the account's sequence number increases by 1.

mDuo13:
(Rebased in the hopes that it fixes some of the broken links.)

PR #728, opened 2019-11-21
AccountDelete transaction reference & related docs

mDuo13:
- New `AccountDelete` transaction definition
- New transaction result codes (`tefTOO_BIG` may get turned into a `ter` though)
- Adjust concept information a little to provide rules for account deletion

PR #731, opened 2019-11-27
Deletable Accounts Transaction Cost

mDuo13:
Honestly I thought this update would be bigger but the rationale was covered in the previous PR (#728) so in the end I only had to add the entry to the special costs table.


scottschurr (approved):
:+1:

PR #734, opened 2019-11-28
DeletableAccounts: clarify re-creating accounts

mDuo13:


content/concepts/payment-system-basics/accounts/accounts.md
```diff
@@ -73,6 +73,8 @@ To be deleted, an account must meet the following requirements:
 - The account must own fewer than 1000 objects in the ledger.
 - The [AccountDelete transaction][] must pay a special [transaction cost][] equal to at least the [owner reserve](reserves.html) for one item (currently 5 XRP).
 
+After an account has been deleted, it can be re-created in the ledger through the normal method of [creating accounts](#creating-accounts). An account that has been deleted and re-created is no different than an account that has been created for the first time.
```
scottschurr, line 76: In general looks good.

It might be nice to say something like, "the account has none of its old settings" or something like that.  I'm expecting folks to be quite surprised by the way recreated accounts work.  Also, you know best, but I wonder if we could do without the hyphen in "re-created"?

PR #739, opened 2019-12-04
subscribe: new master_key field in validations stream

mDuo13:
Documents https://github.com/ripple/rippled/pull/3083

Wish I had a "real" example for a subscription message, but I wasn't able to catch one at the same time as the `amendments` field (which is only on flag ledgers...)


seelabs (approved):
:+1:

PR #740, opened 2019-12-04
fix: transfer fee example 1%

alexprut:
Fixing the transfer fee example, 1% of 2€ is 0,02

0,5% is confusing when looking at the example that follows where the fee is 1%


mDuo13:
That's a good point. Thank you!

mDuo13 (approved):

PR #741, opened 2019-12-04
Typo of the Devnet server IP address

ksato9700:
Fix a typo of the IP address of `s.devnet.rippletest.net`.


mDuo13 (approved):
Wow, good catch there. At first I wondered, "Wait, maybe the old IP address was right at some point?" but then I realized each block can't be more than 255, so it could _only_ be a typo.

PR #744, opened 2019-12-12
Transaction Reference edits: deletable account edge cases

mDuo13:
Add a couple links to the existing descriptions, and clarify the edge case for EscrowFinish to deleted accounts.


content/references/rippled-api/transaction-formats/transaction-types/escrowfinish.md
```diff
@@ -38,6 +38,8 @@ Any account may submit an EscrowFinish transaction.
 
 **Note:** The minimum [transaction cost](transaction-cost.html) to submit an EscrowFinish transaction increases if it contains a fulfillment. If the transaction has no fulfillment, the transaction cost is the standard 10 drops. If the transaction contains a fulfillment, the transaction cost is 330 [drops of XRP][] plus another 10 drops for every 16 bytes in size of the preimage.
 
+In [non-production networks](parallel-networks.html), it may be possible that the destination of an escrow has been [deleted](accounts.html#deletion-of-accounts) after the escrow was created. In this case, an attempt to finish the escrow fails with the result `tecNO_TARGET`, but the escrow object remains unless it has expired normally. If another payment re-creates the destination account, the escrow can be finished successfully. This situation is only possible if the escrow was created before the [fix1523 amendment](known-amendments.html#fix1523) became enabled. No such escrows exist in the production XRP Ledger, so this edge case is not possible on the production XRP Ledger.
```
LoisRP: Suggest:
In [non-production networks](parallel-networks.html), the destination of an escrow might be [deleted](accounts.html#deletion-of-accounts) after the escrow was created. 

Questions and comments:
"the escrow can be finished successfully" - Does this happen automatically when the destination account is re-created? 

"This situation is only possible ..." - Does this mean that no future escrows will have this issue? What is the expected behavior for future escrows if the destination account is re-created?
Do you mean that it is not possible to delete a destination after fix1523?

content/references/rippled-api/transaction-formats/transaction-types/paymentchannelclaim.md
```diff
@@ -47,7 +47,7 @@ The **destination address** of a channel can:
...
+If the [DeletableAccounts amendment](known-amendments.html#deletableaccounts) :not_enabled: is enabled _and_ the [fixPayChanRecipientOwnerDir amendment](known-amendments.html#fixpaychanrecipientownerdir) :not_enabled: was not enabled when the payment channel was created, it is possible that the destination of the payment channel has been [deleted](accounts.html#deletion-of-accounts) and does not currently exist in the ledger. If the destination has been deleted, the source account cannot send XRP from the channel to the destination; instead, the transaction fails with `tecNO_DST`. (And, of course, the deleted account cannot send any transactions to affect the account.) Other uses of this transaction type are unaffected when the destination account has been deleted, including adjusting the channel expiration, closing a channel with no XRP, or removing a channel that has passed its expiration time.
```
LoisRP: Suggestion and question:
(And, of course, the deleted account cannot send any transactions to affect the account.)
"affect the account" - is this referring to the deleted account?
A deleted account cannot send any transactions.

content/references/rippled-api/transaction-formats/transaction-types/escrowfinish.md
scottschurr: @mDuo13 Your explanation looks good to me.  But I've been steeped in this problem for months.

@LoisRP, I assume you're suggesting that @mDuo13's, doc should answer your questions.  I'll let you and him make that call.  But here are answers to your questions.

> "the escrow can be finished successfully" - Does this happen automatically when the destination account is re-created?

No, it won't happen automatically when the destination is recreated.  The escrow finish will have to be re-attempted once the destination is recreated.  FWIW, in general that's how the ledger works.  The data in the ledger sits passively unless a transaction interacts with a piece of data on the ledger while that transaction is running.

> "This situation is only possible ..." - Does this mean that no future escrows will have this issue?

There was a change in the ledger behavior: fix1523.  For escrows created after fix1523 was enabled, the account that is the destination of the escrow cannot be deleted from the ledger.  That change to the ledger is why no future escrows will have the issue.  And the reason the issue will not be present is because the escrow destination cannot be deleted.

Does that help?  @mDuo13 can probably explain it better than I can.

content/references/rippled-api/transaction-formats/transaction-types/paymentchannelclaim.md
scottschurr: @LoisRP, good eyes!  The phrase should probably be, "And, of course, the deleted account cannot send any transactions to affect the payment channel."

scottschurr (approved):
:+1: Your changes look great to me.  There is one correction that @LoisRP pointed out that should probably be made (I missed it until @LoisRP pointed it out).  And I'm sure you and @LoisRP can negotiate which clarifications are appropriate.

PR #745, opened 2019-12-14
Offline Account Setup tutorial

mDuo13:
This tutorial walks through the process of setting up a new account in the XRP Ledger using an air-gapped, offline machine.

I don't have instructions and code samples for doing the same stuff with ripple-lib; I feel like that would be a good addition, but it's a bit too much for the scope of this task at the moment. Still, it should be possible to add some code samples for that eventually.

The sample transactions are validly formed (but the secret key is a placeholder, also used elsewhere in the docs, and not the actual secret for the sending account).

As it turns out, the most straightforward process for doing account setup is _basically_ the same with or without deletable accounts. There is an alternative process, described in [the deletable accounts draft](https://github.com/xrp-community/standards-drafts/issues/8), which reduces the number of round trips between the online and offline machines, but it adds a significant amount of complexity, so I opted not to cover that here. (For that version, you have to estimate what the current ledger index will be at the time you fund your account, then you prepare & use a bunch of no-op transactions to consume sequence numbers leading up to your first meaningful setup transaction. I decided that figuring out how many transactions to submit, tracking their outcome, and spending the transaction costs of those no-op transactions was not worth saving a round trip from the offline machine for instructional purposes.)

### DIFF
--- a/content/tutorials/manage-account-settings/offline-account-setup.md
+++ b/content/tutorials/manage-account-settings/offline-account-setup.md
@@ -1,6 +1,6 @@
 # Offline Account Setup Tutorial
 
-A highly secure [signing configuration](set-up-secure-signing.html) involves keeping an XRP Ledger [account](accounts.html)'s [cryptographic keys](cryptographic-keys.html) securely on an offline, air-gapped machine. After setting up this kind of configuration, you can sign a variety of transactions, transfer only the signed transactions to an online computer, and submit them to the XRP Ledger network without ever exposing your secret key to malicious actors online.
+A highly secure [signing configuration](set-up-secure-signing.html) involves keeping an XRP Ledger [account](accounts.html)'s [cryptographic keys](cryptographic-keys.html) securely on an offline, air-gapped machine. After setting up this configuration, you can sign a variety of transactions, transfer only the signed transactions to an online computer, and submit them to the XRP Ledger network without ever exposing your secret key to malicious actors online.
 
 **Caution:** Proper operational security is necessary to protect your offline machine. For example, the offline machine must be physically located where untrusted people cannot get access to it, and trusted operators must be careful not to transfer compromised software onto the machine. (For example, do not use a USB drive that was previously attached to a network-connected computer.)
 
@@ -29,6 +29,9 @@ Software options for signing on the XRP Ledger include:
 - [Install `rippled`]() from a package (`.deb` or `.rpm` depending on which Linux distribution you use) file, then [run it in stand-alone mode](rippled-server-modes.html).
 - Install [ripple-lib](rippleapi-reference.html) and its dependencies offline. The Yarn package manager, for example, has [recommended instructions for offline usage](https://yarnpkg.com/blog/2016/11/24/offline-mirror/).
 - See also: [Set Up Secure Signing](set-up-secure-signing.html)
+
+You may want to set up custom software to facilitate the process of constructing transaction instructions on the offline machine. For example, your software may track what [sequence number][] to use next, or contain preset templates for certain types of transactions you expect to send.
+
 
 ### {{n.next()}}. Generate cryptographic keys
 
@@ -118,15 +121,17 @@ Loading: "/etc/opt/ripple/rippled.cfg"
 
 ### {{n.next()}}. Enter the sequence number on the offline machine.
 
-Save the account's starting sequence number on the offline machine as its current sequence number. Whenever you sign a new transaction using the offline machine, you will use the current sequence number, then increase the current sequence number by 1.
+Save the account's starting sequence number on the offline machine. Whenever you prepare a transaction using the offline machine, use the saved sequence number, then increase the sequence number by 1 and save the new value.
 
-You can prepare several transactions in advance this way, then transfer the signed transactions to the online machine all at once and submit them. As long as each transaction is validly formed and pays a sufficient [transaction cost](transaction-cost.html), the XRP Ledger network should eventually include those transactions in validated ledgers, keeping the account's sequence number in the shared XRP Ledger in sync with the "current" sequence number you are tracking on the offline machine.
+You can prepare several transactions in advance this way, then transfer the signed transactions to the online machine all at once and submit them. As long as each transaction is validly formed and pays a sufficient [transaction cost](transaction-cost.html), the XRP Ledger network should eventually include those transactions in validated ledgers, keeping the account's sequence number in the shared XRP Ledger in sync with the "current" sequence number you are tracking on the offline machine. (Most transactions get a final, validated result within 15 seconds or less after being submitted to the network.)
+
+Optionally, save the current ledger index to the offline machine. You can use this value to choose an appropriate `LastLedgerSequence` value for upcoming transactions.
 
 
 
 ### {{n.next()}}. Sign initial setup transactions, if any.
 
-On the offline machine, prepare and sign transactions for configuring your account. The details depend on how you intend to you use your account. Some examples of things you might want to do include:
+On the offline machine, prepare and sign transactions for configuring your account. The details depend on how you intend to use your account. Some examples of things you might want to do include:
 
 - [Assign a regular key pair](assign-a-regular-key-pair.html) that you can rotate regularly.
 - [Require destination tags](require-destination-tags.html) so that users can't send you payments without tagging the reason they sent it or the customer it's intended for.
@@ -171,6 +176,8 @@ Loading: "/etc/opt/ripple/rippled.cfg"
 ```
 
 <!-- MULTICODE_BLOCK_END -->
+
+To ensure _all_ transactions have a final outcome within a limited amount of time, provide a [`LastLedgerSequence`](reliable-transaction-submission.html#lastledgersequence) field. This value should be based on the current ledger index (which you must look up from an online machine) and the amount of time you want the transaction to remain valid. Be sure to set a large enough `LastLedgerSequence` value to allow for time spent switching from the online machine to the offline machine and back. For example, a value 256 higher than the current ledger index means that the transaction will be viable for about 15 minutes. For more information, see [Finality of Results](finality-of-results.html) and [Reliable Transaction Submission](reliable-transaction-submission.html).
 
 
 ### {{n.next()}}. Copy transactions to online machine.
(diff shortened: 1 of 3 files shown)


content/tutorials/manage-account-settings/offline-account-setup.md
```diff
@@ -0,0 +1,332 @@
...
+
+
+
+### {{n.next()}}. Confirm account details
+
+When the transaction from the previous step is validated by consensus, your account has been created. From the online machine, you can confirm the status of the account with the [account_info method][]. Make sure the response contains `"validated": true` to confirm that this result is final.
+
+Take note of the sequence number of the account, in the `Sequence` field of the result's `account_data`. You need to know the sequence number to sign transactions from the account in future steps.
+
+If the [DeletableAccounts amendment](known-amendments.html#deletableaccounts) :not_enabled: is enabled, the `Sequence` number of a newly-funded account matches the [ledger index][] when it was funded. Otherwise, a newly funded account's `Sequence` number is always 1.
```
thejohnfreeman, line 89: If / when an amendment becomes enabled, do you make a run through the docs searching for ":not_enabled:" and pick out the right ones to change?

mDuo13: Yes.

content/tutorials/manage-account-settings/offline-account-setup.md
```diff
@@ -0,0 +1,332 @@
+# Offline Account Setup Tutorial
+
+A highly secure [signing configuration](set-up-secure-signing.html) involves keeping an XRP Ledger [account](accounts.html)'s [cryptographic keys](cryptographic-keys.html) securely on an offline, air-gapped machine. After setting up this kind of configuration, you can sign a variety of transactions, transfer only the signed transactions to an online computer, and submit them to the XRP Ledger network without ever exposing your secret key to malicious actors online.
```
LoisRP: Although "highly secure" is intuitively clear, I wonder if "secure" is better.

LoisRP: this configuration

content/tutorials/manage-account-settings/offline-account-setup.md
```diff
@@ -0,0 +1,332 @@
...
+      "validated" : false
+   }
+}
+```
+
+<!-- MULTICODE_BLOCK_END -->
+
+### {{n.next()}}. Enter the sequence number on the offline machine.
+
+Save the account's starting sequence number on the offline machine as its current sequence number. Whenever you sign a new transaction using the offline machine, you will use the current sequence number, then increase the current sequence number by 1.
```
LoisRP: This may be due to my lack of familiarity, but I found this confusing. It sounds like you begin by saving the starting sequence number, and then subsequently the number you need to save is always the last one used.

content/tutorials/manage-account-settings/offline-account-setup.md
```diff
@@ -0,0 +1,332 @@
...
+
+You can prepare several transactions in advance this way, then transfer the signed transactions to the online machine all at once and submit them. As long as each transaction is validly formed and pays a sufficient [transaction cost](transaction-cost.html), the XRP Ledger network should eventually include those transactions in validated ledgers, keeping the account's sequence number in the shared XRP Ledger in sync with the "current" sequence number you are tracking on the offline machine.
```
LoisRP: "should eventually include" - Do we want to give an idea of the timeline here?

content/tutorials/manage-account-settings/offline-account-setup.md
```diff
@@ -0,0 +1,332 @@
...
+
+
+
+### {{n.next()}}. Sign initial setup transactions, if any.
+
+On the offline machine, prepare and sign transactions for configuring your account. The details depend on how you intend to you use your account. Some examples of things you might want to do include:
```
LoisRP: intend to use your account.

content/tutorials/manage-account-settings/offline-account-setup.md
```diff
@@ -0,0 +1,332 @@
...
+- [Enable RequireAuth](become-an-xrp-ledger-gateway.html#enabling-requireauth) so that users can't open [trust lines](trust-lines-and-issuing.html) to you without your permission. If you don't plan to use the XRP Ledger's decentralized exchange or issued currency features, you may want to do this as a precaution.
+- Issued currency [Gateways](become-an-xrp-ledger-gateway.html) may have additional setup, such as:
+    - [Set a TransferRate](become-an-xrp-ledger-gateway.html#transferrate) for users transferring your issued currencies.
+    - [Disallow XRP payments](become-an-xrp-ledger-gateway.html#disallowxrp) if you plan to use this address for issued currencies only.
+
+At this stage, you are only signing the transactions, not submitting them. For each transaction, you must provide all fields, including fields that are normally auto-fillable such as the `Fee` ([transaction cost](transaction-cost.html)) and `Sequence` ([sequence number][]). If you prepare multiple transactions at the same time, you must use sequentially increasing `Sequence` numbers in the order you want the transactions to execute.
```
LoisRP, line 145: What happens if you were to increment by 2 (for example) for each transaction? Does the next transaction (which skips a number) fail?
If incrementing by one, and a transaction fails, can that number ever be reused?

content/tutorials/manage-account-settings/offline-account-setup.md
```diff
@@ -0,0 +1,332 @@
...
+}
+```
+
+<!-- MULTICODE_BLOCK_END -->
+
+You may also find it useful to check the [account_info][account_info method] of the sending account after all transactions have processed. Note the account's current sequence number (`Sequence` field) and, optionally, XRP balance.
+
+For any transactions that failed, you should decide what to do:
+
+- If the transaction failed with a `tefMAX_LEDGER` code, you may need to specify a higher [transaction cost](transaction-cost.html) to get the transaction processed. (This likely indicates that the XRP Ledger network is under load.) You may decide to replace the transaction with a new version that pays a higher cost and has a higher `LastLedgerSequence` parameter (if any).
```
LoisRP, line 299: It sounds like the transaction cost cannot be known in advance? Is there a way to be sure in advance so you don't fail due to insufficient payment for transaction cost?

content/tutorials/manage-account-settings/offline-account-setup.md
```diff
@@ -0,0 +1,332 @@
...
+
+For any transactions you decide to adjust or replace, note the details for when you return to the offline machine.
+
+
+
+### {{n.next()}}. Reconcile offline machine status.
+
+Return to the offline machine and input any changes to its internal tracking, such as:
+
+- Updating the account's current `Sequence` number, if necessary. If all transactions were included in validated ledgers (successfully or with `tec` codes), then the offline machine's tracking should already be accurate. Otherwise, you may need to change the offline machine's stored sequence number to match the value you noted in the previous step.
```
LoisRP: How do you change the stored sequence number? What if you change it to an invalid value?

content/tutorials/manage-account-settings/offline-account-setup.md
```diff
@@ -0,0 +1,332 @@
...
+      }
+   }
+}
+```
+
+<!-- MULTICODE_BLOCK_END -->
+
+**Tip:** If you are submitting more than 10 transactions at a time, you may have more success if you submit them in groups of 10 or less at a time, because the [transaction queue](transaction-queue.html) is limited to 10 transactions from the same sender at a time. After each group of 10 transactions, wait for all the transactions to leave the queue before submitting the next group.
+
+Retry submitting any transactions that failed with a [non-final outcome](finality-of-results.html). There is no chance of the same transaction being processed more than once.
```
ximinez, line 232: It's a stretch, but I am a little worried that someone may misunderstand this second sentence and reconstruct a transaction using a new sequence number, then getting upset that both got processed. Maybe something like "There is no chance of a duplicate submitted transaction being processed more than once, nor a transaction with an already used sequence number."

ximinez (changes requested):

content/tutorials/manage-account-settings/offline-account-setup.md
ximinez: On the XRP ledger a transaction must be submitted with the current sequence number of a given account to make it in to a validated ledger (ignoring Tickets and the fee queue for now). So on the offline machine,
1. You save the current sequence number, X.
2. You create a transaction with sequence number X.
3. You save the new expected sequence number, X+1.
4. Go to step 2.

content/tutorials/manage-account-settings/offline-account-setup.md
ximinez: In principal, if a transaction does not specify a `LastLedgerSequence` field, and no other transaction is submitted for that account with the same sequence number, that transaction could succeed at a completely arbitrary time in the future: days, months, even years. In practice, it's generally going to be less than 15 seconds, but with queueing on a busy network, it could be several minutes. However, because there are so many open-ended possibilities, most places in the documentation and the code leave the time frame intentionally vague, except where we emphasise using a `LastLedgerSequence`.

content/tutorials/manage-account-settings/offline-account-setup.md
ximinez: > What happens if you were to increment by 2 (for example) for each transaction? Does the next transaction (which skips a number) fail?

If you increment by 2, that would mean the XRPL is expecting sequence X, and you submit a transaction with sequence X+1. That transaction would retry (the error code is `terPRE_SEQ`) a few times on the server it was submitted to, in case X gets processed within that time. If X does not come in, then X+1 gets dropped. *However*, that drop can not be guaranteed. A node using a different implementation could easily hold on to that transaction indefinitely such that when X does get processed, it reprocesses X+1 immediately, allowing it to succeed. This is an example of the "open-ended possibilities" I described above.

> If incrementing by one, and a transaction fails, can that number ever be reused?

It depends on how the transaction fails. If it it "fails" with a code starting with `tec`, then the transaction could make it into a validated ledger, and thus that number can never be reused. If it fails with any other code, that failure may not be final, but the sequence number could be reused in a "replacement" transaction.

Note that the result code returned immediately upon submission is always preliminary and could potentially change.

content/tutorials/manage-account-settings/offline-account-setup.md
ximinez: When working offline you can never be completely sure, but as long as you meet the network minimum (currently 10 drops), chances are good that the cost will drop low enough for the transaction to succeed. Increasing the fee increases those odds.

content/tutorials/manage-account-settings/offline-account-setup.md
mDuo13: I chose the phrase "highly secure" to emphasize that this is a higher level of security than the standard secure signing setup instructions in the linked article.

content/tutorials/manage-account-settings/offline-account-setup.md
mDuo13: These instructions assume the user has some custom software set up to track the sequence number on the offline machine, such as writing it to a text file, or a database. Doing so makes sense if you want to construct more than one transaction at a time from the offline machine—and you probably do, if you have a lot of security precautions protecting access to your offline machine. (The alternative is to always require a human to input the next sequence number whenever you construct a transaction offline, remembering to count up for each new transaction.) This step reminds you that, if you are using software to track the sequence number, you may need to adjust it to match what happened when the transactions met the sometimes unpredictable live network.

I'm updating the first step to explicitly call out that you may want to set up custom software for sequence number tracking.

content/tutorials/manage-account-settings/offline-account-setup.md
mDuo13: Still, it's OK to provide some expectations for the typical case. Added:

> (Most transactions get a final, validated result within 15 seconds or less after being submitted to the network.)

mDuo13:
rebased to hopefully get the link fix

PR #746, opened 2019-12-15
Update for ripple-lib 1.5.0

intelliot:


mDuo13:
!rebuild

mDuo13:
I don't really understand why the build is failing but it's definitely not related to this diff.

PR #747, opened 2019-12-17
DeletableAccounts: the current ledger index minus 256 must be greater than the account's current Sequence number

intelliot:
Current wording is incorrect:

> Additionally, an account cannot be deleted if the current ledger index minus 256 is **greater** than the account's current `Sequence` number.

Minimal fix:

> Additionally, an account cannot be deleted if the current ledger index minus 256 is **less** than the account's current `Sequence` number.

(Bold added here for emphasis)

I can easily see how this mistake was made, since the sentence is written as a negative ("cannot"), and the wording in [XLS-7d](https://github.com/xrp-community/standards-drafts/issues/8) is ambiguous (the phrase "at least" combined with "less than" could mean that the value must be lower than "256 less than" or greater than "256 less than", since the words "at least" can refer to either magnitude or direction [as on a number line]).


scottschurr (approved):
:+1: Good catch @intelliot.  Thanks.

mDuo13 (approved):

PR #748, opened 2019-12-27
Update reserves.md

0xCLARITY:
Now that deletable accounts exist, the reserve amount is not permanently locked away.


mDuo13:
This is a good catch, but with one important exception:

DeletableAccounts are not currently available on the XRP Ledger production network. (In fact, they have yet to gain support from 80%+ of validators, so they won't be enabled for at least another 2 weeks.)

I believe this is because Ripple and possibly other validator operators have been slow to update and reconfigure their validators.

mDuo13:
Going to close this for now to keep the open PRs organized. We can re-address this when DeletableAccounts gets enabled

PR #749, opened 2019-12-27
Add entryNotFound error

mDuo13:
ledger_entry can return this error if the request is validly formed but the specified entry doesn't exist.

(See also: https://github.com/ripple/rippled/search?q=entryNotFound&unscoped_q=entryNotFound )


ddibble-rip (approved):
Is it restricted to a ledger object, or can simply a field return this error? If restricted to an object, description is fine.

mDuo13:
Yep, only applies to ledger objects.

mDuo13:
The build failure appears to be a glitch. (It's coming from a page in a different PR that's not even merged yet!) So I'll just merge this now and sort out the rest later if it comes up again.

PR #750, opened 2020-01-03
Improved Peer Management Docs

mDuo13:
- Describe peer reservations in concept page
- Clarify a bunch of other peering details, including pulling some things from the validator tutorial into the concept page
- Add API reference entries for new peer reservation methods
- Add a whole bunch of smallish tutorials related to peering tasks
- Reorganize peering tutorials into their own subcategory
- Reorganize peer-related admin API methods into their own subcategory
- Add a template file to be used when adding new API methods. (The Dactyl config file doesn't mention this file so it doesn't get built into anything, but it's available in the source tree for the repo to make it easier to add new methods to the API reference in the future.)

If you have a Ripple VPN, you can preview the changes on staging:

- [Peer Protocol](https://developers-staging.ripple.com/peer-protocol.html) concept page. **Updated** to add more information about peer reservations, private peering configurations, etc.
- [Run `rippled` as a Validator](https://developers-staging.ripple.com/run-rippled-as-a-validator.html) - **Slight updates** to move some "conceptual" content out to the Peer Protocol page.
- [Configure Peering](https://developers-staging.ripple.com/configure-peering.html) (new tutorial grouping)
    - [Configure a Private Server](https://developers-staging.ripple.com/configure-a-private-server.html) **New**
    - [Forward Ports for Peering](https://developers-staging.ripple.com/forward-ports-for-peering.html) **New**
    - [Manually Connect to a Specific Peer](https://developers-staging.ripple.com/manually-connect-to-a-specific-peer.html) **new**
    - [Set Maximum Number of Peers](https://developers-staging.ripple.com/set-max-number-of-peers.html) **New**
    - [Use a Peer Reservation](https://developers-staging.ripple.com/use-a-peer-reservation.html) **New**
    - (The "Cluster rippled Servers" and "Configure the Peer Crawler" tutorials have been recategorized but are otherwise unchanged and are still available from the same URLs.)
- [Peer Management Methods]() (new API reference grouping)
    - [peer_reservations_add](https://developers-staging.ripple.com/peer_reservations_add.html) **New**
    - [peer_reservations_del](https://developers-staging.ripple.com/peer_reservations_del.html) **New**
    - [peer_reservations_list](https://developers-staging.ripple.com/peer_reservations_list.html) **New**
    - (The `connect` and `peers` methods have been recategorized but have no changes to their contents and they are still available at the same URLs they were before.)
- Changes to the Japanese target are only to fix broken links and add a placeholder telling people to read the English version for the sections that haven't been updated.


content/concepts/the-rippled-server/peer-protocol.md
```diff
@@ -51,22 +77,65 @@ Configuring a server as a private server has several effects:
 
     **Caution:** It is possible to modify a server's source code so that it ignores this request and shares its immediate peers' IP addresses anyway. You should configure your private server to connect only to servers that you know are not modified in this way.
 
+### Pros and Cons of Peering Configurations
+
+To be part of the XRP Ledger, a `rippled` server must be connected to the rest of the open peer-to-peer network. Roughly speaking, there are three categories of configurations for how a `rippled` server connects to the network:
+
+- Using **discovered peers**. This is the default configuration, in which a server connects to any untrusted servers it finds and stays connected as long as those servers
```
ximinez: Incomplete sentence: "[...]stays connected as long as those servers"... as long as the servers do what?

ximinez (changes requested):

mDuo13:
Rebased on master to (hopfully) pull in a fix that'll make the builds work again...

content/references/rippled-api/admin-rippled-methods/peer-management-methods/peer_reservations_add.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+<!-- MULTICODE_BLOCK_END -->
+
+The request includes the following parameters:
+
+| `Field`       | Type   | Description                                         |
+|:--------------|:-------|:----------------------------------------------------|
+| `public_key`  | String | The [node public key][] of the peer reservation to add a reservation for, in [base58][]. |
+| `description` | String | _(Optional)_ A custom description for the peer reservation. The server truncates descriptions longer than 64 characters when it restarts. |
```
thejohnfreeman, line 52: There's the behavior as it exists, and then there's the public interface we want to guarantee to users. I would just change the second sentence to "The description must fit into 64 bytes." We still reserve the right to extend that limit later without breaking anyone. People may discover that the current implementation does not enforce the limit, or only enforces it on restart, but that doesn't matter. The only behavior we promise right now is that every description that fits into 64 bytes will be preserved.

content/references/rippled-api/admin-rippled-methods/peer-management-methods/peer_reservations_add.md
```diff
@@ -0,0 +1,136 @@
...
+
+The response follows the [standard format][], with a successful result containing the following fields:
+
+| `Field`    | Type   | Description                                            |
+|:-----------|:-------|:-------------------------------------------------------|
+| `previous` | Object | _(May be omitted)_ The previous entry for the same [node public key][], if there was already a reservation with the same node public key. This object is formatted as a **Peer Reservation Object**, as described below. |
+
+If there was not a previous entry for the same [node public key][], the `result` object is empty.
+
+#### Peer Reservation Object
```
thejohnfreeman, line 119: The `add` and `list` commands both have this section documenting the structure of a Peer Reservation Object, but `del` is missing. Do we want to copy it to all 3 commands? Do we want to factor it somewhere and reference it from all 3 commands?

content/tutorials/manage-the-rippled-server/configure-peering/use-a-peer-reservation.md
```diff
@@ -0,0 +1,167 @@
+# Use a Peer Reservation
+
+A [peer reservation][] is a setting that makes a `rippled` server always accept connections from a peer matching the reservation. The following steps describe how to use peer reservations to maintain a consistent connection from one server to a specific peer, with the permission of the peer server's administrator.
```
thejohnfreeman: The first sentence seems to define "server" as the node that creates a peer reservation (on itself) and the "peer" as the node that uses the reservation (whose node ID is kept in the Peer Reservation Object). The first part of the second sentence seems consistent with that, but then the second part I find confusing. Either it is calling the server a "peer server", or it incorrectly states that a peer reservation on the server requires the permission of the peer administrator.

mDuo13: Not quite either one.

Which side actually benefits more from a peer reservation depends partially on which side is busier. But also, if it's your own server, you can just force it to try to connect to the peer (as described in this tutorial) and it'll do that... but it only works if the peer has an available slot ready for you, or a reservation.

I wrote this tutorial with the assumption that, for the most part, the "server" following this tutorial would be an ordinary server making an outgoing connection, while the "peer" might be a busy or (semi-)private hub receiving the incoming connection. So it's the peer—the busy server—that would need to add the reservation.

You _can_ also add the reservation on your own server to make sure it saves an outgoing slot, and arguably it's a good idea to add that step to this tutorial, but unless I'm misunderstanding something, you _do_ need the peer's admin to add a reservation (for you) to the peer server.

thejohnfreeman: > I wrote this tutorial with the assumption that, for the most part, the "server" following this tutorial would be an ordinary server making an outgoing connection, while the "peer" might be a busy or (semi-)private hub receiving the incoming connection. So it's the peer—the busy server—that would need to add the reservation.

If this is true, then it conflicts with the wording of the first sentence. It should be reworded to say:

> A peer reservation is a setting that makes a rippled ~server~ peer [the hub] always accept connections from a ~peer~ server [the reader's server] matching the reservation.

However, this wording clashes with the etymology of Peer Reservation, which is named from the perspective of the hub. In that context, the reader's server is called "peer", not the hub. This meaning matches the original wording of the first sentence.

I think the right fix here is to introduce some names other than "peer" and "server". Because they are symmetric, their meaning depends on the reader's perspective, and there seem to be some perspective shifts in the text. Perhaps "local" and "remote" would be better?

I would change "with the permission of" to "with the cooperation of". The first makes me think I'm going to run an admin command on someone else's server.

ryangyoung: One way to clarify some of the potential confusion here might be to add a brief list of terms to the beginning of this article (we should also, at some point, create an overall glossary).

ximinez: Another option would be to change the terms used through the document. Maybe "local peer" and "remote hub".

ryangyoung: Those terms sound reasonable to me, @ximinez but I don't want to weigh in on the color of bike shed. I'm just saying should put an easy-to-read sign on it :) The problem now, I think, is that we're asking the reader to infer some of these definitions and concepts; we should explicitly  define them. 

Also, diagrams are usually helpful for these kinds of things. That's outside the scope of this PR--but maybe worth adding to the backlog, @mDuo13.

content/references/rippled-api/admin-rippled-methods/peer-management-methods/peer_reservations_add.md
mDuo13: That makes sense, but the way you've phrased it makes it seem like it should be an error if you provide a description longer than 64 characters. So I think users would be confused by the current behavior.

I'd rather document how it works now, and update it if and when the implementation changes.

content/references/rippled-api/admin-rippled-methods/peer-management-methods/peer_reservations_add.md
mDuo13: Good catch. It's already factored into an include so I'll just add that bit to the del command.

content/tutorials/manage-the-rippled-server/configure-peering/use-a-peer-reservation.md
mDuo13: I agree with John's idea that we need to introduce terms other than "peer" and "server" since those are symmetric and also overlapping. However, after some thought, I decided that "local" and "remote" still have the same perspective/symmetry problem that "peer" has.

I also think Ryan is on the right track with providing a glossary. So here's what I did:

- I chose the terms **stock server** and **hub server**, where the stock server makes the outgoing connection and the hub server receives the incoming connection.
- I added an explicit description at the beginning defining how I'm using these terms and what configurations these instructions describe.

content/references/rippled-api/admin-rippled-methods/peer-management-methods/peer_reservations_add.md
thejohnfreeman: I think the problem with that is that readers will interpret it as a promise of future behavior. I'd rather consider the lack of error a bug than suggest it is intentional.

PR #753, opened 2020-01-09
Remove warning about fixMasterKeyAsRegularKey

jwbusch:
This warning isn't relevant anymore given that the amendment is now enabled -> https://xrpl.org/known-amendments.html#fixmasterkeyasregularkey


intelliot (approved):
Looks like fixMasterKeyAsRegularKey is not yet enabled on Testnet, but I think this is still OK to merge since it will likely be enabled on Testnet in short order.

mDuo13 (approved):

PR #758, opened 2020-01-13
fix: add tecINSUFF_FEE to transactions-result doc

alexprut:
The `tecINSUFF_FEE` error is not documented in https://xrpl.org/tec-codes.html

As an example consider at block `52617217` transaction `f4ae7d0bd6b96b342e1c91127a6711cc1ee1317cd5f1a7d5f1a376b7dd012789` with `"TransactionResult":"tecINSUFF_FEE"` in https://livenet.xrpl.org/transactions/F4AE7D0BD6B96B342E1C91127A6711CC1EE1317CD5F1A7D5F1A376B7DD012789/raw

As documented in https://github.com/ripple/ripple-dot-net/blob/master/Ripple.Core/Types/EngineResult.cs#L107


mDuo13:
Good catch. I made some changes:

- Clarify what's unusual about tecINSUFF_FEE. (It's a rare case since most of the time the transactions should get stopped by `terINSUF_FEE_B` before they reach a consensus; it's also the only case where a transaction doesn't destroy _exactly_ the amount of XRP specified in the `Fee` field.)
- Put it in the correct place in the table by alphabetical order.

alexprut:
@mDuo13 thanks for the clarification. Let me know if there is something else missing before merging.

mDuo13:
I've asked @ximinez to review my revised phrasing for accuracy before we merge.

ximinez (approved):

PR #761, opened 2020-01-14
Correct tfSetAuth typo

RareData:


mDuo13:
Thanks for the contribution, but that's [not a typo](https://github.com/ripple/rippled/blob/381a1b948b06d9526cc73f14cfc69635fabf8605/src/ripple/protocol/TxFlags.h#L89).

I don't know why the flag is "Set F Auth" but it is.

RareData:
This says otherwise: https://github.com/ripple/xrpl-dev-portal/blob/master/content/references/rippled-api/ledger-data-formats/ledger-object-types/ripplestate.md#ripplestate-flags

RareData:
What's up with this? https://github.com/ripple/rippled/blob/develop/src/test/app/SetTrust_test.cpp#L151

mDuo13:
The RippleState doc has a typo.

That other line of code you linked is correctly using `tfSetfAuth` but the comment above it has the typo.

PR #775, opened 2020-01-23
Currency Formatting synthesis / reorganization

mDuo13:
Before I put in all the work to fix the (likely many) broken links, I wanted to get some reviews to see if the new organizational structure I put in place made sense.

- The bits about currency formatting in **Basic Data Types** have been stripped down to a minimal illustrative example and a link to the detailed currency formatting page.
- I moved the comparison table between XRP and issued currencies near the top of the **Currency Formatting** page and extended it. After doing that I felt like it was redundant to have the minimum and maximum values of the different formats in separate sections, so those details are only in the table now.
- I rephrased the **currency code snippet** to be a little simpler and to link to the expanded definitions in the Currency Formatting page. (This snippet is used in the currency formatting page, where it links to the sections right below it, and is also used in the Data API docs.)
- I moved the nitty-gritty details about the 160-bit currency code format to the **Serialization** page. Unless you're implementing serialization, you probably don't need to know much more about the 160-bit format than "it exists".

If you have a Ripple VPN, you can see the versions I've staged here:

- [Basic Data Types # Specifying Currency Amounts](https://developers-staging.ripple.com/basic-data-types.html#specifying-currency-amounts)
- [Currency Formats](https://developers-staging.ripple.com/currency-formats.html)
- [Serialization # Amount Fields](https://developers-staging.ripple.com/serialization.html#amount-fields)
- [Data API # Currency Code](https://developers-staging.ripple.com/data-api.html#currency-code) (the page may jump to the wrong place because of the code tabs loading in late)

### DIFF
--- a/content/_snippets/data_types/currency_code.md
+++ b/content/_snippets/data_types/currency_code.md
@@ -1,4 +1,6 @@
-There are two kinds of currency code in the XRP Ledger:
+The [`rippled` APIs](rippled-api.html) support two formats of currency code for issued currencies:
 
-* Three-character currency code. We recommend using all-uppercase [ISO 4217 Currency Codes](http://www.xe.com/iso4217.php). However, any combination of the following characters is permitted: all uppercase and lowercase letters, digits, as well as the symbols `?`, `!`, `@`, `#`, `$`, `%`, `^`, `&`, `*`, `<`, `>`, `(`, `)`, `{`, `}`, `[`, `]`, and <code>&#124;</code>. The currency code `XRP` (all-uppercase) is reserved for XRP and cannot be used by issued currencies.
-* 160-bit hexadecimal values, such as `0158415500000000C1F76FF6ECB0BAC600000000`, according to the XRP Ledger's internal [Currency Format](currency-formats.html). This representation is uncommon.
+- **[Standard Currency Codes](currency-formats.html#standard-currency-codes):** As a 3-character string such as `"EUR"` or `"USD"`.
+- **[Nonstandard Currency Codes](currency-formats.html#nonstandard-currency-codes):** As a 160-bit hexadecimal string, such as `"0158415500000000C1F76FF6ECB0BAC600000000"`. This is uncommon.
+
+Currencies with the same code can [ripple](rippling.html) across connected trust lines. Currency codes have no other behavior built into the XRP Ledger.
--- a/content/_snippets/string-number-formatting.md
+++ b/content/_snippets/string-number-formatting.md
@@ -3,5 +3,6 @@ XRP Ledger APIs generally use strings, rather than native JSON numbers, to repre
 * Base-10.
 * Non-zero-prefaced.
 * May contain `.` as a decimal point. For example, ½ is represented as `0.5`. (American style, not European)
-* May contain `E` or `e` to indicate being raised to a power of 10 (scientific notation). For example, `1.2E5` is equivalent to 1.2×10<sup>5</sup>, or `120000`.
+* Negative amounts start with the character `-`.
+* May contain `E` or `e` to indicate being raised to a power of 10 (scientific notation). For example, `1.2E5` is equivalent to 1.2×10<sup>5</sup>, or `120000`. Negative exponents are also possible.
 * No comma (`,`) characters are used.
--- a/content/references/rippled-api/api-conventions/basic-data-types.md
+++ b/content/references/rippled-api/api-conventions/basic-data-types.md
@@ -81,76 +81,27 @@ If you do not specify a ledger, the `current` (in-progress) ledger is chosen by 
 **Note:** Do not rely on this default behavior for specifying a ledger; it is subject to change. Always specify a ledger version in the request if you can.
 
 
-## Currencies
+## Specifying Currency Amounts
 
-There are two kinds of currencies in the XRP Ledger: XRP, and everything else. There are many differences between the two:
+There are two kinds of currencies in the XRP Ledger: XRP, and issued currencies. These two types of currencies are specified in different formats, and they have different amounts of precision and different rounding behavior.
 
-| `XRP`                                                           | Issued Currencies |
-|:----------------------------------------------------------------|:-----------|
-| Has no issuer.                                                  | Always issued by an XRP Ledger account |
-| Specified as a string                                           | Specified as an object |
-| Tracked in [accounts](accountroot.html) | Tracked in [trust lines](ripplestate.html) |
-| Can never be created; can only be destroyed                     | Can be issued or redeemed freely |
-| Maximum value `100000000000` (`1e11`)                           | Maximum value `9999999999999999e80` |
-| Precise to the nearest ["drop"](#xrp) (0.000001 XRP)            | 15 decimal digits of precision, with a minimum nonzero absolute value of `1000000000000000e-96` |
+Some fields, such as the destination `Amount` of a [Payment transaction][], can be either type. Some fields only accept XRP specifically, such as the `Fee` field ([transaction cost](transaction-cost.html)).
 
-**Caution:** The XRP Ledger uses decimal math with different precision than typical floating-point numbers, so currency amounts are always presented as strings.
+XRP is specified as a string containing an integer number of "drops" of XRP, where 1 million drops equals 1 XRP. Issued currencies are instead specified as an object with fields for the decimal amount, currency code, and issuer. For example:
 
-### Specifying Currency Amounts
+- **XRP** - To specify an `Amount` field with a value of 13.1 XRP:
 
-Some API methods require you to specify an amount of currency. Depending on whether you are dealing in the network's native XRP currency or other currency units (called _issuances_), the style for specifying it is very different.
+        "Amount": "13100000"
 
-#### XRP
-[drops of XRP]: #xrp
-[XRP, in drops]: #xrp
+- **Issued Currency** - To specify an `Amount` field with a value of 13.1 FOO issued by or to rf1B...:
 
-Amounts of XRP are represented as strings. (XRP has precision equivalent to a 64-bit integer, but JSON integers are limited to 32 bits, so XRP can overflow if represented in a JSON integer.) XRP is formally specified in "drops", which are equivalent to 0.000001 (one 1-millionth) of an XRP each. Thus, to represent 1.0 XRP in a JSON document, you would write:
+        "Amount": {
+            "value": "13.1",
+            "code": "FOO",
+            "issuer": "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn"
+        }
 
-```
-"1000000"
-```
-
-**Do not specify XRP as an object.**
-
-Unit tests are permitted to submit values of XRP (not drops) with a decimal point - for example, "1.23" meaning 1.23 XRP. All other cases should always specify XRP in drops, with no decimal point: e.g. "1230000" meaning 1.23 XRP.
-
-#### Non-XRP
-
-If you are specifying non-XRP currency (including fiat dollars, precious metals, cryptocurrencies, or other custom currency) you must specify it with a currency specification object. This is a JSON object with three fields:
-
-| `Field`    | Type                       | Description                        |
-|:-----------|:---------------------------|:-----------------------------------|
-| `currency` | String - [Currency Code][] | Arbitrary code for currency to issue. Cannot be `XRP`. |
-| `value`    | String                     | Quoted decimal representation of the amount of currency. This can include scientific notation, such as `1.23e11` meaning 123,000,000,000. Both `e` and `E` may be used. |
-| `issuer`   | String                     | Unique account address of the entity issuing the currency. In other words, the person or business where the currency can be redeemed. |
-
-**Caution:** These field names are case-sensitive.
-
-For example, to represent $153.75 US dollars issued by account `r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59`, you would specify:
-
-```
-{
-    "currency": "USD",
-    "value": "153.75",
-    "issuer": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59"
-}
-```
-
-Unit tests are permitted to submit amounts of non-XRP currencies as a slash-separated string in the format `"amount/currency/issuer"`. All other cases should use the JSON object format above.
-
-#### Specifying Currencies Without Amounts
-
-If you are specifying a non-XRP currency without an amount (typically for defining an order book of currency exchange offers) you should specify it as above, but omit the `value` field.
-
-If you are specifying XRP without an amount (typically for defining an order book) you should specify it as a JSON object with _only_ a `currency` field. Never include an `issuer` field for XRP.
-
-Finally, if the recipient account of the payment trusts multiple issuers for a currency, you can indicate that the payment should be made in any combination of issuers that the recipient accepts. To do this, specify the recipient account's address as the `issuer` value in the JSON object.
-
-### Currency Codes
-[Currency Code]: #currency-codes
-
-{% include '_snippets/data_types/currency_code.md' %}
-<!--{#_ #}-->
+For more information, see [Currency Formats](currency-formats.html).
 
 
 ## Specifying Time
--- a/content/references/rippled-api/api-conventions/basic-data-types.md
+++ b/content/references/rippled-api/api-conventions/basic-data-types.md
@@ -83,7 +83,7 @@ If you do not specify a ledger, the `current` (in-progress) ledger is chosen by 
 
 ## Specifying Currency Amounts
 
-There are two kinds of currencies in the XRP Ledger: XRP, and issued currencies. These two types of currencies are specified in different formats, and they have different amounts of precision and different rounding behavior.
+There are two kinds of currencies in the XRP Ledger: XRP, and issued currencies. These two types of currencies are specified in different formats, with different precision and rounding behavior.
 
 Some fields, such as the destination `Amount` of a [Payment transaction][], can be either type. Some fields only accept XRP specifically, such as the `Fee` field ([transaction cost](transaction-cost.html)).
 
--- a/content/references/rippled-api/api-conventions/currency-formats.md
+++ b/content/references/rippled-api/api-conventions/currency-formats.md
@@ -1,66 +1,112 @@
 # Currency Formats
 
-The XRP Ledger has two kinds of money: [XRP](xrp.html), and [issued currencies](issued-currencies.html). In the XRP Ledger, both types have high precision, although their formats are different.
+The XRP Ledger has two kinds of money: [XRP](xrp.html), and [issued currencies](issued-currencies.html). Both types have high precision, although their formats are different.
 
-## String Formatting
+## Comparison
+
+The following table summarizes some of the differences between XRP and [issued currencies](issued-currencies.html) in the XRP Ledger:
+
+| XRP                                                      | Issued Currencies |
+|:---------------------------------------------------------|:------------------|
+| Has no issuer.                                           | Always issued by an XRP Ledger account |
+| Specified as a string.                                   | Specified as an object |
+| Tracked in [accounts](accountroot.html).                 | Tracked in [trust lines](ripplestate.html). |
+| Can never be created; can only be destroyed.             | Can be issued or redeemed freely. |
+| Minimum value: `0`. (Cannot be negative.)                | Minimum value: `-9999999999999999e80`. Minimum nonzero absolute value: `1000000000000000e-96`.
+| Maximum value `100000000000` (1<sup>11</sup>) XRP. That's `100000000000000000` (10<sup>17</sup>) "drops". | Maximum value `9999999999999999e80` |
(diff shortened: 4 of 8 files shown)


content/_snippets/data_types/currency_code.md
```diff
@@ -1,4 +1,6 @@
-There are two kinds of currency code in the XRP Ledger:
+The [`rippled` APIs](rippled-api.html) support two formats of currency code for issued currencies:
 
-* Three-character currency code. We recommend using all-uppercase [ISO 4217 Currency Codes](http://www.xe.com/iso4217.php). However, any combination of the following characters is permitted: all uppercase and lowercase letters, digits, as well as the symbols `?`, `!`, `@`, `#`, `$`, `%`, `^`, `&`, `*`, `<`, `>`, `(`, `)`, `{`, `}`, `[`, `]`, and <code>&#124;</code>. The currency code `XRP` (all-uppercase) is reserved for XRP and cannot be used by issued currencies.
-* 160-bit hexadecimal values, such as `0158415500000000C1F76FF6ECB0BAC600000000`, according to the XRP Ledger's internal [Currency Format](currency-formats.html). This representation is uncommon.
+- **[Standard Currency Codes](currency-formats.html#standard-currency-codes):** As a 3-character string such as `"EUR"` or `"USD"`.
+- **[Nonstandard Currency Codes](currency-formats.html#nonstandard-currency-codes):** As a 160-bit hexadecimal string, such as `"0158415500000000C1F76FF6ECB0BAC600000000"`. This is uncommon.
```
LoisRP, line 4: "This is uncommon." immediately raises a question for me as to why one would use this, and how uncommon it is.

content/_snippets/string-number-formatting.md
```diff
@@ -3,5 +3,6 @@ XRP Ledger APIs generally use strings, rather than native JSON numbers, to repre
 * Base-10.
 * Non-zero-prefaced.
 * May contain `.` as a decimal point. For example, ½ is represented as `0.5`. (American style, not European)
-* May contain `E` or `e` to indicate being raised to a power of 10 (scientific notation). For example, `1.2E5` is equivalent to 1.2×10<sup>5</sup>, or `120000`.
+* Negative amounts start with the character `-`.
+* May contain `E` or `e` to indicate being raised to a power of 10 (scientific notation). For example, `1.2E5` is equivalent to 1.2×10<sup>5</sup>, or `120000`. Negative exponents are also possible.
 * No comma (`,`) characters are used.
```
LoisRP, line 8: A separate question: commas are typically used in English as a thousands separator, and periods are used as decimal separators. The reverse is true in many other languages. Is it fair to say that we expect all users to adopt the English-language format for numeric currency values?

content/references/rippled-api/api-conventions/basic-data-types.md
```diff
@@ -81,76 +81,27 @@ If you do not specify a ledger, the `current` (in-progress) ledger is chosen by
 **Note:** Do not rely on this default behavior for specifying a ledger; it is subject to change. Always specify a ledger version in the request if you can.
 
 
-## Currencies
+## Specifying Currency Amounts
 
-There are two kinds of currencies in the XRP Ledger: XRP, and everything else. There are many differences between the two:
+There are two kinds of currencies in the XRP Ledger: XRP, and issued currencies. These two types of currencies are specified in different formats, and they have different amounts of precision and different rounding behavior.
```
LoisRP: I'm not certain "amounts of precision" is the right term. How about "levels of precision" or "degree of precision"? A few minutes on Google did not answer this question, but it does not suggest the usage of "amounts of precision". "Amount" could possibly be confused with the currency amount.

content/references/rippled-api/api-conventions/basic-data-types.md
```diff
@@ -81,76 +81,27 @@ If you do not specify a ledger, the `current` (in-progress) ledger is chosen by
...
-#### XRP
-[drops of XRP]: #xrp
-[XRP, in drops]: #xrp
+- **Issued Currency** - To specify an `Amount` field with a value of 13.1 FOO issued by or to rf1B...:
 
-Amounts of XRP are represented as strings. (XRP has precision equivalent to a 64-bit integer, but JSON integers are limited to 32 bits, so XRP can overflow if represented in a JSON integer.) XRP is formally specified in "drops", which are equivalent to 0.000001 (one 1-millionth) of an XRP each. Thus, to represent 1.0 XRP in a JSON document, you would write:
+        "Amount": {
+            "value": "13.1",
+            "code": "FOO",
+            "issuer": "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn"
```
LoisRP, line 101: As a question for my benefit, is this the address starting with "r" that we should always use in examples? Do we have a list of addresses for sample use?

content/references/rippled-api/api-conventions/currency-formats.md
```diff
@@ -1,66 +1,112 @@
...
 
-## XRP Precision
+| XRP                                                      | Issued Currencies |
+|:---------------------------------------------------------|:------------------|
+| Has no issuer.                                           | Always issued by an XRP Ledger account |
+| Specified as a string.                                   | Specified as an object |
+| Tracked in [accounts](accountroot.html).                 | Tracked in [trust lines](ripplestate.html). |
+| Can never be created; can only be destroyed.             | Can be issued or redeemed freely. |
+| Minimum value: `0`. (Cannot be negative.)                | Minimum value: `-9999999999999999e80`. Minimum nonzero absolute value: `1000000000000000e-96`.
+| Maximum value `100000000000` (1<sup>11</sup>) XRP. That's `100000000000000000` (10<sup>17</sup>) "drops". | Maximum value `9999999999999999e80` |
```
LoisRP: A little inconsistent in this table about ending or not ending each entry with a period.

content/references/rippled-api/api-conventions/currency-formats.md
```diff
@@ -1,66 +1,112 @@
...
-4. The remaining 54 bits represent the significant digits as an unsigned integer. When serializing, this value is normalized to the range 10<sup>15</sup> (`1000000000000000`) to 10<sup>16</sup>-1 (`9999999999999999`) inclusive, except for the special case of the value 0. There is a special case for the value 0. In this case, the sign bit, exponent, and mantissa are all zeroes, so the 64-bit value is serialized as `0x8000000000000000000000000000000000000000`.
+The issued currency format can store a wide variety of assets, including those typically measured in very small or very large denominations. This format uses significant digits and a power-of-ten exponent in a similar way to scientific notation. The format supports positive and negative significant digits and exponents within the specified range. Unlike typical floating-point representations of non-whole numbers, this format uses integer math for all calculations, so it always maintains 15 decimal digits of precision. Multiplication and division have adjustments to compensate for over-rounding in the least significant digits.
```
LoisRP, line 87: As a side issue, what if the currency type is such that it should always remain as an integer? For example, I have seen frequent flyer miles mentioned as a possible currency, and those are always expressed as integers.

content/references/rippled-api/api-conventions/currency-formats.md
```diff
@@ -1,66 +1,112 @@
...
+{% include '_snippets/data_types/currency_code.md' %}
+<!--{#_ #}-->
 
-### Standard Currency Codes
 
-The standard currency mapping allocates the bits as follows:
+### Standard Currency Codes
 
-![Standard Currency Code Format](img/currency-code-format.png)
+The standard format for currency codes is a three-character string such as `USD`. This is intended for use with [ISO 4217 Currency Codes](http://www.xe.com/iso4217.php). The following rules apply:
```
LoisRP, line 102: I think this is the obvious implication, but to be clear, someone should not use a three-character string for their own invented currency that already exists on that ISO-4217 list?

content/references/rippled-api/api-conventions/currency-formats.md
```diff
@@ -1,66 +1,112 @@
...
+- Currency codes are case-sensitive.
+- The currency code `XRP` (all-uppercase) is disallowed. Real XRP typically does not use a currency code in the XRP Ledger protocol.
 
-Usually, XRP amounts are not specified with currency codes. In the rare case that a field specifies a currency code for XRP, the currency code's binary format is all zeroes.
+At the protocol level, this format is [serialized](serialization.html#currency-codes) into a 160-bit binary value starting with `0x00`.
 
 ### Nonstandard Currency Codes
 
-You can also issue currency of other types by using a 160-bit (40-character) hexadecimal string such as `015841551A748AD2C1F76FF6ECB0CCCD00000000` as the currency code. To prevent this from being treated as a different currency code type, the first 8 bits MUST NOT be `0x00`.
+You can also issue currency of other types by using a 160-bit (40-character) hexadecimal string such as `015841551A748AD2C1F76FF6ECB0CCCD00000000` as the currency code. To prevent this from being treated as a "standard" currency code, the first 8 bits MUST NOT be `0x00`.
```
LoisRP, line 112: By "other types", do you mean a currency that does not have a 3-character code? If you are creating your own currency, or otherwise using a currency not on the ISO list, should you choose this option?

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -219,6 +219,39 @@ The following diagram shows the serialization formats for both XRP amounts and i
 
 ![XRP amounts have a "not XRP" bit, a sign bit, and 62 bits of precision. Issued currency amounts consist of a "not XRP" bit, a sign bit, an exponent (8 bits), mantissa (54 bits), currency code (160 bits), and issuer (160 bits).](img/serialization-amount.png)
 
+#### Issued Currency Amount Format
+[[Source]](https://github.com/ripple/rippled/blob/35fa20a110e3d43ffc1e9e664fc9017b6f2747ae/src/ripple/protocol/impl/STAmount.cpp "Source")
+
+![Issued Currency Amount Format diagram](img/currency-number-format.png)
+
+The XRP Ledger uses 64 bits to serialize the numeric amount of an issued currency. (This is the `value` field of a currency amount object in JSON format.) This consists of a "not XRP" bit, a sign bit, significant digits, and an exponent, in order:
```
LoisRP: I think that "(This is the `value` field of a currency amount object in JSON format.)" is referring to the `value` field, but it might not be immediately obvious. Suggest:
This numeric amount is the `value` field of a currency amount object in JSON format.

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -219,6 +219,39 @@ The following diagram shows the serialization formats for both XRP amounts and i
...
+2. The sign bit indicates whether the amount is positive or negative. Unlike standard [two's complement](https://en.wikipedia.org/wiki/Two%27s_complement) integers, `1` indicates **positive** in the XRP Ledger format, and `0` indicates negative.
+3. The next 8 bits represent the exponent as an unsigned integer. The exponent indicates the scale (what power of 10 the significant digits should be multiplied by) in the range -96 to +80 (inclusive). However, when serializing, we add 97 to the exponent to make it possible to serialize as an unsigned integer. Thus, a serialized value of `1` indicates an exponent of `-96`, a serialized value of `177` indicates an exponent of 80, and so on.
+4. The remaining 54 bits represent the significant digits as an unsigned integer. When serializing, this value is normalized to the range 10<sup>15</sup> (`1000000000000000`) to 10<sup>16</sup>-1 (`9999999999999999`) inclusive, except for the special case of the value 0. There is a special case for the value 0. In this case, the sign bit, exponent, and mantissa are all zeroes, so the 64-bit value is serialized as `0x8000000000000000000000000000000000000000`.
```
LoisRP: "special case of the value 0. There is a special case for the value 0. " - a bit redundant

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -219,6 +219,39 @@ The following diagram shows the serialization formats for both XRP amounts and i
...
+
+The numeric amount is serialized alongside the [currency code][] and issuer to form a full [issued currency amount](#amount-fields).
+
+#### Currency Codes
+
+At a protocol level, currency codes in the XRP Ledger are arbitrary 160-bit values, except the following values have special meaning:
+
+- The currency code `0x0000000000000000000000005852500000000000` is **always disallowed**. (This is the "standard format" for an issued currency with code "XRP".)
+- The currency code `0x0000000000000000000000000000000000000000` (all zeroes) is **generally disallowed**. Usually, XRP amounts are not specified with currency codes. However, this code is used to indicate XRP in rare cases where a field must specify a currency code for XRP.
```
LoisRP, line 241: To be pedantic, if you use `0x0000000000000000000000000000000000000000` where it is not needed, will you get an error? Or is it that this usage is definitively not preferred?

content/references/rippled-api/api-conventions/serialization.md
```diff
@@ -219,6 +219,39 @@ The following diagram shows the serialization formats for both XRP amounts and i
...
+
+The [`rippled` APIs](rippled-apis.html) support a **standard format** for translating three-character ASCII codes to 160-bit hex values as follows:
+
+![Standard Currency Code Format](img/currency-code-format.png)
+
+1. The first 8 bits must be `0x00`.
+2. The next 88 bits are reserved, and should be all `0`'s.
+3. The next 24 bits represent 3 characters of ASCII.
+    Ripple recommends using [ISO 4217](http://www.xe.com/iso4217.php) codes, or popular pseudo-ISO 4217 codes such as "BTC". However, any combination of the following characters is permitted: all uppercase and lowercase letters, digits, as well as the symbols `?`, `!`, `@`, `#`, `$`, `%`, `^`, `&`, `*`, `<`, `>`, `(`, `)`, `{`, `}`, `[`, `]`, and <code>&#124;</code>. The currency code `XRP` (all-uppercase) is reserved for XRP and cannot be used by issued currencies.
```
LoisRP, line 250: Does this mean that the three-character codes "ABC" and "aBC" and "abc" are each a different currency code?

LoisRP (commented):
Some questions are due to my unfamiliarity with the subject matter. I found this very useful.

content/_snippets/string-number-formatting.md
mDuo13: I wouldn't say we're expecting all users to accept the English-language format. We're using what's basically the native JavaScript format, but putting it in a string just as a precaution. So yeah, it uses `.` for a decimal point like the English-language convention, but it doesn't have thousands separators at all. I think most people writing code have come to terms with that, since it's not like the programming languages have different parsers based on what natural language you speak. (At least, I don't think they do!)

LoisRP: I did a search and found this:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString
I think it's fair to continue doing what we have been doing for now, but consider if there is a need to manage the decimal separators in a more universal way in the future.

content/_snippets/data_types/currency_code.md
mDuo13: I don't know how uncommon it is today. It depends entirely on how people end up using the XRP Ledger.

As for why you'd use it, it would be if you wanted to encode more information into the currency code than just three characters.

For example, if you wanted to trade securities with longer ticker symbols on the XRP Ledger, you could define a standard that translates those ticker symbols (client-side) into the raw hex format for use with the ledger. (Note, there may be legal challenges with doing that, but that's besides the point.) Back in 2014 or so, ripple-lib supported currency codes that had interest built in so that Gold Bullion International could issue gold on the XRP Ledger with [explicit demurrage fees](https://xrpl.org/demurrage.html). There are other possibilities, too, but what it comes down to is, you want currency codes that carry more than 3 letters' worth of information.

content/references/rippled-api/api-conventions/basic-data-types.md
mDuo13: Second question first, we don't have a list.

rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn is an address I control. I use it out of convenience when showcasing examples from the Mainnet. I have another address where I keep most of my XRP but I do use rf1B(..) for various stuff.

Partly that's because of @justmoon's funny anecdote about how he used his real Bitcoin address as the sample destination in the docs for his early Bitcoin JS client and received a surprising amount of money from people who didn't both changing the examples before running them. So, the upside of using your own address in examples is some people might send you free money. The downside is people have also copied example code setting their account's `Domain` field to match mine, so I occasionally get questions like, "Hey, why does this suspicious address say it's owned by you?" 😹 

On the Devnet and Testnet, I generally use the faucet address [rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe](https://testnet.xrpl.org/accounts/rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe) as the default destination for payments so by default the Test XRP gets added back to the supply. For anything else, you can just pull a new address from the faucet as needed and throw it away afterwards; addresses are as numerous and disposable as grains of sand.

Another address you may see in examples is r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59. I don't know whose that one is; it predates my work at Ripple. I think it probably belongs to a former Ripple employee, so we should not use that one in new examples.

Occasionally I'll also use the public issuing addresses of longstanding XRPL businesses like Bitstamp if I need an example of a very active address with a lot going on. I try not to use other random addresses if I don't know who runs them, since I feel like that could direct unwanted attention to people who didn't ask for it. Sure, all addresses and transactions on the XRP Ledger are public information, but it feels rude to just pick specific addresses out of the crowd without permission and draw attention to them.

So, feel free to use your own address, or mine, or generate new addresses as needed.

content/references/rippled-api/api-conventions/currency-formats.md
mDuo13: Good catch. Our convention at Ripple is to always use a period for cases like this in tables and lists, even when the things are not proper sentences.

content/references/rippled-api/api-conventions/currency-formats.md
mDuo13: The XRP Ledger doesn't support that. All issued currencies in the XRP Ledger are always divisible down to the minimum precision of the format. I'll add a note explicitly mentioning that.

content/references/rippled-api/api-conventions/currency-formats.md
mDuo13: It's not really a problem if they do.

The more salient advice is to assume that any issued currency is worthless unless you have reason to believe otherwise.

The XRP Ledger's issued currency features are basically just a big accounting system (you know, a ledger). Regardless of how people want to denominate their balances, the meaning comes from the policies and trust behind those balances, not from any technical details of how things are denoted.

content/references/rippled-api/api-conventions/currency-formats.md
mDuo13: ¯\_(ツ)_/¯

You can make up your own 3-character code if you want. If you're worried about it getting confusing later because it clashes with an actual ISO code, some of the characters available for the XRPL's 3-character codes aren't allowed by the _actual_ ISO spec, so you can pick one that'll never be a valid ISO code. Or you could use the 160-bit option if you want to encode more info (as stated above).

Basically the only downsides to picking the same code that someone else uses for a completely different asset is that if a trader wants to hold both assets in the XRPL's decentralized exchange, they have to be careful not to accidentally enable [rippling](https://xrpl.org/rippling.html) between the two. (Through rippling, third parties can potentially cause the trader to passively trade the two same-name assets at a 1:1 rate.) But rippling is off by default so that shouldn't generally happen.

content/references/rippled-api/api-conventions/serialization.md
mDuo13: Yes.

content/references/rippled-api/api-conventions/serialization.md
mDuo13: It's complicated. You should not do it. But I think there was a bug with rippled at some point where you could use the wrong datatype to specify the currency as `0` and rippled it would accept it. So you might need to be prepared to deserialize it from the ledger even though you can't really use it that way in transactions. I'll try to figure out more precisely what's up there. For now I'm going to leave this as-is.

content/_snippets/string-number-formatting.md
mDuo13: The "locale string" formats are meant for humans to read, but they are sometimes ambiguous and are not suitable as a means of communicating numbers meant for machines to read.

The number format we're talking about here is primarily intended for machines to read and write. We should not add extra considerations like locale if we can avoid it. In any case, this section is describing how the `rippled` APIs behave, so any substantive changes to how the code actually behaves would be [`rippled`](https://github.com/ripple/rippled/) features, not doc changes here.

content/references/rippled-api/api-conventions/basic-data-types.md
mDuo13: I simplified this sentence to just:

> These two types of currencies are specified in different formats, with different precision and rounding behavior.

That's probably for the best since the discussion of precision in this case is not strictly a question of "more" or "less".

mDuo13:
!rebuild

mDuo13:
Since the CI is broken, I ran the link checker by hand.

> Link check report. 374844 links checked.
> Success! No broken links found.

PR #776, opened 2020-01-31
server_info: hostid is an RFC1751 word

intelliot:
The word is not always four letters, but can range from 1 to 4 letters.

https://github.com/ripple/rippled/blob/90d9ca901d06b9da1653526c234bc4b00e27070b/src/ripple/app/misc/NetworkOPs.cpp


mDuo13:
Build failure is a CI problem. This looks good, thanks.

PR #777, opened 2020-02-03
Clarify QualityIn & QualityOut on trustset.md

### LINKED ISSUE #208: Document QualityIn & QualityOut fields
Description of QualityIn and QualityOut fields is very short and vague ("Value(?) incoming balances"):

https://developers.ripple.com/trustset.html

This page

https://web.archive.org/web/20150422102043/https://wiki.ripple.com/Trust_line_quality

contains an explanation but it is deprecated.

_(I edited the Trust Line Quality link to use the wayback machine instead of the broken wiki. --@mduo13)_

_(I edited the TrustSet link to point directly to the doc page on developers.ripple.com. The old ripple.com/build link redirected to the Transaction Formats hub page. -- @jhaaaa)_

mgibson323:
- Fixes #208 
- Adds more clarification to QualityIn and QualityOut on trustset.md

### DIFF
--- a/content/references/rippled-api/transaction-formats/transaction-types/trustset.md
+++ b/content/references/rippled-api/transaction-formats/transaction-types/trustset.md
@@ -31,9 +31,8 @@ Create or modify a trust line linking two accounts.
 | `LimitAmount`.`currency` | String    | (Amount.currency) | The currency to this trust line applies to, as a three-letter [ISO 4217 Currency Code](http://www.xe.com/iso4217.php) or a 160-bit hex value according to [currency format](currency-formats.html). "XRP" is invalid. |
 | `LimitAmount`.`value`    | String    | (Amount.value)    | Quoted decimal representation of the limit to set on this trust line. |
 | `LimitAmount`.`issuer`   | String    | (Amount.issuer)   | The address of the account to extend trust to. |
-| `QualityIn`              | Number    | UInt32            | _(Optional)_ Value incoming balances on this trust line at the ratio of this number per 1,000,000,000 units. A value of `0` is shorthand for treating balances at face value. |
-| `QualityOut`             | Number    | UInt32            | _(Optional)_ Value outgoing balances on this trust line at the ratio of this number per 1,000,000,000 units. A value of `0` is shorthand for treating balances at face value. |
-
+| `QualityIn`              | Number    | UInt32            | _(Optional)_ Value incoming balances on this trust line at the ratio of this number per 1,000,000,000 units. A value of `0` treats balances at face value. A value of `1` is recommended.  |
+| `QualityOut`             | Number    | UInt32            | _(Optional)_ Value outgoing balances on this trust line at the ratio of this number per 1,000,000,000 units. A value of `0` treats balances at face value. Set this value higher than `QualityIn` to charge a percentage for Ripples to this trust line. For example, if `QualityIn` is `1` and `QualityOut` is `1.01`, the charge is 1%. |
 
 ## TrustSet Flags
 


content/references/rippled-api/transaction-formats/transaction-types/trustset.md
```diff
@@ -31,9 +31,8 @@ Create or modify a trust line linking two accounts.
...
 | `LimitAmount`.`value`    | String    | (Amount.value)    | Quoted decimal representation of the limit to set on this trust line. |
 | `LimitAmount`.`issuer`   | String    | (Amount.issuer)   | The address of the account to extend trust to. |
-| `QualityIn`              | Number    | UInt32            | _(Optional)_ Value incoming balances on this trust line at the ratio of this number per 1,000,000,000 units. A value of `0` is shorthand for treating balances at face value. |
-| `QualityOut`             | Number    | UInt32            | _(Optional)_ Value outgoing balances on this trust line at the ratio of this number per 1,000,000,000 units. A value of `0` is shorthand for treating balances at face value. |
-
+| `QualityIn`              | Number    | UInt32            | _(Optional)_ Value incoming balances on this trust line at the ratio of this number per 1,000,000,000 units. A value of `0` treats balances at face value. A value of `1` is recommended.  |
```
mDuo13, line 34: A value of `1` is _NOT_ recommended.

I think you may be mixing up the notional idea behind this setting, and/or the [ripple-lib equivalent](https://xrpl.org/rippleapi-reference.html#trustline), with the actual functionality of this field.

This field is an integer which is considered to be a "per billion" rate. So a value of `1` here would mean that you value incoming balances at a rate of **0.0000001%**!

content/references/rippled-api/transaction-formats/transaction-types/trustset.md
```diff
@@ -31,9 +31,8 @@ Create or modify a trust line linking two accounts.
...
+| `QualityOut`             | Number    | UInt32            | _(Optional)_ Value outgoing balances on this trust line at the ratio of this number per 1,000,000,000 units. A value of `0` treats balances at face value. Set this value higher than `QualityIn` to charge a percentage for Ripples to this trust line. For example, if `QualityIn` is `1` and `QualityOut` is `1.01`, the charge is 1%. |
```
mDuo13, line 35: The example here is also not quite right for the same as with the previous line.

mDuo13 (changes requested):
I apologize for the confusion here; the old documentation is very badly written and this feature is very unintuitive in general, which is why it's still on the backlog. Although #208 was labeled as "Help Wanted", it's not a good beginner task at all. That's my bad.

What I had in mind for #208 was a section in the concept page [Trust Lines and Issuing](https://xrpl.org/trust-lines-and-issuing.html) to explain what trust line settings mean conceptually and why you'd use them.

In order to properly explain trust line quality, one needs to have a pretty strong understanding of the decentralized exchange and currency issuing features of the XRP Ledger, including some idea of why one would want to treat an issued currency at something other than face value. (It's complicated. There are a few reasons, such as differing fees between different issuers, or pricing in the risk that an issuer defaults on the value the issued currency represents.)

content/references/rippled-api/transaction-formats/transaction-types/trustset.md
mDuo13: To be clear, what _is_ recommended is not changing the value from its default of 0.

mgibson323: Okay, great - thanks for clarifying! I got the value of 1 from this page, linked to in the issue: 

https://web.archive.org/web/20150422102043/https://wiki.ripple.com/Trust_line_quality

mDuo13: Yeah, sorry about that. That's the only written doc we have of the thinking behind the feature, but it's pretty lacking, which is why we didn't just port it to the current dev portal.

The idea behind QualityIn/Out is that it represents a rate/percentage, but the way you specify it in the API is in the weird "integer billionths" format because that's more precise for tracking and storing in low-level data structures.

mDuo13:
Closed in favor of a more comprehensive approach to be done later

PR #779, opened 2020-02-06
Updated "The validator claims ownership of the" domain section

mayurbhandary:
Description of how to update the validator's manifest with a domain by generating a new validator token.

### DIFF
--- a/content/references/xrp-ledger-toml.md
+++ b/content/references/xrp-ledger-toml.md
@@ -89,7 +89,7 @@ expires = 2019-03-01T00:00:00.000Z
 
 [[VALIDATORS]]
 public_key = "nHBtDzdRDykxiuv7uSMPTcGexNm879RUUz5GW4h1qgjbtyvWZ1LE"
-attestation="A59AB577E14A7BEC053752ABFE78C3DED6DCEC81A7C41DF1931BC61742BB4FAEAA0D4F1C1EAE5BC74F6D68A3B26C8A223EA2492A5BD18D51F8AC7F4A97DFBE0C"
+attestation = "A59AB577E14A7BEC053752ABFE78C3DED6DCEC81A7C41DF1931BC61742BB4FAEAA0D4F1C1EAE5BC74F6D68A3B26C8A223EA2492A5BD18D51F8AC7F4A97DFBE0C"
 network = "main"
 owner_country = "us"
 server_country = "us"
@@ -97,7 +97,7 @@ unl = "https://vl.ripple.com"
 
 [[VALIDATORS]]
 public_key = "nHB57Sey9QgaB8CubTPvMZLkLAzfJzNMWBCCiDRgazWJujRdnz13"
-attestation="A59AB577E14A7BEC053752FBFE78C3DED6DCEC81A7C41DF1931BC61742BB4FAEAA0D4F1C1EAE5BC74F6D68A3B26C8A223EA249BA5BD18D51F8AC7F4A97DFBE0C"
+attestation = "A59AB577E14A7BEC053752FBFE78C3DED6DCEC81A7C41DF1931BC61742BB4FAEAA0D4F1C1EAE5BC74F6D68A3B26C8A223EA249BA5BD18D51F8AC7F4A97DFBE0C"
 network = "testnet"
 owner_country = "us"
 server_country = "us"
@@ -329,7 +329,7 @@ NmluOEhBU1FLUHVnQkQ2N2tNYVJGR3ZtcEFUSGxHS0pkdkRGbFdQWXk1QXFEZWRGdj
 VUSmEydzBpMjFlcTNNWXl3TFZKWm5GT3I3QzBrdzJBaVR6U0NqSXpkaXRROD0ifQ==     
 ```
 
-Update your `xrp-ledger.toml` file as described in the sample output. See [here](xrp-ledger-toml.html#contents) for an example.
+Update [the contents of your `xrp-ledger.toml` file](#contents) with the `attestation` block, and update the `rippled.cfg` file with the `[validator_token]` block from the sample output.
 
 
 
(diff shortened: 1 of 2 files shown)


content/references/xrp-ledger-toml.md
```diff
@@ -280,16 +280,50 @@ Domain verification requires establishing a two-way link between the domain oper
...
+    - Ensure that you have access to the validator-keys.json file that you created when first setting up your validator. If you have lost your keys or the keys have been compromised, please [revoke your keys](https://xrpl.org/run-rippled-as-a-validator.html#revoke-validator-keys) and generate new keys. 
 
+        Note: Recall that your validator-keys.json file should be stored **in a location not on your validator**. 
+
+    - **In a location not on your validator**, build the [validator-keys-too1](https://github.com/ripple/validator-keys-tool).  
+
+    - Run the following command to generate a new validator token that incorporates your domain and update your `xrp-ledger.toml` and `rippled.cfg` files:
+
+        ```
+        $./validator-keys set_domain your-domain-here.com
```
mDuo13: `your-domain-here.com` is a real domain that appears to serve ads.

The W3C sets aside `example.com` as a domain to use for examples, so please use this instead (and redo the other examples to match).

content/references/xrp-ledger-toml.md
```diff
@@ -280,16 +280,50 @@ Domain verification requires establishing a two-way link between the domain oper
...
+2. The validator claims ownership of the domain: 
 
-    <!-- TODO: Link documentation when it's available. -->
-    <!-- Aside: the old way of doing this was to set the `Domain` field of the
-         XRP Ledger account whose public key matched the validator's public key.
-         Since only the holder of the corresponding private key could send
-         transactions from this address, doing so effectively proved ownership
-         of the validator's private key.
-    -->
+    - Ensure that you have access to the validator-keys.json file that you created when first setting up your validator. If you have lost your keys or the keys have been compromised, please [revoke your keys](https://xrpl.org/run-rippled-as-a-validator.html#revoke-validator-keys) and generate new keys.
```
mDuo13: Links within the site should use the HTML filename (and anchor) only. Please remove the `https://xrpl.org/` prefix. This helps with the translated versions of the site and local testing. It also makes sure that they don't get styled as external links. (Also, if we move domains again, as we've done twice in the last 18 months, it saves us from doing a bunch of work. Hopefully we won't have to any time soon, but it's still good to be prepared.)

content/references/xrp-ledger-toml.md
```diff
@@ -280,16 +280,50 @@ Domain verification requires establishing a two-way link between the domain oper
...
-    -->
+    - Ensure that you have access to the validator-keys.json file that you created when first setting up your validator. If you have lost your keys or the keys have been compromised, please [revoke your keys](https://xrpl.org/run-rippled-as-a-validator.html#revoke-validator-keys) and generate new keys. 
 
+        Note: Recall that your validator-keys.json file should be stored **in a location not on your validator**. 
+
+    - **In a location not on your validator**, build the [validator-keys-too1](https://github.com/ripple/validator-keys-tool).  
+
+    - Run the following command to generate a new validator token that incorporates your domain and update your `xrp-ledger.toml` and `rippled.cfg` files:
+
+        ```
```
mDuo13: Our tooling doesn't like code fences in bulleted lists. It'll work if you remove the backticks and indent the code sample another 4 spaces.

content/references/xrp-ledger-toml.md
```diff
@@ -280,16 +280,50 @@ Domain verification requires establishing a two-way link between the domain oper
...
+
+You should include it in your xrp-ledger.toml file in the
+section for this validator.
+
+You also need to update the rippled.cfg file to add a new
+validator token and restart rippled:
+
+# validator public key: nHDG5CRUHp17ShsEdRweMc7WsA4csiL7qEjdZbRVTr74wa5QyqoF
+
+[validator_token]
```
mDuo13, line 321: We should add a warning reminding people to keep their validator token secret, and not to publish it in the xrp-ledger.toml or post it publicly anywhere. (The token contains the validator's ephemeral private key, so anyone who has the token could impersonate the validator.)

mDuo13 (changes requested):
Thanks for writing this up. I've noted a few things that we should fix before publishing, but this is a great start. Thanks!

mayurbhandary:
Thanks @mDuo13. I made the changes. Let me know what you think.

content/references/xrp-ledger-toml.md
```diff
@@ -325,6 +329,13 @@ NmluOEhBU1FLUHVnQkQ2N2tNYVJGR3ZtcEFUSGxHS0pkdkRGbFdQWXk1QXFEZWRGdj
 VUSmEydzBpMjFlcTNNWXl3TFZKWm5GT3I3QzBrdzJBaVR6U0NqSXpkaXRROD0ifQ==     
 ```
 
+Update your `xrp-ledger.toml` file as described in the sample output. See [here](xrp-ledger-toml.html#contents) for an example.
```
mDuo13: Three things on this line:

1. We should avoid using "here" as link text; it doesn't provide much context about where the link goes. 
2. I think it would be helpful to be more specific about what it means to update "as described in the sample output".
3. This same file is served as `xrp-ledger-toml.html` so we can shorten the link href to just `#contents`.

Taking those into account, here a suggestion of how I'd word it:

```md
Update [the contents of your `xrp-ledger.toml` file](#contents) with the `attestation` block from the sample output.
```

Also, the sample output suggests to me that you need to update the `[validator_token]` in your `rippled` config file. If that's true, we should mention it here; if you can skip that step, we should also mention that here.

mDuo13 (changes requested):
Left one suggestion, and a question about the validator token part of the process. With those addressed I'll be a :+1: .

content/references/xrp-ledger-toml.md
mayurbhandary: Ah that makes sense. Thanks, just pushed up the new changes.

mDuo13:
I'm happy with the changes at this point, but to get the link checker to be happy we'll have to add a redirect to the Japanese language target.

(I'll do that now and merge if the check passes.)

PR #781, opened 2020-02-19
Use the name Testnet instead of altnet

thejohnfreeman:
Before merging, you might want to edit my commit message that I goofed up.


mDuo13:
:+1:

PR #782, opened 2020-02-20
Wallet server

mayurbhandary:
This Pull Request includes the instructions on how to configure a "Wallet Server."


content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
```diff
@@ -0,0 +1,279 @@
+# Run rippled as a Wallet Server
+
+A Wallet Server is a catch-all configuration for rippled. With a Wallet Server, you can submit transactions to the XRP Ledger, access ledger history, and use the latest [tools](https://xpring.io/) to integrate with XRP.
```
mDuo13: I suggest changing "catch-all" to something less like slang. Maybe "versatile" or "multipurpose"?

mDuo13: Also, put `rippled` in code font in body text (anywhere except in the top-level title of an article)

mDuo13: Third, I'm not sure I agree with linking "tools" to Xpring's site. There's already a "Dev Tools" section of this site, and you have to scroll very far down the Xpring landing page to get to the part where it lists "tools". Maybe just don't make it a link.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
```diff
@@ -0,0 +1,279 @@
...
+
+
+A Wallet Server will:
```
mDuo13: As a general style rule, avoid the use of future tense unless you're specifically describing future events. In this case, I'd preferr phrasing it like this:

> A Wallet Server does all of the following:
>
> - Connects to...
> - Relays...

(etc.)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
```diff
@@ -0,0 +1,279 @@
...
+
+- Connect to a [network of peers](consensus-network.html)
+
+- Relay cryptographically signed [transactions](transaction-basics.html)
+
+- Maintain a local copy of the complete shared global [ledger](ledgers.html)
+
+- Behave as a [validator](run-rippled-as-a-validator.html) by issuing [validation](consensus.html#validation) messages
```
mDuo13: I would simplify this to:

> Participates in the [consensus process](consensus.html) as a validator.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
```diff
@@ -0,0 +1,279 @@
...
+
+1. Manually build and run the `validator-keys` tool, if you don't already have it installed via a `rippled` RPM.
+
+    For information about manually building and running the `validator-keys` tool, see [validator-keys-tool](https://github.com/ripple/validator-keys-tool).
+
+2. Generate a validator key pair using the `create_keys` command.
+
+        $ validator-keys create_keys
+
+      Sample output on Ubuntu:
```
mDuo13: By convention, this line should be indented 4 spaces, not 6. (Same goes for the other non-code-sample bits in the following lines.)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
```diff
@@ -0,0 +1,279 @@
...
+        bjdNTzJmZGtjd1JRSWhBT25ndTlzQUtxWFlvdUorbDJWMFcrc0FPa1ZCK1pSUzZQU2
+        hsSkFmVXNYZkFpQnNWSkdlc2FhZE9KYy9hQVpva1MxdnltR21WcmxIUEtXWDNZeXd1
+        NmluOEhBU1FLUHVnQkQ2N2tNYVJGR3ZtcEFUSGxHS0pkdkRGbFdQWXk1QXFEZWRGdj
+        VUSmEydzBpMjFlcTNNWXl3TFZKWm5GT3I3QzBrdzJBaVR6U0NqSXpkaXRROD0ifQ==
+
+On your Wallet Server:
+
+1. Add `[validator_token]` and its value to your wallet server's `rippled.cfg` file.
+
+
```
mDuo13: We have a pre-written snippet you can use here to remind people where to find their config file. To use it:

```
    {% include '_snippets/conf-file-location.md' %}<!--_ -->
```

(The comment at the end is just to fix syntax highlighting in some editors, which think the underscore in the file path should be the start of an italic block.)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
```diff
@@ -0,0 +1,279 @@
...
+3. Use the `server_info` command to get information about your wallet server to verify that validation has been enabled.
+
+        $ rippled server_info
+
+      - The `pubkey_validator` value in the response should match the `public_key` in the `validator-keys.json` file that you generated for use with your wallet server.
+
+      - The `server_state` value should be _**proposing**_.
+
+**Security Tip:** Change the permissions on your `rippled.cfg` file to be more restrictive. On Linux it is recommended to be `0600`. You can do this with `chmod 0600 rippled.cfg`
```
mDuo13: This doesn't show up as a callout, but I think it'd be better if it did. (Only the magic words "Tip", "Note", "Caution", and "Warning" are converted into callouts automatically.) Here's how I'd do it:

> **Tip:** To improve security, restrict access to the `rippled.cfg` file. To set the recommended level of security, run the command `chmod 0600 rippled.cfg`.

(The only supported operating systems for production use are Linux distros, so you don't need to specifically call that out.)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
```diff
@@ -0,0 +1,279 @@
...
+
+## 3. Connect to the network
```
mDuo13: My understanding of the use case we're targeting with this configuration is that of a business that's used to running individual blockchain servers but doesn't want to have to run multiple `rippled` instances for different purposes. In that case, I think this entire section adds unnecessary complexity, and they should connect using discovered peers (which is the default).

If it's really important that we describe the other ways of connecting to the network for a wallet server, we should separate out the "Connect to the network" steps as either a snippet or a separate tutorial rather than copy-pasting the steps into redundant versions. I can help with that if necessary, but as I said, I think it makes more sense to differentiate this tutorial from the other validator tutorial by making this the simpler one with more assumptions built in—such as that you'll connect using discovered peers.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
```diff
@@ -0,0 +1,279 @@
...
+
+    With `[peer_private]` enabled, `rippled` ignores any connections suggested by the `[ips]` stanza. If you need to connect to an IP currently in your `[ips]` stanza, put it in the `[ips_fixed]` stanza instead, but _**only**_ if you can expect them to behave responsibly as described in step 1.
+
+3. Restart `rippled`.
+
+        $ sudo systemctl restart rippled.service
+
+
+
+## 5. Verify your network connection
```
mDuo13: The numbering jumps from 3 to 5 here. If you remove the "Connect to the network" configuration as I've suggested, this would be step 3 instead.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
mDuo13: Finally, we should be consistent about whether "Wallet Server" is capitalized or not. (There are 8 instances in this file with it capitalized and 59 instances with it not capitalized.)

I think it is not a proper noun, and none of our other server modes/models, such as "stock server", "validator", "full-history server", and so on, are capitalized as proper nouns, so we should make it lowercase everywhere except in the top-level header, which is title case. (The other headings, for various steps, are sentence case, so it doesn't need to be capitalized there.)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
```diff
@@ -0,0 +1,279 @@
...
+
+- Use your validator key to claim ownership of a domain.
+
+Creating this link establishes strong evidence that you own both the validator key and the domain. Providing this evidence is just one aspect of [being a good validator](#1-understand-the-traits-of-a-good-validator).
+
+To provide domain verification:
+
+1. Choose a domain name you own that you want to be publicly associated with your wallet server. You must run a public-facing HTTPS server on port 443 of that domain and you must have access to the private key file associated with that server's TLS certificate. (Note: [TLS was formerly called SSL](https://en.wikipedia.org/wiki/Transport_Layer_Security).) As a precaution against DDoS attempts, your domain name should not resolve to the ip address of your wallet server.
+
+2. Serve an [`xrp-ledger.toml`](xrp-ledger-toml.html) file at your domain, and complete the [domain verification](https://xrpl.org/xrp-ledger-toml.html#domain-verification) steps.
```
mDuo13: Change the "domain verification" URL to be a relative URL (`xrp-ledger-toml.html#domain-verification` only) so that it doesn't display as an "external" link.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
```diff
@@ -0,0 +1,279 @@
...
+
+        Provide the value returned in the **Domain Signature** field of the Google Form.
+
+5. After submitting the completed Google Form, you'll receive an email from XRP Charts that tells you whether your domain verification succeeded or failed. If your domain verification succeeds, XRP Charts lists your wallet server and domain in its [Validator Registry](https://xrpcharts.ripple.com/#/validators).
+
+<!--{ ***TODO: For the future - add a new section or separate document: "Operating a Trusted Validator" -- things that you need to be aware of once your validator has been added to a UNL and is participating in consensus. We should tell the user what to expect once they are listed in a UNL. How to tell if your validator is participating in the consensus process? How to tell if something isn't right with your validator - warning signs that they should look out for? How to tell if your validator has fallen out of agreement - what is the acceptable vs unacceptable threshold? Maybe provide a script that will alert them when something is going wrong.*** }-->
```
mDuo13: :+1: 

I've opened #783 to make sure this task doesn't get lost.

dactyl-config.yml
```diff
@@ -2360,6 +2360,16 @@ pages:
         targets:
             - en
 
+    -   md: tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
+        html: run-rippled-as-a-wallet-server.html
+        funnel: Docs
+        doc_type: Tutorials
+        category: Manage the rippled Server
+        subcategory: Configuration
+        blurb: A catch-all configuration for anyone integrating XRP. #TODO:translate
```
mDuo13: Make sure to update this blurb to match the word choice from the intro sentence if you change "catch-all"

mDuo13 (changes requested):
Thanks for drafting this up.

I'd like to avoid having a bunch of very almost-identical content copy-pasted from the validator tutorial, so we may want to restructure the two guides a bit. I also think this guide could probably be further simplified. It might make sense to split some of the nested steps into separate pages that can be referenced by both guides.

Maybe the right way to go about this is to separate out the most basic instructions for enabling validation—generating keys and adding the token to the config file—into a stand-alone tutorial called "Enabling validation". Maybe dividing the validator config into smaller piecemeal steps will help achieve the same thing.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
MarkusTeufelberger: I strongly advise against having a server configured as validator and at the same time exposing ANY functionality to the broader internet (especially unauthenticated API access). There is no reason at all to validate on a server that only has the purpose to be an API endpoint.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
MarkusTeufelberger: I disagree with the naming. There is nothing "wallet-like" about a server like this. It is an API endpoint for wallets. Maybe call it an API server, a data endpoint...

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
mayurbhandary: @MarkusTeufelberger That's a good point. I will make sure to include a warning to convey that this server is not intended to be accessible by the public.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-wallet-server.md
mayurbhandary: I think we should keep the link because the layout of xpring.io will change over time, but the overall theme is that it's where developers should go to for tools. The first subheading in xpring.io says "Xpring Platform provides developer tools, services, and programs to integrate money into your apps" so I think it's still appropriate. Although, I do see what you're saying about having to scroll down to actually get to the tools.

mayurbhandary: The naming refers to the idea that wallets and wallet-like services will use a wallet server.

MarkusTeufelberger: They can use any other server too, this is just a server that's exposing the various APIs `rippled` has to offer for whatever purpose (with some questionable "everything at once" configuration). A fulll history server for example would be much better suited for wallets and wallet-like services than one configured as proposed here.

mayurbhandary: Agreed that you could use any other server too. This configuration is meant to clear any confusion about what can and can't be done through the rippled server. Bucketing the server modes into "Stock Server," "Validator," and "Full History Node" has caused confusion about this in the past. By wrapping it all together into a "Wallet Server," the hope is to eliminate that confusion. I was considering including a section on how to vary the stored history, but didn't want to suggest full history due to the disk size requirements.

mDuo13 (approved):
After fixing one nit with a relative URL, this looks good to go to me. :+1:

PR #784, opened 2020-02-25
Validator domain checker

mayurbhandary:
This is a tool to check that the new [domain verification](https://xrpl.org/xrp-ledger-toml.html#domain-verification) steps were completed properly. The new steps allow anyone who sees a validator's manifest to verify its domain. 

This tool verifies the domain by parsing the manifest to obtain the domain and public key. Then it goes to the domain and accesses a TOML file that should contain an attestation signed by the validator's private key. The signature is verified and the result is displayed.


intelliot:
Does this check the domain's ssl cert? I think the old code for this is here: https://github.com/ripple/validator-domain-verifier/blob/master/lib/verify.js

assets/js/domain-verifier-checker.js
```diff
@@ -0,0 +1,164 @@
+//Browserify was used to create domain-verifier-bundle.js so that we can use 'require' in the browser
```
intelliot: Consider running this file through something like Prettier so that the formatting is consistent. Currently, a lot of different styles are mixed together (semicolons, no semicolons, inconsistent spacing, etc.) [Example - could probably just copy this from the Prettier playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEB6VAhAThA7gZziwEsAzATwAJcBDfSgV0IBNKYJKws4b5LmIAWxrEoAWgBuRMsSJiARgyjMANnAB0AK3r4OMABa9qcTjSiNClAOTcAjg2LcrlUW30n5OAkUoAdKP6QUPgwnBDMcGCUALyUdg7cABQ2xAAOqWoKojRY5GKQEWBWAJQA3IHQIZQ0zMzc+PgAwuGRMXFw9o5wySTpmTV1cA35LUVlFcGhANZw5M2FbfFdPWkZcGIz5KkiWPgl5QFQQVUAKgCSAAoAym1WADypAHyN7mBTLqRuJqTEai70NDAMAYNBUKio+ggITgrCMBhMAFUAEoAGWq8ggUgANJxXu9yBAGFhKIQsFIsL5fFZ6AAJE4na4kuAwGCiADmAOUnCIrJ+YF4cBxZlYwhmJKJJgJRKZZJ8qRwEmIEXo8PanW4rDuNEo+m4pGivhAAA8sKkxGpmGy5OxBCp1PoYLaAMSQXZiQjA1KG54AeSRN3cNSI6juqBoj1DTysE1OPoAsiiAPoXACCJxpbUNqHUuDgYI2UDwUFQJrNFqtWHUNpUhoOx1CjRRKauV0TAHEfT6ACKZkDyGpWyj9y3rfAMMBgIb4WsxhtNluJjApnuxQ3Dwfr9bMMwVmeHdAnfTEeikJRA4jQSiioZfSgqY8wFWGUKEmCpBihVUhXhMdT+U9HKyl7XiiEBsgAorAuSJPARowDiWA0LgxSUMA-iUBhd7MneYHoZhZCUIkiHIaheGYThbJtAAJMkdz3hGqD0SU6g0OkCDMCcEDJE6KhgcxrGpOxMFwHBxRkZQAC+lB5pYaHmORFHUbR9GhkxxRViJMDCXBADUhp+CA6kCexnHcbxbIlOJEnieZ6j1BAKhSG0AHntA2kwChckKe0wJYOYNH3Pg2xQAxQVmI8zEOra7lGWxyimeZYnyRh1nJT5RLmOZ-ipegADM6h+MWqAAGKiKwqoSKCSq8BAxIIDAuQuOYqonPGaI-H8wqUOSZBUF+xBslAP7cJQECfKqvDwN+QFQH+Q34OQRyUC5M2UNsuxwImpYqIm1aJNuMA0Di77yPeYCJpsibuEax0MKdxDnZsOKCFONBWp5eFqKE60sJ92HmQAjG0IFgZBDXkIkhoXDk+DspQrUJvwvA0OoqOGklGHg6RaU-TCbQIyi+3IxjmGA3ZQwOVIkMgFcCKNI04EtujLG1I0Kh0PgiSNs2rYdt2JNSfyMBgPoiRwB9aVk-ZjndOLLPMGzHNc-OrZLl2JMYdwvmHCl-h4ccDkaOZiS48wSV-aElX3gdtWJvVJA3rEpvqAAaimKJnF2aZ+lcBx4QRiRW9V7BYHbUGyPgEvkQHACEKZYIh5DqMe8eJ4HVU26H9sR8UUcKSDEFQRDhouxnNW7EgzPS1ThoAOo4FAlEwFsJhiCSkIMCorDyCYh2nesOSJ8zAyKw0ys84uy4a5J0kqLJ4kYV93VlyHiakISXKxKQoKEAvy21YRxBtAADJQpQuJQdzL9b5dhw1EfqGojcGOfxA6TpxRyXvi-Yak7yxEHTOd8Hb4AANrEAALqgKsCdM6F1ZhWAgXrNK0dSAm3-k7O6cDNif2Qd5cigDb7ryUKwWIDUGBwG-qTbCk0hiHVWgAletts5DHAVAqwtDpoXigIgvB+DMa5C8vwzClViQ9QoG0TY8xIjqHERDOgYBiDEF2hAK6IkaCJBeg0N64ssScPodwrEsCHrwPIGoo009vKpWEaYYW+hCLiyETYguYNoKGi7EIEQ5gXbSD5LwbhlBioiAtFXCmMsxZGVZuzMe3MFxq0sQpVKVCMIBzkbglB+CXFF2ph44QrgfEkD8atK445JwwhhMPKJStYm807OrZJkkGkyTgE44RWTwY5M8fk3xD1-GXiCb8Cphl5aj05jUye9SMnkWsfgmZ0y96x0IavYhyh0n8PaW4kAh4TBLIPpsfexIDDHivGYMgdDqB0EoIWUIKzWCuBam1ZagzKkK2iWMlWEyElzMafJGZzTWkbOLiAAAchwUuN8Q70GKhvZgoTdDhMNAMlQEoXmjPHnEqeeFUpJKKgAJjmugLskQhCpChL3dwJyoBnKqOwUa8hDr3IpQIPJ5gurGKiJsAlqAESWHhLyjgXAeB8FVFo-AOi3BGHwB3LuOoaBOR7ggEkA0oB43kH1CluysDUjWiQSqfB9kwQpfovpUB1L+HQG2PQegmVdNZVyORt5YaDWGiYMat5jWrR+HmO5zUjXpBwPKYgApr7BwPvbJOhUVoBNxomYQVLSB0MSGszC4l6yUuukpKwTo43UpgMxK2SabLYTjW0AoMjCgtE0WYa608l7stMeY4GZhoEXCwQ9AA0ggpBktsKKFIJ8WIKrcCUAwAwftRATZtserMcxOIrDXWYuwAAUlcH0wKk3qAOtqQqRbQjMq8W0a6KjEwKKUVWqA0DcleMQbW3+U7TFtAGPUJoox1AIDLaCiIrb7pgESH20g09xIHiPPQY5IrXqDgMEYWgOhlWqvVTsphWr6CBr1SYTlu6rwQZMKuEAoD92iDEB67hCheLyEriAHSBGoB6WQIaHS9bNi0aQSAVNlRQhEhUL2B0b58BIHQPpHS-BbWUCEwTZMaYaTILY5MRSsRAXJBeJEKYcNnBCc46J6wqN1CWR1phKiLFNA0CNIkVp5FONIEYFgFQWIqFbpOC3SzVhYJ5ts1M8UE4pyWajW5Lded+G2Wrt0KwxUfQImBV2fiVSYkfL5pMmxMbtoqOilu26P6G3XTS9g2Yz1sMJJSm5-hRAcBYG82eGaiRNC2CNLqHE00mA4mK7Vfz+D8C4GIHYwi9XI7YxsRhfklhnPEBeq+Kwld3PeWK7cc4cZwJhZONGCbClPA8CmA0-rdATAcIxFgPN42+uYSm7EKwKYMB+hOOBSL63MIrZoGtpbG3BtNa1ftg70kE63HAkiJEfpFtvaHNwO712Ijb07jAV7B2jvWHC+24FPpa7Ar+8I753lAthKpsVyJrzqkfPiSxUg8AsCJHONcfLPzplJRxea1AlBgVkGbjqESYZ8CKKPu4FQgkK6FXQDx1IfH0C4EFzmXK9kiSTnUJAQQqAjOVRZ70GAYgRJEEUYQfA0u5UKPl+6e+jdFdGmV8edYuKAAc6hUj6C9FAHz5gj3sBPSzs9IQsAAySl-GAojGdGgwrEJ3AMqwQCuDrtkSa-bu5yCSBqtwrCh-XsSdOxJzCxGPufcwV9rqPwQGyF+VzROxFxRLXwMBC9O9z5QQPJBG7qFIDgQQLwcjSJNjDOAZxYCJHT2OeQTvEhQBxPnnEAMABsudQ9JIasyDKEesAHBxTzlkfP+OoEF7gYXousDi8l+r2XXA0gK6V2vw3auZea539rivbI9cG8IGIXFAB2M3Fv-zlYCae5RdvroaKd67vWYfiSDyBrEUBCBGPA+ePHPJPHELjH3BqDPZ+fQFPS+O8eA9+FCBeN3QvD3DNWIYFBgQQHuInJ3CXQwLAaRFMLSU1dSdgcvdkRIQfcYAIIvQ6BOP3d8KVNvESOgwvH5QvLWCfP-LQCAUQZIfYbKZBdAP3bnVAQPHIT8DVHpIWaNHAScBoagdwX1EwJgHweqIgAESlXNOaa3WVZQNQRMDvQQDrMWKQWAFrOASwmAM3bgWwolUHFQLSOgjCcSAKcmMcFwsQVkGANQZiKVPAQtNKAKJ0eoMHQIyEXAEI8iMI8yZiOAQQVIZuWIwqciGNHNBNEIEPEQqAEALEEACAFI7hfAZAUAQePAaGbgYIZAEAOVAQ5gAovsRCN4ZkK4bYRRRuZAchQUEAKKFQWuI8KaTouAK4BAWGVkRUZuOosABoZo0QUkGAC4RCNkYQZAbeOePo7QI0bAQEGYGAK4GgF6FEUQOADYnebY-AI0K4dkNQAARQYAgHgAuK2MKNxiwDqP7B7hUGaMDVgFriVAMGQGN2PneJwEIHrlYjqPlCGCICkGaPsGeLgBWOKLKJQDoDEBVXKSaMKKWG4BWLenWKQE2MIEKMIDMJ6KwAoXJLuLgEeORNeLJJAD7kBOYGBKQFxUKIamCXZGaEEGJJAELBVWaI0JOBoE7yZL6IkAoRbwiFgCuG3xSJTGUED3IDUClIkgkiAA)

mayurbhandary: Thanks for the tip! I used the prettier output.

intelliot (changes requested):
`Error: Cannot find module 'ripple-binary-codec' from '...'`

I think you need some additional instructions about how to get the dependencies.

I wonder if it may make sense to put the build instructions in a separate area or section. They are needed when editing the domain-verifier-checker, but are not necessary for building the docs site in general. Maybe @mDuo13 can provide a better idea here.

mayurbhandary:
Ah right you have to install the libraries before doing Browserify. I will include that in the build instructions.

mayurbhandary:
@intelliot I updated and added the build instructions for the tool under the build instructions for the website.

intelliot (approved):
It's a little unfortunate that we need to use Browserify and commit the browserified file, and it's 1.4 MB (!). It's not minified.

It's also not deterministic. I did the build myself and got an entirely different bundle: basically the same code but with different ordering of imports.

We might look into whether the necessary functionality is exported from ripple-lib (and change ripple-lib to export any needed functionality if it is not). Then we could use the minified ripple-lib build (similar to ripple-lib-1.4.1-min.js - which is 733 KB).

assets/js/domain-verifier-checker.js
```diff
@@ -0,0 +1,179 @@
...
```
mDuo13, line 1: I'm not exactly a fan of shipping a 1.3mb javascript file (`domain-verifier-bundle.js`) so maybe we can look into a way to cut that back. Even ripple-lib, which I think depends on these same requirements and then some, only weighs in at like 700kb.

assets/js/domain-verifier-checker.js
```diff
@@ -0,0 +1,179 @@
...
+
+const codec = require("ripple-binary-codec");
+const addressCodec = require("ripple-address-codec");
+const keyCodec = require("ripple-keypairs");
+
+const TIPS =
+  '<p>Check if the file is actually hosted at the URL above, check your server\'s HTTPS settings and certificate, and make sure your server provides the required <a href="xrp-ledger-toml.html#cors-setup">CORS header.</a></p>';
```
mDuo13: It looks like this warning is copy-pasted from the toml checker JS file. I have a feeling it's not as appropriate in this context.

tool/template-validator-domain-verifier.html
```diff
@@ -0,0 +1,32 @@
+{% extends "template-base.html" %}
+
+{% block main %}
+<section class="container-fluid p-3">
+  <h1>Domain Verification Checker</h1>
+
+  <p>This tool allows you to verify that <a href="https://xrpl.org/xrp-ledger-toml.html#domain-verification">domain verification</a> is properly configured. </p>
```
mDuo13: Use a relative, not absolute, link here.

mDuo13 (commented):
I need a little more time to actually test out the verifier, but this _mostly_ looks good.

However, to reduce complexity for the user, I was thinking... shouldn't it be possible to add domain verification directly to the existing TOML checker, rather than making a new tool for it?

In theory, at least, the toml file should have everything you need to verify the domain: the domain itself is in the URL, the validator public key is in the toml file, and so is the attestation. (Technically, all fields are optional, but this would operate under a rule like, "in order to provide domain verification, the following fields are required")

That way, users don't need to actually find/export their server's manifest at all. If the toml checker shows that the domain is verified, that seems like it would be enough.

tool/template-validator-domain-verifier.html
mayurbhandary: I embarrassingly keep making this mistake. Sorry, I will make it right!

assets/js/domain-verifier-checker.js
mayurbhandary: Ah yup it needs to be slightly modified for this tool. Thanks for catching.

mayurbhandary:
> However, to reduce complexity for the user, I was thinking... shouldn't it be possible to add domain verification directly to the existing TOML checker, rather than making a new tool for it?
> 
> In theory, at least, the toml file should have everything you need to verify the domain: the domain itself is in the URL, the validator public key is in the toml file, and so is the attestation. (Technically, all fields are optional, but this would operate under a rule like, "in order to provide domain verification, the following fields are required")

That is true, but this also checks that the manifest was correctly updated to include the domain. In practice, we'd only be able to fully verify the domain if we saw a manifest with the domain section updated. Someone might potentially sign the message "[domain-attestation-blob:example.com:publickey]" using the validator-keys tool to make the attestation without updating the manifest.

mayurbhandary:
I minified the bundle and included new build instructions.

assets/js/domain-verifier-checker.js
```diff
@@ -0,0 +1,193 @@
...
+          if (verify) {
+            makeLogEntry("Domain Verification Succeeded").addClass(CLASS_GOOD);
+          } else {
+            makeLogEntry("Domain Verification Failed").addClass(CLASS_BAD);
+          }
+        }
+      }
+      if (!validator_found) {
+        makeLogEntry(
+          "The validator key for this manifest was not found in the TOML file"
```
mDuo13, line 90: I've found an interesting failure case that shows this error message. If we can, I'd like to fix the tool to work around this, but we could alternatively tweak this error message to compensate.

The error case is this:

1. Serve a valid toml file but with the wrong validator public key (in my case it was my old hardcoded key that predates `validator-keys-tool`)
2. Run this checker. It downloads the tool, but finds the wrong validator key and reports this error message.
3. Update the toml file with the correct validator key
4. Run the checker again. Rather than downloading the updated toml file, the browser loads the file from cache, so this tool doesn't see the correct key. This error shows up again.

The workaround for the client is to reload without using the cache.

A change we could probably make on the client side is to add a unique query parameter to each AJAX request to make sure it bypasses the cache, e.g. `?v=123` with the value of the query parameter changing every time you press the "Check Domain Verification" button.

assets/js/domain-verifier-checker.js
```diff
@@ -0,0 +1,193 @@
...
+  $("#log").empty();
+
+  parse_manifest();
+}
+
+//1.
+//Start the verification process when the user enters a manifest.
+$(document).ready(() => {
+    $('#manifest-entry').submit(handle_submit)
+})
```
mDuo13: Nit: all files should end in newlines

tool/template-validator-domain-verifier.html
mDuo13: Still waiting for this fix.

mDuo13 (changes requested):
I tested this out and was able to get the tool to report that my domain is successfully verified (though my validator is currently only running in standalone mode because I don't have bandwidth to spare right now).

While I understand the idea about checking that the manifest is properly updated, this tool doesn't actually get there.

1. It needs instructions for how to get your manifest. For example, you can do this:

        $ validator-keys show_manifest hex

2. Getting the manifest as described in step 1 has the same problem you mentioned as purely relying on the toml file: it doesn't actually verify that you've updated your `rippled` server's config with the appropriate token.

    What _would_ verify that is the `manifest` admin method that's supposedly coming in `rippled` v1.5.0. But since we don't have that yet, we have to make do.

3. I still think it would be very helpful to update the existing toml checker tool to verify the domain. Even if it can't tell you that the validator configuration is correct, it can tell you, yes, the person who controls the domain definitely also controls the keys for a given validator address, which is basically what domain verification is about.

assets/js/domain-verifier-checker.js
mayurbhandary: I added a query parameter that increments every time the button is pressed. Thanks for noticing this and for the suggestion.

mDuo13:
Merged as 711513b5ab4df1b3f5bb7ce98d4f465f297588ed

PR #786, opened 2020-02-27
xrp-ledger-toml.md - fix typo

intelliot:
validator-keys-too1 -> validator-keys-tool


mDuo13:
Good catch. I completely couldn't see the typo until I looked very closely!

PR #788, opened 2020-02-27
Added entry in the the validators table for attestation

mayurbhandary:
Added an entry in the validators table for the attestation block used for domain verification.

### DIFF
--- a/content/references/xrp-ledger-toml.md
+++ b/content/references/xrp-ledger-toml.md
@@ -169,6 +169,7 @@ For _each_ `[[VALIDATORS]]` entry, you MAY provide any of the following fields:
 | Field        | Type   | Description                                          |
 |:-------------|:-------|:-----------------------------------------------------|
 | `public_key` | String | The master public key of your primary validator, encoded in the XRP Ledger's base58 format (typically, this starts with `n`). |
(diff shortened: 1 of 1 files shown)


content/references/xrp-ledger-toml.md
```diff
@@ -169,6 +169,7 @@ For _each_ `[[VALIDATORS]]` entry, you MAY provide any of the following fields:
 | Field        | Type   | Description                                          |
 |:-------------|:-------|:-----------------------------------------------------|
 | `public_key` | String | The master public key of your primary validator, encoded in the XRP Ledger's base58 format (typically, this starts with `n`). |
+| `attestation`| String | The signed attestation message that is output by the `set_domain` command in the [validator-keys-tool](https://github.com/ripple/validator-keys-tool)
```
mDuo13: Rather than linking the validator keys tool here maybe it would make more sense to link the domain verification steps in this same page? Also, I think it would make more sense to focus on what the attestation is for, not what tool provides it.

Suggested revised description:

> A signed message, in hexadecimal, indicating that the same entity runs this validator and the domain serving this TOML file. For more information, see [Domain Verification](#domain-verification).

mDuo13 (changes requested):

content/references/xrp-ledger-toml.md
mayurbhandary: Thanks! I made the change.

PR #789, opened 2020-03-01
Revert "Added warning about validator-keys.json file"

chalermchai29:
Reverts ripple/xrpl-dev-portal#787

### DIFF
--- a/content/references/xrp-ledger-toml.md
+++ b/content/references/xrp-ledger-toml.md
@@ -296,8 +296,7 @@ Domain verification requires establishing a two-way link between the domain oper
 
         
             $./validator-keys set_domain example.com
-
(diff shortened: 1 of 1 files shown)


mDuo13:
Why?

mDuo13:
If the reason this was opened was because of the syntax error causing the warning to be treated as a code block rather than a callout, that has been fixed by 3aae2cee40e1c01b40010511e2ec83a6acf127dc.

If there is another justification for this PR, please provide more information and reopen.

content/references/xrp-ledger-toml.md
```diff
@@ -296,8 +296,7 @@ Domain verification requires establishing a two-way link between the domain oper
 
         
             $./validator-keys set_domain example.com
```
chalermchai29, line 298: 3aae2ce

PR #796, opened 2020-03-07
Augmented submit

mDuo13:
Updates the `submit` method reference with the new fields added in https://github.com/ripple/rippled/pull/3125

Known issues that may affect these docs: 
- <s>https://github.com/ripple/rippled/issues/3285</s> - fixed in rc2
- https://github.com/ripple/rippled/issues/3284 - won't be fixed in v1.5.0 (docs in this PR are accurate to how it'll be released)


content/references/rippled-api/public-rippled-methods/transaction-methods/submit.md
```diff
@@ -79,9 +79,9 @@ The request includes the following parameters:
 | `seed_hex`     | String  | _(Optional)_ Secret key of the account supplying the transaction, used to sign it. Must be in hexadecimal format. If provided, you must also specify the `key_type`. Cannot be used with `secret`, `seed`, or `passphrase`. |
 | `passphrase`   | String  | _(Optional)_ Secret key of the account supplying the transaction, used to sign it, as a string passphrase. If provided, you must also specify the `key_type`. Cannot be used with `secret`, `seed`, or `seed_hex`. |
 | `key_type`     | String  | _(Optional)_ Type of cryptographic key provided in this request. Valid types are `secp256k1` or `ed25519`. Defaults to `secp256k1`. Cannot be used with `secret`. **Caution:** Ed25519 support is experimental. |
-| `fail_hard`    | Boolean | (Optional, defaults to false) If true, and the transaction fails locally, do not retry or relay the transaction to other servers |
+| `fail_hard`    | Boolean | _(Optional)_ If `true`, and the transaction fails locally, do not retry or relay the transaction to other servers. The default is `false`. [Updated in: rippled 1.5.0][] |
```
LoisRP, line 82: What is the consequence if you do retry or relay?

content/references/rippled-api/public-rippled-methods/transaction-methods/submit.md
```diff
@@ -79,9 +79,9 @@ The request includes the following parameters:
...
 | `offline`      | Boolean | (Optional, defaults to false) If true, when constructing the transaction, do not try to automatically fill in or validate values. |
-| `build_path`   | Boolean | _(Optional)_ If provided for a Payment-type transaction, automatically fill in the `Paths` field before signing. You must omit this field if the transaction is a direct XRP-to-XRP transfer. **Caution:** The server looks for the presence or absence of this field, not its value. This behavior may change. |
+| `build_path`   | Boolean | _(Optional)_ If provided for a Payment-type transaction, automatically fill in the `Paths` field before signing. You must omit this field if the transaction is a direct XRP-to-XRP transfer. **Caution:** The server looks for the presence or absence of this field, not its value. This behavior may change. <!-- https://github.com/ripple/rippled/issues/3272 --> |
```
LoisRP: "automatically fill in the `Paths` field before signing. " - This might be clearer in context, but here I got confused.

content/references/rippled-api/public-rippled-methods/transaction-methods/submit.md
```diff
@@ -261,6 +261,14 @@ The response follows the [standard format][], with a successful result containin
 | `engine_result_message` | String  | Human-readable explanation of the transaction's preliminary result |
 | `tx_blob`               | String  | The complete transaction in hex string format |
 | `tx_json`               | Object  | The complete transaction in JSON format  |
+| `accepted`              | Boolean | The value `true` indicates that the transaction was applied, queued, broadcast, or kept for later. A value of `false` indicates that none of those happened, so the transaction cannot possibly succeed as long as you do not submit it again and have not already submitted it another time. [New in: rippled 1.5.0][] |
```
LoisRP: "The value `true`"  not parallel to "A value of `false`".

content/references/rippled-api/public-rippled-methods/transaction-methods/submit.md
mDuo13: It's kind of a long story, actually. If you don't retry or relay then any provisional failure becomes a permanent failure _and_ the server won't propose the transaction for inclusion in any validated ledgers.

By contrast, the default is that many transactions result in [`tec` codes](https://xrpl.org/tec-codes.html), and they are still shared with the network for inclusion in a validated ledger. Depending on the cause of the failure and what other transactions are happening, a transaction that provisionally failed may or may not result in success when (re)applied as part of a consensus-approved transaction set that builds (what eventually becomes) a validated ledger. But a lot of times it just fails again in the validated ledger, which means you destroy the XRP transaction cost (which is minimally inconvenient) and use up a sequence number (which may be preferable if you're sending several transactions in a short amount of time).

mDuo13: If you want to read more about this stuff, the [Reliable Transaction Submission](https://xrpl.org/reliable-transaction-submission.html#background) doc summarizes part of the problem, and the in-depth ["Validation" section in the Consensus article](https://xrpl.org/consensus.html#validation) may help you understand why that is.

content/references/rippled-api/public-rippled-methods/transaction-methods/submit.md
```diff
@@ -256,15 +256,23 @@ The response follows the [standard format][], with a successful result containin
...
-
-**Caution:** If this command results in an error messages, the message can contain the secret key from the request. (This is not a problem if the request contained a signed tx_blob instead.) Make sure that these errors are not visible to others.
+| `accepted`              | Boolean | _(Omitted in sign-and-submit mode)_ The value `true` indicates that the transaction was applied, queued, broadcast, or kept for later. A value of `false` indicates that none of those happened, so the transaction cannot possibly succeed as long as you do not submit it again and have not already submitted it another time. [New in: rippled 1.5.0][] |
+| `account_sequence_available` | Number | _(Omitted in sign-and-submit mode)_ The next [Sequence Number][] available for this account after all pending and [queued](transaction-queue.html) transactions. [New in: rippled 1.5.0][] |
+| `account_sequence_next` | number  | _(Omitted in sign-and-submit mode)_ The next [Sequence Number][] for this account after all pending transactions, not counting transactions in the queue. [New in: rippled 1.5.0][] |
```
cjcobb23: What is the difference between pending and queued? Might be useful to state the difference here; otherwise I think the description of these two can be confusing.

content/references/rippled-api/public-rippled-methods/transaction-methods/submit.md
```diff
@@ -256,15 +256,23 @@ The response follows the [standard format][], with a successful result containin
...
+| `broadcast`             | Boolean | _(Omitted in sign-and-submit mode)_ The value `true` indicates this transaction was broadcast to peer servers in the peer-to-peer XRP Ledger network. (Note: if the server has no peers, such as in [stand-alone mode](rippled-server-modes.html#reasons-to-run-a-rippled-server-in-stand-alone-mode), the server uses the value `true` for cases where it _would_ have broadcast the transaction.) The value `false` indicates the transaction was not broadcast to any other servers. [New in: rippled 1.5.0][] |
+| `kept`                  | Boolean | _(Omitted in sign-and-submit mode)_ The value `true` indicates that the transaction was kept to be retried later. [New in: rippled 1.5.0][] |
+| `queued`                | Boolean | _(Omitted in sign-and-submit mode)_ The value `true` indicates the transaction was put in the [Transaction Queue](transaction-queue.html), which means it is likely to be included in a future ledger version. [New in: rippled 1.5.0][] |
```
cjcobb23, line 298: What is the difference between queued and kept to be retried later? Might be useful to state the difference here.

cjcobb23 (commented):
Left some minor comments

content/references/rippled-api/public-rippled-methods/transaction-methods/submit.md
mDuo13: Rephrased to hopefully make it clearer.

content/references/rippled-api/public-rippled-methods/transaction-methods/submit.md
mDuo13: The difference is probably more complicated than can be explained here. I've rephrased this one slightly, and I'll let people read more in the queue documentation if they really need to.

> ... after all transactions that have been provisionally applied, but not transactions in the [queue](https://xrpl.org/transaction-queue.html).

content/references/rippled-api/public-rippled-methods/transaction-methods/submit.md
mDuo13: I believe the difference is that a `kept` transaction provisionally failed (and _not_ with a tec code) but it was cached to be retried over the next few ledgers. By comparison, a queued transaction was judged "likely to succeed" but wasn't applied yet because the open ledger was already full (and the transaction didn't meet the open ledger cost).

But I don't totally understand how the cache for failed transactions works (maybe @ximinez could explain it better)? In any case, I'd rather not complicate things by going into that level of detail in this parameter table.

mDuo13:
For the record, the only broken link (9 instances) is to the rippled v1.5.0 release tag, which should resolve itself when the release is published.

ghost (commented):
1. There was another field added not covered by this doc, validated_ledger_index, see https://github.com/ripple/rippled/pull/3125
2. May be it makes a sense to add new fields to an example.
3. "The current open ledger cost after processing this transaction." - before processing of this transaction. If actual fee was this value or larger then the transaction would be accepted.

PR #797, opened 2020-03-10
account_tx: fix CLI syntax.

mDuo13:
The `limit` and `offset` parameters in the CLI were swapped, and the example had the wrong syntax for `binary`/`count`/`descending` on the commandline.

This should fix the example up a little bit, at least until https://github.com/ripple/rippled/issues/3279 or https://github.com/ripple/rippled/issues/2926 touch up this area of the code.


LoisRP (commented):
For binary/count/descending, use the parameter name for true and omit for false.

cjcobb23 (approved):
looks good to me

PR #804, opened 2020-03-25
Fix incorrect number in cryptographic-keys.md

Frassle:
33 bytes + 4 bytes + 4 bytes = 41 bytes, not 40


mDuo13 (approved):
I was so used to working with 40-byte (160-bit) values that it completely slipped past my notice that this wouldn't be 160 bits. D'oh!

PR #808, opened 2020-03-26
Add "Configure gRPC" tutorial

mDuo13:
Adds a quick tutorial for enabling gRPC on your server. (This will, among other things, let you use Xpring SDK with your own rippled instance.)

### DIFF
--- a/content/tutorials/manage-the-rippled-server/configuration/configure-grpc.md
+++ b/content/tutorials/manage-the-rippled-server/configuration/configure-grpc.md
@@ -1,6 +1,6 @@
 # Configure gRPC
 
-The `rippled` server has an experimental binary API compatible with the [gRPC standard](https://grpc.io/). Currently, this API provides a subset of the full [`rippled` API](rippled-api.html). You can enable the gRPC API on your server with a new configuration stanza. [New in: rippled 1.5.0][]
+The `rippled` server has an experimental [gRPC API](https://grpc.io/). Currently, this API provides a subset of the full [`rippled` API](rippled-api.html). You can enable the gRPC API on your server with a new configuration stanza. [New in: rippled 1.5.0][]
 
 **Caution:** gRPC support in `rippled` v1.5.0 is experimental. Configuration settings and API formats are likely to have breaking changes in forthcoming versions.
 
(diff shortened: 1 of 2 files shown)


content/tutorials/manage-the-rippled-server/configuration/configure-grpc.md
```diff
@@ -0,0 +1,55 @@
+# Configure gRPC
+
+The `rippled` server has an experimental binary API compatible with the [gRPC standard](https://grpc.io/). Currently, this API provides a subset of the full [`rippled` API](rippled-api.html). You can enable the gRPC API on your server with a new configuration stanza. [New in: rippled 1.5.0][]
```
cjcobb23:
```suggestion
The `rippled` server has an experimental [gRPC](https://grpc.io/) API. Currently, this API provides a subset of the full [`rippled` API](rippled-api.html). You can enable the gRPC API on your server with a new configuration stanza. [New in: rippled 1.5.0][]
```
I wouldn't call it a binary API. gRPC doesn't explicitly call it a binary API, and a lot of the methods in our gRPC API have a binary flag, just as the JSON API does. I think calling it a binary API while simultaneously having binary flags in the API is confusing.
I think the previous phrasing of the sentence makes it seem like we rolled our own tech solution that is compatible with gRPC, but in actuality we are running an actual gRPC server within rippled.

Also, I wonder if it would be useful to spell out what subset of the API that we support via gRPC. We support tx, account_tx, submit, fee and account_info. What do you think?

cjcobb23 (changes requested):
Looks good. I think the first sentence should be rephrased (left a suggestion), but otherwise looks good.

content/tutorials/manage-the-rippled-server/configuration/configure-grpc.md
mDuo13: I'll rephrase the first part.

This document is not the right place to list what subset of the API is in gRPC. The gRPC overview doc (still forthcoming) will be the right place for that.

cjcobb23 (approved):
Looks good

content/tutorials/manage-the-rippled-server/configuration/configure-grpc.md
```diff
@@ -0,0 +1,55 @@
+# Configure gRPC
```
LoisRP, line 1: Worth explaining the initialism? (general-purpose Remote Procedure Call)

content/tutorials/manage-the-rippled-server/configuration/configure-grpc.md
```diff
@@ -0,0 +1,55 @@
...
+
+1. In your `rippled`'s config file, add a `[port_grpc]` configuration stanza.
+
+        [port_grpc]
+        port = 50051
+        ip = 0.0.0.0
+
+    The configurable fields are as follows:
+
+    - `port` field defines the port the server listens on for gRPC connections from client applications. The recommended port is `50051`.
```
LoisRP, line 27: `port` field - for consistency with `ip`, I would leave out field.

LoisRP (commented):
LGTM

PR #809, opened 2020-03-27
tx method: new min_ledger & max_ledger fields

mDuo13:
Documents https://github.com/ripple/rippled/pull/3145 .

Adds the new fields, and a caution about misunderstanding them. (They only affect the txnNotFound response and only if you specify both of them. If your server has the transaction, it'll still find it regardless of what `min_ledger` and `max_ledger` you use.)

Also adds a more detailed description of the txnNotFound response and the new `searched_all` field.

### DIFF
--- a/content/references/rippled-api/public-rippled-methods/transaction-methods/tx.md
+++ b/content/references/rippled-api/public-rippled-methods/transaction-methods/tx.md
@@ -11,7 +11,7 @@ An example of the request format:
 
 *WebSocket*
 
-```
+```json
 {
   "id": 1,
   "command": "tx",
@@ -21,7 +21,7 @@ An example of the request format:
 ```
 *JSON-RPC*
 
-```
+```json
 {
     "method": "tx",
     "params": [
@@ -34,7 +34,7 @@ An example of the request format:
 ```
 *Commandline*
 
-```
+```sh
 #Syntax: tx transaction [binary]
 rippled tx E08D6E9754025BA2534A78707605E0601F03ACE063687A0CA1BDDACFCD1698C7 false
 ```
@@ -48,7 +48,11 @@ The request includes the following parameters:
 | `Field`       | Type    | Description                                        |
 |:--------------|:--------|:---------------------------------------------------|
 | `transaction` | String  | The 256-bit hash of the transaction, as hex.       |
-| `binary`      | Boolean | (Optional, defaults to false) If true, return transaction data and metadata as hex strings instead of JSON |
+| `binary`      | Boolean | _(Optional)_ If `true`, return transaction data and metadata as binary [serialized](serialization.html) to hexadecimal strings. If `false`, return transaction data and metadata as JSON. The default is `false`. |
+| `min_ledger`  | Number  | _(Optional)_ If provided, search for the transaction in ledger versions whose [ledger index][] is this value or higher. If provided, you should also specify `max_ledger`. [New in: rippled 1.5.0][] |
+| `max_ledger`  | Number  | _(Optional)_ If provided, search for the transaction in ledger versions whose [ledger index][] is this value or lower. If provided, you should also specify `min_ledger`. [New in: rippled 1.5.0][] |
+
+**Caution:** This command may successfully find the transaction even if it is included in a ledger _outside_ the range of `min_ledger` to `max_ledger`.
 
 ## Response Format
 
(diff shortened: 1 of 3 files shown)


content/references/rippled-api/public-rippled-methods/transaction-methods/tx.md
```diff
@@ -48,7 +48,11 @@ The request includes the following parameters:
 | `Field`       | Type    | Description                                        |
 |:--------------|:--------|:---------------------------------------------------|
 | `transaction` | String  | The 256-bit hash of the transaction, as hex.       |
-| `binary`      | Boolean | (Optional, defaults to false) If true, return transaction data and metadata as hex strings instead of JSON |
+| `binary`      | Boolean | _(Optional)_ If `true`, return transaction data and metadata as binary [serialized](serialization.html) to hexadecimal strings. If `false`, return transaction data and metadata as JSON. The default is `false`. |
+| `min_ledger`  | Number  | _(Optional)_ If provided, search for the transaction in ledger versions whose [ledger index][] is this value or higher. If provided, you should also specify `max_ledger`. [New in: rippled 1.5.0][] |
+| `max_ledger`  | Number  | _(Optional)_ If provided, search for the transaction in ledger versions whose [ledger index][] is this value or lower. If provided, you should also specify `min_ledger`. [New in: rippled 1.5.0][] |
```
cjcobb23: I don't think the description in these two rows is correct. "If provided, search for the transaction in ledger versions whose index is this value of higher". This isn't actually what happens, just as the "Caution" below notes. It searches the entire transaction database, regardless of what min_ledger and max_ledger are. It only uses min_ledger and max_ledger if it doesn't find the transaction, basically to report if there are any gaps in the requested range.

cjcobb23 (changes requested):
Looks good, just left one requested change. Let me know what you think.

content/references/rippled-api/public-rippled-methods/transaction-methods/tx.md
seelabs: I agree with @cjcobb23; it acts as he describes above. Nice comment @cjcobb23!

mDuo13: I'm aware of this, but I struggled a lot with how to accurately and concisely characterize what these fields _actually_ do.

Something like, "If the transaction is not found, provides an [upper/lower] bound on the range of ledgers you would like the server to confirm that it searched" would be more accurate, but it takes some mental gymnastics to actually parse that. I'll keep thinking on how to describe it.

mtrippled:
LGTM :+1:

cjcobb23 (approved):
Looks good, much clearer now

seelabs (approved):
:+1:

content/references/rippled-api/public-rippled-methods/transaction-methods/tx.md
```diff
@@ -197,11 +201,79 @@ The response follows the [standard format][], with a successful result containin
...
+```
+
+<!-- MULTICODE_BLOCK_END -->
+
 ## Possible Errors
 
 * Any of the [universal error types][].
 * `invalidParams` - One or more fields are specified incorrectly, or one or more required fields are missing.
 * `txnNotFound` - Either the transaction does not exist, or it was part of an older ledger version that `rippled` does not have available.
+* `excessiveLgrRange` - The `min_ledger` and `max_ledger` fields of the request were more than 1000 apart.
```
LoisRP: "were more than 1000 apart." - I'm pretty sure this means 1001 and above causes the error, but wanted to confirm.

content/references/rippled-api/public-rippled-methods/transaction-methods/tx.md
```diff
@@ -197,11 +201,79 @@ The response follows the [standard format][], with a successful result containin
...
+* `invalidLgrRange` - The specified `min_ledger` was larger than the `max_ledger`, or one of those parameters was not a valid ledger index.
```
LoisRP: If you set the two to be equal, does that result in a single ledger being searched?

content/references/rippled-api/public-rippled-methods/transaction-methods/tx.md
LoisRP: "were more" - "are more" seems to be more common to use, although I understand the time sequencing issue here.

content/references/rippled-api/public-rippled-methods/transaction-methods/tx.md
mDuo13: Yes

content/references/rippled-api/public-rippled-methods/transaction-methods/tx.md
mDuo13: [Yes](https://github.com/ripple/rippled/blob/b54d672710227e494dd774c3ca3a4bc0a9976233/src/ripple/rpc/handlers/Tx.cpp#L120).

PR #810, opened 2020-03-28
New upcoming/current amendment block warnings

mDuo13:
Documents https://github.com/ripple/rippled/pull/3225.

### DIFF
--- a/content/references/rippled-api/api-conventions/response-formatting.md
+++ b/content/references/rippled-api/api-conventions/response-formatting.md
@@ -120,7 +120,7 @@ Example warning:
 ]
 ```
 
-This warning indicates that the one or more [amendments](amendments.html) to the XRP Ledger protocol are scheduled to become enabled if they maintain support from trusted validators, but the current server does not have an implementation for those amendments. If those amendments become enabled, the current server will become [amendment blocked](amendments.html#amendment-blocked), so you should [upgrade to the latest `rippled` version](install-rippled.html) as soon as possible.
+This warning indicates that the one or more [amendments](amendments.html) to the XRP Ledger protocol are scheduled to become enabled, but the current server does not have an implementation for those amendments. If those amendments become enabled, the current server will become [amendment blocked](amendments.html#amendment-blocked), so you should [upgrade to the latest `rippled` version](install-rippled.html) as soon as possible.
 
 The server only sends this warning if the client is [connected as an admin](get-started-with-the-rippled-api.html#admin-access).
 
@@ -131,7 +131,7 @@ This warning includes a `details` field with the following fields:
 | `expected_date`     | Number | The time that the first unsupported amendment is expected to become enabled, in [seconds since the Ripple Epoch][]. |
 | `expected_date_UTC` | String | The timestamp, in UTC, when the first unsupported amendment is expected to become enabled. |
 
-Due to the variation in ledger close times, these times are approximate. It is also possible that the amendment fails to maintain support from >80% of validators until the specified time, and does not become enabled at the expected time. The server will not become amendment blocked as long as the unsupported amendments do not become enabled.
+Due to the variation in ledger close times, these times are approximate. It is also possible that the amendment fails to maintain support from >80% of validators until the specified time, and does not become enabled at the expected time. The server will not become amendment blocked so long as the unsupported amendments do not become enabled.
 
 
 ### 1002. This server is amendment blocked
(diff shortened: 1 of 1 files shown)


content/references/rippled-api/api-conventions/response-formatting.md
```diff
@@ -87,3 +89,84 @@ HTTP Status: 200 OK
...
+      "expected_date" : 864030,
+      "expected_date_UTC" : "2000-Jan-11 00:00:30.0000000 UTC"
+    },
+    "id" : 1001,
+    "message" : "One or more unsupported amendments have reached majority. Upgrade to the latest version before they are activated to avoid being amendment blocked."
+  }
+]
+```
+
+This warning indicates that the one or more [amendments](amendments.html) to the XRP Ledger protocol are scheduled to become enabled if they maintain support from trusted validators, but the current server does not have an implementation for those amendments. If those amendments become enabled, the current server will become [amendment blocked](amendments.html#amendment-blocked), so you should [upgrade to the latest `rippled` version](install-rippled.html) as soon as possible.
```
LoisRP: "they maintain support from trusted validators" - This may be explained elsewhere, but it's not clear to me how this happens.

content/references/rippled-api/api-conventions/response-formatting.md
```diff
@@ -87,3 +89,84 @@ HTTP Status: 200 OK
...
+
+The server only sends this warning if the client is [connected as an admin](get-started-with-the-rippled-api.html#admin-access).
+
+This warning includes a `details` field with the following fields:
+
+| Field               | Value  | Description                                   |
+|:--------------------|:-------|:----------------------------------------------|
+| `expected_date`     | Number | The time that the first unsupported amendment is expected to become enabled, in [seconds since the Ripple Epoch][]. |
```
LoisRP, line 131: (Being pedantic here) - When is Number used and when is Integer used? Is it possible to have a non-integer number here?

content/references/rippled-api/api-conventions/response-formatting.md
```diff
@@ -87,3 +89,84 @@ HTTP Status: 200 OK
...
+| `expected_date_UTC` | String | The timestamp, in UTC, when the first unsupported amendment is expected to become enabled. |
+
+Due to the variation in ledger close times, these times are approximate. It is also possible that the amendment fails to maintain support from >80% of validators until the specified time, and does not become enabled at the expected time. The server will not become amendment blocked as long as the unsupported amendments do not become enabled.
```
LoisRP: (pedantic again) - as long as - I'd suggest "so long as" as that is used in the sense of a continuing condition to be satisfied.

content/references/rippled-api/api-conventions/response-formatting.md
```diff
@@ -87,3 +89,84 @@ HTTP Status: 200 OK
...
+  {
+    "id" : 1002,
+    "message" : "This server is amendment blocked, and must be updated to be able to stay in sync with the network."
+  }
+]
+```
+
+This warning indicates that the server is [amendment blocked](amendments.html#amendment-blocked) and can no longer remain synced with the XRP Ledger.
+
+The server administrator must [upgrade `rippled`](install-rippled.html) to a version that supports the activated amendments.
```
LoisRP, line 152: "must upgrade" - I think the implication is that if you don't upgrade, nothing will work, but it's not clear specifically how or when things fall apart.

content/references/rippled-api/api-conventions/response-formatting.md
mDuo13: It's explained on the linked `amendments.html` page.

Maybe the word "sustain" would be better than "maintain" here.

mDuo13: I decided to simplify the explanation here and rely on the amendments page to explain it in more detail.

The sentence now reads as follows:

> This warning indicates that the one or more [amendments](https://xrpl.org/amendments.html) to the XRP Ledger protocol are scheduled to become enabled, but the current server does not have an implementation for those amendments.

content/references/rippled-api/api-conventions/response-formatting.md
mDuo13: I'm phasing out the use of "Integer" for purely JSON data structures, as JSON doesn't make a distinction between number types; it has only one number type.

content/references/rippled-api/api-conventions/response-formatting.md
mDuo13: The linked "amendment blocked" description describes in detail exactly what doesn't work. (It's a lot of things.)

mDuo13:
Looks like the broken links are not related to this PR (except the expected v1.5.0 tag)

ximinez (commented):
I didn't get to this before it was merged, but everything looks good.

PR #822, opened 2020-04-03
Display manifest

mayurbhandary:
Manifests contain the following public information about validators:

Domain
Master Public Key
Ephemeral Public Key
Sequence number of manifest

The domain verification checker takes in a manifest to verify the domain and is a perfect place to display the contents of a manifest. This PR adds this functionality.


mDuo13:
I squashed and rebased the commits on the latest master to pull in the latest fixes

intelliot:
It's pretty difficult to review this since most of the changes look like whitespace changes. Also the tabs look huge on github's interface; usually I just use 2 spaces to indent (and this is enforced by the prettifier).

mDuo13:
Looks like it works, but I noticed a few things:

- The new `validator_info` command in rippled displays your manifest in base64. Can we extend this checker to accept base64? (Not in this PR, but in a later task.)
- I made a minimal change to support lowercase hexadecimal. Not sure I used the same browserify settings.
- The README needs to be updated with browserify (not webpack) instructions

mDuo13:
Did a find/replace to change tabs to 2 spaces each in that file. Hopefully that helps reviewability. There are still some whitespace changes, but not as many.

intelliot (commented):
I think this is replaced by https://github.com/ripple/xrpl-dev-portal/pull/825

PR #824, opened 2020-04-13
Fully document ledger stream response

intelliot:
Add listing of the fields included in the response when subscribing to the `ledger` stream.

### DIFF
--- a/content/references/rippled-api/public-rippled-methods/subscription-methods/subscribe.md
+++ b/content/references/rippled-api/public-rippled-methods/subscription-methods/subscribe.md
@@ -109,11 +109,11 @@ An example of a successful response:
 
 The response follows the [standard format][]. The fields contained in the response vary depending on what subscriptions were included in the request.
 
-* `accounts` and `accounts_proposed` - No fields returned
+* `accounts` and `accounts_proposed` - No fields returned.
 * *Stream: server* - Information about the server status, such as `load_base` (the current load level of the server), `random` (a randomly-generated value), and others, subject to change.
-* *Stream: transactions*, *Stream: transactions_proposed*, *Stream: validations*, and *Stream: consensus* - No fields returned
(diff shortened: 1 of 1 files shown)


content/references/rippled-api/public-rippled-methods/subscription-methods/subscribe.md
```diff
@@ -109,11 +109,11 @@ An example of a successful response:
...
-* `books` - No fields returned by default. If `"snapshot": true` is set in the request, returns `offers` (an array of offer definition objects defining the order book)
+* *Stream: transactions*, *Stream: transactions_proposed*, *Stream: validations*, and *Stream: consensus* - No fields returned.
+* *Stream: ledger* - Information about the ledgers on hand and current fee schedule. Except for `type` and `txn_count`, this includes the same fields as a [ledger stream message](#ledger-stream): `fee_base` (current base fee for transactions in XRP), `fee_ref` (current base fee for transactions in fee units), `ledger_hash` (hash of the latest validated ledger), `ledger_index` (index of the latest validated ledger), `ledger_time` (close time of the latest validated ledger), `reserve_base` (minimum reserve for accounts), `reserve_inc` (owner reserve for each object), and `validated_ledgers` (ranges of ledgers that the server has available).
```
mDuo13: - I think this is too much info to provide at this point. (The existing doc makes the poor decision to list off fields.) It would be better to cut it off after the link to the #ledger-stream anchor.
- It seems like a bug to me that the response here is missing `txn_count`. Aside from `type` (which is provided at the top level), the contents of `result` _should_ exactly match the `ledger` stream messages. I've opened https://github.com/ripple/rippled/issues/3354 to track this.

```suggestion
* *Stream: ledger* - Information about the ledgers on hand and current fee schedule. This includes the same fields as a [ledger stream message](#ledger-stream), except that it omits the `type` and `txn_count` fields.
```

mDuo13 (changes requested):
Good catch, and thanks for cleaning up the inconsistencies with periods while you're at it.

intelliot:
I should’ve mentioned that I opened an issue as well - https://github.com/ripple/rippled/issues/3349 - although your issue is more clear :)

mDuo13 (approved):

PR #831, opened 2020-04-28
Known Amendments as of 2020-04-28

mDuo13:
- Updates statuses of amendments that currently have a majority
- Adds blog posts for some other features that are open for voting
- Adds description for fix1781 (@seelabs please review this part)

### DIFF
--- a/content/concepts/consensus-network/amendments/known-amendments.md
+++ b/content/concepts/consensus-network/amendments/known-amendments.md
@@ -3,18 +3,19 @@
 
 The following is a comprehensive list of all known amendments and their status on the production XRP Ledger:
 
-| Name                            | Introduced | Status                              |
-|:--------------------------------|:-----------|:------------------------------------|
+| Name                            | Introduced | Status                        |
+|:--------------------------------|:-----------|:------------------------------|
 | [CryptoConditionsSuite][]       | TBD        | [In Development: TBD]( "BADGE_LIGHTGREY") |
 | [OwnerPaysFee][]                | TBD        | [In Development: TBD]( "BADGE_LIGHTGREY") |
 | [Tickets][]                     | TBD        | [In Development: TBD]( "BADGE_LIGHTGREY") |
-| [fixQualityUpperBound][]        | v1.5.0     | [Open for Voting: TBD]( "BADGE_80d0e0") |
-| [RequireFullyCanonicalSig][]    | v1.5.0     | [Open for Voting: TBD]( "BADGE_80d0e0") |
-| [DeletableAccounts][]           | v1.4.0     | [Open for Voting: TBD]( "BADGE_80d0e0") |
-| [fixCheckThreading][]           | v1.4.0     | [Open for Voting: TBD]( "BADGE_80d0e0") |
-| [fixPayChanRecipientOwnerDir][] | v1.4.0     | [Open for Voting: TBD]( "BADGE_80d0e0") |
-| [Checks][]                      | v0.90.0    | [Open for Voting: TBD]( "BADGE_80d0e0") |
-| [FlowCross][]                   | v0.70.0    | [Open for Voting: TBD]( "BADGE_80d0e0") |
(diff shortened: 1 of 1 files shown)


content/concepts/consensus-network/amendments/known-amendments.md
```diff
@@ -3,18 +3,19 @@
 
 The following is a comprehensive list of all known amendments and their status on the production XRP Ledger:
 
-| Name                            | Introduced | Status                              |
-|:--------------------------------|:-----------|:------------------------------------|
+| Name                            | Introduced | Status                        |
+|:--------------------------------|:-----------|:------------------------------|
+| [fix1781][]                     | v1.6.0-b1  | [In Development: TBD]( "BADGE_LIGHTGREY") |
 | [CryptoConditionsSuite][]       | TBD        | [In Development: TBD]( "BADGE_LIGHTGREY") |
```
carlhua, line 9: This is no longer going to be published right? @seelabs  Do we want to do something about this line?

carlhua (approved):
LGTM

content/concepts/consensus-network/amendments/known-amendments.md
mDuo13: What's no longer going to be published?

content/concepts/consensus-network/amendments/known-amendments.md
```diff
@@ -339,6 +340,18 @@ The fix1623 amendment has no effect on [CheckCash transactions][] for a fixed am
...
+[fix1781]: #fix1781
+
+| Amendment ID                                                     | Status    |
+|:-----------------------------------------------------------------|:----------|
+| 25BA44241B3BD880770BFA4DA21C7180576831855368CBEC6A3154FDE4A7676E | In Development |
+
+Fixes a bug where certain XRP endpoints were not checked when detecting circular paths.
+
+Without this amendment, in some circumstances it may be possible to create and execute "circular" payments where the same account sends and receives the same currency. With this amendment, those payments fail with the [`temBAD_PATH_LOOP` result code](tem-codes.html) instead.
+
```
seelabs: I'd reword this slightly so it talks about payment paths and path steps rather than the "same account". We disallow such payment even if different accounts are involved. Maybe
```
Without this amendment, it was possible to have a payment path where the input to the path
is XRP, and an intermediate path step also output XRP. This is a "circular" payment and it should 
have been disallowed. With this amendment, those payments fail with the
[`temBAD_PATH_LOOP` result code](tem-codes.html) instead.
```

content/concepts/consensus-network/amendments/known-amendments.md
seelabs: The project may be resurrected at some point, but there are no short term plans to merge cryptoconditions. Up to you if you want to remove the line.

mDuo13: I'll wait until there's a `rippled` beta that removes the amendment.

content/concepts/consensus-network/amendments/known-amendments.md
mDuo13: Reworded slightly to use present tense and to clarify _why_ that qualifies as a loop:

> Without this amendment, it is possible to have a [payment path](paths.html) where the input to the path is XRP, and an intermediate path step also outputs XRP. This is a "circular" payment, because it could be completed more efficiently with a shorter path that skips everything up to the intermediate step that outputs XRP.

seelabs: I'd leave off the "it could be completed more efficiently", that's not the motivation for the rule.

The reason we disallow payments like this has to do with the inner workings of the payment engine. We first "execute" the path backwards to see how much we need to put in to get the desired amount out, then we "execute" the path forward (because we may not be able to get the desired amount out). If we had circular payments, then one step on the path could defund another step on the path and paths would act differently in the forward direction than in the reverse direction. To keep the code simple, we do not allow such cases. I would not put that in the explanation (that's too far into the weeds), but wanted to explain the rational for the rule.

PR #835, opened 2020-05-07
Update terminology: XRPL Classic Address

intelliot:
These are now called XRPL classic addresses


mDuo13 (approved):
:+1: 

There are probably a lot more places to go about instituting this terminology change, but this is a good start.

PR #837, opened 2020-05-08
DeletableAccounts enabled

mDuo13:
Now that DeletableAccounts has become enabled, I've updated the various mentions in the docs regarding its status.

(I also caught a few cases in the Japanese-translated docs that hadn't been updated for MultiSignReserve and fixed those.)


scottschurr (approved):
:+1: All these changes look great to me.  Thanks for all the upkeep!

ximinez (changes requested):
Would it be appropriate to add the new `deletion_blockers_only` field to `account_objects` to this PR, or should that be a different one?

mDuo13:
@ximinez Different PR for that.

PR #840, opened 2020-05-13
/vl/ method

mDuo13:
Documents the `/vl/` method introduced by https://github.com/ripple/rippled/pull/3072 including the content of validator lists.


content/references/rippled-api/validator-list.md
```diff
@@ -0,0 +1,124 @@
+# Validator List Method
+
+The validator list method is a special API endpoint for fetching the latest version of a current trusted validator list a `rippled` server is using. This often represents the exact list of validators a server trusts. [New in: rippled 1.5.0][]
```
gbarr01: Is it grabbing the latest version (as in number)? Or that latest VL itself? 

If it's the former, consider this tweak: 

> The validator list method is a special API endpoint that fetches the latest version of a trusted validator list in use by a `rippled` server.

If it's the latter: 

> The validator list method is a special API endpoint that fetches the latest trusted validator list in use by a `rippled` server.

content/references/rippled-api/validator-list.md
```diff
@@ -0,0 +1,124 @@
...
+
+A "manifest" contains information uniquely identifying a person or organization involved in the consensus process, either a **validator** or a **list publisher**. A validator's manifest contains the _public_ information from that [validator's token](run-rippled-as-a-validator.html#3-enable-validation-on-your-rippled-server). A list publisher's manifest provides information about the list publisher. Both are typically encoded to binary in the XRP Ledger's standard [binary serialization format](serialization.html). (There is no standard JSON representation of a manifest.)
+
+One of the main purposes of manifests relates to rotating validator keys. When a validator changes its ephemeral key pair, the validator publishes a new manifest with to share its new ephemeral public key, using the validator's master key pair to sign the manifest to prove its authenticity. A validator uses its ephemeral key pair to sign validations as part of the [consensus process](consensus.html) and uses its master key pair only to sign new manifests and similar.
```
gbarr01: Couple typos here: 

```
< publishes a new manifest with to share its new ephemeral public key
---
> publishes a new manifest to share its new ephemeral public key
```


Not sure if "similar" should be removed or something added at the end:

```
< uses its master key pair only to sign new manifests and similar.
---
> uses its master key pair only to sign new manifests.
```

gbarr01 (commented):
One suggestion, two typos. Otherwise, looks great.

ximinez (approved):

content/references/rippled-api/validator-list.md
mDuo13: Since a server can use more than one list, I don't want to say "the latest... list" but I see your point. Going with:

>... that fetches a current, trusted validator list ...

content/references/rippled-api/validator-list.md
mDuo13: With "and similar" I was thinking about special cases like the old domain verification process. But now that we have the new process, I guess it's not really necessary. Instead I added a note about how manifests relate to validator tokens.

mDuo13:
Looks like the CI is not getting the PR links quite right but nothing's broken for the actual site.

PR #841, opened 2020-05-13
Migrate to Dactyl v0.10 locale tools

mDuo13:
Moves the strings to `.po` files, and switch to using `{% trans %}` tags in the templates.

This builds the exact same output (Dactyl v0.10+ required) but should be a bit easier to manage in the future, especially since `.po` files are a standard format that professional translators are usually familiar with.

Another nice thing about this is the locale files can be reused across multiple targets (in this case, the "Data API Only" and XRP-API only" targets don't need the strings copied into their target definitions) and Dactyl won't raise an error if you try to build a target that doesn't have all the string fields defined (e.g. building ad-hoc markdown files using the existing templates).


ryangyoung (approved):
This looks good, afaict. I'm excited about this new framework!

PR #845, opened 2020-05-18
Diagnosing: getRippledInfo script

mDuo13:
Mentions the `getRippledInfo` script (where it is, how to run it, what to expect) in the "Diagnosing Problems" page.

Documents https://github.com/ripple/rippled/pull/3260

### DIFF
--- a/content/tutorials/manage-the-rippled-server/troubleshooting/diagnosing-problems.md
+++ b/content/tutorials/manage-the-rippled-server/troubleshooting/diagnosing-problems.md
@@ -143,13 +143,13 @@ To use the script:
 
 2. Look over the output file for sensitive information.
 
-    Although the script attempts to scrub sensitive information from the information it collects, it's important to look it over to be sure before posting the output publicly. For example, the script outputs detailed information about your server hardware, and you may want to remove some sections for privacy reasons. Use a text editor to read the output file and to remove anything you don't want to post.
+    The script attempts to scrub sensitive information from the output, such as validator keys or tokens. However, you should still check the output before posting publicly, as a precaution. For example, the script outputs detailed information about your server hardware, and you may want to remove some sections for privacy reasons. Use a text editor to read the output file and to remove anything you don't want to post.
 
         nano /tmp/ripple_info.Xo8Xr/rippled_info.md
 
 3. Upload the output file where others can see it.
 
-    For example, you can upload a file directly to [GitHub Gist](https://gist.github.com/), [Pastebin](https://pastebin.com/), or a similar service. If you are running `rippled` on a remote server, you may find it easier to first transfer the file to a machine with a web browser, using `scp` or a similar tool.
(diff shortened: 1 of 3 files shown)


content/tutorials/manage-the-rippled-server/troubleshooting/diagnosing-problems.md
```diff
@@ -112,6 +115,43 @@ It is normal for a `rippled` the server to print many warning-level (`WRN`) mess
...
+          any sensitive information from this file, but you
+          should verify before posting.
+        ####################################################
+
+    The script collects the output of numerous commands and writes them to a temporary file. The filename is randomized with a string of letters and numbers (case-sensitive), for example: `/tmp/ripple_info.Xo8Xr/rippled_info.md`
+
+
+2. Look over the output file for sensitive information.
+
+    Although the script attempts to scrub sensitive information from the information it collects, it's important to look it over to be sure before posting the output publicly. For example, the script outputs detailed information about your server hardware, and you may want to remove some sections for privacy reasons. Use a text editor to read the output file and to remove anything you don't want to post.
```
ddibble-rip: Look over is problematic, especially for English as a 2nd language (too similar to overlook). Why not use something more active like the following, and also state what they might want to do with sensitive info they locate. 

View the output file for any sensitive information you may want to remove before you post.

The pronoun "it" in the second half of the first sentence is a bit confusing. Might want to restate that they should are "view the generated output file",  as "it" could be misinterpreted as the script itself. 

How does the script attempt to scrub sensitive information? Does it replace base endpoints with variables? Does it remove host ports?

content/tutorials/manage-the-rippled-server/troubleshooting/diagnosing-problems.md
```diff
@@ -112,6 +115,43 @@ It is normal for a `rippled` the server to print many warning-level (`WRN`) mess
...
+
+        nano /tmp/ripple_info.Xo8Xr/rippled_info.md
+
+3. Upload the output file where others can see it.
+
+    For example, you can upload a file directly to [GitHub Gist](https://gist.github.com/), [Pastebin](https://pastebin.com/), or a similar service. If you are running `rippled` on a remote server, you may find it easier to first transfer the file to a machine with a web browser, using `scp` or a similar tool.
```
ddibble-rip: Delete "For example" and change "a" to "the"

You can upload the file directly to [GitHub Gist](https://gist.github.com/), [Pastebin](https://pastebin.com/), or a similar service.

ddibble-rip: It might be good to also state what type of information is in the output file (or procvide a snippet of information from the output file). ATM, it simply says "system" information. Is that services that are part of the rippled installation? The servers and db that rippled is connected to?

mDuo13: It's honestly a lot of information. And I would expect the exact scope may change without warning in future releases. Seems like some of it depends on whether the script can collect it from your installation (in other words, the script just omits stuff it can't query) as well.

I was thinking of attaching a sample output I'm not really sure I have a good sample to work with. In any case, people should be looking over the output before they upload it, so they'll see what it actually collected then.

manojsdoshi (approved):
LGTM

PR #854, opened 2020-06-11
Changing pubkey to public_key

manojsdoshi:
Passing `pubkey` as a request parameter doesn't do anything. We need to pass in `public_key` to get the desired result.


mDuo13:
Good catch. I was fooled by [an inaccurate comment in the code](https://github.com/ripple/rippled/blob/f00f263852c472938bf8e993e26c7f96f435935c/src/ripple/rpc/handlers/CrawlShards.cpp#L35), but [you're right, it's actually public_key](https://github.com/ripple/rippled/blob/f00f263852c472938bf8e993e26c7f96f435935c/src/ripple/rpc/handlers/CrawlShards.cpp#L60).

mDuo13 (approved):

PR #857, opened 2020-06-21
Update xrp-ledger-toml.md

kchaot:
Added the Universal Payment Identifier to the section groups PRINCIPALS and ACCOUNTS. A better way to send or receive payments.


mDuo13:
Interesting idea! I'd like to see a bit more consensus from the community that this is the correct way/place to convey this information, but I think this structure at least makes sense. I'm also curious if people have come up with any use cases that might prompt someone to put their PayID into their xrp-ledger.toml.

mDuo13:
Given the lack of comments either way, I'm going to close this for now. Feel free to reopen it if we see more traction on this.

PR #858, opened 2020-06-23
Update install for latest distro releases

mDuo13:
Default to Focal Fossa, list our other supported releases.


mDuo13:
Per @manojsdoshi:

> Ubuntu 19.04 (Disco Dingo) End of Life reached on January 23 2020 so no point mentioning it in our docs. We cant do a fresh install on Ubuntu 19.04 since we cant install ca-certificates on Ubuntu 19.04 any more

PR #859, opened 2020-06-24
RequireFullyCanonicalSig expected

mDuo13:
Also syncs the Japanese translation of known amendments up to the latest status.

(blog post for RFCS is the expected URL, but not live yet)


mDuo13:
As an aside, here's an EN←→JA glossary for amendment statuses.

| English Term | Japanese Term |
|---|---|
| TBD                       | 未定 |
| In Development          |   開発中 |
| Open for Voting           | 投票中 |
| Expected                 |  予定 |
| Enabled                   | 有効 |
| Vetoed                    | 禁止 |
| Removed in vX.Y.Z   |       vX.Y.Zで削除 |

PR #865, opened 2020-07-09
Freezes: touch up "No Freeze" language

mDuo13:
Hopefully makes this section read a little better. Fixes DOC-1297, regarding this passage. Doesn't assume that issuers are businesses.

### DIFF
--- a/content/concepts/issued-currencies/freezes.md
+++ b/content/concepts/issued-currencies/freezes.md
@@ -69,7 +69,7 @@ The No Freeze setting has two effects:
 * The issuing address can no longer enable Individual Freeze on trust lines to any counterparty.
 * The issuing address can still enable Global Freeze to enact a global freeze, but the address cannot _disable_ Global Freeze.
 
-The XRP Ledger cannot force an issuer to honor the obligations that its issued funds represent, so No Freeze does not protect customers from an issuer defaulting on its obligations. However, No Freeze ensures that the Global Freeze feature is not used unfairly against some customers.
+The XRP Ledger cannot force an issuer to honor the obligations that its issued funds represent, so No Freeze does stop an issuer from defaulting on its obligations. However, No Freeze ensures that an issuer does not use the Global Freeze feature unfairly against specific users.
 
 The No Freeze setting applies to all currencies issued to and from an address. If you want to be able to freeze some currencies but not others, you should use different addresses for each currency.
 
--- a/content/concepts/issued-currencies/freezes.md
+++ b/content/concepts/issued-currencies/freezes.md
@@ -60,16 +60,16 @@ An address can always enable the Global Freeze setting. However, if the address 
 
 ## No Freeze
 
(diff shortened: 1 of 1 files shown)


content/concepts/issued-currencies/freezes.md
```diff
@@ -60,16 +60,16 @@ An address can always enable the Global Freeze setting. However, if the address
 
 ## No Freeze
 
-The **No Freeze** feature is a setting on an address that permanently gives up the ability to freeze counterparties' issued currencies. A business can use this feature to treat its issued funds as "more like physical money" in the sense that the business cannot interfere with customers trading it among themselves.
+The **No Freeze** feature is a setting on an address that permanently gives up the ability to freeze issued currencies arbitrarily. An issuer can use this feature to treat its issued funds as "more like physical money" in the sense that the issuer cannot interfere with counterparties trading the currency among themselves.
```
ryangyoung, line 63: Why "arbitrarily"?

content/concepts/issued-currencies/freezes.md
```diff
@@ -60,16 +60,16 @@ An address can always enable the Global Freeze setting. However, if the address
...
 Reminder: XRP already cannot be frozen. The No Freeze feature only applies to other currencies issued in the XRP Ledger.
 
-The NoFreeze setting has two effects:
+The No Freeze setting has two effects:
 
 * The issuing address can no longer enable Individual Freeze on trust lines to any counterparty.
 * The issuing address can still enable Global Freeze to enact a global freeze, but the address cannot _disable_ Global Freeze.
 
-The XRP Ledger cannot force a financial institution to honor the obligations that its issued funds represent, so giving up the ability to enable a Global Freeze cannot protect customers. However, giving up the ability to _disable_ a Global Freeze ensures that the Global Freeze feature is not used unfairly against some customers.
+The XRP Ledger cannot force an issuer to honor the obligations that its issued funds represent, so No Freeze does not protect customers from an issuer defaulting on its obligations. However, No Freeze ensures that the Global Freeze feature is not used unfairly against some customers.
```
ryangyoung: If we're moving away from "business" and "financial institution" to "issuer" (which makes sense), we should probably change other instances in this article. There's at least one more on [line 33](https://gitlab.ops.ripple.com/ripplenet/docs/product-docs/-/blob/knowledge-center/documents/ripplenet/tools-services/test-harness/th-sender-tests.md#L3). 

Similarly, does it make sense to use "customer" if we're not talking about businesses anymore?

ryangyoung (commented):
Had a couple of questions, but otherwise looks like a good start to clarifying tricky, subtle language issues.

content/concepts/issued-currencies/freezes.md
mDuo13: The word "arbitrarily" represents the intended use of No Freeze and the (somewhat unintuitive) way that intended use is reflected in its functionality. Namely, that the issuer can't use individual freeze, or some kind of on-and-off use of global freeze, to temporarily stop the issued currency from being used for specific purposes or at specific times; but they can nevertheless enable a global freeze to (permanently!) stop the issued currency from being used by everyone for everything—a drastic step that's still available to help if there's a serious security issue, but is too broad/blunt/cumbersome to use for anything less.

PR #868, opened 2020-07-10
Adds flow-of-funds diagram to list-xrp-as-an-exchange.md

rmurphy23:
Changes include:

- Adding a Flow of Funds diagram to the Flow of Funds section of list-xrp-as-an-exchange.md

Context: https://ripplelabs.atlassian.net/browse/DOC-3101 

@ryangyoung can you review this?
@mDuo13 can you review this?


mDuo13:
Looks like the image itself didn't get added? You might need to explicitly `git add` the file...

rmurphy23:
> Looks like the image itself didn't get added? You might need to explicitly `git add` the file...

Weird I explicitly added the the .png before the commit and I just attempted to explicitly add it again and it says my branch is up to date; I'll probably make a different branch tomorrow.

rmurphy23:
Capture

mDuo13:
!rebuild

mDuo13:
Oh, I'm seeing it now. Maybe it was on my end. (BTW, the above message is to re-run the link checker, because it looks like the previous failure was a fluke.)

mDuo13:
OK, so now that I can see the diagram:

1. Do you have the source file you made the diagram from? We try to commit those to the repo if possible (in `_content/img-sources/`) so that future editors can maintain/update the diagram. If you used an online tool or something, then it's fine to just leave a comment with a link in the doc near where you include the diagram (as long as it's possible for other editors to get access from the link)
2. Instead of Ripple triskelions, the diagram should use the XRP "X" mark.
3. The diagram is a bit confusing with regards to the actual flow of funds. For example, depositing XRP at an exchange typically means sending a transaction (on the XRP Ledger) to the exchange's cold (or hot) wallet; however, the diagram has the deposit on the opposite side from the hot/cold wallet. I'd be happy to talk you through my thoughts in more detail in a call or something.

mDuo13:
P.S. the link checker failures are because Wikipedia deleted the Base58 article. Not related to this PR.

rmurphy23:
@mDuo13 Let me know what you think of the updated diagrams.

mDuo13:
I'd like to revisit this later with a different approach; I don't think this PR represents the best path forward now.

PR #869, opened 2020-07-14
KA: v1.5.0 amendments enabled, add v1.6.0-b8 amendments

mDuo13:
- New fixAmendmentMajorityCalc and NegativeUNL amendments added in v1.6.0-b8. (These descriptions are stubs, to be expanded later.)
- Updates statuses of fixQualityUpperBound and RequireFullyCanonicalSig amendments now that they're live
- Same updates on the Japanese version of the page
- Changes the Japanese version to use UNIX-style line breaks (linefeed-only) instead of Windows-style (carriage return + linefeed)

### DIFF
--- a/content/concepts/consensus-network/amendments/known-amendments.md
+++ b/content/concepts/consensus-network/amendments/known-amendments.md
@@ -3,15 +3,17 @@
 
 The following is a comprehensive list of all known amendments and their status on the production XRP Ledger:
 
+**Tip:** This list is updated manually. For a live view of amendment voting, see the [XRPScan Amendments Dashboard](https://xrpscan.com/amendments).
+
 | Name                            | Introduced | Status                        |
 |:--------------------------------|:-----------|:------------------------------|
-| [fixAmendmentMajorityCalc][]    | v1.6.0-b8  | [In Development: TBD]( "BADGE_LIGHTGREY") |
-| [NegativeUNL][]                 | v1.6.0-b8  | [In Development: TBD]( "BADGE_LIGHTGREY") |
-| [HardenedValidations][]         | v1.6.0-b5  | [In Development: TBD]( "BADGE_LIGHTGREY") |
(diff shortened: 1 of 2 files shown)


content/concepts/consensus-network/amendments/known-amendments.md
```diff
@@ -504,6 +518,16 @@ Without this amendment, the owner reserve for a SignerList ranges from 15 to 50
 With this amendment enabled, the owner reserve for a new SignerList is 5 XRP, regardless of the number of signers. The reserve requirement for previously-created SignerList objects remains unchanged. To reduce the reserve requirement of SignerList objects created before this amendment was enabled, use a [SignerListSet transaction](signerlistset.html) to replace the SignerList after this amendment has been enabled. (The replacement can be identical to the previous version.)
 
 
+## NegativeUNL
```
nbougalis, line 523: We should not document at this time. There's no corresponding amendment for it at this time.

nbougalis (changes requested):

content/concepts/consensus-network/amendments/known-amendments.md
mDuo13: Aha, I looked at [the wrong `-b8` PR](https://github.com/ripple/rippled/pull/3459). The [right `-b8` PR](https://github.com/ripple/rippled/pull/3483) doesn't list it. My source code crawler found [the commented-out amendment](https://github.com/ripple/rippled/blob/7b048b423e8ae08a54018a89231d050b9f562855/src/ripple/protocol/impl/Feature.cpp#L135) and added it to the list of things "in development".

Maybe it's a good idea to leave this documented but change the "introduced" version to TBD?

nbougalis: >Maybe it's a good idea to leave this documented but change the "introduced" version to TBD?

I think that's _probably_ fine.

PR #878, opened 2020-07-30
`token_sequence` must be incremented in `validator-keys.json` backups

jscottbranson:
Docs currently indicate that validators should backup `validator-keys.json`, and not edit the backup. This causes problems when validator operators generate new validation tokens using backups of the `validator-keys.json` file. Specifically:
1. Validator operator copies backup of `validator-keys.json` to from backup location to validator
2. Operator generates a new token on validator
3. Operator deletes `validator-keys.json` from the validator

If the validator operator does not then increment the "token_sequence" in the backed up `validator-keys.json`, the next time the operator repeats the above three steps, their validations will be ignored by the network, as the resultant token will have a sequence number that is <= the last observed number.


content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -72,7 +72,7 @@ In a location **not** on your validator:
...
...tore the generated `validator-keys.json` key file in a secure, offline, and recoverable location, such as an encrypted USB flash drive. Do not store keys on the validator where you intend to use the keys. If your `secret_key` is compromised, [revoke the key](https://github.com/ripple/validator-keys-tool/blob/master/doc/validator-keys-tool-guide.md#key-revocation) immediately. Do not modify the contents of `validator-keys.json`, except to increment the `token _sequence` after generating a new validation token using a copy of the file. For example, if you copy your secure, backed up `validation-keys.json` to your validator to generate a new token, generating the token will increase the `token_sequence` field. If you do not update the `token_sequence` field in your secure backup, then the next token you generate will have the same or lower sequence number as the token you are currently using. The network will ignore validations that have a sequence number that is less than the previously highest observed number. Alternately, simply copy `validator-keys.json` to your backup location every time you generate a new validation token, so your backup version has the correct `token_sequence`.
```
mDuo13: 1. You should not need to copy your `validation-keys.json` file to your validator. You should, ideally, run the validator keys tool on a separate machine to update the keys.

2. I don't want to instruct people to update the `token_sequence` manually except in special cases, because I'd like these instructions to be future-proof against changes in the validator keys tool. For example, @nbougalis was thinking tokens could switch to a time-based instead of sequence-based system for marking new tokens. There's probably a separate place (troubleshooting?) to cover the [special case](https://github.com/ripple/xrpl-dev-portal/issues/879).

2. I'm suggesting a couple edits to remove some words like [simply](https://justsimply.dev/) as a best practice.

```suggestion
      **Warning:** Store the generated `validator-keys.json` key file in a secure, offline, and recoverable location, such as an encrypted USB flash drive. Do not store keys on the validator where you intend to use the keys. If your `secret_key` is compromised, [revoke the key](https://github.com/ripple/validator-keys-tool/blob/master/doc/validator-keys-tool-guide.md#key-revocation) immediately. Do not modify the contents of `validator-keys.json`, except to update the backup after generating a new token. If you generate more than one token from the same backup without updating, the network ignores the later tokens because they use the same `token_sequence` number.
```

mDuo13 (changes requested):

mDuo13:
Thank you for pointing out this issue, by the way.

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
jscottbranson: All three points make sense, and I like how much more concisely you get the message across.

jscottbranson:
my bad - didn't mean to click requesting your review. I'm still good with your language.

mDuo13 (approved):

mDuo13 (approved):

PR #880, opened 2020-07-31
WIP: Adds invariant-checking.md

rmurphy23:
Changes include:

- Adds invariant-checking.md
- Adds content to invariant-checking.md
- Configures dactyl-config.yml to support invariant-checking.md

Context: https://github.com/ripple/xrpl-dev-portal/issues/587


content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,404 @@
...
+- [No Bad Offers](#no-bad-offers)
+- [No Zero Escrow](#no-zero-escrow)
+- [Valid New Account Root](#valid-new-account-root)
+
+
+### Transaction Fee Check
+
+- **Condition(s) Checked:**
+
+    - [Trnsaction fees](https://xrpl.org/rippleapi-reference.html#transaction-fees) should never be negative nor larger than the transaction itself.
```
ximinez: Typo: "Transaction fees"

Clarify: "not larger than specified by the transaction itself."

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,404 @@
...
+        ReadView const&,
+        beast::Journal const&);
+};
+```
+
+### Account Roots Not Deleted
+
+- **Condition(s) Checked:**
+    
+    - An account ledger entry cannot be removed.
```
ximinez: An account ledger entry can be removed by an AccountDelete transaction, but this invariant checks that exactly 1 is deleted by a successful AccountDelete.

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,404 @@
...
+        TER const,
+        XRPAmount const,
+        ReadView const&,
+        beast::Journal const&);
+};
+```
+
+### Ledger Entry Types Match
+
+- **Condition(s) Chceked:**
```
ximinez: Typo: "Checked"

ximinez (changes requested):

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,404 @@
+# Invariant Checking
+
+This article provides a high-level overview of Invariant Checking, why it exist, how it works, and lists active invariant checks.
+
+When building applications on the XRP Ledger, it is crucial to understand the idea of Invariant Checks so as not to be surprised by specific error codes.
```
mDuo13: I would say this is not the main reason to understand the idea of invariant checks. In fact, I would go so far as to say it's _not_ crucial to understand invariant checks when building applications on the XRP Ledger, similar to how it is not crucial to understand the concept of a copilot to ride an airplane.

```suggestion
Like many safety features, we all hope that invariant checking never actually needs to do anything. However, it can be useful to understand the XRP Ledger's invariants because they define hard limits on the XRP Ledger's transaction processing, and to recognize the problem in the unlikely event that a transaction fails because it violated an invariant check.
```

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,404 @@
+# Invariant Checking
+
+This article provides a high-level overview of Invariant Checking, why it exist, how it works, and lists active invariant checks.
```
mDuo13: We should use consistent form and capitalization with regards to "Invariant Checks" "Invariant Checking". 

I'm leaning towards leaving both in lowercase, and preferring the word "invariants" instead of the phrase "invariant checks" (in part so as to head off any confusion about [the other type of Checks](https://xrpl.org/checks.html)...)

Some definitions that may help guide consistent usage:

- An _invariant_ is a rule that should always, without exception, be true. For example, "New XRP cannot be created."
- In the XRP Ledger, _invariant checking_, is a system where code automatically confirms that transaction processing does not break the invariants. If a transaction's execution would break an invariant, the invariant checking system fails the transaction.

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,404 @@
...
+
+When building applications on the XRP Ledger, it is crucial to understand the idea of Invariant Checks so as not to be surprised by specific error codes.
+
+## Introduction
+
+Invariant checking is a protection enhancement to the XRP Ledger and reinforces the critical properties of the XRP Ledger.
```
mDuo13: This sentence could be a lot clearer and more specific.

```suggestion
Invariant checking is a safety feature of the XRP Ledger. It consists of a set of checks, separate from normal transaction processing, that guarantee that certain _invariants_ hold true across all transactions.
```

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,404 @@
...
+
+Invariant checks should not trigger, but they ensure the XRP Ledger's integrity from bugs yet to be discovered or even created.
+
+
+## Why it Exists
+
+Nevertheless, why do we need invariant checks?
```
mDuo13: The header already asks this question so you can omit this line.

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,404 @@
...
+
+- The overarching code for the XRP Ledger is complicated and vast; therefore, there is a high potential for code to execute incorrectly.
+- The cost of incorrectly executing a transaction is high and not acceptable by any standards.
+
+
+## How it Works
+
+A second layer of code runs automatically and in real-time after each transaction completes. It then examines the changes it made for correctness before the results are committed to the ledger. After every transaction runs, the invariant checker runs. Problematic transactions are marked with a **tecINVARIANT_FAILED** result code and are included in the ledger as having done nothing.
```
mDuo13: 1. "It then examines the changes it made..." — confusing antecedents for "it" here
2. I'm trying to put transaction result codes in `monospace` consistently. Among other reasons, it exempts them from the spelling checker. So let's do that with `tecINVARIANT_FAILED` here and in the next paragraph.
3. The 1st and 3rd sentences are closely related, so let's combine them.

```suggestion
The invariant checker is a second layer of code that runs automatically in real-time after each transaction. Before the transaction's results are committed to the ledger, the invariant checker examines those changes for correctness. If the transaction's results would break one of the XRP Ledger's strict rules, the invariant checker rejects the transaction. Transactions that are rejected this way have the result code `tecINVARIANT_FAILED` and are included in the ledger with no effects.
```

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,404 @@
...
+
+The invariant checker is then invariantly checked to ensure the invariant checker's failure is not due to a bug. If this second level check fails, transactions are marked with a **tefINVARIANT_FAILED** result code. The transaction is in its final state and is not included in the ledger.
```
mDuo13: Most people, even software engineers, won't easily understand what the phrase "invariantly checked" means here. I checked with Nik and I think I can suggest a better explanation of what's going on here:

```suggestion
To include the transaction in the ledger with a `tec`-class code, some minimal processing is necessary. If this minimal processing still breaks an invariant, the transaction fails with the code `tefINVARIANT_FAILED` instead, and is not included in the ledger at all.
```

Also, despite the "tef" code, tefINVARIANT_FAILED is not a guaranteed final state. Since it only happens if there's a bug (likely a pretty specific one) there's a chance that the changes to the ledger could change the circumstances and cause the bug not to arise again, so resubmitting the transaction could conceivably result in success or a "normal" tec failure. So I left out the "in its final state" sentence.

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,404 @@
...
+
+
+## Active Invariant Checks
+
+Following is a list of active checks that the invariant checker runs against each transaction on the XRP Ledger.
```
mDuo13: As a style thing, I suggest using "The following is..." instead of "Following is..." and ending with a colon instead of a period here. But we can also simplify this sentence more:

```suggestion
The XRP Ledger checks all the following invariants on each transaction:
```

content/concepts/consensus-network/invariant-checking.md
mDuo13: For links to other xrpl.org docs, please use the HTML filename (and anchor, if appropriate) only, not a fully qualified URL.

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,404 @@
...
+        std::shared_ptr<SLE const> const&);
+
+    bool
+    finalize(
+        STTx const&,
+        TER const,
+        XRPAmount const,
+        ReadView const&,
+        beast::Journal const&);
+};
```
mDuo13: It's not useful to include a snippet of C++ code without context like this. It can be helpful to link to the relevant source code (see the "[Source]" links on many other pages for an example). For the scope and context of this page I don't think this code needs to be in the page at all.

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,404 @@
...
+```
+
+### XRP Not Created
+
+- **Condition(s) Checked:**
+    
+    - A transaction must not create XRP and should only destroy the XRP [fee](https://xrpl.org/rippleapi-reference.html#transaction-fees)).
```
mDuo13: 1. As mentioned above, please use a relative URL, not an absolute/fully-qualified URL here.
2. In the documentation we try to avoid using the word "fee" for the [transaction cost](https://xrpl.org/transaction-cost.html) because people tend to assume that a fee is paid to someone in particular (it's actually destroyed). Also, that's the more relevant concept link for this idea.


```suggestion
    - A transaction must not create XRP and should only destroy the XRP [transaction cost](transaction-cost.html)).
```

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,404 @@
...
+        ReadView const&,
+        beast::Journal const&);
+};
+```
+
+### XRP Balance Checks
+
+- **Condition(s) Checked:**
+    
+    - An account's XRP balance must be of type XRP and take a value between 0 and INITIAL_XRP drops, inclusive.
```
mDuo13: Instead "and take a value between 0 and INITIAL_XRP drops, inclusive" we should write something an ordinary person could understand without more background knowledge. 

```suggestion
    - An account's XRP balance must be of type XRP, and it cannot be less than 0 or more than 100 billion XRP exactly.
```

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,404 @@
...
+        ReadView const&,
+        beast::Journal const&);
+};
+```
+
+### Ledger Entry Types Match
+
+- **Condition(s) Chceked:**
+    
+    - Corresponding modified ledger entries should match in type and added entries should be a [valid type](https://xrpl.org/transaction-types.html#transaction-types).
```
mDuo13: Wrong link. "Ledger entry types" are the same thing as ["ledger object types"](https://xrpl.org/ledger-object-types.html). Also, this would make more sense if there were some more context about what's "corresponding" to what in this one. I think it means the objects that the transaction modified, as identified by their IDs, before and after executing the transaction? Not sure if it's exactly looking at the [metadata](https://xrpl.org/transaction-metadata.html).

mDuo13 (changes requested):
Not all the way through, but I think this will need significant changes. I've suggested a few things but I really think the doc needs more original text to provide context and understanding around individual checks, and I don't think the code snippets are helping here.

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,152 @@
+# Invariant Checking
+
+This article provides a high-level overview of invariant checking, why it exist, how it works, and lists active invariants.
```
mDuo13: I feel this sentence is unnecessary. The table of contents tells the user the same information. We can cut to the chase and start describing invariant checking right away.

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,152 @@
...
+
+Like many safety features, we all hope that invariant checking never actually needs to do anything. However, it can be useful to understand the XRP Ledger's invariants because they define hard limits on the XRP Ledger's transaction processing, and to recognize the problem in the unlikely event that a transaction fails because it violated an invariant check.
+
+## Introduction
+
+Invariant checking is a safety feature of the XRP Ledger. It consists of a set of checks, separate from normal transaction processing, that guarantee that certain _invariants_ hold true across all transactions.
```
mDuo13: This should be the first sentence of the article. I'd move it up to replace the first sentence.

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,152 @@
...
+
+Invariants should not trigger, but they ensure the XRP Ledger's integrity from bugs yet to be discovered or even created.
+
+| Term | Description |
+|-----------|-------------|
+| Invariant | A rule that should always, without exception, be true. For example, "New XRP cannot be created". |
+| Invariant Checking | In the XRP Ledger, a system where code automatically confirms that transaction processing does not break the invariants. If a transaction's execution would break an invariant, the invariant checking system fails the transaction. |
```
mDuo13: I meant these definitions to be for us as writers, not for the reader. This doesn't have to be in the article. (If we had a glossary, that would be a different story, but we don't.)

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,152 @@
...
+
+
+## Why it Exists
+
+- The overarching code for the XRP Ledger is complicated and vast; therefore, there is a high potential for code to execute incorrectly.
```
mDuo13:
```suggestion
- The source code for the XRP Ledger is complicated and vast; there is a high potential for code to execute incorrectly.
```

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,152 @@
...
+- The cost of incorrectly executing a transaction is high and not acceptable by any standards.
```
mDuo13, line 13: It would be good to go into more detail on what this means. Maybe not as in-depth as possible, but I think it's not immediately obvious to the reader why "the cost... is high".

A transaction being processed incorrectly could have devastating consequences for the XRP Ledger. Don't quite this exactly, but depending on the type of exploit and other details, some possibilities include:

- Intentional abuse of a bug might allow a user to profit by causing harm to others. They might be able to steal money in the system similar to what happened [in Ethereum's DAO exploit](https://dzone.com/articles/the-dao-development-flaws-a-smart-contract-that-cr). Or the ability to create more XRP could lead to unbounded inflation.
- In a particularly severe case, incorrect transaction execution could create invalid or corrupt data that later consistently crashes servers in the network by sending them into an "impossible" state. This could potentially halt the entire network.
- The value of money in a system is directly correlated to people's trust in that system. Incorrect transaction processing would undermine and devalue the entire XRP Ledger.
    - (Conversely, the fact that invariant checking exists provides value to the entire XRP Ledger because it makes it a more reliable system.)

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,152 @@
...
+The invariant checker is a second layer of code that runs automatically in real-time after each transaction. Before the transaction's results are committed to the ledger, the invariant checker examines those changes for correctness. If the transaction's results would break one of the XRP Ledger's strict rules, the invariant checker rejects the transaction. Transactions that are rejected this way have the result code `tecINVARIANT_FAILED` and are included in the ledger with no effects.
+
+To include the transaction in the ledger with a `tec`-class code, some minimal processing is necessary. If this minimal processing still breaks an invariant, the transaction fails with the code `tefINVARIANT_FAILED` instead, and is not included in the ledger at all.
+
+
+## Active Invariants
+
+The XRP Ledger checks all the following invariants on each transaction:
+
+[[Source]](https://github.com/ripple/rippled/blob/023f5704d07d09e70091f38a0d4e5df213a3144b/src/ripple/app/tx/impl/InvariantCheck.h#L92 "Source")
```
mDuo13, line 32: For consistency with other pages, let's move the source links to be under the headers for the checks they implement

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,152 @@
...
+[[Source]](https://github.com/ripple/rippled/blob/023f5704d07d09e70091f38a0d4e5df213a3144b/src/ripple/app/tx/impl/InvariantCheck.h#L275 "Source")
+
+- [No Zero Escrow](#no-zero-escrow)
+
+[[Source]](https://github.com/ripple/rippled/blob/023f5704d07d09e70091f38a0d4e5df213a3144b/src/ripple/app/tx/impl/InvariantCheck.h#L300 "Source")
+
+- [Valid New Account Root](#valid-new-account-root)
+
+
+### [Transaction Fee Check](#active-invariants)
```
mDuo13: Let's not make the headers into links. In general, having more markup in headers adds complications and it's not necessary here.

```suggestion
### Transaction Fee Check
```

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,152 @@
...
+
+- **Invariant Condition(s):**
+    - [Transaction fees](rippleapi-reference.html#transaction-fees) should never be negative nor larger than the transaction itself.
```
mDuo13:
```suggestion
    - The [transaction cost](transaction-cost.html) amount must never be negative, nor larger than the cost specified in the transaction.
```

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,152 @@
...
+### [XRP Not Created](#active-invariants)
+
+- **Invariant Condition(s):**
+    - A transaction must not create XRP and should only destroy the XRP [transaction cost](transaction-cost.html).
+
+
+### [Account Roots Not Deleted](#active-invariants)
+
+- **Invariant Condition(s):**
+    - An account ledger entry can be removed by an AccountDelete transaction, but this invariant checks that exactly 1 is deleted by a successful AccountDelete.
```
mDuo13:
```suggestion
    - An [account](accounts.html) cannot be deleted from the ledger except by an [AccountDelete transaction][].
    - A successful AccountDelete transaction always deletes exactly 1 account.
```

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,152 @@
...
+### [No Bad Offers](#active-invariants)
+
+- **Invariant Condition(s):**
+    - [Offers](offer.html#offer) should be for non-negative amounts and must not be XRP to XRP.
+
+
+### [No Zero Escrow](#active-invariants)
+
+- **Invariant Condition(s):**
+    - An [escrow](escrow-object.html) entry must hold a quantity of XRP between 0 and 99.99 billion.
```
mDuo13:
```suggestion
    - An [escrow](escrow-object.html) entry must hold more than 0 XRP and less than 100 billion XRP.
```

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,152 @@
...
+
+
+### [Valid New Account Root](#active-invariants)
+
+- **Invariant Condition(s):**
+    - A new [account root](accountroot.html) must be the consequence of a payment.
+    - A new account root must have the right starting [sequence](basic-data-types.html#account-sequence).
+    - A new [account](accounts.html) may not create more than one new account root.
```
mDuo13:
```suggestion
    - A transaction must not create more than one new [account](accounts.html).
```

mDuo13 (changes requested):
It's looking close. I think there were some miscommunications on my last review with some things I wasn't intending to be included verbatim in the doc, and now that I see it better I have some formatting suggestions and wording clarifications.

mDuo13 (approved):
:+1:

mDuo13:
The link failure was because the page is expected to exist in all targets. Hopefully with my edit to add it to the Japanese target, the link checker will now succeed. (Per our [translation plan](https://github.com/ripple/xrpl-dev-portal/issues/526), new docs are added in English to all targets until they're translated. The untranslated doc automatically shows a blurb that says it's not available in their language yet.)

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -0,0 +1,149 @@
...
+### Ledger Entry Types Match
+
+- **Invariant Condition(s):**
+    - Corresponding modified ledger entries should match in type and added entries should be a [valid type](ledger-object-types.html).
+
+
+### No XRP Trust Lines
+
+- **Invariant Condition(s):**
+    - [Trust lines](trust-lines-and-issuing.html#trust-lines-and-issuing) using XRP are not allowed.
```
mDuo13: We don't need to have the page-title anchor in this link. (It confuses the link checker on the Japanese target and does nothing useful on the English target.)

```suggestion
    - [Trust lines](trust-lines-and-issuing.html) using XRP are not allowed.
```

mDuo13 (changes requested):
Hopefully these changes should get the link checker to pass

PR #899, opened 2020-09-10
Consistent use of AccountID in transaction types

mDuo13:
The [Serialization docs](https://xrpl.org/serialization.html) describe the AccountID internal format; this PR fixes the the transaction types docs  to consistently use "AccountID" and not "Account" to refer to this type.


seelabs (approved):
:+1: LGTM

PR #901, opened 2020-09-17
Qa calvin

Calvinjwala:
- Fixed layout for the impact, calculator, and overview page
- Added net-new assets and replaced some older ones
- Style changes per requests in the QA doc
- Copy changes per Figma


tool/template-learn-uses.html
```diff
@@ -72,22 +72,22 @@ <h2 class="mb-10">{% trans %}Digital Cryptocurrency Wallets{% endtrans %}</h2>
...
-      <div class="col-md-4 offset-md-1">
+      <div class="col-md-4 offset-md-2">
         <h2 class="mb-10">{% trans %}Exchanges{% endtrans %}</h2>
         <p class="mb-10 normal">{% trans %}Market makers and individuals use exchanges to trade XRP and other digital assets. Exchanges are often also intermediaries in cross-border payments among financial institutions.{% endtrans %}</p>
         <a href="exchanges.html" class="btn btn-outline-secondary">{% trans %}Explore Exchanges{% endtrans %}</a>
       </div>
     </section>
 
-    <section class="row last-section">
+    <section class="row mb-50">
```
mDuo13: This undoes some changes I put in specifically to make the spacing at the bottom of the page match Gratas' request. Is that intentional?

tool/template-calculator.html
```diff
@@ -16,35 +16,38 @@
...
 <div class="fullwidth-bg">
-  <div class="container">
+  <div class="container mt-30">
 
     <section class="row mb-49">
       <h6 class="section-marker">{% trans %}Sustainability{% endtrans %}</h6>
-      <div class="col-sm-4 d-sm-none d-md-block sustainability-bg">
-        &nbsp;
+      <div class="col-sm-4 d-sm-none d-md-block">
+        <img class="mw-100" src="./img/green/green-graphic.png">
```
mDuo13: For decorative images such as this one, we should always [provide empty alt text so that screen readers do not read the file name aloud](https://www.w3.org/WAI/tutorials/images/decorative/).

(There are many other cases of this throughout the new pages; we should comb through and fix them at some point.)

mDuo13 (changes requested):
Left a couple comments. I plan to take a closer look tomorrow.

tool/template-learn-uses.html
Calvinjwala: Did not know that. We can leave as is, but the spacing is more or less the same with one less class

tool/template-calculator.html
Calvinjwala: Sounds good, I can take a look at that

tool/template-learn-uses.html
```diff
@@ -82,7 +82,7 @@ <h2 class="mb-10">{% trans %}Exchanges{% endtrans %}</h2>
       </div>
     </section>
 
-    <section class="row mb-50">
+    <section class="row lsat-section">
```
mDuo13: Typo?

tool/template-calculator.html
```diff
@@ -25,7 +25,7 @@
     <section class="row mb-49">
       <h6 class="section-marker">{% trans %}Sustainability{% endtrans %}</h6>
       <div class="col-sm-4 d-sm-none d-md-block">
-        <img class="mw-100" src="./img/green/green-graphic.png">
+        <img class="mw-100" src="./img/green/green-graphic.png" alt="{% trans %}Sustainability icon{% endtrans %}">
```
mDuo13, line 28: I think for cases like this it's better to have an actually empty alt text since the image is purely decorative. (Read the link from my previous comment for details)

mDuo13 (changes requested):

mDuo13 (approved):

PR #902, opened 2020-09-18
Change "broken" calculator anchors to data attrs

mDuo13:
Anchor links to d-crypto etc. were not really broken but were used in
JavaScript; this changes them to data attributes so the link checker
doesn't report them as broken.


Calvinjwala (approved):
Looks good, thanks for cleaning up the spacing

PR #913, opened 2020-10-05
Japanese translations for 2020 refresh

mDuo13:
- Add Japanese translations for new pages
- Updated fonts for Japanese to Noto Sans JP instead of Makinas 4.
- Adds translations for top nav where necessary
- Incorporate font CSS into core stylesheet & provide fonts directly
    - This allows the site to be viewed locally if you're offline, and also reduces the total number of requests needed to load the site


mDuo13:
The remaining broken links are not related to this PR.

PR #926, opened 2020-11-10
MemoData HEX must be UPPERCASE

WietseWind:
Hex must be UPPERCASE or you'll end up with:

`Unexpected error ?? is not a valid hex representation`
(?? = the first two chars where a lower case alpha hex char is present, eg. `2f`)

Credits: @KoenPaas


mDuo13:
The rippled APIs do not require uppercase hex. Confirmed it myself. I suspect this error comes from whatever programming language or library you're using to do the encoding?

WietseWind:
Oops, I was convinced it was rippled but it was ripple-binary-codec. Never mind, sorry for wasting your time.

Issue #914, opened 2020-10-06
Harbor Wallet links not functional


LoisRP:
As noted in this forum post, the Harbor Wallet links at the bottom for Windows and OSX no longer work.

https://forum.xpring.io/t/xrpl-org-update/191

https://www.secureblockchains.com/harbor-xrp-wallet/

PR #932, opened 2020-11-18
updating copy on calc page

Calvinjwala:
Minor copy change on the Carbon Calculator Page


mDuo13:
Build failure is from #914, not this, so :+1:

PR #939, opened 2020-12-02
telINSUF_FEE_P: clarify description

mDuo13:
Should hopefully clarify questions like the one asked in https://github.com/ripple/rippled/issues/3594


content/references/rippled-api/transaction-formats/transaction-results/tel-codes.md
```diff
@@ -16,7 +16,7 @@ These codes indicate an error in the local server processing the transaction; it
...
 | `telCAN_NOT_QUEUE_FULL` | The transaction did not meet the [open ledger cost](transaction-cost.html) and the server did not queue this transaction because this server's transaction queue is full. You could increase the `Fee` and try again, try again later, or try submitting to a different server. The new transaction must have a higher transaction cost, as measured in [fee levels](transaction-cost.html#fee-levels), than the transaction in the queue with the smallest transaction cost. [New in: rippled 0.70.2][] |
 | `telFAILED_PROCESSING` | An unspecified error occurred when processing the transaction. |
-| `telINSUF_FEE_P`       | The `Fee` from the transaction is not high enough to meet the server's current [transaction cost](transaction-cost.html) requirement, which is derived from its load level. |
+| `telINSUF_FEE_P`       | The `Fee` from the transaction is not high enough to meet the server's current [transaction cost](transaction-cost.html) requirement, which is derived from its load level. The individual server is too busy to process your transaction right now, but may cache it and automatically retry later. |
```
ximinez: This code also gets returned if the transaction pays less than the network-defined minimum. e.g. `Fee: 1` when the base fee is 10 drops. There's room for improvement, but a starting point is:
> The `Fee` from the transaction is not high enough to meet the server's current [transaction cost](transaction-cost.html) requirement, which is derived from its load level and network fee requirements. If the transaction cost was otherwise sufficient, the individual server is unable to process your transaction right now, but may cache it and automatically retry later.

ximinez (changes requested):

ximinez (approved):

PR #941, opened 2020-12-07
ripple-lib: update to 1.9.0

intelliot:


mDuo13 (approved):
Dang, I _just_ started upgrading some of the dev portal's samples to use v1.8.2 earlier today. That's an easy (separate) change, though.

PR #953, opened 2021-01-20
"Use Tickets" tutorial

mDuo13:
Adds an interactive tutorial ([`use-tickets.md`](https://github.com/ripple/xrpl-dev-portal/pull/953/files#diff-9c5c7d8b0ea65fe4faa6fd58985b11b73b9625220af00999d32081a525f696b3)) showing how to use Tickets to send transactions.

(EDITED: it's been updated to use Devnet)

Secondary features of this PR:

- Minor style improvements to interactive tutorials
- Fixes a bug in the interactive tutorials JavaScript that caused a bug when step names had more than one space in them
- Updates the version(s) of ripple-lib used in various places and puts the version number in the config file for easier updates going forward
- Moves the "Generate" (Testnet faucet) step of interactive tutorials to a reusable snippet
- Moves interactive_steps filter config to frontmatter (requires [Dactyl v0.13.2](https://github.com/ripple/dactyl/releases/tag/v0.13.2))
- Uses YAML anchor/extend syntax to move certain redundant fields of target definitions into a single shared definition

Staged preview (Ripple VPN required): https://developers-staging.ripple.com/use-tickets.html (last commit pushed: eb0b9a66e519d636d142d9bd4db1c1cfd2669fac )


content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+Since JavaScript works in the web browser, you can read along and use the interactive steps without any setup.
+
+Tickets must be enabled. At this time, the [TicketBatch amendment][] :not_enabled: is only available on Devnet.
+
+
+
+## Steps
+{% set n = cycler(* range(1,99)) %}
+
+### {{n.next() }}. Get Credentials
```
intelliot:
```suggestion
### {{n.next()}}. Get Credentials
```

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+
+To interact with the XRP Ledger, you need a address and secret key, and some XRP. You can get all of these on the [Devnet](parallel-networks.html) using the following interface: <!--TODO: change to Testnet eventually -->
```
intelliot: If you have some XRP then you probably have an address and secret key. There are some exceptions, but they are rare, so I think we can simplify to:
```suggestion
To transact on the XRP Ledger, you need some XRP. For development purposes, you can get some free test XRP on the [Devnet](parallel-networks.html) using the following interface: <!--TODO: change to Testnet eventually -->
```

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+{% include '_snippets/generate-step.md' %}
+
+When you're [building actual production-ready software](production-readiness.html), you'll instead use an existing account, and manage your keys using a [secure signing configuration](set-up-secure-signing.html).
+
+
+### {{n.next()}}. Connect to Network
+
+You must be connected to the network to submit transactions to it.
+
+The following code sample instantiates a new RippleAPI instance and connects to one of the public XRP Devnet servers that Ripple runs:
```
intelliot: Feel free to reject this, but we could make this a bit less wordy:
```suggestion
The following code uses a RippleAPI instance to connect to a public XRP Devnet server:
```

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+  <strong>Connection status:</strong>
+  <span id="connection-status">Not connected</span>
+  <div id='loader-connect' style="display: none;"><img class='throbber' src="assets/img/xrp-loader-96.png"></div>
+</div>
+{{ end_step() }}
+
+
+### {{n.next()}}. Check Sequence Number
+
+If you already have at least one Ticket available in the ledger, you can skip this step. Otherwise, you need to get ready to create some Tickets for your other transactions to use.
```
intelliot:
```suggestion
(If you already have at least one Ticket available in the ledger, you can skip this step.) You need to get ready to create some Tickets for your other transactions to use.
```

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+<button id="check-sequence" class="btn btn-primary connection-required"
+  title="Complete all previous steps first" disabled>Check Sequence Number</button>
+<div id="check-sequence-output"></div>
+{{ end_step() }}
+
+
+
+### {{n.next()}}. Prepare and Sign TicketCreate
+
+If you already have at least one Ticket available in the ledger, you can skip this step. Otherwise, you need to prepare the transaction to create some Tickets.
```
intelliot:
```suggestion
(If you already have at least one Ticket available in the ledger, you can skip this step.) You need to prepare the transaction that will create some Tickets.
```

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+
+Construct a [TicketCreate transaction][] the sequence number you determined in the previous step. Use the `TicketCount` field to set how many Tickets to create. For example, this example [prepares a transaction](rippleapi-reference.html#preparetransaction) that would make 10 Tickets:
```
intelliot:
```suggestion
Construct a [TicketCreate transaction][] using the sequence number you determined in the previous step. Use the `TicketCount` field to specify how many Tickets to create. For example, this example [prepares a transaction](rippleapi-reference.html#preparetransaction) that would make 10 Tickets:
```

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+<button id="prepare-and-sign" class="btn btn-primary connection-required"
+  title="Complete all previous steps first" disabled>Prepare & Sign</button>
+<div id="prepare-and-sign-output"></div>
+{{ end_step() }}
+
+
+
+### {{n.next()}}. Submit TicketCreate
+
+If you already have at least one Ticket available in the ledger, you can skip this step. Otherwise, you need to send a transaction to create some Tickets.
```
intelliot:
```suggestion
(If you already have at least one Ticket available in the ledger, you can skip this step.) You need to send the transaction to create some Tickets.
```

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+
+Submit the signed transaction blob that you created in the previous step. Take note of the latest validated ledger index at the time of submission, so you can set a lower bound on what ledger versions the transaction could be validated in. For example:
+
+```js
+let prelim_result = await api.request("submit", {"tx_blob": tx_blob})
```
intelliot, line 147: nit: As of ES6 (aka ES2015) we can use a [shorthand property name](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer):
```suggestion
let prelim_result = await api.request("submit", {tx_blob})
```

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+console.log("Preliminary result:", prelim_result)
+const min_ledger = prelim_result.validated_ledger_index
+```
+
+**Warning:** Be sure that you **DO NOT UPDATE** the `min_ledger` value. It is safe to submit a signed transaction blob multiple times (the transaction can only execute at most once), but when you look up the status of the transaction you should use the earliest possible ledger index the transaction could be in, _not_ the validated ledger index at the time of the most recent submission. Using the wrong minimum ledger value could cause you to incorrectly conclude that the transaction did not execute. For best practices, see [Reliable Transaction Submission](reliable-transaction-submission.html).
```
intelliot:
```suggestion
**Warning:** Be sure that you **DO NOT UPDATE** the `min_ledger` value. It is safe to submit a signed transaction blob multiple times (the transaction can only execute at most once), but when you look up the status of the transaction you should use the earliest possible ledger index that the transaction could be in, _not_ the validated ledger index at the time of the most recent submission. Using the wrong minimum ledger value could cause you to incorrectly conclude that the transaction did not execute. For best practices, see [Reliable Transaction Submission](reliable-transaction-submission.html).
```

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+  </tr>
+</table>
+{{ end_step() }}
+
+
+### (Optional) Intermission
+
+The power of Tickets is that you can send other transactions during this time, and generally carry on with your account's normal business during this time. If you are planning on using a given Ticket, then as long as you don't use that Ticket for something else, you're free to continue using your account as normal. When you want to send the transaction using a Ticket, you can do that in parallel with other transactions using regular sequence numbers or different Tickets, and submit any of them in any order when you're ready.
+
+**Tip:** You can come back here to send Sequenced transactions between or during any of the following steps, without interfering with the success of your Ticketed transaction.
```
intelliot, line 239: Awesome!

intelliot (changes requested):
Very well done! This is a great tutorial. It may make sense to add a brief summary/recap at the end, perhaps with a mention of why this is useful with multi-signing, but it's not strictly necessary.

content/tutorials/manage-account-settings/use-tickets.md
mDuo13: > If you have some XRP then you probably have an address and secret key.

I hear a lot of people say they have XRP, but what they mean is they hold have some at a custodial exchange, which may even give them a deposit address (even more likely if X-addresses gain more adoption) but not a secret.

content/tutorials/manage-account-settings/use-tickets.md
mDuo13: That's fascinating, and totally new to me. I think the wordier traditional syntax might be a bit more obvious/clear for people who aren't familiar with this new piece of syntactic sugar. Seems like it would be easy to get confused or mix it up (forget the braces, rename the variable and not understand why it's not working) if you're not familiar with the idea.

content/tutorials/manage-account-settings/use-tickets.md
intelliot: Good point. Here's a smaller suggested change:
```suggestion
To transact on the XRP Ledger, you need an address and secret key, and some XRP. For development purposes, you can get these on the [Devnet](parallel-networks.html) using the following interface: <!--TODO: change to Testnet eventually -->
```

intelliot (approved):
Amazing 👍

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+}
+
+main()
+```
+
+**Note:** The code samples in this tutorial use JavaScript's [`async`/`await` pattern](https://javascript.info/async-await). Since `await` needs to be used from within an `async` function, the remaining code samples are written to continue inside the `main()` function started here. You can also use Promise methods `.then()` and `.catch()` instead of `async`/`await` if you prefer.
+
+For this tutorial, you can connect directly from your browser by pressing the following button:
+
+{{ start_step("Connect") }}
```
ryangyoung, line 77: I know you mentioned a few caveats related to these buttons in the PR description, but none of the buttons after this one worked for me (tried on Chrome and FF).

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+<div id="check-sequence-output"></div>
+{{ end_step() }}
+
+
+
+### {{n.next()}}. Prepare and Sign TicketCreate
+
+(If you already have at least one Ticket available in the ledger, you can skip this step.) You need to prepare the transaction that will create some Tickets.
+
+Construct a [TicketCreate transaction][] using the sequence number you determined in the previous step. Use the `TicketCount` field to specify how many Tickets to create. For example, this example [prepares a transaction](rippleapi-reference.html#preparetransaction) that would make 10 Tickets:
```
ryangyoung: > For example, this example

Perhaps drop one of the examples here.

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+console.log("Prepared transaction:", prepared.txJSON)
+let max_ledger = prepared.instructions.maxLedgerVersion
+
+let signed = api.sign(prepared.txJSON, "s████████████████████████████")
+console.log("Transaction hash:", signed.id)
+
+let tx_blob = signed.signedTransaction
+```
+
+Take note of the transaction's hash and `LastLedgerSequence` value so you can [be sure whether or not it got validated](reliable-transaction-submission.html) later.
```
ryangyoung:
```suggestion
 Note the transaction's hash and `LastLedgerSequence` value so you can [be sure whether it got validated](reliable-transaction-submission.html) later.
```

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+<button id="prepare-and-sign" class="btn btn-primary connection-required"
+  title="Complete all previous steps first" disabled>Prepare & Sign</button>
+<div id="prepare-and-sign-output"></div>
+{{ end_step() }}
+
+
+
+### {{n.next()}}. Submit TicketCreate
+
+(If you already have at least one Ticket available in the ledger, you can skip this step.) You need to send the transaction to create some Tickets.
```
ryangyoung: What do you think of moving up some the optionality to the beginning of the tutorial? If I'm understanding right, you can skip this and the previous two steps if you've already created a Ticket. As a user with a Ticket already, I would imagine that I'd like to know that I could skip ahead to step 5 if I want to. This is a tricky tutorial, though, so I understand if you already considered and rejected that idea.

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+
+Submit the signed transaction blob that you created in the previous step. Take note of the latest validated ledger index at the time of submission, so you can set a lower bound on what ledger versions the transaction could be validated in. For example:
```
ryangyoung:
```suggestion
Submit the signed transaction blob that you created in the previous step. Note the latest validated ledger index at the time of submission, so you can set a lower bound on what ledger versions the transaction could be validated in. For example:
```

intelliot:
Just a thought -- does it make sense that this tutorial is in the "Manage Account Settings" section?

content/_snippets/generate-step.md
```diff
@@ -0,0 +1,74 @@
...
+    })
+  })
+
+  const EXAMPLE_ADDR = "rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe"
+  const EXAMPLE_SECRET = "s████████████████████████████"
+
+})
+</script>
+
+**Caution:** Ripple provides the [XRP Ledger Testnet](parallel-networks.html) for testing purposes only, and regularly resets the state of the test net along with all balances. As a precaution, Ripple recommends **not** using the same addresses on Testnet and Mainnet.
```
intelliot: Just a nitpick here
```suggestion
**Caution:** Ripple provides the Testnet and Devnet [parallel networks](parallel-networks.html) for testing purposes only, and sometimes resets the state of these test networks along with all balances. As a precaution, Ripple recommends **not** using the same addresses on Testnet/Devnet and Mainnet.
```

content/tutorials/manage-account-settings/use-tickets.md
```diff
@@ -0,0 +1,372 @@
...
+  </tr>
+  <tr id="tx-validation-status">
+  </tr>
+</table>
+{{ end_step() }}
+
+
+### (Optional) Intermission
+
+The power of Tickets is that you can send other transactions during this time, and generally carry on with your account's normal business during this time. If you are planning on using a given Ticket, then as long as you don't use that Ticket for something else, you're free to continue using your account as normal. When you want to send the transaction using a Ticket, you can do that in parallel with other transactions using regular sequence numbers or different Tickets, and submit any of them in any order when you're ready.
```
ryangyoung: > The power of Tickets is that you can send other transactions **during this time**, and generally carry on with your account's normal business **during this time**.

We could stand to lose one "during this time" in the opening sentence. 

The whole paragraph could probably be tightened up a little more, though I know that's easier suggested than done :)

ryangyoung (approved):
I wasn't able to run through all the steps, but this is great. I left a few comments and suggestions that might help readability/usability a bit but nothing that should block this from going live.

mDuo13:
> @intelliot 
> Just a thought -- does it make sense that this tutorial is in the "Manage Account Settings" section?

I've struggled a lot with the categorization/hierarchy of the tutorials. "Use Tickets" doesn't cleanly fall into any of the existing categories, IMO:

- Get Started
- Production Readiness
- Manage Account Settings
- Use Specialized Payment Types
- XRP Ledger Businesses
- Manage the rippled Server

I'm hesitant to keep adding categories, especially ones with only one entry in them, since going beyond about 7 categories is likely to get overwhelming. And I'd be open to renaming and reorganizing some of the existing categories, but we should probably not do this too often since it'll just get people even more lost.

content/tutorials/manage-account-settings/use-tickets.md
mDuo13: I don't like using "Note" as a verb to start a sentence since that feels too similar to "Note: somethingsomething".

Changed to "Record the..." instead

content/tutorials/manage-account-settings/use-tickets.md
mDuo13: The idea I was trying to convey with this is that you don't have to create a ticket each time you want to send a ticketed transaction. It's not quite a one-time setup step since you need to redo it whenever you run out of tickets but it is something you only have to do once in a while.

After some thought, I changed the way I approach this as follows:

- Removed the "You can skip this step" sentence from the early steps
- Added a brief overview to the beginning that categorizes the steps into a few conceptual phases and clarifies how those steps do/don't need to be repeated
- Added a **Tip** to the beginning of the phase where you're actually using a Ticket, clarifying that you can repeat the steps from there to the end as long as you have Tickets left

content/tutorials/manage-account-settings/use-tickets.md
mDuo13: Fixed in 0e33d93

PR #948, opened 2020-12-16
TJ Harris - Draft Pull Request for Take-Home Test Option 1

tjh3:
I made most of the changes with first commit (and provided details of those changes), and then a quick typo fix with the second commit.  Submitting as a draft pull request per instructions in the repo.

PR #957, opened 2021-01-26
ledger_entry and WebSocket Tool updates

mDuo13:
Continues the work of #948, but with more extensive updates:

- Fixes some mistakes in `ledger_entry` description including the `type` field (not actually used in the code as far as I can tell!)
- Updates some examples in the WebSocket tool. In particular, discontinues the use of `r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59` which probably belongs to a former Ripple employee of unknown identity
- Adds examples in the WebSocket tool for various `ledger_entry` modes

### DIFF
--- a/assets/js/apitool-methods-ws.js
+++ b/assets/js/apitool-methods-ws.js
@@ -299,8 +299,9 @@ Request('ledger_entry - DepositPreauth', {
 
 Request("Transaction Methods")
 
-// Signing methods are not provided here because it's a terrible idea to send
-// your secret to another server over the internet.
+// Signing methods are not provided here because they're admin-only by default.
+// (Sending your secret to another server is extremely insecure and could cause
+// someone else to take over your account, steal all your money, etc.)
 
 Request('submit', {
   description: "Submits a transaction to the network to be confirmed and included in future ledgers.",
--- a/assets/js/apitool-methods-ws.js
+++ b/assets/js/apitool-methods-ws.js
@@ -1,3 +1,5 @@
+DEFAULT_ADDRESS_1 = "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn"
+
 Request("Account Methods")
 
 Request('account_channels', {
@@ -6,8 +8,8 @@ Request('account_channels', {
     body: {
       "id": 1,
       "command": "account_channels",
-      "account": "rN7n7otQDd6FczFgLdSqtcsAUxDkw6fzRH",
-      "destination_account": "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn",
+      "account": "rN7n7otQDd6FczFgLdSqtcsAUxDkw6fzRH", //TODO: different address
+      "destination_account": DEFAULT_ADDRESS_1,
       "ledger_index": "validated"
     }
 })
@@ -17,7 +19,7 @@ Request('account_currencies', {
   link: "account_currencies.html",
   body: {
     "command": "account_currencies",
-    "account": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",
+    "account": DEFAULT_ADDRESS_1,
     "strict": true,
     "ledger_index": "validated"
   }
@@ -29,7 +31,7 @@ Request('account_info', {
   body: {
     "id": 2,
     "command": "account_info",
-    "account": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",
+    "account": DEFAULT_ADDRESS_1,
     "strict": true,
     "ledger_index": "current",
     "queue": true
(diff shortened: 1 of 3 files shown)


tjh3:
Agree with pretty much all the changes, except for removal of the (currently) optional `type` field. Made recommendation that field should be made required so that underlying method can check for that field first instead of inferring type of call based on first encountered field name (i.e. to avoid confusing "**Caution:** If you specify more than 1 of these type-specific fields in a request, the server retrieves results for only 1 of them. It is not defined which one the server chooses, so you should avoid doing this." message).

cjcobb23:
> Agree with pretty much all the changes, except for removal of the (currently) optional `type` field. Made recommendation that field should be made required so that underlying method can check for that field first instead of inferring type of call based on first encountered field name (i.e. to avoid confusing "**Caution:** If you specify more than 1 of these type-specific fields in a request, the server retrieves results for only 1 of them. It is not defined which one the server chooses, so you should avoid doing this." message).

The `type` field is not required for `ledger_entry`. I often use `ledger_entry` without specifying a type, in situations where I don't know the type.

mDuo13:
Like I said, from looking at the code, I wasn't able to find a part where the code even _looks at_ the `type` field in this method. (There are some other related methods that _do_ use a `type` field, formatted the same way, to good effect.)

assets/js/apitool-methods-ws.js
```diff
@@ -159,28 +165,148 @@ Request('ledger_data', {
...
+//     "ledger_index": "validated"
+//   }
+// })
+
+
 Request("Transaction Methods")
 
-// Future feature: special case sign/sign_for so you can use those if you're
-// connected to a local server (with big warning)
+// Signing methods are not provided here because it's a terrible idea to send
```
ryangyoung: trivial: Consider softening this a bit by briefly explaining _why_ it's a "terrible" idea (i.e., insecure, you could lose money, etc).

content/references/rippled-api/public-rippled-methods/ledger-methods/ledger_entry.md
```diff
@@ -41,144 +72,491 @@ An example of the request format:
...
-7. `escrow` - Retrieve an [Escrow object](escrow-object.html), which holds XRP until a specific time or condition is met. [New in: rippled 1.0.0][]
-8. `payment_channel` - Retrieve a [PayChannel object](paychannel.html), which holds XRP for asynchronous payments. [New in: rippled 1.0.0][]
-9. `deposit_preauth` - Retrieve a [DepositPreauth object](depositpreauth-object.html), which tracks preauthorization for payments to accounts requiring [Deposit Authorization](depositauth.html). [New in: rippled 1.1.0][]
-10. `ticket` - Retrieve a [Ticket object](ticket.html), which records a [sequence number][] set aside for future use. _(Requires the [TicketBatch amendment][])_
 
-If you specify more than one of the above items, the server retrieves only one of them; it is undefined which it chooses.
+### Get AccountRoot Object
 
-The full list of parameters recognized by this method is as follows:
+Retrieve an [AccountRoot object](accountroot.html) by its address. This is roughly equivalent to the [account_info method][].
```
ryangyoung, line 93: Is it worth enumerating any difference(s) between the two methods?

content/references/rippled-api/public-rippled-methods/ledger-methods/ledger_entry.md
```diff
@@ -41,144 +72,491 @@ An example of the request format:
...
+
+<!-- MULTICODE_BLOCK_START -->
+
+*WebSocket*
+
+```json
+{
+  "id": "example_get_escrow",
+  "command": "ledger_entry",
+  "escrow": {
```
ryangyoung, line 372: Would it be worth providing a sample that shows `escrow` as a string as well?

ryangyoung (approved):

content/references/rippled-api/public-rippled-methods/ledger-methods/ledger_entry.md
mDuo13: I don't really think so? The main difference is that account_info has some more options, as described in the doc, for fetching other stuff. There are some other inconsequential things, like `ledger_entry` puts its results in a `node` field rather than an `account_data` field (since it doesn't only return account data) and includes copies the ledger object ID to a higher-level field. Not really worth going into IMO.

content/references/rippled-api/public-rippled-methods/ledger-methods/ledger_entry.md
mDuo13: Most or all of the alternate objects can be specified using their ledger object ID as a string. This is pretty much redundant with specifying the ledger object ID as a string in the `index` field instead, so I don't think it's worth doubling the number of examples to show.

PR #960, opened 2021-02-01
Removing google analytics tag

Calvinjwala:
- Removing the duplicate GA tag that can be found in Google Tag Manager
- Moved the Linker code over to GTM


mDuo13:
This look suspiciously like it would add a regression. Remember when Google Analytics was broken for like a month after the relaunch? [This commit](https://github.com/ripple/xrpl-dev-portal/commit/518e0fa90f2ff53f2947e8b12982a8e3d7fc7364) is the one that fixed it.

Calvinjwala:
@mDuo13 - we can avoid consolidating if you think this will cause issues. Let's revisit this later - probably not worth breaking at the moment.

alloynetworks:
What is the purpose of the Google Analytics tracker? And who controls the account? There is no reason for these analytics to be available, if indeed desirable at all, to only one party.

PR #961, opened 2021-02-03
Update freezes.md

alloynetworks:
Removed "not even Ripple". There is no special privilege that requires Ripple to be mentioned here.

### DIFF
--- a/content/concepts/issued-currencies/freezes.md
+++ b/content/concepts/issued-currencies/freezes.md
@@ -14,7 +14,7 @@ There are three settings related to freezes:
 * [**Global Freeze**](#global-freeze) - Freeze all counterparties.
 * [**No Freeze**](#no-freeze) - Permanently give up the ability to freeze individual counterparties, as well as the ability to end a global freeze.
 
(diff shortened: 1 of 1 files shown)


mDuo13:
I understand that the way this page reads [calls to mind Shakespeare](https://en.wikipedia.org/wiki/The_lady_doth_protest_too_much,_methinks) (in a bad way), but the amount of FUD around the freeze feature justifies a little extra wording here.

I'm willing to make this change but I'd like to see another opinion or two before making changes to this hot-button issue page.

Perhaps a better way to handle it would be to link a special FAQ or blog post or something debunking common myths about the freeze feature with specific details (For example, "but Ripple froze Jed's XRP!" No, Ripple asked a custodial exchange to suspend Jed's liquidation of XRP pending legal action, something that could happen with any asset.) That's much more work, though, so I don't want to let the perfect get in the way of the good by waiting for that thing to happen.

mDuo13:
Going to merge this for now and leave #983 to remind us to do the anti-FUD article later.

PR #976, opened 2021-02-26
ripple-lib: update to 1.9.1

intelliot:
and change the blurb slightly: who are we to say that this library is the Official one? :) 

I don't feel strongly about the blurb, so feel free to change it back (or to something else) if you like.


mDuo13 (approved):

mDuo13:
Would also be great to replace `assets/js/ripple-lib-1.9.0.min.js` (and change [this line](https://github.com/ripple/xrpl-dev-portal/blob/e72e4e6c7231ba1bb3551f9967fadcbd99566928/dactyl-config.yml#L69) to point to the new version)

mDuo13:
The build failures are unrelated. I manually checked and these changes are OK.

PR #998, opened 2021-03-10
RPC tool updates

mDuo13:
Updates the long-ignored RPC lookup tool:

- Migrates from ripple-lib v0.11 to 1.9.1
- Use modern async/await instead of the "async.min.js" library's "waterfall" approach.
- Replaces the `account_offers` call with `account_objects` since many new types of objects have been added to the XRP Ledger protocol since this tool was originally introduced
- Changes account_tx pagination to work again (previously it used the deprecated and now removed legacy version with a `count` and `offset` rather than using `marker`)
- Improves the code that summarizes collapsed objects so that it preferentially shows what type of object it is
- Replaces Ripple Charts links with livenet.xrpl.org links
- Removes lots of deprecated stuff


mDuo13:
The only broken link is fixed in master ( 77a40691fb3484c7d054e1185a7b049e92fc7822 )

PR #1001, opened 2021-03-13
Fixed menu layout issues on Safari browser

timothyTheCoder:
Fixed layout issues on Safari browser:
1. Menu language dropdown now has proper styling on Safari
2. Menu dropdown buttons on Safarinow no longer display the outline when clicking on them to open the menu

Fixed layout issue on all browsers:
The jump-to-top had a yellow background and white letters, which makes reading it almost impossible. Changed the background to transparent.


mDuo13:
Was able to borrow a roommate's mac to confirm this does in fact fix the problem, and it looks like everything else is still OK. Thanks! :+1:

PR #1002, opened 2021-03-16
Interactive tutorial improvements

mDuo13:
- Makes a bunch of reusable pieces to simplify the creation of interactive tutorials. For details, see the [interactive tutorials README](https://github.com/mDuo13/xrpl-dev-portal/blob/interactive_tutorial_improvements/tool/INTERACTIVE_TUTORIALS_README.md) (also in this PR).
- Refactors "Use Tickets" and "Send XRP" to use some of these new bits. 
    - As part of this, I moved the bulk of the interactive JavaScript code for "Send XRP" (the code that actually runs on the page)  to its own file, similar to what I did with "Use Tickets".
    - I also moved the example code (the part shown in the code blocks on the page) for "Send XRP" to its own file in the `_code-samples` section, so that it's easier to confirm that it (still) runs completely on its own, and people can see the full picture in one place. While I was at it, I improved the example of how to wait for a transaction to be validated.
        - To support this, I added a Dactyl filter that can display excerpts from a file starting from and ending at particular search strings. "Send XRP" is using it now; my thought is most other interactive tutorials can use it eventually.
- Rewrites "Require Destination Tags" as a test case for the new interactive tutorial bits (and also just because it was a kind of outdated page).
    - Initially, I added a mention of X-addresses to this page, but I decided it was distracting from the main thrust of the page, so I moved that note to the "Source and Destination Tags" concept page instead.
- Minor improvements to the transaction sender along the way. (I was originally thinking of linking out to it for the last "Send Test Payments" bit rather than providing an interactive thing in-page, but ended up deciding against it.)
- Translates some of the text displayed to the user in the Japanese version of the "Send XRP" tutorial and in the pieces that are common across interactive tutorials.


mDuo13:
I've added [a (fairly long) README with all the stuff you need to know to make your own interactive tutorials](https://github.com/mDuo13/xrpl-dev-portal/blob/interactive_tutorial_improvements/tool/INTERACTIVE_TUTORIALS_README.md) (some HTML/JavaScript expertise required).

assets/js/tx-sender.js
```diff
@@ -239,6 +239,17 @@ const set_up_tx_sender = async function() {
...
+    // Wait for the address's funding to be validated so we don't get the wrong
+    // starting sequence number.
+    while (true) {
+      try {
+        await new Promise(resolve => setTimeout(resolve, 1000))
+        await api.request("account_info", {account: pp_issuer_address,
+          ledger_index: "validated"})
+        break
+      } catch(e) {}
+    }
```
mDuo13, line 251: I ran into https://github.com/ripple/rippled/issues/3795 while developing the interactive tutorials, and realized that it's also the reason that the Transaction Sender tool's partial payments setup stopped working. (It was written before Deletable Accounts made this a thing.) Fortunately, the same fix I came up with for the interactive tutorials also works for the transaction sender.

content/references/rippled-api/transaction-formats/transaction-results/transaction-results.ja.md
```diff
@@ -1,8 +1,8 @@
-# 取引の結果
+# トランザクションの結果
```
mDuo13, line 1: This is just a weird thing I noticed while I was working on translating the send-xrp tutorial; this page ("Transaction Results") translates "Transaction" as 取引 instead of トランザクション, which is inconsistent with the usage on most of the rest of the site. So I fixed it.

content/tutorials/manage-account-settings/require-destination-tags.md
```diff
@@ -1,74 +1,140 @@
...
+// Look for "requireDestinationTag": true
+```
+
+{{ start_step("Confirm Settings") }}
+<button id="confirm-settings" class="btn btn-primary previous-steps-required" data-wait-step-name="Wait">Confirm Settings</button>
+<div class="loader collapse"><img class="throbber" src="assets/img/xrp-loader-96.png">Sending...</div>
+<div class="output-area"></div>
+{{ end_step() }}
+
+For further confirmation, you can send test transactions (from a different address) to confirm that the setting is working as you expect it to. If you a payment with a destination tag, it should succeed, and if you send one _without_ a destination tag, it should fail with the error code [`tecDST_TAG_NEEDED`](tec-codes.html).
```
mDuo13, line 130: The setup involved in doing this is kind of complicated (you need another address to send from) so I opted not to provide a code sample in the document.

content/tutorials/use-simple-xrp-payments/send-xrp.ja.md
```diff
@@ -114,54 +81,30 @@ async function doPrepare() {
     "maxLedgerVersionOffset": 75
   })
   const maxLedgerVersion = preparedTx.instructions.maxLedgerVersion
-  console.log("Prepared transaction instructions:", preparedTx.txJSON)
-  console.log("Transaction cost:", preparedTx.instructions.fee, "XRP")
-  console.log("Transaction expires after ledger:", maxLedgerVersion)
+  console.log("準備されたトランザクション指示：", preparedTx.txJSON)
+  console.log("トランザクションコスト：", preparedTx.instructions.fee, "XRP")
+  console.log("トランザクションの有効期限はこのレジャー後：", maxLedgerVersion)
```
mDuo13, line 86: Rather than use the include_code filter in this page, I opted to leave the code samples inlined in the Markdown file but translated. Since the text is different, there's going to be a redundant (possibly out of sync) copy regardless, but I hope it's more helpful if the log messages at least are translated.

assets/js/tutorials/require-destination-tags.js
```diff
@@ -0,0 +1,148 @@
...
+    // let dest_x_address
+    // if (dt) {
+    //   dest_x_address = XCodec.Encode({ account: address, tag: dt, test: true })
+    // } else {
+    //   dest_x_address = XCodec.Encode({ account: address, test: true })
+    // }
+    // call_faucet(faucet_url, dest_x_address)
+    //
+    // // TODO: monitor our target address and report when we receive the tx from
+    // // the faucet. (including ✅ or ❌ as appropriate)
```
mDuo13, line 134: I scrapped the code for using the Faucet to send these test transactions for two reasons:

- I needed to pull in another library on top of ripple-lib to get the relevant X-address functions. This was fixed in [ripple-lib 1.9.3](https://github.com/ripple/ripple-lib/blob/develop/HISTORY.md#193-2021-03-16) in response to my questions about this.
- I realized it would be tricky to report the results of a given transaction since the Faucet doesn't tell us anything about the transaction itself. But it might be doable using the account subscriptions pattern I later added to the "Send XRP" tutorial.

mDuo13:
!rebuild

mDuo13:
Squashed and rebased to get link fixes and resolve merge conflicts.

PR #1003, opened 2021-03-16
Add Feedback Widget

jakeatdocforce:
-Change trackingId to live when ready.
- Location of assets/js will change in future.


mDuo13:
I think we need to move the position of the widget. It's overlapping the (Back to) "Top" button.


One other thing we may want to consider is excluding the homepage or other landing/table of contents pages. To me, at least, it feels weird to ask if it was useful since those are just kind of a waypoint to the actually useful pages. But maybe just aggregating the relevant content is something people would consider useful enough to click the thumbs-up?

jakeatdocforce:
Fixed position to not cover that button.  
I plan on adding more customization coming up. Better control over placement and styling when adding the widget.

mDuo13:
The widget runs off the screen on mobile. Screenshots:

[Firefox on Android](https://user-images.githubusercontent.com/7515597/111708474-26abba00-8803-11eb-9058-459e3d813c84.png)
[Chrome on Android](https://user-images.githubusercontent.com/7515597/111708476-27dce700-8803-11eb-8a72-4478fc636cce.png)

mDuo13:
Was concerned about the thumbs not showing up when I tried it on my mobile device but realized that's just because of how my test environment is set up. The way I'm using my phone to check it, the site is at `{my IP}/xrpl/` instead of at the web server root, so the `/asset/` (rather than `asset/`) paths don't exist. That shouldn't be a problem in production.

The edge of the widget still runs off the screen a little on my roommate's iPhone 7 (would be the same on the 8, I think) and in Firefox Dev Tools' presets for some other phones' screen sizes. It's still usable/readable on at least most of them (the Galaxy S9 one gets cut down to "as this page helpful?"), so it's probably worth getting this out there at this point and then iterating.

jakeatdocforce:
Working on better layout and handling of different sizes in future PRs. Also asset handling is being revised.

PR #1004, opened 2021-03-17
Add getting started with xrpl-py tutorial

ryangyoung:
Complete draft of a basic getting started tutorial, including sample code. 

See a preview of this on staging: https://developers-staging.ripple.com/get-started-using-python.html

PR #1009, opened 2021-03-27
Send XRP - add Python examples

mDuo13:
With the xrpl-py library almost at 1.0.0, I've extended the "Send XRP" tutorial to have code samples in Python. (The interactive bits, of course, still have to use JavaScript to run in the browser.)

### DIFF
--- a/content/_code-samples/send-xrp/send-xrp.py
+++ b/content/_code-samples/send-xrp/send-xrp.py
@@ -1,0 +1,58 @@
+# Connect ----------------------------------------------------------------------
+from xrpl.clients import JsonRpcClient
+testnet_url = "https://s.altnet.rippletest.net:51234"
+client = JsonRpcClient(testnet_url)
+
+# Get credentials from the Testnet Faucet -----------------------------------
+faucet_url = "https://faucet.altnet.rippletest.net/accounts"
+print("Getting a new account from the Testnet faucet...")
+from xrpl.wallet import generate_faucet_wallet
+test_wallet = generate_faucet_wallet(client, debug=True)
+
+# Prepare transaction ----------------------------------------------------------
+from xrpl.ledger import get_latest_validated_ledger_sequence
+current_validated_ledger = get_latest_validated_ledger_sequence(client)
+min_ledger = current_validated_ledger + 1
+max_ledger = min_ledger + 20
+
+from xrpl.models.transactions import Payment
+my_payment = Payment(
+    account=test_wallet.classic_address,
+    amount="22000000",
+    destination="rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe",
+    last_ledger_sequence=max_ledger,
+    sequence=test_wallet.next_sequence_num,
+    fee="10",
+)
+print("Payment object:", my_payment)
+print(f"Can be validated in ledger range: {min_ledger} - {max_ledger}")
+
+# Sign transaction -------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


content/_code-samples/send-xrp/send-xrp.py
```diff
@@ -0,0 +1,74 @@
+# Example Credentials ----------------------------------------------------------
+from xrpl.wallet import Wallet
+test_wallet = Wallet(seed="sn3nxiW7v8KXzPzAqzyHXbSSKNuN9")
```
mvadari: You don't have to use keyword args here

content/_code-samples/send-xrp/send-xrp.py
```diff
@@ -0,0 +1,74 @@
...
+    tx_response = xrpl.transaction.get_transaction_from_hash(tx_id, client)
+    if tx_response.is_successful():
+        if tx_response.result.get("validated"):
+            print("Got validated result!")
+            break
+        else:
+            print(f"Results not yet validated... "
+                  f"({current_validated_ledger}/{max_ledger})")
+    if current_validated_ledger > max_ledger:
+        print("max_ledger has passed. Last tx response:", tx_response)
```
mvadari: Is there a reason this is being done by hand instead of via `send_reliable_submission`?

mDuo13: To fit into the existing [Send XRP document](https://xrpl.org/send-xrp.html), which demonstrates the preliminary submit result in one step and the final result later, I matched the process to the existing steps.

content/_code-samples/send-xrp/send-xrp.py
mDuo13: Since this is some introductory-level example code, I explicitly added the keyword to make it clearer what's happening here.

content/tutorials/use-simple-xrp-payments/send-xrp.md
```diff
@@ -116,13 +169,25 @@ The signing API also returns the transaction's ID, or identifying hash, which yo
...
 
-Of course, if the same transaction was previously submitted, it could already be in a previous ledger. (It can't succeed a second time, but you may not realize it succeeded if you aren't looking in the right ledger versions.)
+- **JavaScript:** Use the [submit() method](rippleapi-reference.html#submit) to submit a transaction to the network. Use the [`getLedgerVersion()` method](rippleapi-reference.html#getledgerversion) to get the latest validated ledger index.
+- **Python:** Use the [xrpl.transaction.submit_transaction_blob() method](https://xrpl-py.readthedocs.io/en/latest/source/xrpl.transaction.html#xrpl.transaction.submit_transaction_blob) to submit a transaction to the network. Use the [xrpl.ledger.get_latest_validated_ledger_sequence() method](https://xrpl-py.readthedocs.io/en/latest/source/xrpl.ledger.html#xrpl.ledger.get_latest_validated_ledger_sequence) to get the latest validated ledger index.
```
mDuo13: Should probably use the new autofill capabilities and update this line

content/_code-samples/send-xrp/send-xrp.py
mDuo13: I added some notes (to the code and the page) telling people to use the reliable submit methods where they can

mDuo13:
The broken link is to a page added by #1004, so it should be automatically resolved when both are merged.

PR #1011, opened 2021-03-29
Add python example for secure signing

ryangyoung:
Adds example code for signing a transaction with [xrpl-py](https://github.com/XRPLF/xrpl-py) and updates text accordingly.

### DIFF
--- a/content/_code-samples/secure-signing/py/sign-payment.py
+++ b/content/_code-samples/secure-signing/py/sign-payment.py
@@ -1,32 +1,20 @@
-from xrpl.core import keypairs 
-
-# Define network client
-from xrpl.clients import JsonRpcClient
-JSON_RPC_URL = "https://s.altnet.rippletest.net:51234/"
-client = JsonRpcClient(JSON_RPC_URL)
-
 # Define signer address
+import os  
+my_secret = os.getenv("MY_SECRET")
 from xrpl.wallet import Wallet
-test_wallet_1 = Wallet(seed="shrPSF6vV3v3yoND9J3NeKks6M3xd")
-print(test_wallet_1.classic_address) # "raaFKKmgf6CRZttTVABeTcsqzRQ51bNR6Q"
-
-# Prepate the tranaaction
-from xrpl.ledger import get_latest_validated_ledger_sequence
-current_validated_ledger = get_latest_validated_ledger_sequence(client)
-min_ledger = current_validated_ledger + 1
-max_ledger = min_ledger + 20
+wallet = Wallet(seed="MY_SECRET")
+print(wallet.classic_address) # "raaFKKmgf6CRZttTVABeTcsqzRQ51bNR6Q"
 
 from xrpl.models.transactions import Payment
-# not working in testing
-# from xrpl.utils import xrp_to_drops
+from xrpl.utils import xrp_to_drops
 my_payment = Payment(
     account=test_wallet_1.classic_address,
-    amount="2200000",
+    amount=xrp_to_drops(22),
+    fee=xrp_to_drops(10),
     destination="rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe",
-    sequence=test_wallet_1.next_sequence_num,
+    sequence="16126889",
 )
 print("Payment object:", my_payment)
-print(f"Can be validated in ledger range: {min_ledger} - {max_ledger}")
 
 # Sign transaction -------------------------------------------------------------
 import xrpl.transaction 
--- a/content/_code-samples/secure-signing/py/sign-payment.py
+++ b/content/_code-samples/secure-signing/py/sign-payment.py
@@ -10,9 +10,9 @@ from xrpl.utils import xrp_to_drops
 my_payment = Payment(
     account=test_wallet_1.classic_address,
     amount=xrp_to_drops(22),
-    fee=xrp_to_drops(10),
+    fee="10",
     destination="rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe",
-    sequence="16126889",
+    sequence=16126889,
 )
 print("Payment object:", my_payment)
 
--- a/content/_code-samples/secure-signing/py/sign-payment.py
+++ b/content/_code-samples/secure-signing/py/sign-payment.py
@@ -1,0 +1,34 @@
+from xrpl.core import keypairs 
+
+# Define network client
+from xrpl.clients import JsonRpcClient
+JSON_RPC_URL = "https://s.altnet.rippletest.net:51234/"
+client = JsonRpcClient(JSON_RPC_URL)
+
+# Define signer address
+from xrpl.wallet import Wallet
+test_wallet_1 = Wallet(seed="shrPSF6vV3v3yoND9J3NeKks6M3xd")
+print(test_wallet_1.classic_address) # "raaFKKmgf6CRZttTVABeTcsqzRQ51bNR6Q"
+
+# Prepate the tranaaction
+from xrpl.ledger import get_latest_validated_ledger_sequence
+current_validated_ledger = get_latest_validated_ledger_sequence(client)
+min_ledger = current_validated_ledger + 1
+max_ledger = min_ledger + 20
+
+from xrpl.models.transactions import Payment
+# not working in testing
+# from xrpl.utils import xrp_to_drops
+my_payment = Payment(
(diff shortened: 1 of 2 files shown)


content/_code-samples/secure-signing/py/sign-payment.py
```diff
@@ -0,0 +1,34 @@
+from xrpl.core import keypairs 
+
+# Define network client
+from xrpl.clients import JsonRpcClient
+JSON_RPC_URL = "https://s.altnet.rippletest.net:51234/"
+client = JsonRpcClient(JSON_RPC_URL)
+
+# Define signer address
+from xrpl.wallet import Wallet
+test_wallet_1 = Wallet(seed="shrPSF6vV3v3yoND9J3NeKks6M3xd")
```
mDuo13: Leaving the secret hardcoded in the file like this is not actually that secure. The example JavaScript code pulls the secret from environment variables, which is a common technique for managing secrets of scripts, especially when you need to use them from fully-automated scripts/servers/bots.

Here's that JavaScript example:

```js
const from_address = process.env['MY_ADDRESS']
const secret = process.env['MY_SECRET']
```

The Python equivalent would be:

```py
import os
my_secret = os.getenv("MY_SECRET")
wallet = Wallet(seed=my_secret)
```

(Fun fact: there's a ([real example of this in txsplainbot](https://github.com/mDuo13/txsplain/blob/7b88660199970aa61e27256b3ae2f078d1f0c1d5/bot.py#L7).)

Of course, the environment variable thing is only as secure as whatever thing actually sets the environment variable before running the script. If you put `export MY_SECRET=shrPSF6vV3v3yoND9J3NeKks6M3xd` in your `.bashrc` or equivalent file, it's not a whole lot more secure than having it hard-coded (but it _is_ more flexible since you could run the same code with a different secret).

Another way to do it in Python would be to ask the user to type in a secret key using [`getpass()`](https://docs.python.org/3/library/getpass.html)—that way, they could have the key saved encrypted in a password manager and only paste it into the terminal when needed. It's not useful for automated systems, but it's a good basic-level way of getting the secret from the user that can stand-in for more specific/complicated equivalents that might come from more advanced (read: graphical) user interfaces.

content/tutorials/get-started/set-up-secure-signing.md
```diff
@@ -74,14 +74,27 @@ If you use a client library not published by Ripple, make sure it uses proper, s
...
+*Python*
+
+```py
+{% include '_code-samples/secure-signing/py/sign-payment.py' %}
+```
+
+<!-- MULTICODE_BLOCK_END -->
+
+
 For greater security, you can load your secret keys from a management tool such as [Vault](https://www.vaultproject.io/).
```
mDuo13, line 98: I opened up #1014 to track an effort to document how to do something like this in more detail

content/_code-samples/secure-signing/py/sign-payment.py
```diff
@@ -0,0 +1,34 @@
+from xrpl.core import keypairs 
+
+# Define network client
+from xrpl.clients import JsonRpcClient
+JSON_RPC_URL = "https://s.altnet.rippletest.net:51234/"
+client = JsonRpcClient(JSON_RPC_URL)
```
mDuo13: If you notice, the ripple-lib example doesn't connect to the network. That's intentional, since the most secure signing takes place on a machine that's not online at all. The equivalent is possible with xrpl-py.

content/_code-samples/secure-signing/py/sign-payment.py
```diff
@@ -0,0 +1,34 @@
...
+# not working in testing
+# from xrpl.utils import xrp_to_drops
+my_payment = Payment(
+    account=test_wallet_1.classic_address,
+    amount="2200000",
+    destination="rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe",
+    sequence=test_wallet_1.next_sequence_num,
+)
+print("Payment object:", my_payment)
+print(f"Can be validated in ledger range: {min_ledger} - {max_ledger}")
```
mDuo13: This statement is inaccurate because your `Payment()` transaction doesn't have a `last_ledger_sequence` field. I would just drop this.

content/_code-samples/secure-signing/py/sign-payment.py
```diff
@@ -0,0 +1,34 @@
...
+
+# Prepate the tranaaction
+from xrpl.ledger import get_latest_validated_ledger_sequence
+current_validated_ledger = get_latest_validated_ledger_sequence(client)
+min_ledger = current_validated_ledger + 1
+max_ledger = min_ledger + 20
+
+from xrpl.models.transactions import Payment
+# not working in testing
+# from xrpl.utils import xrp_to_drops
```
mDuo13: It just got merged, so you may need to clear out some `__pycache__` folders and `pip install --upgrade` the library

content/_code-samples/secure-signing/py/sign-payment.py
```diff
@@ -0,0 +1,34 @@
...
+# Define signer address
+from xrpl.wallet import Wallet
+test_wallet_1 = Wallet(seed="shrPSF6vV3v3yoND9J3NeKks6M3xd")
+print(test_wallet_1.classic_address) # "raaFKKmgf6CRZttTVABeTcsqzRQ51bNR6Q"
+
+# Prepate the tranaaction
+from xrpl.ledger import get_latest_validated_ledger_sequence
+current_validated_ledger = get_latest_validated_ledger_sequence(client)
+min_ledger = current_validated_ledger + 1
+max_ledger = min_ledger + 20
```
mDuo13: For offline signing you can skip this entire section (you have to either manually specify the `last_ledger_sequence`, as in the ripple-lib example, or omit it)

content/_code-samples/secure-signing/py/sign-payment.py
```diff
@@ -0,0 +1,34 @@
...
+
+from xrpl.models.transactions import Payment
+# not working in testing
+# from xrpl.utils import xrp_to_drops
+my_payment = Payment(
+    account=test_wallet_1.classic_address,
+    amount="2200000",
+    destination="rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe",
+    sequence=test_wallet_1.next_sequence_num,
```
mDuo13: This won't always work depending on how/when you made the `Wallet()` instance and what the corresponding account has done on the ledger. If you, for example, paste in a seed from your password manager, it won't know what sequence number you're at until you explicitly look it up.

For purposes of this example code, we can just hard-code a sequence number. We should also hardcode a fee so that the transaction can be signed fully offline.

mDuo13 (changes requested):
Left some suggestions to make the example signing code more secure and closer in function to the ripple-lib example.

content/_code-samples/secure-signing/py/sign-payment.py
```diff
@@ -1,32 +1,20 @@
...
-# from xrpl.utils import xrp_to_drops
+from xrpl.utils import xrp_to_drops
 my_payment = Payment(
     account=test_wallet_1.classic_address,
-    amount="2200000",
+    amount=xrp_to_drops(22),
+    fee=xrp_to_drops(10),
     destination="rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe",
-    sequence=test_wallet_1.next_sequence_num,
+    sequence="16126889",
```
mDuo13: I believe the sequence should be a number, not a string.

```suggestion
    sequence=16126889,
```

content/_code-samples/secure-signing/py/sign-payment.py
```diff
@@ -1,32 +1,20 @@
...
 
 from xrpl.models.transactions import Payment
-# not working in testing
-# from xrpl.utils import xrp_to_drops
+from xrpl.utils import xrp_to_drops
 my_payment = Payment(
     account=test_wallet_1.classic_address,
-    amount="2200000",
+    amount=xrp_to_drops(22),
+    fee=xrp_to_drops(10),
```
mDuo13: This would be a 10 XRP fee. I don't think you want that.

```suggestion
    fee="10"
```

mDuo13 (changes requested):

mDuo13:
Broken link looks like an unrelated flap (https://pastebin.com/ was down briefly, works again now)

mDuo13 (approved):

PR #1012, opened 2021-03-29
yum install/upgrade notes for 1.7.0

mDuo13:
@alloynetworks noted that several people had reported problems with upgrading to 1.7.0 on CentOS. These instructions should help alleviate those issues. (It's not clear if/how the different order of instructions in the repofile are helping, but this specific version has had several people corroborate that they work, so I've changed them just in case.)

### DIFF
--- a/content/tutorials/manage-the-rippled-server/installation/update-rippled-manually-on-centos-rhel.md
+++ b/content/tutorials/manage-the-rippled-server/installation/update-rippled-manually-on-centos-rhel.md
@@ -8,6 +8,18 @@ These instructions assume you have already [installed `rippled` from the `yum` r
 
 To update manually, complete the following steps:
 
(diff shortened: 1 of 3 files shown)


content/tutorials/manage-the-rippled-server/installation/update-rippled-manually-on-centos-rhel.md
```diff
@@ -8,6 +8,18 @@ These instructions assume you have already [installed `rippled` from the `yum` r
 
 To update manually, complete the following steps:
 
+1. If you are upgrading to `rippled` 1.7.0 from an earlier version, re-add the the repository to get Ripple's updated GPG key. Otherwise, skip this step:
```
ryangyoung:
```suggestion
1. If you are upgrading to `rippled` 1.7.0 from an earlier version, re-add the repository to get Ripple's updated GPG key. Otherwise, skip this step:
```

ryangyoung (approved):
👍

PR #1015, opened 2021-03-30
Known Amendments: 1.6.0 amendment x3 expected

mDuo13:
Updates the table with the current status. These amendments ping-ponged back and forth for a while but their majorities seem stable now so it's pretty important for people to upgrade.

The blog post is the expected URL where a post (currently in review) will go.


intelliot (approved):
Awesome. Thanks for keeping this page up-to-date. I refer to it frequently.

mDuo13:
Link failures are already fixed in master (this was branched before that though)

PR #1016, opened 2021-03-31
Revert "AccountDelete is now active"

q411:
Reverts ripple/xrpl-dev-portal#842


mDuo13:
Was this opened by mistake? The original PR was correct and is still accurate.

PR #1020, opened 2021-04-05
Remove references to legacy domain verification.

mayurbhandary:
Removes references to the old google form for domain verification and directs new validator operators to the livenet explorer.


mDuo13 (approved):
:tada: Glad to finally be rid of this!

PR #1037, opened 2021-04-07
Fix duplicate SigningPubKey in Pseudo-Transactions page

mDuo13:
originally messed up in https://github.com/ripple/xrpl-dev-portal/commit/15b15b7ac00a095f945e803445cba270db10fa45#diff-86530e87bb83057cc9962054e4b68c814eb452b37896739453a8b9d4f7138b57L13 (though the original was wrong too, since it was supposed to be TxnSignature, not Signature)

[source](https://github.com/ripple/rippled/blob/master/src/ripple/protocol/impl/SField.cpp#L209)


mvadari (approved):
Thanks!

PR #1044, opened 2021-04-21
Navigation tech improvements

mDuo13:
- Use hierarchy fields (from Dactyl v0.10+) for page hierarchy
- Use virtual pages (Dactyl v0.14.0+) instead of special cases for top nav
- Use Dactyl's built-in templates (updated in v0.14.0+) for sidebar tree nav and child display on landing pages
- Remove the now-unnecessary "category", "funnel", "doc_type" etc. fields.
- Start moving page metadata to frontmatter where applicable
- Associated SCSS changes.


mDuo13:
So, the link failures were caused by a real problem, which is partly due to Dactyl's built-in templates, which should be fixed in Dactyl v0.14.2, but Jenkins seems to still be using the old templates. I've seen a problem like this before and I don't fully understand why it happens, but I think basically it means the old templates are saved somewhere that the build machine uses those instead of the new ones. So before we merge this, I want to see if I can figure out a fix or workaround.

mDuo13:
OK, the problem may not have been in it using a cached template, but one of the custom templates in this repo hadn't been updated too. Hopefully 6bd04c3 fixes that. When I built with that and tested the link checker locally, the only failure was http://world.std.com/~reinhold/diceware.html (the original Diceware site), which appears to actually be down at the moment. That's unrelated to this PR, so I'm going to leave it for now in the hopes it comes back online.

PR #1049, opened 2021-04-23
Fix search showing behind widget

jakeatdocforce:
Moved search height above. This also will cover the Top button. 
Lowered widget height.
Added close X to widget
Step2 cancel button closes vs moving to next step.


mDuo13:
I suspect it may be difficult to press the "X" button without hitting one of the other buttons next to it (esp. thumbs-down), but this is still an improvement.

mDuo13 (approved):

PR #1050, opened 2021-04-25
Add .circleci/config.yml

sirpenev:


mDuo13:
For security reasons, we have a policy against using CircleCI, but also, it seems like you've only put the basic "hello world" script together here? If you're interested in adding CI to this project, perhaps try using GitHub Actions, and feel free to open a PR when you've got something a little more complete.

Issue #1022, opened 2021-04-05
"References" landing page needs work


mDuo13:
The [References landing page](https://xrpl.org/references.html) is a very confusing landing page that could provide a lot better direction and context around what reference content exists on xrpl.org.

There are four top-level items, all with confusing names, and they aren't particularly representative of what people might want to do with the XRP Ledger:

- **rippled API Reference** → most of the good stuff is buried in here. Transactions! Ledger data! The API methods that are most likely to persist into the future! I mean, there are a lot of [regrettable inconsistencies in the rippled APIs](https://github.com/ripple/rippled/issues?q=is%3Aissue+is%3Aopen+label%3A%22API+Change%22) but they're definitely the most fundamental bedrock of XRP Ledger programming—everything else is built on top of them, so they have to stick around.
    - It could make sense to split certain parts off as a "Protocol Reference" that's relevant to every language and interface, separate from the parts that are specific to calling WebSocket and JSON-RPC. Transactions and Ledger Data Formats are the most obvious ones but there are probably some more things that could/should be split off. There will always be some that don't cleanly fit into one or the other bucket though.
- **XRP-API** → a REST interface built on top of rippled, which is another service you have to stand up independently, never really finished and now approaching one year since significant dev work was done on it. But it has an appealing name, so it's likely to draw people's attention away from the other stuff. The blurb for it is totally misleading and confusing (I'm actually not sure at the moment where it's getting that phrase. I think maybe it's inheriting it from the target? Oops.)
- **RippleAPI Reference** → almost the exact same name as the first thing, but this one's a client library for JavaScript/TypeScript. That's confusing enough, but the actual documentation for this library is also a bit of a mess; it's all in one page, there's no concise list/table of methods, there are a bunch of older interfaces and data structures that are probably considered deprecated to some degree or other, and so on. (The library itself is about due for a breaking-change v2 that pares back a bunch of the deprecated stuff.) Unfortunately since we pull the reference documentation directly from the ripple-lib repository, these changes are a little more challenging to implement.
- **Ripple Data API v2** → This is the third "Ripple API" on this page, although at least this one has a key word "Data" in there to tell you something useful about it. Unfortunately, it's deprecated!
- **xrp-ledger.toml file** → "One of these things is not like the others." Where the other ones are mostly specifications for programming interfaces, this is a document format. 

[...]

PR #1052, opened 2021-05-01
Restructures references section.

mDuo13:
- Removes deprecated references and adds client libraries as described in #1022.
- Renames client library references by name and groups them under a Client Libraries landing page. Takes the table of client libraries from the Set Up Secure Signing page and moves it there.
- Splits some protocol-level reference information into a separate reference from the rippled API methods
- Adds external link markers to virtual pages in the left nav
- Tweaks references landing to respect the `nav_omit: true` parameter.
- Removes the animation on hovering external links (this was left over from the previous design)


mDuo13:
Before: 

After:

intelliot (approved):
haven't tried this out in detail, but overall (looking at screenshots) LGTM

PR #1057, opened 2021-05-04
Fix Get Started w/ Python missing from nav

mDuo13:
#1044 accidentally caused "Get Started Using Python" to disappear from the left nav because it doesn't have a `parent` attribute assigned. This adds that field so it appears in the nav where it should again. 

I wrote a quick script check for other instances of the same and there were no other pages that qualified. (All pages that didn't have a `parent` were either placeholder/redirects, or the homepage.)


mDuo13:
Link failures are unrelated (Boost lib link in the macOS build instructions)

Issue #1045, opened 2021-04-21
Fix wide tables on mobile


mDuo13:
Certain tables (like [Special Addresses](https://xrpl.org/accounts.html#special-addresses)) run off the side of the screen on mobile layouts, and you can't zoom or scroll horizontally to see the whole thing. We should find a way to make these tables legible on mobile.

mDuo13:
Some other spots to check any proposed solution on:

- [Deposit Authorization semantics table](https://xrpl.org/depositauth.html#precise-semantics) (scroll down a little)
- [Open, Closed, and Validated Ledgers](https://xrpl.org/ledgers.html#open-closed-and-validated-ledgers)
- [Transaction Results](https://xrpl.org/transaction-results.html#transaction-results) (this one is actually OK with the current style)
- [tec Codes](https://xrpl.org/tec-codes.html)
- [Unit Tests](https://xrpl.org/commandline-usage.html#unit-tests)
- [PaymentChannelCreate](https://xrpl.org/paymentchannelcreate.html)

mDuo13:
Fixed by #1062

PR #1062, opened 2021-05-10
#1045 Tables on mobile WIP

jakeatdocforce:
One option is to stack on small devices. Hide current table headers and add headers to rows.
Will need to add divs above tables to id each table for naming.  Only 2 tables in this as a test.
Other tables will hide headers completely and full width content.


mDuo13:
Tried building it and I'm not seeing much difference on mobile in the "Special Accounts" table. The table cells have a little bit more space but aren't getting stacked.

jakeatdocforce:
Should look like this. 

Screen Shot 2021-05-10 at 4 15 57 PM

mDuo13:
First of all, the overall approach looks a little fragile. Having to add a unique div and some JS code for each wide table is not optimal, though I imagine some of that could probably be made more generic with some JS and maybe a Dactyl filter.

Second, the result still isn't working quite right, as you can see "Black Hole" and "Yes" are overlapping in this image:



Actually, I did some searching and found out that [Bootstrap's `table-responsive` class](https://getbootstrap.com/docs/4.0/content/tables/#responsive-tables) makes the table horizontally scrollable, which is probably what we want for most or all tables in the docs. So we can probably either copy the style rules for that for `.content table` or we can find a way to add that class to tables in the content, like with a Dactyl filter.

jakeatdocforce:
correct I don't think its a useable or scaleable strategy.  Was thinking of a separate way using JS to make more generic for all tables. 
But if scrolling is acceptable that would be fastest/easiest to implement.

tool/filter_css_tables.py
```diff
@@ -0,0 +1,18 @@
+################################################
+## Add class to tables
+## Author: Rome Reginelli
```
mDuo13: Hey, take credit for your own work here! :laughing:

mDuo13 (changes requested):

tool/filter_css_tables.py
jakeatdocforce: haha done

mDuo13 (approved):

mDuo13:
Rebuilt and it looks good now.

PR #1066, opened 2021-05-11
Adjust colors and elements for print.

jakeatdocforce:
Code colors might need to change from the green if needed.


mDuo13:
For some reason the text seems to run off the side of the page for me. This happens in Firefox even with the page size set to Letter, unless I check "Fit to page width." In Chrome it happens sometimes when the page size is smaller than Letter, but I can't seem to reproduce consistently.

When there are multiple code samples, the print CSS used to leave the names of the language tabs as plain text above the code samples. I'm guessing that got broken when we moved some of the code tab stuff to a Dactyl filter that runs compile-time instead of doing it in JS. So it might not be easy to fix, but if we could add the tab labels back, that would be good. Also, as you mentioned, the colors of the syntax highlighting are a bit pale.

The "Expand/Collapse" button on code samples should be hidden in print.

We might also want to hide `.interactive-block` elements since those don't do anything of value in print. (The [Send XRP page](https://xrpl.org/send-xrp.html) is an example with those.)

Not the highest priority, but the [theme-aware diagrams styles](https://github.com/ripple/xrpl-dev-portal/pull/885) should probably apply light-mode colors in print instead of the current dark theme styles. (Actual light mode is still TBD, but we can at least unswap black and white.) A good test case is the [Consensus](https://xrpl.org/consensus.html) diagrams and their white text.

styles/_print.scss
```diff
@@ -4,9 +4,9 @@
   .multicode > div {
     display: block !important;
   }
-  .multicode > ul {
-    display: none !important;
-  }
+  // .multicode > ul {
+  //   display: none !important;
+  // }
```
mDuo13: Let's not leave obsolete, commented-out code in place.

```suggestion
```

mDuo13 (approved):

PR #1067, opened 2021-05-11
Update jQuery, styles deps

mDuo13:
- Update jQuery. The site's currently on an ancient, unsupported version of jQuery. We don't depend on anything that had breaking changes so the latest version (3.6.0) is just a drop-in replacement.
- Update the yarn lockfile for SCSS and rebuild the CSS. This upgrades us to Bootstrap 4.6.0, lodash 4.17.21, etc. (The lodash one fixes a security issue, though I think we aren't affected by the security issue regardless.)
- Remove some obsolete code from the PDF template.


mDuo13:
Build failure is due to the problem fixed in #1063

PR #1071, opened 2021-05-13
Remove more stuff

mDuo13:
- remove Data API tool (Data API is deprecated with no ongoing support)
- update dev tools landing page (EN and JA) to remove data API tool
- remove target for building XRP API ref only or Data API ref only (no longer relevant following #1052)
- Make tx sender template translatable, & partially update translation strings


SirJosh1987 (commented):
My bad

SirJosh1987 (commented):
Got it

PR #1073, opened 2021-05-13
Ledger close times details

mDuo13:
I've written about this before [on XRPChat](https://www.xrpchat.com/topic/35474-cant-seem-to-understand-the-ledger_time-field-in-ledger/?tab=comments#comment-848038) but it's come up a few more times so I realized I should clarify in the concept docs.


intelliot (approved):

ximinez (approved):
It makes sense to me.

PR #1074, opened 2021-05-14
Add vl.xrplf.org to list providers

alloynetworks:
Add https://vl.xrplf.org as a list provider.


ximinez:
Should Coil's VL be on that list?

alloynetworks:
> Should Coil's VL be on that list?

Certainly does make sense.

mDuo13 (approved):

PR #1080, opened 2021-05-19
9Update offercreate.md

Yonat29:
rail permit cluster pool skate now team grab nasty answer alarm wink


mDuo13:
Not sure if this is a mistake or spam, but this is not a constructive pull request.

PR #1088, opened 2021-05-28
Clarify that issuing `feature` via the commandline requires 'accept' or 'reject' booleans.

jscottbranson:
The `feature` command requires `vetoed` to be set to 'true' or 'false' for websockets/rpc commands, however, the commandline requires either 'accept' or 'reject', and I believe the documentation could benefit from explicitly mentioning this inconsistency.


mDuo13:
This pattern is not specific to this method, but is a convention of the commandline API in general. You see something similar in a lot of other methods, too. For example, [account_info](https://xrpl.org/account_info.html)'s `strict` parameter has to be specified as `strict` on the commandline, but `true`/`false` in JSON. Same with `tx` or `full` in the [ledger method](https://xrpl.org/ledger.html).

So, while I don't think it's a bad thing to mention it, I worry it might also be confusing to add this in just this one place, since it implies this case is different than the others.

llattanaja (approved):

jscottbranson:
As always, @mDuo13 you make an outstanding point.

Right now, I think the convention is to use the command line examples to show that the request patterns for the command line are different, so I guess it is a bit redundant to document the same thing in the examples and in the body text.

However, I do think it would make things slightly more clear/easier to have the differences between the command line and JSON requests more explicitly discussed in the text body.

If others are agreeable to documenting the differences between command line and JSON requests in the actual text, I would be willing to go through all the commands and make additions similar to this PR. Otherwise, it probably makes sense to leave it as is.

Cheers!

mDuo13:
I would be in support of clarifying the differences in the body text of other methods too, especially if you're volunteering to do it! :smile: 

Small suggestion, by the way, if you do: instead of saying "Use ... on the commandline," phrase it as "On the commandline, use ..."—this lets readers do the equivalent of boolean short-circuiting the sentence, and stop reading if that's not relevant to them.

SirJosh1987 (approved):

content/references/rippled-api/admin-rippled-methods/status-and-debugging-methods/feature.md
```diff
@@ -65,7 +65,7 @@ The request includes the following parameters:
 | `Field`   | Type    | Description                                            |
 |:----------|:--------|:-------------------------------------------------------|
 | `feature` | String  | _(Optional)_ The unique ID of an amendment, as hexadecimal; or the short name of the amendment. If provided, limits the response to one amendment. Otherwise, the response lists all amendments. |
-| `vetoed`  | Boolean | (Optional; ignored unless `feature` also specified) If true, instructs the server to vote against the amendment specified by `feature`. If false, instructs the server to vote in favor of the amendment. |
+| `vetoed`  | Boolean | (Optional; ignored unless `feature` also specified) If true, instructs the server to vote against the amendment specified by `feature`. If false, instructs the server to vote in favor of the amendment. Use 'accept' or 'reject rather than 'true' or 'false' when using the commandline. |
```
SirJosh1987: accept

SirJosh1987 (approved):

content/references/rippled-api/admin-rippled-methods/status-and-debugging-methods/feature.md
intelliot:
```suggestion
| `vetoed`  | Boolean | (Optional; ignored unless `feature` also specified) If true, instructs the server to vote against the amendment specified by `feature`. If false, instructs the server to vote in favor of the amendment. On the commandline, use 'accept' or 'reject rather than 'true' or 'false'. |
```

mDuo13 (approved):

PR #1090, opened 2021-05-31
Light mode

jakeatdocforce:
Toggle is there for testing. Will need to be moved to final spot in nav. 
Colors will likely need to be tweaked when an updated design is ready. 
On load will check for system preference first and save that.  If user interacts with toggle that will take priority.
This could result in the default for most users being the light theme.  Might be useful to test how many use dark vs light. 
New light overrides are in the _light-theme sass file. 

Also added a watch command for testing sass builds during dev.


assets/js/theme-switch.js
```diff
@@ -0,0 +1,64 @@
+// Check user prefers color, toggle light/dark, save state
+// Based partly on https://github.com/vinorodrigues/bootstrap-dark
+
+$(document).ready(function() {
```
mDuo13: Most of this functionality shouldn't be behind a `$(document).ready()` callback. That causes a flash of the default theme on each new page as it loads, which switches over when it's done. I tested this and got it working with just the bind function in the ready callback. (Feel free to cherry-pick this commit: 9b2563d0f65da830d4fe9905be97dbdcae8df899 )

styles/light/_light-theme.scss
```diff
@@ -0,0 +1,170 @@
...
+
+// Colors
+$navbar-light-color: $dark;
+$navbar-light-hover-color: $secondary;
+$navbar-light-active-color: $primary;
+
+// Nav logo. 
+
+.logo {
+  content: url(/assets/img/XRPLedger_DevPortal-black.svg);
```
mDuo13: This does not work for me in Firefox. (It does work in Chrome.)

styles/xrpl.scss
```diff
@@ -53,3 +53,19 @@ $font-family-sans-serif: -apple-system, system-ui, 'Roboto', sans-serif;
...
+
+//Mixin to conditionaly include other themes
+// @import "light/_mixins.scss";
+
+// Default to dark theme with light being alt
+// $color-scheme-alt: light;
+
+// @include prefers-color-scheme($color-scheme-alt) {
+//     @import "light/_light-theme.scss";
+// }
```
mDuo13, line 67: Are these comments accurate? If they're not just leftover copy-paste stuff, I'd like a little more clarity on what this means.

mDuo13 (changes requested):
Can we also set this up to change out images when you toggle themes? If nothing else, there's probably a way to do it with some CSS background image trickery. For the actual image(s) to replace, you can do a quick-and-dirty placeholder. Like, here's a replacement for `img/graphics/validator.png` (used in overview.html):
validator-dark

assets/js/theme-switch.js
jakeatdocforce: Fixed.  I was just keeping the code in the previous block but you are correct it can be outside that.

styles/xrpl.scss
jakeatdocforce: This was/is for a different way to handle the theme changes.  
If we went this direction it would load and respect only the users system pref.   Basically would load only if user had a light pref set.  But not able to handle user toggle on page.  As the content wouldn't be loaded.  Media query for css.  

Left in thinking it could be discussed as a potential option if you didn't need or want user to select manually.   
Looks like Github for example has a setting in your prefs for overriding the system settings and will add separate themes on multiple conditions.   
Unsure how far down that rabbit hole we go.

SirJosh1987 (approved):

mDuo13:
Closing in favor of #1119.

PR #1098, opened 2021-06-11
Codesample tabs

jakeatdocforce:
Save a preferred language on pages with multiple code samples.  
Will change all tabs to match that language if available.


mDuo13:
Looks pretty great! Found at least one significant bug though:

If you go to (your locally-built equivalent of [Example CheckCreate Preparation](https://xrpl.org/send-a-check.html#example-checkcreate-preparation) there's a tab whose name has some characters that cause a syntax error in the selector syntax:

```
Uncaught Error: Syntax error, unrecognized expression: .multicode[id='code-0']>UL>LI>A[class=jsonrpc,websocket,orcommandline]
    jQuery 7
    setTabsToLang http://xrpl.local/assets/js/multicodetab.js:67
    showSlectedTab http://xrpl.local/assets/js/multicodetab.js:91
    jQuery 2
    showSlectedTab http://xrpl.local/assets/js/multicodetab.js:82
    minitabs http://xrpl.local/assets/js/multicodetab.js:110
    jQuery 9
    minitabs http://xrpl.local/assets/js/multicodetab.js:107
    jQuery 2
    minitabs http://xrpl.local/assets/js/multicodetab.js:101
    <anonymous> http://xrpl.local/send-a-check.html:71
    jQuery 13
jquery-3.6.0.min.js:2:13639
    jQuery 7
    setTabsToLang http://xrpl.local/assets/js/multicodetab.js:67
    showSlectedTab http://xrpl.local/assets/js/multicodetab.js:91
    jQuery 2
    showSlectedTab http://xrpl.local/assets/js/multicodetab.js:82
    minitabs http://xrpl.local/assets/js/multicodetab.js:110
    jQuery 9
    minitabs http://xrpl.local/assets/js/multicodetab.js:107
    jQuery 2
    minitabs http://xrpl.local/assets/js/multicodetab.js:101
    <anonymous> http://xrpl.local/send-a-check.html:71
    jQuery 13
```

Looks like `jsonrpc,websocket,orcommandline` really is the name of the class, so I think you'll need to strip more punctuation out in the code that generates that.

Also, this is not a blocker, but less than ideal behavior I noticed: if you tab to a specific language (say, Python) and then, on a page that doesn't have that language (say, one of the API ref pages that has WS/JSON-RPC/CLI options) and you select a tab there, it forgets that you preferred Python on the first page. That is not _unreasonable_ behavior, but it would be better if it had, like, a brief history of tabs you'd selected so that, failing the first, it could pick the second- or third-most recent tab name you had selected.

jakeatdocforce:
Yep looks like the tab name had commas in. Stripped those out now.  
So currently it only saves the one language name.  Potentially we could save both page and language pref.  Or a list of the most selected language. Then like you say go down the list checking for those.  
Might run into issues if  say python and javascript are on the list and you change to a different language.  Does that resort them all? New selected language at the top? And what if you wanted python on one page and javascript on the other where both pages have both?  
Safest might be to match page url with language?

SirJosh1987 (approved):

mDuo13 (approved):
Revised version looks good. (And the CI failure is just because it's coming from a fork.)

PR #1099, opened 2021-06-16
macOS build: build boost yourself again

intelliot:
I was not able to install boost@1.75 using homebrew:
```
➜  brew install boost@1.75
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
Error: No available formula or cask with the name "boost@1.75".
==> Searching for a previously deleted formula (in the last month)...
Error: No previously deleted formula found.
==> Searching taps on GitHub...
Error: No formulae found in taps.
```

Downloading boost 1.75.0 and building it seems to work.

### DIFF
--- a/content/tutorials/manage-the-rippled-server/installation/build-run-rippled-macos.md
+++ b/content/tutorials/manage-the-rippled-server/installation/build-run-rippled-macos.md
@@ -27,13 +27,11 @@ For development purposes, run `rippled` as a non-admin user, not using `sudo`.
 
 0. Use Homebrew to install dependencies.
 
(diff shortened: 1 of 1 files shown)


mDuo13:
_\*sigh\*_ I guess that nice period of not having to build it was short-lived.

mDuo13 (approved):

mDuo13:
Also, I'm not sure what's wrong with the CI; it's failing to after it does almost everything, and then it fails at a step where it's trying to get the head commit ID from the pull request info. Might be that GitHub updated something that the action we're using isn't ready for. In any case, the link checker passes and the preview looks good.

PR #1103, opened 2021-06-23
Update build-run-rippled-macos.md

solman-gh:


SirJosh1987 (approved):

SirJosh1987 (approved):

mDuo13 (approved):
cloning via SSH is more flexible and powerful but for this use case, I agree HTTPS is fine and it might be simpler to set up.

PR #1113, opened 2021-06-28
Ripple data API alternatives

WietseWind:
Add alternative (as most questions I received so far are about simple things like getting account balance or transaction history, HTTP Post to JSON RPC would do.

Would it make sense to add a quick ready to use snippet __like__ this (not exactly this one)?
https://runkit.com/wietsewind/xrpl-json-rpc

image

### DIFF
--- a/content/references/data-api.md
+++ b/content/references/data-api.md
@@ -10,3 +10,10 @@ nav_omit: true
 **Warning:** The Ripple Data API v2 is deprecated with no ongoing support. Please use the native [XRP Ledger HTTP APIs](rippled-api.html) instead.
 
 For information on the old Data API, see the [rippled-historical-database repository](https://github.com/ripple/rippled-historical-database).
+
+## Alternatives
+
+Most common operations like requesting account information and balances, transaction history, etc. 
+can be sent to self hosted or public XRP Ledger nodes. Data can be retrieved using a [WebSocket connection](get-started-using-http-websocket-apis.html#websocket-api) or [JSON RPC (HTTP POST)](.org/get-started-using-http-websocket-apis.html#json-rpc).
+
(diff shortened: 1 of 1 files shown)


content/references/data-api.md
```diff
@@ -10,3 +10,10 @@ nav_omit: true
 **Warning:** The Ripple Data API v2 is deprecated with no ongoing support. Please use the native [XRP Ledger HTTP APIs](rippled-api.html) instead.
 
 For information on the old Data API, see the [rippled-historical-database repository](https://github.com/ripple/rippled-historical-database).
+
+## Alternatives
+
+Most common operations like requesting account information and balances, transaction history, etc. 
+can be sent to self hosted or public XRP Ledger nodes. Data can be retrieved using a [WebSocket connection](get-started-using-http-websocket-apis.html#websocket-api) or [JSON RPC (HTTP POST)](.org/get-started-using-http-websocket-apis.html#json-rpc).
```
mDuo13:
```suggestion
For most common operations, like requesting account balances or transaction history, you can query a self-hosted or [public XRP Ledger server](public-servers.html) using a [WebSocket connection](get-started-using-http-websocket-apis.html#websocket-api) or [JSON-RPC (HTTP POST)](get-started-using-http-websocket-apis.html#json-rpc).
```

mDuo13 (changes requested):
Thanks, and yes, that's a good point. I've seen this too—people try to use the Data API because they don't realize that JSON-RPC is just a plain HTTP POST.

I like the idea of making a big list of examples demonstrating a simple JSON-RPC call, like the one you linked, for a variety of programming languages, and putting them all in the Getting Started with the HTTP APIs page.

Anyway I fixed a typo in the link and rephrased the text slightly.

SirJosh1987 (commented):
Fixed?

PR #1115, opened 2021-06-29
Include reference to the xrpintel API as alternate data source

movitto:
As a followup to #1113, we're proposing to add reference to the xrpintel API which provides similar functionality to that which the Data API v2 used to including list of accounts and known gateways. This functionality is useful to seed custom datasets and tools based on the XRPL.


SirJosh1987 (approved):

mDuo13 (approved):
Excellent point, and a great resource to highlight here.

PR #1116, opened 2021-06-30
Update capacity-planning.md

mtrippled:


SirJosh1987 (approved):

mDuo13 (approved):
It's a bummer to have to increase the requirements just after we blogged about how we reduced the memory requirements, but it has been a long time since we last increased them and the network _is_ bigger than it used to be...

stininger1:
How do I get this to my account

PR #1121, opened 2021-07-09
Template fixes: faucet, xrp overview

mDuo13:
- Remove broken tabs in XRP test faucets sidebar (#1098 broke them)
- On overview page, dynamically show the most recent amount of XRP remaining in escrow

### DIFF
--- a/template/page-xrp-faucets.html.jinja
+++ b/template/page-xrp-faucets.html.jinja
@@ -3,40 +3,18 @@
 {% block bodyclasses %}page-test-net{% endblock %}
 
 {% block right_sidebar %}
-<div class="card">
-  <div class="card-header">
     <h4>Testnet Servers</h4>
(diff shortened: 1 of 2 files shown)


SirJosh1987 (approved):

intelliot (approved):
I believe we are hoping to deprecate the Data API, so we may want to consider what replaces it for the purpose of tracking Ripple's escrows.

SirJosh1987 (approved):

mDuo13:
> I believe we are hoping to deprecate the Data API, so we may want to consider what replaces it for the purpose of tracking Ripple's escrows.

Yes, well, as long as whatever replaces it has HTTP API access or something along those lines, we can update to that. I guess that'll have to be a factor in the discussion around deprecating the Data API.

PR #1122, opened 2021-07-15
ripple-lib: update to 1.9.7

intelliot:
includes the updates from https://github.com/ripple/ripple-lib/pull/1459


mDuo13 (approved):
Thanks! I've updated the copy of ripple-lib in this repo so the interactive pages that use it are also on the latest version.

PR #1127, opened 2021-07-24
Updated feedback widget

jakeatdocforce:
Removed old widget js
Added new widget js
Updated embed with new format for options.


mDuo13:
I'm still seeing a mysterious warning in the Firefox console about cookies that's coming from the updated script. (Even after building commit ebc062e19551645f7f3bd90236c8e0f3e0123baa .)

```text
Some cookies are misusing the recommended “SameSite“ attribute
Cookie “_c_” will be soon rejected because it has the “SameSite” attribute set to “None” or an invalid value, without the “secure” attribute. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite feedback-widget.js:8:254272
Cookie “_c_” will be soon rejected because it has the “SameSite” attribute set to “None” or an invalid value, without the “secure” attribute. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite feedback-widget.js:8:254272
```

Unfortunately in the minified JS it's pretty difficult to tell where this is coming from, but we should track it down and make sure that the feedback widget and its dependencies are not misusing cookies.

mDuo13:
OK, since we've confirmed that the cookie is coming from the Google Analytics bits (already there in the other one) we can merge.

PR #1130, opened 2021-07-28
reserves.md - accounts can now be deleted

intelliot:
Also: removed "and the full ledger history on disk" because I don't think reserve requirements really reduce the size of the ledger history. (But I could be persuaded.) Transaction fees are the main way that ledger history is kept from growing excessively large.

### DIFF
--- a/content/concepts/payment-system-basics/accounts/reserves.md
+++ b/content/concepts/payment-system-basics/accounts/reserves.md
@@ -9,19 +9,19 @@ top_nav_grouping: Popular Pages
 ---
 # Reserves
 
-The XRP Ledger applies _reserve requirements_, in XRP, to protect the shared global ledger from growing excessively large as the result of spam or malicious usage. The goal is to constrain the growth of the ledger to match improvements in technology so that a current commodity-level machine can always fit the current ledger in RAM and the full ledger history on disk.
+The XRP Ledger applies _reserve requirements_, in XRP, to protect the shared global ledger from growing excessively large as the result of spam or malicious usage. The goal is to constrain the growth of the ledger to match improvements in technology so that a current commodity-level machine can always fit the current ledger in RAM.
 
-To submit transactions, an address must hold a minimum amount of XRP in the shared global ledger. You cannot send this XRP to other addresses. To fund a new address, you must send enough XRP to meet the reserve requirement.
+To have an account object, an address must hold a minimum amount of XRP in the shared global ledger. You cannot send this XRP to other addresses. To fund a new address, you must send XRP to that address in an amount that satisfies the minimum reserve requirement.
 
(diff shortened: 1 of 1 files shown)


content/concepts/payment-system-basics/accounts/reserves.md
```diff
@@ -9,19 +9,19 @@ top_nav_grouping: Popular Pages
...
 
-The XRP Ledger applies _reserve requirements_, in XRP, to protect the shared global ledger from growing excessively large as the result of spam or malicious usage. The goal is to constrain the growth of the ledger to match improvements in technology so that a current commodity-level machine can always fit the current ledger in RAM and the full ledger history on disk.
+The XRP Ledger applies _reserve requirements_, in XRP, to protect the shared global ledger from growing excessively large as the result of spam or malicious usage. The goal is to constrain the growth of the ledger to match improvements in technology so that a current commodity-level machine can always fit the current ledger in RAM.
 
-To submit transactions, an address must hold a minimum amount of XRP in the shared global ledger. You cannot send this XRP to other addresses. To fund a new address, you must send enough XRP to meet the reserve requirement.
+To have an account object, an address must hold a minimum amount of XRP in the shared global ledger. You cannot send this XRP to other addresses. To fund a new address, you must send XRP to that address in an amount that satisfies the minimum reserve requirement.
```
mDuo13: > To fund a new address, you must send XRP to that address in an amount that satisfies the minimum reserve requirement.

This is much wordier than the original text. How about this instead?

```suggestion
To have an account, an address must hold a minimum amount of XRP in the shared global ledger. You cannot send this XRP to other addresses. To fund a new address, you must send that address enough XRP to meet the reserve requirement.
```

content/concepts/payment-system-basics/accounts/reserves.md
```diff
@@ -9,19 +9,19 @@ top_nav_grouping: Popular Pages
...
 
-The current minimum reserve requirement is **20 XRP**. (This is the cost of an address that owns no other objects in the ledger.) Each new [account](accounts.html) must set aside this much XRP, which cannot be recovered or sent to others.
+As determined by [Fee Voting](fee-voting.html), the current minimum reserve requirement is **20 XRP**. (This is the cost of an address that owns no other objects in the ledger.) Each new [account](accounts.html) must set aside this much XRP. Some of this XRP can be recovered by [deleting the account](accounts.html#deletion-of-accounts).
```
mDuo13: Let's cut straight to the point about the current value, then explain later how it can change:

```suggestion
The current minimum reserve requirement is **20 XRP**. (This is the cost of an address that owns no other objects in the ledger.) Each new [account](accounts.html) must set aside this much XRP. Some of this XRP can be recovered by [deleting the account](accounts.html#deletion-of-accounts).

The [Fee Voting](fee-voting.html) process can change the reserve requirement if enough validators agree to new reserve settings.
```

mDuo13 (changes requested):

intelliot:
good suggestions  👍

PR #1133, opened 2021-08-03
add top banner for Apex event

jakeatdocforce:
Add top banner for Apex event. 
Added styles in separate file so no original files would be changed.


mDuo13:
The banner breaks [the offset for anchor links](https://github.com/XRPLF/xrpl-dev-portal/blob/master/styles/_top-nav.scss#L542). It also doesn't look good in light mode:



FWIW, you can build with the theme toggler enabled by running this command:

```sh
dactyl_build --vars '{"light_theme_enabled": true}'
```

mDuo13 (approved):

PR #1142, opened 2021-08-10
fix: include last_close in the server_info/server_state response formats table

mukulljangid:
Attribute exists in the example request and should be included in the table below.


mukulljangid:
@mDuo13 and @manojsdoshi Please review this. Thanks!

mDuo13 (approved):
I was a bit confused because `last_close` has been there in the server_info docs since the beginning and the field has been reported by the API in both commands for 7 years. But yeah, good catch, the server_state docs were missing it from the table.

I realized that the server_info / server_state examples were a bit old, so I updated the example responses (for both) to non-admin responses from s1.ripple.com.

I also clarified the child fields of `last_close` since those details were missing for both, and I added a couple links to clarify the node public key.

mDuo13:
The CI is failing on stuff that is already fixed in the master branch. Merging.

PR #1144, opened 2021-08-16
First pass copy to clipboard

jakeatdocforce:
Works on main docs pages.  
* Issues on get started page where there are multi step hidden code blocks.  
* Needs styling. 
* Does this need to be added for any 'output' blocks?  Probably not but they have the class codehilight so will get the button added.  Unless some way to designate those as not needing the button.


SirJosh1987 (approved):

mDuo13:
Functionally, it seems to work well. Visually, it's obviously a placeholder.

Let's use a clipboard icon and style it like the code expand/collapse buttons. Something like this should work:

```html
<button class="btn btn-outline-secondary"><i class="fa fa-clipboard"></i>
```

Also, place it next to the expand/collapse button (in a [button group](https://getbootstrap.com/docs/4.6/components/button-group/) if possible). While you're in there, I think the code that decides whether a particular code block needs an expand/collapse button broke with some other update (maybe the change in syntax highlighters?) so maybe you could look into how that can be fixed/improved.

EDIT: P.S. also, please add the button to the permalink and curl syntax dialogs of the [websocket API tool](https://xrpl.org/websocket-api-tool.html).

mDuo13 (changes requested):
Looks like good progress. Don't forget to add the appropriate Bootstrap classes to the button (probably `btn btn-outline-secondary`) so it gets styled appropriately. And I guess the work to put it in a toolbar it with the expand/collapse button is still in progress?

jakeatdocforce:
@mDuo13  Added both copy and expand btns to a group in the top right of the box.  And also the websocket tool. 
Expand doesn't work still on non multi tabs yet.  The way code is grouped and named will need some restructuring for all of it. And also detecting if a hidden element has scrollbars has been tricky.  
Some options could be to toggle the expand, get h/w, then close again but that could cause delays or flashes for user.  
Or we could add the buttons only when a user selects the tab vs adding all on load or in preprocess

SirJosh1987 (approved):

mDuo13:
Merged as https://github.com/XRPLF/xrpl-dev-portal/commit/6c5b953a4bee07e476c511b301c40d60cd7acb39 et al.

PR #1155, opened 2021-08-27
update to robots.txt file

Calvinjwala:
Updating the robots.txt file for crawling

### DIFF
--- a/template/robots.txt.jinja
+++ b/template/robots.txt.jinja
@@ -1,1 +1,5 @@
+User-agent: *
+
+Allow: /
+
 Sitemap: http://www.xrpl.org/sitemap.txt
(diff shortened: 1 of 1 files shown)


template/robots.txt.jinja
```diff
@@ -1 +1,5 @@
+User-agent: *
+
+Allow: /
```
mDuo13: Seems like the `Allow` directive [is specific to Googlebot](https://moz.com/learn/seo/robotstxt), and if we just want to tell bots they can crawl the whole site, this syntax is more universal:

```suggestion
Disallow: 
```

mDuo13 (changes requested):

PR #1157, opened 2021-08-30
ledger method: add validated field to table

### LINKED ISSUE #1156: Discrepancy between example and format description for `ledger` API call
Relevant page: https://xrpl.org/ledger.html

Under `Response Format`, the example has a field called `validated`, but the delineation of the different fields does not list `validated` as a field that should be there. Which is correct?

mDuo13:
Fixes #1156.

I spent some time trying to figure out when the field was added to the response, but didn't reach a definitive conclusion. I'm unclear if it was added as part of the Reporting Mode changes or if it's been there for many years, so I didn't add a "New in" label.


mvadari (approved):
Thanks!

PR #1160, opened 2021-09-01
Rename content/concepts/payment-system-basics/accounts/reserves.md to…

mooha12:


mDuo13:
Looks like this may have been a mistake? The filenames are meaningful, and not the right place for what looks like an XRP Ledger address.

PR #1164, opened 2021-09-08
Max fee values: use superscripts

mDuo13:
The existing table inconsistently uses ** or ^ to indicate exponents. So I changed it to use inlined HTML to produce superscripts instead.


intelliot (approved):
LGTM

PR #1171, opened 2021-09-19
Remove reference to increase in value of XRP

bharathchari:
It is undesirable to mention that the burning of fees increases value of XRP. No price reference will be made on xrpl.org.


bharathchari:
Sorry @mDuo13 - I wanted to merge this with some urgency but it seems to fail the checks.

PR #1175, opened 2021-09-22
Add example of Ledger Close Time determination

movitto:
Expands upon documentation providing example of how Ledger Close Time determination works

### DIFF
--- a/content/concepts/payment-system-basics/ledgers.md
+++ b/content/concepts/payment-system-basics/ledgers.md
@@ -56,6 +56,28 @@ Since new ledger versions usually close about every 3 to 5 seconds, these rules 
 
 Generally speaking, the ledger cannot make any time-based measurements that are more precise than the close time resolution. For example, to check if an object has passed an expiration date, the rule is to compare it to the close time of the parent ledger. (The close time of a ledger is not yet known when executing transactions to go into that ledger.) This means that, for example, an [Escrow](escrow.html) could successfully finish at a real-world time that is up to about 10 seconds later than the time-based expiration specified in the Escrow object.
 
+### Example
+
+Lets say the last close time is 12:00:00:
+
+**Current consensus round**
+
+- A validator individually determines that the next close time should be 12:00:03
+- This is then rounded down based on the close time resolution to 12:00:00
+- But since this is not greater than the last close time, it's incremented to 12:00:01
+- The validators then negotiate and agree that the next close time is 12:00:01
+
+**Next consensus round**
+
+- In the next round validators project that the next close time should be 12:00:04
+- This is rounded down based on the close time resolution to 12:00:00
+- But again this is not greater than the last close time, so it's incremented to 12:00:02
+
+**Next consensus round after that**
+
+- Next validators determine that the next close time should be 12:00:05
+- Which is rounded up based on the close time resolution to 12:00:10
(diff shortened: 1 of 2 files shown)


thejohnfreeman:
Yes, that is how it works. Here is the code that adjusts the close time:

https://github.com/ripple/rippled/blob/8a25f3282426f382e1933242ef875d5aaf3e31e9/src/ripple/app/consensus/RCLConsensus.cpp#L465-L478

The first branch is when the close time is effectively "zero" signaling there was no consensus close time. In that case, it uses the previous close time plus one second, as you write.

The second branch adjusts when the close time is agreed in consensus. That calls this function:

https://github.com/ripple/rippled/blob/8a25f3282426f382e1933242ef875d5aaf3e31e9/src/ripple/consensus/LedgerTiming.h#L152-L165

Ignore the early return. It returns the maximum of a rounded close time and the previous close time plus one second. The rounding looks like this:

https://github.com/ripple/rippled/blob/8a25f3282426f382e1933242ef875d5aaf3e31e9/src/ripple/consensus/LedgerTiming.h#L129-L139

If the `closeResolution` is 10 seconds, then this effectively rounds to the nearest 10th second.

thejohnfreeman (approved):

mDuo13:
Per @nbougalis:

> The steps are a little mixed up:
> 1) Validators individually determines the proposed close time and include that time in their proposals. Let's assume 2 validators are proposing a close time of 12:00:03, and one is proposing a close time of 12:00:04.
> 2) Validators examine other's proposals to determine the consensus time. In this case, the resulting consensus time is 12:00:03.
> --- Voting process over. Everything else is deterministic. ---
> 3) The validators first "round down" this value to the nearest "interval" (a value also negotiated by consensus, currently 10 seconds) resulting a consensus time of 12:00:00. But since this value is identical to the previous ledger's close time, validators deterministically execute code that "adjusts" the close time to be precisely 1 second greater than the last ledger's close time.
> 4) The "adjusted" consensus time is 12:00:01.

Which in is the opposite order of how I thought it worked, but makes sense.

thejohnfreeman:
I don't think I would characterize it as "round down". It's a "round", the typical rounding where if it meets or crosses the half-interval it rounds up to the next interval. Right?

intelliot (approved):

wojake (approved):

mDuo13:
> I don't think I would characterize it as "round down". It's a "round", the typical rounding where if it meets or crosses the half-interval it rounds up to the next interval. Right?

Based on [this line of code](https://github.com/ripple/rippled/blob/8a25f3282426f382e1933242ef875d5aaf3e31e9/src/ripple/consensus/LedgerTiming.h#L138), I think it is actually a "round down" or "floor" operation using the mod (`%`) operator:

```
return closeTime - (closeTime.time_since_epoch() % closeResolution);
```

But I may be reading that wrong. The comment seems to suggest it rounds up too.

thejohnfreeman:
You have to take [these two lines](https://github.com/ripple/rippled/blob/8a25f3282426f382e1933242ef875d5aaf3e31e9/src/ripple/consensus/LedgerTiming.h#L137-L138) together, which is a conventional round. Rounding numbers in the half-open range [10.5, 11.5) to 11 is like adding 0.5 and flooring.

PR #1180, opened 2021-10-01
Update top banner to grants wave 2

mDuo13:
Apex is over, but wave 2 of the XRPL grants has just opened, so this updates the top banner to send people there.


intelliot (approved):
🎉

PR #1182, opened 2021-10-04
Add DestinationNode to the code sample

DennisDawson:
DestinationNode is listed in the parameter descriptions, but missing from sample code. I've added the entry, copying the OwnerNode value for now. I'm assuming that the value is a placeholder with the correct number of digits. If we can put in a more realistic value, that would probably be better.
This is to address issue https://github.com/XRPLF/xrpl-dev-portal/issues/892

### DIFF
--- a/content/references/protocol-reference/ledger-data/ledger-object-types/paychannel.md
+++ b/content/references/protocol-reference/ledger-data/ledger-object-types/paychannel.md
@@ -32,6 +32,7 @@ For an example of using payment channels, see the [Payment Channels Tutorial](us
     "CancelAfter": 536891313,
     "SourceTag": 0,
     "DestinationTag": 1002341,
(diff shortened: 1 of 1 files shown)


mDuo13 (approved):
Good point. I wonder if the original ticket missed that `DestinationNode` is in the parameter table because the fields are not really sorted properly. (They _should_ be organized with all required fields in alphabetical order, followed by all optional fields in alphabetical order.)

Incidentally, `"0000000000000000"` actually is a realistic value for a `DestinationNode` since that means that the channel is in the first "page" of the destination address's owner directory. (Each page of an owner directory tracks up to 32 objects owned by / attached to an account.)

P.S. the build failing is a [known issue](https://github.com/XRPLF/xrpl-dev-portal/issues/1176) with our CI.

PR #1194, opened 2021-10-12
Add details for SignerWeight and SignerQuorum

DennisDawson:
Per https://github.com/XRPLF/xrpl-dev-portal/issues/685, add some detail and description around the SignerWeight and SignerQuorum properties.

### DIFF
--- a/content/concepts/payment-system-basics/accounts/multi-signing.md
+++ b/content/concepts/payment-system-basics/accounts/multi-signing.md
@@ -17,7 +17,7 @@ Benefits of multi-signing include:
 * You can delegate the power to send transactions from your address to a group of people, who can control your address if you are unavailable or unable to sign normally.
 * ... and more.
 
-## Signer Lists
+## SignerLists
 
 Before you can multi-sign, you must create a list of which addresses can sign for you.
 
@@ -25,19 +25,31 @@ The [SignerListSet transaction][] defines which addresses can authorize transact
 
 ### SignerWeight
 
-You can assign a weight to each signer in the signer list. The weight represents the relative authority of the signer to other signers on the list. The higher the value, the more authorization authority. Individual SignerWeight values cannot exceed 2<sup>16</sup>-1.
+You can assign a weight to each signer in the SignerList. The weight represents the relative authority of the signer to other signers on the list. The higher the value, the more authorization authority. Individual SignerWeight values cannot exceed 2<sup>16</sup>-1.
 
 ### SignerQuorum
 
-The SignerQuorum value is the minimum SignerWeight total required to authorize a transaction. The SignerQuorum value must be attainable. The SignerQuorum must be greater than 0 but less than or equal to the sum of the SignerWeight values in the SignerList, or all transactions fail.  
+The SignerQuorum value is the minimum SignerWeight total required to authorize a transaction. The SignerQuorum value must be attainable. The SignerQuorum must be greater than 0 but less than or equal to the sum of the SignerWeight values in the SignerList.  
 
-For example, you might assign a SignerQuorum of 3 to to an account. You assign your CEO a SignerWeight of 3, Vice Presidents a SignerWeight of 2, and Directors  a SignerWeight of 1. To approve a transaction for this account requires the approval of 3 Directors (total weight of 3), 1 Vice President and 1 Director (total weight of 3), 2 Vice Presidents (total weight of 4), or the CEO (total weight of 3). The SignerWeight and SignerQuorum allow you to set an appropriate level of scrutiny for each transaction, based on the relative trust and authority relegated to responsible members of your organization. 
+### Examples Using SignerWeight and SignerQuorum
+
+For a typical use case, you might have a shared account with a SignerQuorum of 1, then give all participants a SignerWeight of 1. A single approval from any one of them is all that is required to approve a transaction.
+
+For a very important account, you might set the SignerQuorum to 3, with 3 participants that have a SignerWeight of 1. All of the participants must agree and approve each transaction.
+
+Another account might also have a SignerQuorum of 3. You assign your CEO a SignerWeight of 3, 3 Vice Presidents a SignerWeight of 2 each, and 3 Directors a SignerWeight of 1 each. To approve a transaction for this account requires the approval of all 3 Directors (total weight of 3), 1 Vice President and 1 Director (total weight of 3), 2 Vice Presidents (total weight of 4), or the CEO (total weight of 3).
+
+In each of the previous three use cases, you would disable the master key without configuring a regular key, so that multi-signing is the only way of authorizing transactions.
+
+There might be a scenario where you create a multi-signing list as a "backup plan." The account owner normally uses a regular key for their transactions (not a multi-signing key). For safety, the owner adds a signer list containing 3 friends, each with a weight of 1. Then the owner sets the SignerQuorum to 3. If the account owner were to lose the private key, they can ask their friends to multi-sign a transaction to replace the regular key.
+
+SignerWeight and SignerQuorum allow you to set an appropriate level of oversight for each transaction, based on the relative trust and authority relegated to responsible participants who manage the account. 
 
 ## Sending Multi-Signed Transactions
 
 To successfully submit a multi-signed transaction, you must do all of the following:
 
-* The address sending the transaction (specified in the `Account` field) must have a [signer list in the ledger](signerlist.html). For instructions on how to do this, see [Set Up Multi-Signing](set-up-multi-signing.html).
+* The address sending the transaction (specified in the `Account` field) must have a [SignerList in the ledger](signerlist.html). For instructions on how to do this, see [Set Up Multi-Signing](set-up-multi-signing.html).
 * The transaction must include the `SigningPubKey` field as an empty string.
 * The transaction must include a [`Signers` field](transaction-common-fields.html#signers-field) containing an array of signatures.
 * The signatures present in the `Signers` array must match signers defined in the SignerList.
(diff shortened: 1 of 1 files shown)


content/concepts/payment-system-basics/accounts/multi-signing.md
```diff
@@ -21,7 +21,17 @@ Benefits of multi-signing include:
 
 Before you can multi-sign, you must create a list of which addresses can sign for you.
```
mDuo13, line 22: So, this is a preexisting problem in the documentation, but I noticed that we're mixing and matching the terms "SignerList" and "signer list" in the following paragraphs (and probably elsewhere—sorry, that's my bad). 

As a convention, I've been trying to refer to [objects in the ledger state](https://xrpl.org/ledger-object-types.html) in title case matching the code name for the object—Escrow, Offer, etc.¹—so let's use introduce the term SignerList here and use it consistently throughout.

```suggestion
Before you can multi-sign, you must create a list of which addresses can sign for you. This list is stored in the ledger as a SignerList.
```


¹ RippleState is the exception because it's confusing. We usually talk about "trust lines", which are often described as being unilateral, but the storage format (RippleState) is a bilateral format (i.e. if you have a unilateral trust line you really have a bilateral RippleState with one half default/disabled; and if you have a matching pair of incoming/outgoing trust lines they're represented as a single shared RippleState object. Like I said, it's confusing.)

content/concepts/payment-system-basics/accounts/multi-signing.md
```diff
@@ -21,7 +21,17 @@ Benefits of multi-signing include:
...
+The [SignerListSet transaction][] defines which addresses can authorize transactions from your address. You can include 1 to 8 addresses in a SignerList. The SignerList cannot include the sender's address and there can be no duplicate entries. You can control how many signatures are needed, in which combinations, by using the *SignerWeight* and *SignerQuorum* values of the SignerList.
+
+### SignerWeight
+
+You can assign a weight to each signer in the signer list. The weight represents the relative authority of the signer to other signers on the list. The higher the value, the more authorization authority. Individual SignerWeight values cannot exceed 2<sup>16</sup>-1.
+
+### SignerQuorum
+
+The SignerQuorum value is the minimum SignerWeight total required to authorize a transaction. The SignerQuorum value must be attainable. The SignerQuorum must be greater than 0 but less than or equal to the sum of the SignerWeight values in the SignerList, or all transactions fail.
```
mDuo13: > or all transactions fail.

This sounds dramatic, and I think it's not what you mean here.

It's not possible to set a SignerQuorum that's higher than the sum of weights in a list. If you tried, the SignerListSet transaction to define such a quorum would fail. Since that's the case, I think the sentence is clearest if we just remove this bit.

```suggestion
The SignerQuorum value is the minimum SignerWeight total required to authorize a transaction. The SignerQuorum value must be attainable. The SignerQuorum must be greater than 0 but less than or equal to the sum of the SignerWeight values in the SignerList.  
```

content/concepts/payment-system-basics/accounts/multi-signing.md
```diff
@@ -21,7 +21,17 @@ Benefits of multi-signing include:
...
+
+You can assign a weight to each signer in the signer list. The weight represents the relative authority of the signer to other signers on the list. The higher the value, the more authorization authority. Individual SignerWeight values cannot exceed 2<sup>16</sup>-1.
+
+### SignerQuorum
+
+The SignerQuorum value is the minimum SignerWeight total required to authorize a transaction. The SignerQuorum value must be attainable. The SignerQuorum must be greater than 0 but less than or equal to the sum of the SignerWeight values in the SignerList, or all transactions fail.  
+
+For example, you might assign a SignerQuorum of 3 to to an account. You assign your CEO a SignerWeight of 3, Vice Presidents a SignerWeight of 2, and Directors  a SignerWeight of 1. To approve a transaction for this account requires the approval of 3 Directors (total weight of 3), 1 Vice President and 1 Director (total weight of 3), 2 Vice Presidents (total weight of 4), or the CEO (total weight of 3). The SignerWeight and SignerQuorum allow you to set an appropriate level of scrutiny for each transaction, based on the relative trust and authority relegated to responsible members of your organization.
```
mDuo13:
```
and Directors  a SignerWeight of 1.
```

Extra space after "Directors" and I think it would be good to clarify that you mean a weight of 1 _each_ for some number of directors.
```suggestion
For example, you might assign a SignerQuorum of 3 to to an account. You assign your CEO a SignerWeight of 3, Vice Presidents a SignerWeight of 2, and three Directors a SignerWeight of 1 each. To approve a transaction for this account requires the approval of 3 Directors (total weight of 3), 1 Vice President and 1 Director (total weight of 3), 2 Vice Presidents (total weight of 4), or the CEO (total weight of 3). The SignerWeight and SignerQuorum allow you to set an appropriate level of scrutiny for each transaction, based on the relative trust and authority relegated to responsible members of your organization. 
```

> The SignerWeight and SignerQuorum allow you to set an appropriate level of scrutiny for each transaction

I'm not sure what it means to set a level of scrutiny for each transaction. I guess maybe what you're getting at is that it lets you set how many people must look at the transaction before approving it?

mDuo13 (changes requested):
This is a good start. I left a couple specific suggestions for things to tweak, and I had some other more general comments:

- Can you look over the whole file and apply some more consistency around the use of "SignerList" vs "signer list"?
- The example with the CEO/directors is a good one, but I'd like to see some additional examples, to demonstrate the flexibility of the system so people don't get stuck with a single idea of how it should be used. Here are some more ideas:
    - A shared account where any one any of several people could approve any transaction? (say, quorum of 1 with 5 signers, each with a weight of 1)
    - An important account—for example, the issuer of an issued token—where _all_ the signers have to approve changes. (say, quorum of 3 with 3 signers each with a weight of 1)
    - A "backup plan" multi-signing list where the account owner normally uses their regular key (not multi-signing) to authorize transactions, but adds a signer list containing 3 friends (quorum of 3, 3 signers, each weight 1) so that if they ever lose their own private key they can ask their friends to multi-sign a transaction to replace the regular key.
- It may also be worth mentioning as a reminder in some of these examples that the master key should be disabled (and no regular key configured) so that multi-signing is the _only_ way of authorizing transactions on some of these accounts. Obviously that doesn't apply to the "backup plan" example.

content/concepts/payment-system-basics/accounts/multi-signing.md
DennisDawson: I've updated the text to address these suggestions.

content/concepts/payment-system-basics/accounts/multi-signing.md
DennisDawson: I've updated the doc to address these suggestions.

content/concepts/payment-system-basics/accounts/multi-signing.md
DennisDawson: I've updated the doc to address these suggestions.

content/concepts/payment-system-basics/accounts/multi-signing.md
```diff
@@ -17,17 +17,39 @@ Benefits of multi-signing include:
...
+
+### Examples Using SignerWeight and SignerQuorum
+
+For a typical use case, you might have a shared account with a SignerQuorum of 1, then give all participants a SignerWeight of 1. A single approval from any one of them is all that is required to approve a transaction.
+
+For a very important account, you might set the SignerQuorum to 3, with 3 participants that have a SignerWeight of 1. All of the participants must agree and approve each transaction.
+
+Another account might also have a SignerQuorum of 3. You assign your CEO a SignerWeight of 3, 3 Vice Presidents a SignerWeight of 2 each, and 3 Directors a SignerWeight of 1 each. To approve a transaction for this account requires the approval of all 3 Directors (total weight of 3), 1 Vice President and 1 Director (total weight of 3), 2 Vice Presidents (total weight of 4), or the CEO (total weight of 3).
+
+In each of the previous three use cases, you would disable the master key without configuring a regular key, so that multi-signing is the only way of authorizing transactions.
```
mDuo13: Here's an appropriate link for people who want more context:

```suggestion
In each of the previous three use cases, you would disable the master key without configuring a regular key, so that multi-signing is the only way of [authorizing transactions](transaction-basics.html#authorizing-transactions).
```

content/concepts/payment-system-basics/accounts/multi-signing.md
```diff
@@ -17,17 +17,39 @@ Benefits of multi-signing include:
...
+
+There might be a scenario where you create a multi-signing list as a "backup plan." The account owner normally uses a regular key for their transactions (not a multi-signing key). For safety, the owner adds a signer list containing 3 friends, each with a weight of 1. Then the owner sets the SignerQuorum to 3. If the account owner were to lose the private key, they can ask their friends to multi-sign a transaction to replace the regular key.
```
mDuo13: The bit about "then the owner sets" make it seem like it's a second step, but it happens at the same time, so I suggest rephrasing it as follows:

```suggestion
There might be a scenario where you create a multi-signing list as a "backup plan." The account owner normally uses a regular key for their transactions (not a multi-signing key). For safety, the owner adds a SignerList containing 3 friends, each with a weight of 1, and a SignerQuorum of 3. If the account owner were to lose the private key, they can ask their friends to multi-sign a transaction to replace the regular key.
```

mDuo13 (approved):
Looking good. I left a couple more minor suggestions but they're not urgent.

Also, I think I've fixed the issues with the CI lately, optionally, if you rebase this PR onto the latest master branch, we can test having it staged and having the link checker run automatically. Happy to help walk you through the commands to do that if you'd like.

PR #1204, opened 2021-10-18
Update freeze code examples

JST5000:
* Migrate all code to xrpl.js
* Use the styling from https://github.com/XRPLF/xrpl-dev-portal/blob/xrpljs2.0/content/_code-samples/require-destination-tags/require-destination-tags.js
* Split up check-global-freeze-no-freeze.js into two files
* Add pause + disable steps to set-global-freeze.js (And individual freeze)


content/_code-samples/freeze/check-global-freeze.js
```diff
@@ -0,0 +1,36 @@
+// Dependencies for Node.js.
+// In browsers, use <script> tags as in the example demo.html.
+if (typeof module !== "undefined") {
+    // gotta use var here because const/let are block-scoped to the if statement.
+    var xrpl = require('xrpl')
+  }
```
JST5000, line 6: @mDuo13 Is this conditional necessary for these examples? (I don't fully understand why it's there in the example code)

content/_code-samples/freeze/check-individual-freeze.js
```diff
@@ -1,37 +1,58 @@
...
+
+// Connect -------------------------------------------------------------------
+async function main() {
+  console.log("Connecting to Mainnet...")
+  const client = new xrpl.Client('wss://s1.ripple.com')
+  await client.connect()
+
+  client.on('error', (errorCode, errorMessage) => {
+    console.log(errorCode + ': ' + errorMessage)
+  })
```
JST5000, line 16: @mDuo13 This was in the previous version of this code, what's the purpose of this? (So I can add an explanatory comment or remove if it's no longer necessary)

content/_code-samples/freeze/set-global-freeze-with-checks.js
```diff
@@ -0,0 +1,73 @@
...
+  console.log("Requesting an address from the Testnet faucet...")
+  const { wallet, balance } = await client.fundWallet()
+
+  // Prepare an AccountSet transaction to enable global freeze ------------------
+  const accountSetTx = {
+    TransactionType: "AccountSet",
+    Account: wallet.address,
+    // Set a flag to turn on a global freeze on this account
+    SetFlag: xrpl.AccountSetAsfFlags.asfGlobalFreeze
+  }
```
JST5000, line 21: @mDuo13 Since this is a JS code example, should I add `validate(tx)`? (I think it's best practice for JS users to validate the shape of any transactions they make)

content/_code-samples/freeze/check-global-freeze.js
mDuo13: This conditional lets the file run unmodified in both Node.js and web browsers. (`module` and `require` are both not available in browsers normally).

content/_code-samples/freeze/check-individual-freeze.js
mDuo13: It's a general-purpose thing to write errors to the console to help with debugging. I think we can remove it.

content/_code-samples/freeze/set-global-freeze-with-checks.js
mDuo13: Good idea.

mDuo13 (approved):
In general, I think these look reasonable, and you identified some good improvements to make post-launch. For now, I'm going to merge to try and get these incorporated into the tutorials, with a plan of revisiting them after we publish the initial wave of xrpl.js 2.0 docs.

PR #1210, opened 2021-10-21
Update xrpljs2-migration-guide.md

intelliot:
- Add note that an exception from `submitAndWait()` does not guarantee that the transaction was not validated
- Small formatting improvements

### DIFF
--- a/content/references/xrpljs2-migration-guide.md
+++ b/content/references/xrpljs2-migration-guide.md
@@ -81,9 +81,9 @@ console.log(trustlines.result)
 
 ## Transaction Submission
 
-In xrpl.js there are specific helper functions for signing and submitting transactions and waiting for the XRP Ledger blockchain to confirm those transactions' final outcomes. Options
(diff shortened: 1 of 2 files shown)


content/references/xrpljs2-migration-guide.md
```diff
@@ -125,7 +125,7 @@ const prepared = await client.autofill({
 })
 ```
 
-Like the old prepare methods, `Client.autofill()` provides a reasonable `LastLedgerSequence` value by default. To prepare a transaction _without_ a `LastLedgerSequence` field, pass it with the value `null`:
+Like the old prepare methods, `Client.autofill()` provides a reasonable `LastLedgerSequence` value by default. To prepare a transaction _without_ a `LastLedgerSequence` field, set "LastLedgerSequence" to the value `null`:
```
mDuo13:
```suggestion
Like the old prepare methods, `Client.autofill()` provides a reasonable `LastLedgerSequence` value by default. To prepare a transaction _without_ a `LastLedgerSequence` field, provide a `LastLedgerSequence` with the value `null`:
```

mDuo13 (approved):
Good point, thanks.

PR #1213, opened 2021-10-21
Update freeze code samples to match xrpl.js release syntax

JST5000:
Also addressed comments from https://github.com/XRPLF/xrpl-dev-portal/pull/1204

### DIFF
--- a/content/_code-samples/freeze/check-global-freeze.js
+++ b/content/_code-samples/freeze/check-global-freeze.js
@@ -11,10 +11,6 @@ async function main() {
   const client = new xrpl.Client('wss://s1.ripple.com')
   await client.connect()
 
-  client.on('error', (errorCode, errorMessage) => {
-    console.log(errorCode + ': ' + errorMessage)
-  })
-
   const my_address = 'rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn';
 
   // Request account info for my_address to check account settings ------------
--- a/content/_code-samples/freeze/check-global-freeze.js
+++ b/content/_code-samples/freeze/check-global-freeze.js
@@ -3,6 +3,7 @@
 if (typeof module !== "undefined") {
   // gotta use var here because const/let are block-scoped to the if statement.
   var xrpl = require('xrpl')
+  var { AccountRootFlags } = require('xrpl/dist/npm/models/ledger')
 }
   
 // Connect -------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


content/_code-samples/freeze/check-no-freeze.js
```diff
@@ -1,33 +1,29 @@
...
+  const response = await client.request(
+    {command: 'account_info', account: my_address })
+  const settings = response.result
 
-    // Request account info for my_address to check account settings ------------
-    const response = await client.request(
-      {command: 'account_info', account: my_address })
-    const settings = response.result
+  console.log('Got settings for address', my_address);
+  console.log('No Freeze enabled?', (settings.noFreeze === true))
```
mDuo13: Sorry, I should've noticed earlier: this is broken. `noFreeze` was a RippleAPI-specific name. To make this work with xrpl.js 2.0, you need to parse the account flags and look for lsfNoFreeze. Like this:

```
  const flags = xrpl.parseAccountRootFlags(response.result.account_data.Flags)
  console.log('Got settings for address', my_address, flags)
  console.log("No Freeze enabled?", (flags.lsfNoFreeze === true))
```

content/_code-samples/freeze/check-global-freeze.js
```diff
@@ -1,36 +1,33 @@
...
+
+  // Request account info for my_address to check account settings ------------
+  const response = await client.request(
+    {command: 'account_info', account: my_address })
+  const settings = response.result
+
+  console.log('Got settings for address', my_address);
+  console.log('Global Freeze enabled?',
+              ((settings.account_data.Flags & AccountRootFlags.lsfGlobalFreeze) 
+              === AccountRootFlags.lsfGlobalFreeze))
```
JST5000: @mDuo13 I'm not sure if this will run when `module === 'undefined'` - Is there a different way to set this up so that it runs either way (Or does this work?)

JST5000: A simple, and less satisfying, method to avoid this is to just set a local variable to the flag value and add a comment for where we got it from (Ex. `const lsfGlobalFreeze = 0x00400000`)

mDuo13: You're right that won't run in browsers where `require` isn't available. For AccountRoot objects I think the easiest way is to use `xrpl.parseAccountRootFlags()`.

Like, here's code that'll run in browser (provided you have the appropriate `<script>` tag) or in Node.js:

```js
const account_info = await api.request({
    "command": "account_info",
    "account": address,
    "ledger_index": "validated"
})
console.log(account_info)
const flags = xrpl.parseAccountRootFlags(account_info.result.account_data.Flags)
console.log(`Got settings for address ${address}:
    ${pretty_print(flags)}`)
if (flags.lsfGlobalFreeze) {
  console.log("Global Freeze is enabled.")
} else {
  console.log("Global Freeze Tag is disabled.")
}
```

But I think you could also use `xrpl.LedgerEntry.AccountRootFlags` since [that's exported](https://js.xrpl.org/modules/LedgerEntry.html).

mDuo13 (changes requested):
Let's try to make the code run in browsers and Node.js if at all possible.

Otherwise it looks good (and I'd like to get this out ASAP since it fixes some publicly-visible code samples that won't work as intended right now)

mDuo13 (approved):

PR #1220, opened 2021-10-26
Clarify that you can generate keys in any normal way

### LINKED ISSUE #1191: Regular key docs - clarify that you can generate key in all the normal ways
Per @ledhed2222:

> In [Assign a Regular Key Pair](https://xrpl.org/assign-a-regular-key-pair.html), it has users using the `wallet_propose` method to create the regular key. I think instead it should explain that you can generate the key pair however you normally would, since most users will not have access to that method and, even if they do, it may seem like that is the ONLY way to create a regular key pair 

Some other ways to create key pairs that can be used to include regular keys:

Python using xrpl-py:

```py
keypair = xrpl.wallet.Wallet.create()
print("seed:", keypair.seed)
print("classic address:", keypair.classic_address)
```

JavaScript using xrpl.js:

```js
const keypair = new xrpl.Wallet()
console.log("seed:", keypair.seed)
console.log("classic address:", keypair.classicAddress)
```

Java using xrpl4j:

```java
WalletFactory walletFactory = DefaultWalletFactory.getInstance();
Wallet keypair = walletFactory.randomWallet(true).wallet();
System.out.println(keypair)
System.out.println(keypair.privateKey().get())
```

amarantha-k:
Fix #1191


content/tutorials/manage-account-settings/assign-a-regular-key-pair.md
```diff
@@ -22,9 +22,15 @@ This tutorial walks through the steps required to assign a regular key pair to y
 
 ## 1. Generate a Key Pair
 
+Generate a key pair that you'll assign to your account as a regular key pair.
+
+You can generate the key pair using one the following methods.
```
mDuo13: The phrase "one of the following methods" is a bit misleading here. Because the first such method has 3 tabs in the example, it's easy to misread those 3 options as being the "following methods" when really they represent just one option and not even the preferred one.

For this context, I think it would make sense to have all the options as tabs in the same block. The WebSocket/JSON-RPC/Commandline options have a response format, but I think it would be OK to remove that. Also, if you name the tabs just by language, then the code we have to persist the user's preferred programming language will work.

The one challenge is that the names of the fields, and available formats, are not the same in all methods. so it might be useful to call out on a language-by-language basis which fields you should care about. (Sadly, you can't put these things into the tabs as currently implemented.)

mDuo13 (changes requested):
You're on the right track, but I think this presentation makes it easy to get confused about which options are alternatives of each other.

mDuo13 (approved):

PR #1222, opened 2021-10-28
Code samples page

mDuo13:
Add a page to highlight some XRPL code samples that are available in the source code for this repo.

- [x] Add a filter to parse READMEs for each sample. The first header in each code sample is the title, and the first paragraph is the description that appears in the code sample card.
    - [x] Add and update READMEs for all code samples accordingly.
- [x] Move existing code samples so that they're always grouped into subfolders by language. The filter uses these subfolders to populate the icons in the card headers.
    - [x] Update references to the moved samples.

The styles and details of the new page still need some work. Known issues include:

- [ ] Fix unwanted borders/shading on the card header/footer, especially on light mode.
- [ ] Properly center the cards in mobile view.
- [ ] Choose final placement of the page in the hierarchy. For now it's directly under Docs.
- [ ] (Maybe) Change spacing of the top header so it looks more balanced alongside the usual Docs navigation sidebar.
- [ ] Add button and section with steps describing how to add a new code sample.


mDuo13:
Replaced by #1258.

PR #1224, opened 2021-11-03
Update xrpljs2-migration-guide.md

diadal:
fix xrpl.Wallet.fromSeed is not a constructor


diadal:
fix xrpl.Wallet.fromSeed is not a constructor

mDuo13 (approved):

PR #1226, opened 2021-11-06
Hindi translations 

TusharPardhe:
Added Hindi language translations


mDuo13:
This is impressive! Unfortunately I don't have the fonts or the language knowledge to understand Hindi, but I'll try to see if I can get this working.

amarantha-k (commented):
The translated text looks good.

mDuo13 (approved):
So, regrettably, I don't think we'll be able to put these translations to good use, for a couple reasons:

1. The `.pot` files haden't been updated since before the last major site update, so this PR is effectively a translation the old version of the site. (Well, some top-level pages and the template elements of the old site; the documentation contents are in their own files.)

    As a result, only a few parts of the site (words and phrases that stayed the same after the redesign) are updated, like this:

    

2. We don't have the resources at the moment to commit to keeping _any_ translations updated. Even just keeping track of everything that changes in the English version is quite a bit of work.

Unfortunately, I don't think that it is much benefit to offering a Hindi version of the site if barely any of the current version is translated. That said, there isn't a harm in merging these translations into the repo so that we have a starting point in the future if it becomes feasible to offer more translations again.

I updated the config and updated the strings so they're at least more accurate to the current version of the site, but I didn't put in place the necessary changes to actually deploy a Hindi version of the site.

PR #1228, opened 2021-11-07
Filter () and : in code tabs

### LINKED ISSUE #1141: wallet_propose example disappears when switching request format
In the wallet_propose documentation (https://xrpl.org/wallet_propose.html#code-0-1) there is a commandline example of how to call the function. If you click any of the 4 format types (Websocket, JSON-RPC with and without key type) the example disappears. Refreshing the page brings back the original example. The format for the four formats is also different than the same buttons on a similar function (Ex. https://xrpl.org/validation_create.html), so maybe this is caused by a legacy component being used on wallet_propose's documentation page?

Before:
Before

After: (There's no longer any example, and the text formats oddly)
After

Note: Clicking on any type other than "Commandline" does not bring up any examples, and clicking "Commandline" DOES bring up a different example than the one initially shown on the page:
Clicking Commandline

For reference: I am using Chrome Version 92.0.4515.131 (Official Build) (x86_64)

jakeatdocforce:
Fixes #1141

### DIFF
--- a/tool/filter_multicode_tabs.py
+++ b/tool/filter_multicode_tabs.py
@@ -67,7 +67,7 @@ def filter_soup(soup, mode="html", **kwargs):
                 label = "Code Sample %d-%d" % (index1, index2)
             linkback.string = label
             #add class with name of tab label
-            label_class = label.replace(" ","").replace("-","").replace(",","").lower()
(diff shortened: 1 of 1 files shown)


mDuo13 (approved):
Looks good on staging. Thanks!

PR #1235, opened 2021-11-12
Fix area for links on FAQ page

fersmi:
The active parts of the links are misplaced on the FAQ page. From a user perspective, this is confusing.


wojake (approved):

mDuo13:
As far as I can tell, the links are already in the right place in production and this PR breaks them (tested in both Firefox and Chrome on Linux).

Before:


After (note the expand/collapse chevron overlapping the actual answer text):


Can you please clarify what you were trying to fix with these changes? From there we can figure out how to fix it without breaking stuff that's currently functional.

fersmi:
Are you sure, that you are looking at the current state of the code?
Pull request fixed two things:
1) Interactive area of tag _a_ (see video). Tag _a_ is displayed as inline, but tag _h4_ is displayed as block. It's invalid insert tag _h4_ to _a_. Only _a_ tag in _h4_ is valid.
2) When sending a url with a id question fragment, the question remains unpacked. It isn't clear what question and answer is sent. See eg. https://www.xrpl.org/faq.html#what-purpose-does-xrp-serve

Before
https://user-images.githubusercontent.com/1818297/143059596-5f946134-77f9-425e-a15d-afcdfd0c11ec.mp4

After
https://user-images.githubusercontent.com/1818297/143059742-0a36b2d6-8ba3-4e80-b3cb-5b0e30f3541a.mp4

mDuo13 (approved):
I understand what you mean now. As for the problem with the carets being in the wrong place, it was because the updated compiled CSS needed to be committed to the repo. (I've done that now.) Now that it looks like it works properly, I 'm happy with this. :+1: 

>  It's invalid insert tag h4 to a. Only a tag in h4 is valid.

This is [no longer true](https://www.bennadel.com/blog/3693-anchor-tags-can-contain-block-level-elements-as-of-html5.htm) as of HTML5.

I think having a larger clickable area for the links is usually good because people don't have to be as precise to click it. It's a shame that the area between the text and the caret is not clickable with your version. However, I agree that the clickable area is _too big_ with the previous version, and yours is an improvement because the large `:before` offset should not be part of the link.

> When sending a url with a id question fragment, the question remains unpacked. It isn't clear what question and answer is sent. 

Excellent improvement, thank you.

PR #1236, opened 2021-11-14
Update get-acct-info.js

jphovila:
The call to initiate a new wallet (through fundWallet) must not include parameters.
The previous version of the code gives error: ReferenceError: Cannot access 'test_wallet' before initialization


wojake (approved):

mDuo13:
Actually, [you _can_ pass a Wallet instance as `fundWallet(wallet)`](https://js.xrpl.org/classes/Client.html#fundWallet) to have the library request that the faucet send the funds to that wallet's address. But, you're correct that the code as written fails with a ReferenceError, because it doesn't instantiate the Wallet instance first.

mDuo13 (approved):

PR #1242, opened 2021-11-18
Upgrade script no longer restarts the server. [v1.8.0]

mDuo13:
- Add a note about the new upgrade script not automatically restarting the server (not sure if this applies to to the upgrade _to_ 1.8.0 itself or only future upgrades)
- Removed `$` from most shell commands to facilitate easier copying
- Attempted to update the Japanese versions too. Pardon my non-fluency there (but I figured it's more important to say approximately the right thing awkwardly than to be fluent but outdated)


seelabs (approved):
:+1:

PR #1248, opened 2021-11-20
Amendment default votes and retirement

mDuo13:
- Adds mention of unconditionalized amendments ([XLS-11d](https://github.com/XRPLF/XRPL-Standards/discussions/19))
- Documents https://github.com/ripple/rippled/pull/3877

### DIFF
--- a/content/concepts/consensus-network/amendments/amendments.md
+++ b/content/concepts/consensus-network/amendments/amendments.md
@@ -225,9 +225,9 @@ TrustSetAuth
 ```
 
 
-## Retiring Amendments
+## Retiring Legacy Code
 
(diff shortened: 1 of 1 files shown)


content/concepts/consensus-network/amendments/amendments.md
```diff
@@ -223,6 +225,15 @@ TrustSetAuth
 ```
 
 
+## Retiring Amendments
```
JST5000: This title makes me think that we're talking about how to make amendments stop working, but it seems to be actually referring to "forgetting" code from before a certain amendment. Maybe something like "Retiring Old Behavior" would be more accurate, but it's also not that intuitive...

JST5000 (approved):
LGTM

PR #1251, opened 2021-11-23
Update Recommended Specs

alloynetworks:
Increase recommended production specifications to more accurately reflect the current needs

### DIFF
--- a/content/tutorials/manage-the-rippled-server/installation/system-requirements.md
+++ b/content/tutorials/manage-the-rippled-server/installation/system-requirements.md
@@ -16,7 +16,7 @@ A `rippled` server should run comfortably on commodity hardware, to make it inex
     - Development: Mac OS X, Windows (64-bit), or most Linux distributions
 - CPU: 64-bit x86_64, 2+ cores
 - Disk: Minimum 50 GB for the database partition. SSD strongly recommended (minimum 1000 IOPS, more is better)
-- RAM: 8 GB+
+- RAM: 16 GB+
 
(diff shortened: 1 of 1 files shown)


content/tutorials/manage-the-rippled-server/installation/system-requirements.md
```diff
@@ -25,12 +25,12 @@ Amazon EC2's `m3.large` VM size may be appropriate depending on your workload. A
...
 
-For best performance in enterprise production environments, Ripple recommends running `rippled` on bare metal with the following characteristics:
+For best performance in enterprise production environments, it is recommended to run `rippled` on bare metal with the following characteristics:
 
 - Operating System: Ubuntu (LTS) or CentOS or RedHat Enterprise Linux (latest release)
-- CPU: Intel Xeon 3+ GHz processor with 4 cores and hyperthreading enabled
-- Disk: SSD (10,000 IOPS or better)
-- RAM: 32 GB
+- CPU: Intel Xeon 3+ GHz processor with 8 cores and hyperthreading enabled
+- Disk: SSD / NVMe(10,000 IOPS or better)
```
mDuo13, line 32: At Ripple we've been thinking of bumping the IOPS number as well since 10,000 is pretty low for modern hardware. Can follow up with that.

mDuo13 (approved):
I've been meaning to do a slightly more thorough update. One thing that occurred to me would probably make a big difference is to move the "Recommended" specs first and put the minimum specs later (with a note that the minimum specs may not maintain sync to mainnet reliably through periods of high load)

But this is a good start and we can do the rest later.

PR #1254, opened 2021-11-29
Owner reserve reference patch

Silkjaer:
Updated references from the old owner reserve of 5 XRP to 2 XRP.


wojake (approved):

mDuo13 (approved):
Thanks! Thought I caught most of these but there were a bunch. Wonder if maybe we should set some variables at the target level to fill in all these types of mentions so we can update them in one place next time there's a change in reserves...

PR #1255, opened 2021-12-01
Update 'validators' method response

nixer89:
fix response examples -> "validators" method returns the validator pub key (nH...) and not the node pub key (n9...)

The given examples are the encoded "public_key" values of the UNLs from XRPLF and Ripple


wojake (approved):

amarantha-k (approved):
lgtm

PR #1259, opened 2021-12-04
Update public-servers.md

XRP-DEV:
ADDED sologenic as public websocket node


mDuo13:
Found a public statement from Sologenic confirming that this API is open to the public! 

> We created a large #XRPL node to handle #Sologenic transactions. This node can also be used by the public:
>
> wss://x1.sologenic.org
> Ledgers in history: 3,000,000
>
> We hope this helps relieve the pressure on other XRPL nodes. We'll launch more nodes in the future.

https://twitter.com/realSologenic/status/1466942499243507715

mDuo13 (approved):

PR #1262, opened 2021-12-07
Update AccountDelete transaction cost

N3TC4T:


intelliot (approved):
We may also consider saying the cost is 1 owner reserve and linking to a page that shows the current owner reserve. It may change again in the future

mDuo13 (approved):

wojake:
@intelliot What if we just create variables for these sort of stuff (e.g: account reserves, fees,etc...) and use them, this way, we wouldn't need to change all of them every time it changes. This probably has been suggested many times before.

PR #1263, opened 2021-12-08
Minor updates to install docs

mDuo13:
- Update mentions of Ubuntu 16.04 which is past end-of-life
- Fewer mentions of "rippled" name (so that eventually when it gets renamed to `xrpld` or whatever, the transition is easier)
- Remove $ from copy-pasteable commands
- Other small wording cleanup

### DIFF
--- a/content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
+++ b/content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
@@ -7,13 +7,9 @@ labels:
 ---
 # Build and Run rippled on Ubuntu
 
-`rippled` is the core peer-to-peer server that manages the XRP Ledger. A `rippled` server can connect to a network of peers, relay cryptographically signed transactions, and maintain a local copy of the complete shared global ledger.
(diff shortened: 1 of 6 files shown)


content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
```diff
@@ -38,7 +34,7 @@ These instructions use Ubuntu's APT (Advanced Packaging Tool) to install `ripple
 
 4. Install CMake.
 
-    Version 1.7.2 of `rippled` requires CMake 3.9.0 or higher. For the purposes of this tutorial, we used CMake 3.13.3.
+    Version 1.8.1 of `rippled` requires CMake 3.9.0 or higher. For the purposes of this tutorial, we used CMake 3.13.3.
```
amarantha-k, line 37: Could we use a variable instead of manually updating the version number?

amarantha-k (approved):
Just one minor suggestion. Otherwise LGTM

content/tutorials/manage-the-rippled-server/installation/build-run-rippled-ubuntu.md
mDuo13: We could, but it's not very helpful since we often need to update these compatibility statements with the release anyway, and it could be misleading if we update the variable without checking all the places it's used. Without a variable, it always provides accurate, if possibly out-of-date, information if we don't update it.

PR #1270, opened 2021-12-10
Add xrplcuster to Full History providers

alloynetworks:
Modify providers of Full History access and also specify that providers may restrict access.


wojake (changes requested):
Line 51,
> operated by community members

Can we change this to:
> operated by the XRP Ledger Foundation and the XRPL community members

mDuo13 (approved):

PR #1272, opened 2021-12-16
xrpl.js migration guide updates

mDuo13:
Addresses some feedback from https://github.com/XRPLF/xrpl.js/issues/1831.

- Link all the methods in the table of equivalents
- Add example of generating a new keypair
- Fix a couple cases where the table of equivalents didn't have the correct place for the equivalent (stuff that moved during dev)
- I believe `xrpl.deriveAddress()` [should be there](https://github.com/XRPLF/xrpl.js/issues/1871), but isn't exported properly, so I provided a workaround.


content/references/xrpljs2-migration-guide.md
```diff
@@ -227,36 +253,36 @@ In ripple-lib 1.x all methods and properties were on instances of the `RippleAPI
...
+| `getFee()` | (Removed - see Notes column) | Use [`Client.autofill()`](https://js.xrpl.org/classes/Client.html#autofill) to provide a sensible [transaction cost][] automatically, or use `Client.request({"command": "fee"})` to look up information about the current transaction cost (in _drops of XRP_). |
+| `getLedgerVersion()` | [`Client.getLedgerIndex()`](https://js.xrpl.org/classes/Client.html#getLedgerIndex) | |
+| `getTransaction()` | [`Client.request()`](https://js.xrpl.org/classes/Client.html#request) | Use [`Client.request()`](https://js.xrpl.org/classes/Client.html#request) to call the [tx method][] instead. **Warning:** Unlike `getTransaction()`, the `tx` method can return [results that are not validated and final](#validated-results). Be sure to look for `"validated": true` in the response object before taking action in response to a transaction. |
```
JST5000, line 271: Is `[tx method][]` supposed to have a link? (Context "to call the [tx method][] instead. **Warning:**..."

content/references/xrpljs2-migration-guide.md
```diff
@@ -227,36 +253,36 @@ In ripple-lib 1.x all methods and properties were on instances of the `RippleAPI
...
 | `getAccountObjects(address, options)` | (Removed - see Notes column) | Use [`Client.request()`](https://js.xrpl.org/classes/Client.html#request) to call the [account_objects method][] instead. **Warning:** Unlike `getAccountObjects()`, `account_objects` can return [results that are not validated and final](#validated-results). |
 | `getPaymentChannel()` | (Removed - see Notes column) | Use [`Client.request()`](https://js.xrpl.org/classes/Client.html#request) to call the [ledger_entry method](ledger_entry.html#get-paychannel-object) instead. **Warning:** Unlike `getPaymentChannel()`, `ledger_entry` can return [results that are not validated and final](#validated-results). |
 | `getLedger()` | (Removed - see Notes column) | Use [`Client.request()`](https://js.xrpl.org/classes/Client.html#request) to call the [ledger method][] exactly. **Warning:** Unlike `getLedger()`, `ledger` can return [ledgers that are not validated and final](#validated-results). |
 | `parseAccountFlags()` | `xrpl.parseAccountRootFlags()` | Now a static method on the module. |
```
JST5000:
```suggestion
| `parseAccountFlags()` | [`xrpl.parseAccountRootFlags()`](https://js.xrpl.org/modules.html#parseAccountRootFlags) | Now a static method on the module. |
```

JST5000 (approved):

content/references/xrpljs2-migration-guide.md
JST5000: Wasn't quite sure what the criteria for "worth a link" vs not worth a link was.

content/references/xrpljs2-migration-guide.md
mDuo13: This is one of the Markdown reference links. Standard markdown syntax allows for something like this:

```
The [tx method][] is a link.

[tx method]: https://xrpl.org/tx.html
```

Which produces this:

> The [tx method][] is a link.
>
> [tx method]: https://xrpl.org/tx.html

In our case, we have a single file that provides the link definitions for a bunch of things including all the API methods and gets included using an `{% include '_snippets/rippled-api-links.md' %}` macro at the end of the file. The macro is one of Dactyl's extensions on standard Markdown, so the GitHub preview doesn't show it correctly, but it works when you actually build the site.

P.S. There is a bug in the original Markdown, which python-markdown also has, where if the text of your reference link contains `monospace` syntax, the reference link doesn't match, so that's why most of the examples in this file can't by formatted using reference links.

PR #1274, opened 2021-12-17
Graphics for NFT topics

DennisDawson:


mDuo13:
combining this with #1275

PR #1275, opened 2021-12-17
Add NFT Tutorial

DennisDawson:


mDuo13:
I'd like to see some more clarifications on prerequisites and availability (e.g. NFTs are only on Devnet for now, and you need an updated client library to use some of the functions). And I think we can make some organizational improvements after the holidays, like putting the new ledger entry types and transaction types into the References.

And it looks like you fixed the image URLs, so the rest is good for now.

mDuo13 (approved):

PR #1296, opened 2022-01-14
Fix nftokenmint.md JSON example

SFsourDoh:
correct bad JSON formatting and add required field "TokenTaxon": 0


mDuo13 (approved):
Thanks for catching this!

PR #1302, opened 2022-01-22
"Build a Desktop Wallet in Python" tutorial

mDuo13:
Adds sample code and detailed instructions for building a basic desktop wallet application using xrpl-py and wxPython.

To-do list:

- [x] Refactor or remove `x_owned_objects.py` which was branched off a previous version of the code and demonstrates listing all the objects an account owns.
- [x] Test on other operating systems and add any changes needed to work on Mac and Windows.

PR preview: [Build a Desktop Wallet in Python (preview)](https://xrplf.github.io/xrpl-dev-portal/pr-preview/wallet_tutorial/build-a-desktop-wallet-in-python.html)

Explored, but won't do for now: Add mention of Xrplorer's Advisory List. While that's a good thing to add, the process for getting API keys to the advisory list is in flux right now so this is something to revisit when that's more stable.


wojake (changes requested):
build-a-desktop-wallet-in-python.md file, add a warning section:
This wallet isn't recommended to be used as a main wallet to hold all your currencies as it's insecure compared to other ready made wallets, **use at your own risk**.
(for some reason, i cant edit the md file)

ximinez:
> build-a-desktop-wallet-in-python.md file, add a warning section: This wallet isn't recommended to be used as a main wallet to hold all your currencies as it's insecure compared to other ready made wallets, **use at your own risk**. (for some reason, i cant edit the md file)

I haven't looked over this PR, but this comment caught my attention. I think it would be very valuable to have a brief explanation as to _how_ this wallet is less secure than other wallets. I also think it would be worth mentioning that a user can set a more secure regular key and disable the master key if they do want to keep using it.

mDuo13:
I don't think this wallet is inherently less secure than other wallets. It keeps the seed in regular memory, I suppose, whereas some other wallets might use OS-hardened memory to hold it? And there's no persistence—you have to actually save the seed somewhere else like a password manager, which is a reasonably good option for security. But the wallet described here never sends your keys out to any server, nor displays them, nor uses them to sign anything other than the transactions you send. Sure, it doesn't support regular keys or multi-signing—that would be a nice little upgrade—but in some other ways it's more secure than, say, a random wallet you downloaded from the internet where you don't know what's in the source code.

Basically, this wallet—since you're building it yourself—is as secure as you make it, and if other wallets are more secure we should lay out specifically how they are.

wojake:
I don't really build desktop based wallets but if I'm not mistaken, most software wallets have a certain algorithm that they use to cipher the key, this way malicious bots (similar to how antivirus's scan work) wouldn't be able to scan the raw private key when the computer is compromised, it'll look like random jargon to them unless they find the right algorithm and use it to decipher the key, which is very unlikely. I think @WietseWind knows more about this (XUMM).

>Why this wallet is insecure compared to other ready-made wallets (e.g: Exodus, XUMM)?

It doesn't have an algorithm to cipher/decipher the priv. key

mDuo13:
@wojake I think you may be thinking about how the wallets protect the key "at rest" when it's stored on disk and not in use. You always have to decipher the key to its original format in memory in order to sign transactions. In this case, the wallet itself doesn't store any secrets on disk at all; if you use a password manager to save your keys, then it's as secure as any password you have for those.

That said, I'm curious to know more about what types of additional precautions various wallets take to protect secret keys so that we can provide documentation or guidelines for how to implement a range of precautions that are appropriate for various use cases.

PR #1303, opened 2022-01-23
TokenIDs -> TokenOffers

ihomp:
I think it's actually called "TokenOffers"

### DIFF
--- a/content/references/protocol-reference/transactions/transaction-types/nftokencanceloffer.md
+++ b/content/references/protocol-reference/transactions/transaction-types/nftokencanceloffer.md
@@ -17,7 +17,9 @@ The `NFTokenCancelOffer` transaction can be used to cancel existing token offers
 {
   	"TransactionType": "NFTokenCancelOffer",
   	"Account": "ra5nK24KXen9AHvsdFTKHSANinZseWnPcX",
-  	"TokenIDs": "000100001E962F495F07A990F4ED55ACCFEEF365DBAA76B6A048C0A200000007"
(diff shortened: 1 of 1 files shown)


Gusti21111995 (approved):

mDuo13 (approved):
That is true, and also, the example needs to be updated as well—the internal type is VECTOR256 which means the JSON type should be an array of offer IDs, not just one.

mDuo13:
I believe this comes from a shortcoming in the original spec for this transaction, which is written as if an TokenID and NFTokenOffer object ID are the same thing; they are not. A TokenID identifies a non-fungible token; an NFTokenOffer object ID identifies an offer to trade such a token. This transaction cancels NFT offers, not NFTs themselves.

For now I provided a different, random ID since the IDs in `TokenOffers` contents are not the same as `TokenID`. It would be good to update the example with an actual NFTokenOffer object ID in the future.

For what it's worth, here's the part of the source that defines which fields those are: [TxFormats.cpp](https://github.com/ripple/rippled/blob/9f04c34e5e3c5040e4a4ade8eccdca972edb2412/src/ripple/protocol/impl/TxFormats.cpp#L308-L314). And here's the part that defines what the type & JSON field name are: [SField.cpp](https://github.com/ripple/rippled/blob/9f04c34e5e3c5040e4a4ade8eccdca972edb2412/src/ripple/protocol/impl/SField.cpp#L255).

scottschurr (approved):
:+1:  Looks right to me.

PR #1309, opened 2022-02-03
Update autobridging.md

wojake:
Add bridge currency use case + system -> function

XRP is the native currency of the XRPL but it's mainly used (in this example) as a bridge currency.
The auto bridging function is not a system in my opinion, it's not a network nor a software, it's a function on the XRPL.


content/concepts/decentralized-exchange/autobridging.md
```diff
@@ -8,9 +8,9 @@ labels:
...
+Any [Offer](offers.html) in the XRP Ledger's [decentralized exchange](decentralized-exchange.html) that would exchange two non-XRP currencies could potentially use [XRP](xrp.html) as an intermediary currency in a synthetic order book. This is because of _auto-bridging_, which serves to improve liquidity across all currency pairs by using XRP when doing so is cheaper than trading those currencies directly. This works because of XRP's nature as a native cryptocurrency to the XRP Ledger and XRP's bridge currency use case. Offer execution can use a combination of direct and auto-bridged offers to achieve the best total exchange rate.
```
mDuo13, line 11: I get that you're trying to clarify that one of XRP's purposes is as a bridge currency, but when you phrase it this way, you have the cause and effect backwards.

XRP is a better bridge currency _because_ it's possible to use in autobridging. It's possible to use XRP in autobridging because, as the native currency of the XRPL, everyone can hold it and send it to each other directly. Autobridging would not be possible with other tokens as the bridge because those tokens aren't native to the XRPL. The fact that XRP has a use as a bridge currency is a result of these properties, not a cause.

mDuo13 (changes requested):
I don't really see the case for this change. Was there some event in particular that inspired this?

content/concepts/decentralized-exchange/autobridging.md
wojake: Ohhhh... I see, alright, I'll close the PR.
Thank you for clarifying it in such detail.

PR #1314, opened 2022-02-10
Fix links (2022-02-09)

mDuo13:
- Add exceptions for some sites that block scripts like our link checker.
- Change video play links so they don't get detected as broken. (Fix related JS so it doesn't scroll you to the top when you click videos.)
- Fix some typos in the Japanese translations.
- Touch up a couple of NFT transaction reference pages where the inlined HTML was causing some Markdown not to get parsed, and a few links were using full URLs rather than relative links (so they were being treated as "external" even though they're part of the same site).


content/references/rippled-api/peer-port-methods/validator-list.md
```diff
@@ -29,7 +29,6 @@ request:
 
     The validator list contents are signed with a separate cryptographic key, so you can verify their integrity regardless of the TLS certificate used.
 
-**Tip:** Since this request uses the GET method, you can test this request using the URL bar of your web browser. For example, <https://s1.ripple.com:51235/vl/ED2677ABFFD1B33AC6FBC3062B71F1E8397C1505E1C42C64D11AD1B28FF73F4734> requests Ripple's recommended list from one of Ripple's public servers. <!-- SPELLING_IGNORE: s1 -->
```
mDuo13, line 32: This link no longer works now that s1 is reporting servers only. (Same for the Peer Protocol page.) The principle behind this callout is still true, but without an easy example that uses a non-self-signed certificate, it's a lot less useful of a note, so I elected to remove it.

PR #1317, opened 2022-02-15
fix: link to new consensus mechanism explainer

intelliot:
The old video is no longer available. The new video, hosted by the XRPLF, introduces the concept of the consensus mechanism.


amarantha-k (approved):
LGTM with the updated link

PR #1320, opened 2022-02-16
Support pages in subfolder

mDuo13:
Pages that exist in subfolders such as `/tutorials/` are now OK, as long as they have this line in the frontmatter:

```yaml
prefix: "/"
```

This ensures that nav elements including top nav, sidebar, breadcrumbs, etc. all work, and that the appropriate JavaScript loads as intended.


mDuo13:
Looks like the build failure is caused by the two commits running jobs concurrently, so it's only the CI, not the site, that's broken.

mDuo13:
That does remind me, subfolders don't work well with our CI setup though since there are more and different folders above the "base" of the site.

PR #1342, opened 2022-02-22
Contributing: document theme-aware diagram tech

mDuo13:
Documents the theme-aware diagrams tech we're using and provides some guidance on making diagrams that work well with it.


DennisDawson (changes requested):
This is written in an informal style, which might be appropriate for the audience. Keeping in mind international contributors, I would avoid contractions and informal terms like "docs." Not a hill I'll die on.

Line 209 - I'd start a new paragraph with "To include a theme-aware diagram in the documentation..." because the previous two sentences are not parallel with the task.

Line 215 - I'd replace "may" with "can." 
Line 221 - I'd replace "may" with "might."
Line 221 - typo at the end of the sentence ("digrams")
Line 224 - I don't use exclamation points in technical documentation. 
Line 225 - I'd avoid ending a sentence with a preposition - for example, "...use one of the colors mapped by the theme."

mDuo13:
Re: ending a sentence with a preposition. [Microsoft Style Guidance](https://docs.microsoft.com/en-us/style-guide/grammar/prepositions#placement-in-the-sentence) on this subject is that it's OK where it improves readability. In my opinion, "use one of the colors the theme has mappings for" is clearer than "use one of the colors mapped by the theme."

Made the other changes as suggested.

DennisDawson (approved):

PR #1288, opened 2022-01-10
Tutorial to run rippled in reporting mode

amarantha-k:

PR #1352, opened 2022-03-02
Reporting mode docs 

legleux:
I really didn't know how deep to go here. I usually like copy-pastable instructions like this but reporting mode should probably be attempted by slightly more knowledgeable rippled users so maybe that's not necessary.

The linux instructions procedure has an install CMake step but the macOS doesn't.
Then there's reference to `$HOME` one spot then it's `/home/ubuntu` elsewhere.
Minor stuff like inconsistent reference to the `rippled-example.cfg` filename or the location to even find it (`rippled-example.cfg` will never be in `/etc/opt/ripple` but `rippled.cfg` may be on a rippled installation.)

This note had me thinking bc this is only true if you're not running Cassandra on the same machine of course.

>     **Note:** If you choose to use Cassandra as the database, the disk requirements for `rippled` will be lower as the data will not be stored on your local disk.  

 
I could have done more but maybe it'd be more confusing.
Why even hint to someone this could (should) be run on Windows? 
Maybe the CMake line should be removed and just put in prerequisites like "a compatible version of CMake?"

### DIFF
--- a/content/tutorials/manage-the-rippled-server/installation/build-run-rippled-in-reporting-mode.md
+++ b/content/tutorials/manage-the-rippled-server/installation/build-run-rippled-in-reporting-mode.md
@@ -116,7 +116,6 @@ NuDB is installed as part of your `rippled` build setup and does not require any
 
     *macOS*
 
-        <!-- Ensure that a compatible version of CMake is installed -->
 	cmake -B build -G "Unix Makefiles" -Dreporting=ON -DCMAKE_BUILD_TYPE=Debug
         cmake --build build --parallel $(nproc)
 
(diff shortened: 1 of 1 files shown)


amarantha-k (commented):
Thanks for the review updates - they look good. 
- I'd added the "Install Postgres on Windows" when incorporating review feedback and would second taking it out if users should not be running on Windows platform.
-  Would appreciate your input on the following question from the original PR: https://github.com/XRPLF/xrpl-dev-portal/pull/1288#discussion_r785225877

legleux:
> * Would appreciate your input on the following question from the original PR: [#1288 (comment)](https://github.com/XRPLF/xrpl-dev-portal/pull/1288#discussion_r785225877)

I think the answer to that is "yes" but CJ would be the authority.

It looks like the \ on line 119 trashed the mac part

amarantha-k (approved):

PR #1354, opened 2022-03-03
NFTokenMint - clarify URI is hex

mDuo13:
The example as depicted returns an error like this one:

```
"error_message": "Field 'tx_json.URI' has invalid data.",
```

This is confusing since the doc says that the contents are not checked for validity. The actual cause is that the field needs to be hexadecimal (as with other arbitrary blob/text fields in XRPL transactions)

### DIFF
--- a/content/references/protocol-reference/transactions/transaction-types/nftokenmint.md
+++ b/content/references/protocol-reference/transactions/transaction-types/nftokenmint.md
@@ -27,7 +27,7 @@ If the transaction is successful, the newly minted token is owned by the account
   "TokenTaxon": 0,
   "Flags": 2147483659,
   "Fee": 10,
-  "URI": "ipfs://bafybeigdyrzt5sfp7udm7hu76uh7y26nf4dfuylqabf3oclgtqy55fbzdi",
+  "URI": "697066733a2f2f62616679626569676479727a74357366703775646d37687537367568377932366e6634646675796c71616266336f636c67747179353566627a6469",
   "Memos": [
         {
             "Memo": {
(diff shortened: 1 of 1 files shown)


content/references/protocol-reference/transactions/transaction-types/nftokenmint.md
```diff
@@ -27,7 +27,7 @@ If the transaction is successful, the newly minted token is owned by the account
   "TokenTaxon": 0,
   "Flags": 2147483659,
   "Fee": 10,
-  "URI": "ipfs://bafybeigdyrzt5sfp7udm7hu76uh7y26nf4dfuylqabf3oclgtqy55fbzdi",
+  "URI": "697066733a2f2f62616679626569676479727a74357366703775646d37687537367568377932366e6634646675796c71616266336f636c67747179353566627a6469",
```
mDuo13, line 30: This hex represents the same original string. Here's some example of how to encode/decode hex using Python for future reference:

```py
>>> uri = "ipfs://bafybeigdyrzt5sfp7udm7hu76uh7y26nf4dfuylqabf3oclgtqy55fbzdi"
>>> from binascii import hexlify
>>> hexlify(uri)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: a bytes-like object is required, not 'str'
>>> hexlify(uri.encode("ascii"))
b'697066733a2f2f62616679626569676479727a74357366703775646d37687537367568377932366e6634646675796c71616266336f636c67747179353566627a6469'
>>> from binascii import unhexlify
>>> unhexlify(b'697066733a2f2f62616679626569676479727a74357366703775646d37687537367568377932366e6634646675796c71616266336f636c67747179353566627a6469')
b'ipfs://bafybeigdyrzt5sfp7udm7hu76uh7y26nf4dfuylqabf3oclgtqy55fbzdi'
```

It's safe to encode URIs using ASCII because they are restricted to a subset of ASCII-valid characters. For [IRIs](https://www.w3.org/International/iri-edit/draft-duerst-iri.html) it's a different story but we can set that aside for now.

DennisDawson (approved):
This is fine for a quick fix.  Duplicate of https://ripplelabs.atlassian.net/browse/DOC-3769

PR #1355, opened 2022-03-04
Dd nft jira fixes 20220303

DennisDawson:
Addressing Jira issues RIPD-1796, DOC-3769, TOK-52


content/references/protocol-reference/transactions/transaction-types/accountset.md
```diff
@@ -71,6 +71,7 @@ The available AccountSet flags are:
 | Flag Name          | Decimal Value | Corresponding Ledger Flag | Description   |
 |:-------------------|:--------------|:--------------------------|:--------------|
 | `asfAccountTxnID`  | 5             | (None)                    | Track the ID of this account's most recent transaction. Required for [`AccountTxnID`](transaction-common-fields.html#accounttxnid) |
+| `asfAuthorizedMinter` :not_enabled:| 10         | (None)       | Enable to allow an account to mint tokens on behalf of another account. Specify the authorized account in the `minter` field of the [AccountRoot](accountroot.html) object. |
```
mDuo13: In addition to the `:not_enabled:` icon, there should be some link or mention here to explain what this experimental field is from / part of (the NFT preview)

mDuo13: Some other notes about this field:

- I think I've convinced the team to remove this flag from the implementation since it is an unnecessary complication. So the final version probably won't have this at all, but it's unknown how long until that change makes it to the NFT-Devnet.
- `Minter` is case-sensitive and needs to be capitalized
- The description could better contextualize who's authorizing whom. Specifically, "Allow another account to mint non-fungible tokens on this account's behalf." (Also: say NFTs rather than just "tokens" since this does not apply to fungible tokens.)

mDuo13 (changes requested):
The changes to AccountSet still need to add mention of the `Minter` field. (It should be part of the fields table on lines 32–42.)

content/references/protocol-reference/transactions/transaction-types/accountset.md
```diff
@@ -35,6 +35,7 @@ An AccountSet transaction modifies the properties of an [account in the XRP Ledg
 | [Domain](#domain) | String          | Blob              | _(Optional)_ The domain that owns this account, as a string of hex representing the ASCII for the domain in lowercase. [Cannot be more than 256 bytes in length.](https://github.com/ripple/rippled/blob/55dc7a252e08a0b02cd5aa39e9b4777af3eafe77/src/ripple/app/tx/impl/SetAccount.h#L34) |
 | `EmailHash`      | String           | Hash128           | _(Optional)_ Hash of an email address to be used for generating an avatar image. Conventionally, clients use [Gravatar](http://en.gravatar.com/site/implement/hash/) to display this image. |
 | `MessageKey`     | String           | Blob              | _(Optional)_ Public key for sending encrypted messages to this account. To set the key, it must be exactly 33 bytes, with the first byte indicating the key type: `0x02` or `0x03` for secp256k1 keys, `0xED` for Ed25519 keys. To remove the key, use an empty value. |
+| `Minter`         | String           | Blob              | _(Optional)_ Sets an alternate account that is allowed to mint NFTokens on this account's behalf using NFTokenMint's `Issuer` field. |
```
mDuo13:
```suggestion
| `Minter` :not_enabled: | String           | Blob              | _(Optional)_ Sets an alternate account that is allowed to mint NFTokens on this account's behalf using NFTokenMint's `Issuer` field. This field is part of the experimental XLS-20 standard for non-fungible tokens. |
```

mDuo13 (changes requested):
The `Minter` field definition should also be flagged as experimental. Otherwise looks good

mDuo13 (approved):
Fixes look good.

PR #1356, opened 2022-03-05
Events

jakeatdocforce:


SirJosh1987 (approved):
Can someone help me get unblocked from GitHub

mDuo13:
- The hover effect is visually confusing on the "Filter By" checkboxes under Upcoming Events and Past Events. When I click a box to uncheck it, it's almost indistinguishable from being actually checked. (After looking closely, I noticed that the opacity is slightly lower on the hover effect, but it was subtle and unintuitive.) It's also confusingly asymmetrical: hovering an unchecked box shows a "preview checked" effect, but hovering a checked box does not show a corresponding "preview unchecked" effect. IMO we should just remove the hover effect entirely.
- The labels "Meetups" and "Hackathons" are not functioning properly. Clicking the text should check/uncheck the box. Normally if you get the IDs and `for` attribute correct for html `<label>` elements, this should work automatically. Not sure what's wrong, but it's not behaving the way it should for me (confirmed in both Firefox 97 and Chromium 99 on Linux).
- The "New Year, New NFT" event doesn't have a link yet?
- The event cards are still dark in a light mode view. They're legible, but they don't match with the design for the light cards used elsewhere in the site. Also, after you click one, it enters the "active" state, which is illegible. (You have to switch back from the new tab it opens to see this though.) Screenshot of this:

    


Mobile layout looks good.

mDuo13:
Fixed version looks just right. Can you switch this to be not a draft PR, and resolve the merge conflicts? If we can do that without any regressions, I'm good to merge & deploy this.

mDuo13 (approved):
Revised version looks great! :tada:

PR #1361, opened 2022-03-08
fix NFTokenCancelOffer code snippets

khancode:
The [NFT documentation](https://xrpl.org/nftoken-tester-tutorial.html#the-canceloffer-function) and [NFToken Tester](https://github.com/XRPLF/xrpl-dev-portal/blob/master/content/_code-samples/nftoken-tester/js/nftoken-tester.html#L220) are incorrect when constructing the NFTokenCancelOffer transaction object. The correct property name should be TokenOffers rather than TokenIDs.

[xrpl.js Github issue](https://github.com/XRPLF/xrpl.js/issues/1936)

### DIFF
--- a/content/_code-samples/nftoken-tester/js/nftoken-tester.html
+++ b/content/_code-samples/nftoken-tester/js/nftoken-tester.html
@@ -211,7 +211,7 @@ async function cancelOffer() {
 	console.log("Connected to Sandbox")
 
 	const tokenOfferID = tokenOfferIndex.value
-  const tokenOffers = [tokenOfferID]
+	const tokenOffers = [tokenOfferID]
 
  // Prepare transaction -------------------------------------------------------
   const transactionBlob = {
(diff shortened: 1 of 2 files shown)


content/_code-samples/nftoken-tester/js/nftoken-tester.html
```diff
@@ -210,14 +210,14 @@
 	await client.connect()
 	console.log("Connected to Sandbox")
 
-	const tokenID = tokenOfferIndex.value
-	const tokenIDs = [tokenID]
+	const tokenOfferID = tokenOfferIndex.value
+  const tokenOffers = [tokenOfferID]
```
mDuo13: Indentation looks wrong here? I guess it's because the file originally uses tab characters rather than spaces.

(We've been following a style rule of always using spaces instead of tabs, but I guess that rule is not documented in the contributor guidelines. We should fix that too...)

DennisDawson (approved):
LGTM

PR #1363, opened 2022-03-09
Update look-up-transaction-results.md

silvertrojan:
Should it be 10^(-16) instead of 10^16?


mDuo13:
The original is correct as written. The "factor of" is a measurement of scale—rounding occurs when there is a _large_ difference between the two, not a small difference.

PR #1367, opened 2022-03-10
Added a CheckCreate Transaction Api Response

justTil:
Checks are in general rare to find on the ledger, so I created a CheckCreate Transaction on the Testnet, so other developers can see, how a CheckCreate would look like.

### DIFF
--- a/content/references/protocol-reference/transactions/transaction-types/checkcreate.md
+++ b/content/references/protocol-reference/transactions/transaction-types/checkcreate.md
@@ -27,6 +27,104 @@ Create a Check object in the ledger, which is a deferred payment that can be cas
 }
 ```
 
+## {{currentpage.name}} Testnet Transaction
+```json
+{
+  "id": 1,
+  "result": {
+    "Account": "rJ21pxdiBbJ7To89HwZUFZUEbQceQiwwb2",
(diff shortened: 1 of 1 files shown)


mDuo13:
While I agree that having an example of a processed transaction with metadata can be somewhat useful, I don't know about putting it there, since it buries some important parts like the parameter table further down the page, and we don't have examples like this for any of the other transaction types so the inconsistency might confuse people.

Maybe a better option would be to add a permalink to look up a CheckCreate transaction on a Mainnet full-history server, like [this one](https://xrpl.org/websocket-api-tool.html?server=wss%3A%2F%2Fxrplcluster.com%2F&req=%7B%22id%22%3A%22example_checkcreate%22%2C%22command%22%3A%22tx%22%2C%22transaction%22%3A%224E0AA11CBDD1760DE95B68DF2ABBE75C9698CEB548BEA9789053FCB3EBD444FB%22%2C%22binary%22%3Afalse%7D).

justTil:
You're right; it does bury some vital information. However, I like the idea of adding a link to a CheckCreate transaction in the Mainnet, which would also be helpful for other, if not all, transaction types.

wojake (approved):

mDuo13:
Opened #1421 for the preferred approach to solving this problem.

PR #1368, opened 2022-03-10
Contributor Guidelines: Clarify use of spaces for indentation

mDuo13:
We've been (largely) following this rule but it has not been formally documented until now.


intelliot (approved):

CONTRIBUTING.md
```diff
@@ -127,6 +127,7 @@ Use the following conventions when creating a page:
 - Don't use any formatting (like _italics_ or `code font`) in the title of the page
 - When in doubt, follow [Ciro Santilli's Markdown Style Guide (Writability Profile)](https://cirosantilli.com/markdown-style-guide/).
 - Landing pages should be in subfolders and should have the same filename as the folder. For example, the landing page of the "Accounts" page group should be `accounts/accounts.md` with the HTML filename `accounts.html`. **Don't** use `index.md`.
+- Don't use tab characters for indentation in Markdown or code samples. Use **4 spaces per indent**, except in **JavaScript** code samples, which should use 2 spaces per indent.
```
amarantha-k, line 130: Suggestion: For clarity and future reference, we should include the reason why we recommend not using tabs for indentation.

mDuo13: For posterity: tabs are discouraged because they can display as different widths depending on how you view them. (A tab is often displayed as 8, 4, or occasionally 2 columns' worth of white space. A space is always 1 column wide.)

PR #1370, opened 2022-03-16
Fixed Public Key Generation Description

justTil:
Hi there,
To use the [PaymentChannelCreate](https://xrpl.org/paymentchannelcreate.html) Method, you need to create a Public Key. However, I believe the description is now incorrect/not clear enough, so I changed it.

I used the PaymentChannelCreate method in the Testnet, with a Public Key that was 66 Characters long, derived from the sources' private key (secret), and hashed it with the SHA-512 algorithm. Then I took the first 66 Characters and used them as the public key, and the PaymentChannel was created. 

**Why 66 Characters?**
In the Documentation for [PaymentChannelCreate](https://xrpl.org/paymentchannelcreate.html) you are using this Public Key: _"32D2471DB72B27E3310F355BB33E339BF26F8392D5A93D3BC0FC3B566612DA0F0A"_, which is 66 characters long, and from what I believe is derived from your secret key. Shorter versions (64 characters -> half of 128) are not working.

The documentation may be improved with my description to clarify how to generate the Public Key/how it is generated.


mDuo13:
I believe this change is incorrect, and I would be surprised if you can actually redeem claims signed against this public channel.

The key needs to be 33 bytes of hexadecimal but it's _not_ a SHA-512 hash; you have to do elliptic curve cryptography to derive the public key from the private key. For secp256k1 keys you use the "compressed" format which is 33 bytes. For Ed25519 keys, the public key is 32 bytes but you prefix it with the byte `0xED` so that it's 33 bytes in total. See [Key derivation](https://xrpl.org/cryptographic-keys.html#key-derivation) for details.

mDuo13:
Opened #1373 to address the root cause behind this issue without introducing mistakes.

PR #1371, opened 2022-03-17
Hide top banner. Move events card to past

jakeatdocforce:
Hide top events banner. 
Comment styles.
Renamed new style devportal2022-v2
Deleted old unneeded styles
Moved upcoming event to past event group


jakeatdocforce:
New change to update banner instead of hide.

PR #1372, opened 2022-03-17
Update info in top banner

jakeatdocforce:
Update info to xrplgrants.org
Renamed new style devportal2022-v2
Deleted old unneeded styles
Moved upcoming event to past event group


mDuo13 (approved):
LGTM. Though, do you think we should display past events with the most recent first? Not sure how many the plan is to keep at a time, but if we keep putting recent ones at the bottom, it'll involve more and more scrolling to see what are probably the most relevant ones?

PR #1373, opened 2022-03-17
Clarify payment channel key derivation

mDuo13:
This is an attempt to clarify the confusion from #1370.


justTil:
What do you exactly mean by "This can be any secp256k1 or Ed25519 public key". Any key that is derived from the Accounts' secret?

wojake:
@justTil: The XRPL uses 2 cryptographic algorithms as of now, `secp256k1` &  `Ed25519`.

When you're generating a private key using the `wallet_propose` method, you can specify which algorithm you're intending on using:
```
{
    "command": "wallet_propose",
    "seed": "snoPBrXtMeMyMHUVTgbuqAfg1SUTb",
    "key_type": "secp256k1" <-
}
```

If you want to learn more about how key pairs are generated on a high-level structure, see here:
1. https://xrpl.org/cryptographic-keys.html#secp256k1-key-derivation
2. https://xrpl.org/cryptographic-keys.html#ed25519-key-derivation

amarantha-k (approved):

justTil:
@mDuo13 & @mDuo13 you both ofc. were right with the key derivation and I looked into it and created the key pairs correctly, and it worked. So thank you for the help, and maybe if the change you made help people in the future to grasp it even faster.

Thanks again :D

PR #1344, opened 2022-02-23
Rename "Issued Currencies" to Tokens

mDuo13:
Replaces #1318. (Moved to XRPLF repo so that we can have a staged preview.)

Changes the wording around "issued currencies" to "tokens" throughout the docs, with various related touch-ups and rewording. Moves some pages around in ways I hope will also improve SEO.

Fixes #984, #1034, and #983.

PR #1374, opened 2022-03-17
Data Types: issued currency→token

mDuo13:
Updates a couple mentions of the words "issued currency" that didn't get changed in #1344.


wojake:
Are we going to call them IOUs, issued currencies, or tokens? They're all the same, I don't get why we're calling them using different terms.

amarantha-k (approved):
lgtm

amarantha-k (commented):
Link checker shows broken links; PTAL and resolve before merging.

mDuo13:
> Are we going to call them IOUs, issued currencies, or tokens? They're all the same, I don't get why we're calling them using different terms.

We are calling them "Tokens". The point of this PR is to move away from some of the previous usage.

PR #1381, opened 2022-03-28
Remove duplicate definition for the term sidechain

amarantha-k:


mvadari (approved):

content/concepts/consensus-network/federated-sidechains.md
```diff
@@ -35,14 +35,12 @@ Transactions within the sidechain are not visible to the servers on the mainchai
...
  
 _Federator_: A server on the sidechain that listens for triggering transactions on both the main chain and the side chain. Each federator has a signing key associated with it that is used to sign transactions. A transaction must be signed by a quorum of federators before it can be submitted. Federators are responsible for creating and signing valid response transactions, collecting signatures from other federators, and submitting transactions to the main chain and side chain.
 
 _Mainchain_: The blockchain where assets originate and where assets are held while being used on the side chain. For most sidechains, the mainchain is the XRP Ledger Mainnet, Testnet, or Devnet.
```
mDuo13:
```suggestion
_Mainchain_: The blockchain where assets originate and where assets are held while being used on the sidechain. For most sidechains, the mainchain is the XRP Ledger Mainnet, Testnet, or Devnet.
```

content/concepts/consensus-network/federated-sidechains.md
```diff
@@ -35,14 +35,12 @@ Transactions within the sidechain are not visible to the servers on the mainchai
...
+_Sidechain_: An XRP Ledger sidechain is another blockchain based on XRP Ledger technology. A _federated_ sidechain provides a way to transfer assets from a mainchain to the sidechain and back. A sidechain can use an exact copy of the XRP Ledger's protocol or it can make changes, including to the [consensus algorithm](consensus.html), [transaction types](transaction-types.html), and other rules. Sidechains have separate history, rules, and validators than the mainchain. Proxy assets are issued in the sidechain, with the equivalent assets held by a door account on the mainchain. Proxy assets on the sidechain can be sent back to the mainchain and unlocked from the control of the federators.
  
 _Federator_: A server on the sidechain that listens for triggering transactions on both the main chain and the side chain. Each federator has a signing key associated with it that is used to sign transactions. A transaction must be signed by a quorum of federators before it can be submitted. Federators are responsible for creating and signing valid response transactions, collecting signatures from other federators, and submitting transactions to the main chain and side chain.
```
mDuo13:
```suggestion
_Federator_: A server on the sidechain that listens for triggering transactions on both the mainchain and the sidechain. Each federator has a signing key associated with it that is used to sign transactions. A transaction must be signed by a quorum of federators before it can be submitted. Federators are responsible for creating and signing valid response transactions, collecting signatures from other federators, and submitting transactions to the mainchain and sidechain.
```

mDuo13 (changes requested):
Lists should always have some sort of order to them; I suggest putting these terms in alphabetical order. If it were possible to introduce simpler concepts and build up to the other ones, that would define a good order to introduce terms in. Unfortunately, all the terms here are defined in terms of the other terms here, so there's no clear learning order. Maybe they could be reworded more to clarify that?

At least having them in alphabetical order would make it easier to find the definition you're looking for and would have made it obvious that there was a duplicate entry.

Also, I noticed a few cases where treatment of "mainchain" and "sidechain" (with or without a space was inconsistent and tagged those.

wojake (approved):
LGTM

amarantha-k:
@mDuo13 The terms are now in alphabetical order. Looks like you need re-review and approve before this PR can be merged.

wojake (changes requested):
Why must it be in an alphabetical order? I think the list should be in a relevance order. The easier/basic concepts should be on top of the list and so on...

mDuo13 (approved):

PR #1379, opened 2022-03-25
Document NFT-related API methods

DennisDawson:
TOK-57: Add Docs for NFT Methods (https://ripplelabs.atlassian.net/browse/TOK-57) account_nfts, nft_buy_offers, nft_sell_offers

PR #1382, opened 2022-03-28
WebSocket Tool: support NFT methods

mDuo13:
Opening this PR mostly to showcase the staging preview

- Add example syntax for the 3 NFT methods.
- Add the ability to specify the network without the rest of a permalink. NFT method "try it!" links can use this to have the tool use the NFT-Devnet by default.
- Add the ability to mark methods as not enabled on Mainnet.


mDuo13:
Changing this to be based on #1379 so it can be merged in

PR #1383, opened 2022-03-29
Add new events 

JakeatRipple:
Updated upcoming events list. 
Sorted by date.


template/component-top-nav.html.jinja
```diff
@@ -1,9 +1,9 @@
-<div class="top-banner fixed-top">
+<!-- <div class="top-banner fixed-top">
```
mDuo13: I suggest we put these in a Jinja comment so the comment is just empty in the source code for the final site, instead of leaving a full commented-out block.

```suggestion
<!--{# <div class="top-banner fixed-top">
```

template/component-top-nav.html.jinja
```diff
@@ -1,9 +1,9 @@
...
   <div class="d-flex justify-content-center ">
     <span><p class="mb-0">{% trans %}Wave 3 of the XRPL Grants program is open!{% endtrans %}</p></span>
     <span><a href="https://xrplgrants.org/?utm_medium=referral&utm_source=xrpl&utm_campaign=banner" class=" btn btn-outline-secondary" target="_blank">Apply</a></span>
   </div>
-</div>
+</div> -->
```
mDuo13: Per the above suggestion of putting this in a Jinja comment

```suggestion
</div> #}-->
```

mDuo13 (approved):
Looks OK, but two suggestions. One below, the other I can't comment on the line but I suggest you change line 262 as follows:

```jinja
{% for event in past|reverse %}
```

That way past events display with most recent first.

mDuo13 (approved):

PR #1385, opened 2022-04-05
Add missing `Owner` field to NFTokenBurn page

JST5000:
When burning a token which has `lsfBurnable` enabled as the original minter, the `Owner` field is required to specify which NFTokenPages to search for the NFTokenID. 

Here's the original issue which prompted this discovery: https://github.com/XRPLF/xrpl.js/issues/1947


DennisDawson (approved):
This change looks good to me.

PR #1387, opened 2022-04-11
Fix broken example on NFTokenMint + explain Issuer

JST5000:
In response to this thread on Discord:
* Initial problem: https://discord.com/channels/886050993802985492/890685878068969472/962675085557235742
* Response: https://discord.com/channels/886050993802985492/890685878068969472/963111448995831838

This attempts to:
1. Simplify the initial example to just work without any setup (Fixing issues which make copy + pasting simply not work)
2. Add a section on minting with another issuer to handle that specific case, with a more complicated example that also simply works to copy + paste. (And putting the "Set the MinterAccount" instruction at the top rather than as a foot note.


DennisDawson (approved):
I made some style edits, appreciate the help!

PR #1395, opened 2022-04-20
Quickstart: Use buttons for nav

mDuo13:
I noticed that the quickstart has these links at the bottom for navigating to the next or previous steps but it doesn't look that nice in the way it ends up getting formatted. I was able to make a quick improvement by taking advantage of the buttonize filter to make them stand out more. I also added ← and → arrows to emphasize whether the links are going forward or back.

The end result is something like this:


DennisDawson (approved):
These changes look good. I'm not certain that we'll want to keep the buttons in the long run. They're less important in the docs, since we have the menu right there in numerical order. However, as we add more modules, there might be more than one option for "Next." (For example, _Next_ might go to NFTokens or to digital currency management.) Let's be flexible as we build out.

PR #1397, opened 2022-04-20
Conceptual doc for NFT transfers

DennisDawson:


content/concepts/tokens/non-fungible-token-transfers.md
```diff
@@ -0,0 +1,82 @@
...
+ - include_code
+labels:
+ - Non-fungible Tokens, NFTs
+status: not_enabled
+---
+
+# Trading NFTokens on the XRP Ledger
+{% include '_snippets/nfts-disclaimer.md' %}
+
+You can transfer NFTokens between accounts on the XRP Ledger. You can transfer `NFTokens` by offering to buy or sell `NFTokens`, or accepting offers from other accounts. You can even give away `NFTokens` by offering to sell them at a price of 0.  All offers are created using the `NFTokenCreateOffer` transaction.
```
mDuo13: I suggest linking the reference page for the transaction type. If you include the `tx-type-links.md` snippet somewhere in this file, then you can even use a predefined reference link for it, like so:

```suggestion
You can transfer NFTokens between accounts on the XRP Ledger. You can transfer `NFTokens` by offering to buy or sell `NFTokens`, or accepting offers from other accounts. You can even give away `NFTokens` by offering to sell them at a price of 0.  All offers are created using [NFTokenCreateOffer transactions][].
```

Enabling that and other shortcuts is why I have this block at the end of a lot files:

```
<!--{# common link defs #}-->
{% include '_snippets/rippled-api-links.md' %}			
{% include '_snippets/tx-type-links.md' %}			
{% include '_snippets/rippled_versions.md' %}
```

content/concepts/tokens/non-fungible-token-transfers.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+## Sell Offers
+
+
+### Create a Sell Offer
+
+As the owner of a `NFToken object`, you can create a sell offer using a `NFTokenCreateOffer` transaction with a _Flags_ setting of _1_. You provide the _NFTokenID_ and the _Amount_ you are willing to accept in payment. You can optionally specify an _Expiration_ date, after which the offer is no longer valid, and a _Destination_ account, which is the only account that is allowed to purchase the `NFToken`.
```
mDuo13: Stylistically, the use of monospace and italics here is at odds with the rest of our docs. Monospace, not italics, should be used when referring to verbatim fields from the protocol or API, but should not encompass prose words like "object". 

An exception is transaction types, which are typically presented as links to the transaction reference, but can also be written in plain text. (This is partly because there is a specific glitch with Markdown reference links that contain monospace.)

Also, as a best practice, you should refer to flags by their names rather than their values.

```suggestion
As the owner of a `NFToken` object, you can create a sell offer using a [NFTokenCreateOffer transaction][] with the `tfSellToken` flag. You provide the `NFTokenID` and the `Amount` you are willing to accept in payment. You can optionally specify an `Expiration` date, after which the offer is no longer valid, and a `Destination` account, which is the only account that is allowed to purchase the `NFToken`.
```

content/concepts/tokens/non-fungible-token-transfers.md
```diff
@@ -0,0 +1,82 @@
...
+
+To purchase a `NFToken` that is offered for sale, you use a `NFTokenAcceptOffer` transaction. You provide the owner account and specify the `nft_offer_index` of the `NFTokenOffer` object you choose to accept.
+
+
+## Buy Offers
+
+
+### Create a Buy Offer
+
+Any account can offer to buy a `NFToken` on the XRP Ledger. You can create a buy offer using `NFTokenCreateOffer` with a _Flags_ setting of _null_. You provide the _Owner_ account, _NFTokenID_, and the _Amount_ of your offer.
```
mDuo13: It's not _quite_ accurate to say "a _Flags_ setting of _null_".

- `null` is a specific data type in JSON, which is not allowed for the Flags field in the `rippled` APIs. You can omit the field when constructing a transaction via JSON, which will cause it to be provided with a default value (possibly `0` though that depends on the interface).
- It is possible to specify other flags in the `Flags` field including global flags (currently the only one is tfFullyCanonicalSig, but this could change in the future), and still have the NFTokenCreateOffer treated as a buy.

I would say the most accurate way to phrase it would be either "with no NFTokenCreate Offer flags" or "without the `tfSellToken` flag".

Also, my notes on monospace font also apply here. And you can probably omit "on the XRP Ledger" since that's understood from context.

```suggestion
Any account can offer to buy a `NFToken`. You can create a buy offer using [NFTokenCreateOffer][] _without_ the `tfSellToken` flag. You provide the `Owner` account, `NFTokenID`, and the `Amount` of your offer.
```

content/concepts/tokens/non-fungible-token-transfers.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+### Accept a Buy Offer
+
+Use the `NFTokenAcceptOffer` transaction to transfer a `NFToken`. Provide the `nft_offer_index` and the owner account to complete the transaction.
```
mDuo13: It's unclear from context what the `nft_offer_index` is or where you get one. Is that a specific field from some API?

content/concepts/tokens/non-fungible-token-transfers.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+## Trading Modes
+
+When trading `NFTokens`, you can choose between a _direct_ transaction between a buyer and seller or a _brokered_ transaction, where a third party account matches sell and buy offers to arrange the trade.
```
mDuo13: I'm uneasy about `NFTokens` with the "s" in the monospace font since that's not a verbatim part of the object name.

Writing it with the s outside of the monospace block, e.g. `NFToken`s, might be even worse. So, I don't have a specific recommendation or action item here, but we may want to brainstorm if there is a better way of handling this, or if "eh, it's fine" can be our official policy on this.

content/concepts/tokens/non-fungible-token-transfers.md
```diff
@@ -0,0 +1,82 @@
...
+
+Trading in direct mode gives the seller control over the transfer. The seller can either post `NFTokens` for purchase by anyone, or sell NFTokens to a specific destination account. The seller receives the entire purchase price for the `NFToken`.
```
mDuo13: There's a non-monospace "NFTokens" in the middle sentence. That should probably be changed for consistency, though per my above comment I'm not sure which way to be consistent is better.

content/concepts/tokens/non-fungible-token-transfers.md
```diff
@@ -0,0 +1,82 @@
...
+
+If a `NFToken` creator has the time and patience to seek out the right buyers, the creator keeps all proceeds from the sale. This works fine for a creator who sells few `NFTokens` at variable prices.
+
+On the other hand, creators might not want to spend their time selling their creations when they could spend the time creating. Instead of handling each individual sale, the sales work can be turned over to a third-party broker account.
+
+Using a broker offers several advantages. For example:
+
+
+
+* The broker can act as an agent, working to maximize the selling price of the `NFToken`. IF the broker is paid a percentage of the sale price, the higher the price, the more the broker can earn.
```
mDuo13: Not sure why there are 3 blank lines between the intro paragraph and the list here. Probably 1 line would be better, so they don't seem intentionally separated.

I think the "F" in "IF" being uppercase is probably a typo?

And finally, I think "can earn" is unnecessary given that this is already a hypothetical due to the opening "if". 

```suggestion
Using a broker offers several advantages. For example:

* The broker can act as an agent, working to maximize the selling price of the `NFToken`. If the broker is paid a percentage of the sale price, the higher the price, the more the broker earns.
```

content/concepts/tokens/non-fungible-token-transfers.md
```diff
@@ -0,0 +1,82 @@
...
+* The broker can act as a curator, organizing `NFTokens` based on a niche market, price point, or other criteria. This can attract groups of buyers who might not otherwise discover a creator’s work.
+* The broker can act as a marketplace, such as Opensea.io, to handle the auction process at the application layer.
```
mDuo13: Not sure it's wise to mention a specific marketplace (Opensea.io) here, especially when they aren't using XRPL. Consider removing that part.

content/concepts/tokens/non-fungible-token-transfers.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+### Brokered Sale Workflows
+
+In the most straightforward workflow, a creator mints a new `NFToken`. The creator initiates a sell offer, entering the minimum acceptable sale price and setting the broker as the destination. Potential buyers make bids for the `NFToken`, setting the broker as the destination for the bid. The broker selects a winning bid and completes the transaction, taking a broker’s fee. The broker then cancels any remaining buy offers for the `NFToken`.
```
mDuo13: I don't completely understand the chain of custody from this explanation.

When the broker "completes sale using NFTokenAcceptOffer" does the broker actually become the owner of the NFToken at any point? Or is setting the broker as the destination just a technical detail that allows them to use the NFTokenAcceptOffer to direct the ownership of the token to the right bidder?

mDuo13: Also, can the bidders cancel their own offers if the broker doesn't do so? (I would imagine that they have to be able to—otherwise lazy/malicious brokers could leave people with dangling objects consuming their reserve, yeah?)

mDuo13 (approved):
Good explanation of some important stuff, though I did have some questions about brokered mode (of course). I suggested some formatting changes which are up to you, and I think I spotted a typo or two. Assuming those are addressed this is a :+1: from me.

content/concepts/tokens/non-fungible-token-transfers.md
DennisDawson: The mention of Opensea.io came from Jazzi, so I'm okay with it.

content/concepts/tokens/non-fungible-token-transfers.md
DennisDawson: The broker never takes possession of the NFToken. They find a buyer and complete the transaction, transferring the NFToken from the seller to the buyer. The funds go directly from the buyer to the seller, less the broker fee, which goes directly from the buyer to the broker at the time of the sale. 

Bidders can cancel their own buy offers (and should), but best practices are for the brokers to cancel all of the losing offers for the reasons you mention.

content/concepts/tokens/non-fungible-token-transfers.md
DennisDawson: For whatever reason, in working with the call I found that omitting the flags field or setting it to 0 caused an error. Only setting the flag to null worked (and it continues to work in the code sample). We can come back to this at some future point.

content/concepts/tokens/non-fungible-token-transfers.md
DennisDawson: `nft_offer_index` is the revised name of the `index` field. It's the NFTokenOfferID. That needs clarification.

content/concepts/tokens/non-fungible-token-transfers.md
DennisDawson: Not sure I agree with the style, but I'll abide by it so that this document can go out. We can circle back on this.

content/concepts/tokens/non-fungible-token-transfers.md
DennisDawson: Reworded to avoid the use of NFTokens.

PR #1399, opened 2022-04-24
Update parallel-networks.md

wojake:
Change Hooks link from V1 to V2 *"betanet"*.

Different network, newer and up-to-date features, the V1 testnet would still be up for some time tho.


amarantha-k (approved):
Thanks for adding the v2 Testnet link.

wojake:
They're calling the v2 testnet, "betanet". Should we use that instead?

intelliot:
> They're calling the v2 testnet, "betanet". Should we use that instead?

Open to it, but where do you see that? https://hooks-testnet-v2.xrpl-labs.com/ says `Testnet`

wojake:
Please do not push this PR yet, there are a few features that they want to add in and the doc isn't quite finished. Let's push this PR when they officially announce it on an open medium. My bad for opening up a PR this early.

mDuo13:
Converted to draft to accurately reflect the status (so we don't merge these too early) until upstream is ready with the new Hooks docs & whatnot

mDuo13 (approved):

PR #1401, opened 2022-05-05
Update send-a-conditionally-held-escrow.md

ObiajuluM:
Added python example for generating escrow condition and fulfillment


mDuo13 (approved):
I made some changes:

- Fixed some typos
- Put the two languages in tabs
- Made the Python version use a cryptographically-random preimage rather than a hardcoded preimage (matches the JS code)

PR #1404, opened 2022-05-16
Add missing word to ripplestate doc

JST5000:
I'm not certain if "given" or "issued" is a better word for this, but I think "A negative balance that the low account has tokens to the high account" is not a correct sentence. 

Based on the doc, it seems negative balance implies that the high account has tokens on this trustline. (While a positive balance indicates the low account has tokens)

### DIFF
--- a/content/references/protocol-reference/ledger-data/ledger-object-types/ripplestate.md
+++ b/content/references/protocol-reference/ledger-data/ledger-object-types/ripplestate.md
@@ -54,7 +54,7 @@ A `RippleState` object has the following fields:
 |-----------------|-----------|---------------|-------------|
 | `LedgerEntryType` | String    | UInt16 | The value `0x0072`, mapped to the string `RippleState`, indicates that this object is a RippleState object. |
 | `Flags`           | Number    | UInt32 | A bit-map of boolean options enabled for this object. |
-| `Balance`         | Object    | Amount | The balance of the trust line, from the perspective of the low account. A negative balance indicates that the low account has tokens to the high account. The issuer in this is always set to the neutral value [ACCOUNT_ONE](accounts.html#special-addresses). |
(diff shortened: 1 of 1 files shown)


content/references/protocol-reference/ledger-data/ledger-object-types/ripplestate.md
```diff
@@ -54,7 +54,7 @@ A `RippleState` object has the following fields:
 |-----------------|-----------|---------------|-------------|
 | `LedgerEntryType` | String    | UInt16 | The value `0x0072`, mapped to the string `RippleState`, indicates that this object is a RippleState object. |
 | `Flags`           | Number    | UInt32 | A bit-map of boolean options enabled for this object. |
-| `Balance`         | Object    | Amount | The balance of the trust line, from the perspective of the low account. A negative balance indicates that the low account has tokens to the high account. The issuer in this is always set to the neutral value [ACCOUNT_ONE](accounts.html#special-addresses). |
+| `Balance`         | Object    | Amount | The balance of the trust line, from the perspective of the low account. A negative balance indicates that the low account has given tokens to the high account. The issuer in this is always set to the neutral value [ACCOUNT_ONE](accounts.html#special-addresses). |
```
mDuo13, line 57:
```suggestion
| `Balance`         | Object    | Amount | The balance of the trust line, from the perspective of the low account. A negative balance indicates that the high account holds tokens issued by the low account. The issuer in this object is always set to the neutral value [ACCOUNT_ONE](accounts.html#special-addresses). |
```

mDuo13 (commented):
Good catch. I think "given" is not quite the right word because it comes with the connotation of "free" and also that the account has taken direct action to make that happen. It's entirely likely that one of the accounts may have purchased the tokens in the DEX, in which case the issuer did neither of those things!

I suggested a slightly bigger rewording that I think works better.

PR #1406, opened 2022-05-19
Add jobs link to nav

JakeatRipple:


dactyl-config.yml
```diff
@@ -4100,6 +4100,13 @@ pages:
         targets:
             - en
 
+    -   name: XRPL Jobs
+        html: http://jobs.xrpl.org/
```
mDuo13: Let's upgrade this to HTTPS as a good practice.

```suggestion
        html: https://jobs.xrpl.org/
```

mDuo13 (approved):
Left one nitpick, but :+1: as long as that's resolved.

One other question: is it intentional that we're not putting the link in the the Japanese translation? I can kind of see the sense in that if English is a requirement for all the jobs, but it does feel a little odd to have different navigation in the two site versions. (Hopefully it won't break the spacing/styling of the menus to have that, but I think it'll be OK.)

dactyl-config.yml
JakeatRipple: Agreed. I should have added that

JakeatRipple:
@mDuo13  Do we have a translation for the Japanese version?  Otherwise I see your point if it's a requirement for English.

mDuo13:
I don't think jobs.xrpl.org has a translation, no. But it might still be weird to not link it from the Japanese version?

JakeatRipple:
@mDuo13  added the link to Japanese target

PR #1414, opened 2022-06-02
Updates for js2.2.3

DennisDawson:
Update Quickstart tutorial for changes in language, add Broker Sale lesson.


content/tutorials/quickstart/create-accounts-send-xrp.md
```diff
@@ -243,7 +228,7 @@ Otherwise, populate the operational account fields.
 ```
 
 
-Insert the seed values for both accounts as they are created to the **Seeds** field as a convenience. You can copy the values and store them offline, then paste them into the **Seeds** field to retrieve the accounts with the `getAccountsFromSeeds()` function.
+Insert the seed values for both accounts as they are created to the **Seeds** field as a convenience. You can copy the values and store them offline. when you reload this form or another in this tutorial, copy and paste them into the **Seeds** field to retrieve the accounts with the `getAccountsFromSeeds()` function.
```
mDuo13: Looks like a missing capital letter at least. Not sure if there's a bigger typo going on here.

```suggestion
Insert the seed values for both accounts as they are created to the **Seeds** field as a convenience. You can copy the values and store them offline. When you reload this form or another in this tutorial, copy and paste them into the **Seeds** field to retrieve the accounts with the `getAccountsFromSeeds()` function.
```

content/tutorials/quickstart/create-accounts-send-xrp.md
```diff
@@ -539,18 +524,24 @@ Create a standard HTML form to send transactions and requests, then display the
...
+  
+<!-- ************************************************************** -->
+<!-- ********************** The Form ****************************** -->
+<!-- ************************************************************** -->
+
   <body>
     <h1>Token Test Harness</h1>
     <form id="theForm">
-      Choose your network:  
+      Choose your ledger instance:
```
mDuo13, line 544: I suggest rolling this change back. The term "network" is more consistent with the usage across our other docs, and "ledger instance" is likely to be interpreted to mean an instance of the ledger data type.

content/tutorials/quickstart/create-trustline-send-currency.md
```diff
@@ -53,7 +53,8 @@ To create a trustline between accounts:
 
 3. Enter a [currency code](https://www.iban.com/currency-codes) in the **Currency** field.
 4. Enter the maximum transfer limit in the **Amount** field.
-5. Click **Create Trustline**.
+5. Enter the destination account value in the **Destination** field.
+6. Click **Create Trustline**.
```
mDuo13, line 57: Per our style guide, "trust line" is two words. (The corresponding UI element should also change.)

content/tutorials/quickstart/mint-and-burn-nftokens.md
```diff
@@ -5,7 +5,8 @@ blurb: Quickstart step 3, mint and burn NFTokens.
 labels:
   - Quickstart
   - Tokens
-  - Non-fungible tokens, NFTs
+  - Non-fungible tokens
+  - NFTs
```
mDuo13: Changing the label to two labels requires changes to the CSS (to ensure both labels have appropriate colors) and to the dactyl config file to make sure both labels have landing pages.

To get this PR out the door ASAP, I suggest rolling this change back (in all places where it appears). We can address the issue more thoroughly in a follow-up.

```suggestion
  - Non-fungible tokens, NFTs
```

content/tutorials/quickstart/mint-and-burn-nftokens.md
```diff
@@ -20,7 +21,7 @@ This example shows how to:
 3. Delete (Burn) a NFToken.
 
 
-![Test harness with mint NFToken fields](img/quickstart13.png)
+![Test harness with mint NFToken fields](img/quickstart8.png)
```
mDuo13, line 23: It might be more effective to have descriptive rather than numbered file names for these images so that you don't have to renumber them every time you update the tutorial. Just a suggestion, no action needed on this now.

content/tutorials/quickstart/mint-and-burn-nftokens.md
```diff
@@ -122,15 +114,15 @@ Connect to the ledger and get the account wallets.
...
-          const operational_wallet = xrpl.Wallet.fromSeed(operationalSeedField.value)
-          const client = new xrpl.Client(net)
-        await client.connect()
-          results += '\nConnected. Minting NFToken.'
-          document.getElementById('standbyResultField').value = results
+  results = 'Connecting to ' + getNet() + '....'
+  document.getElementById('standbyResultField').value = results
+  let net = getNet()
+  const standby_wallet = xrpl.Wallet.fromSeed(standbySeedField.value)
+  const client = new xrpl.Client(net)
```
mDuo13, line 120: The blurb on line 111 above says "account wallets" but this code only gets one wallet now. Not sure which one should be fixed.

content/tutorials/quickstart/mint-and-burn-nftokens.md
```diff
@@ -219,51 +220,51 @@ Disconnect from the ledger.
...
 
-      async function getTokens() {
+async function getTokens() {
 ```
 
 
 Connect to the ledger and get the account wallet.
 
 
 ```
```
mDuo13, line 229: Another suggestion not important immediately: these code samples are a good use case for the `include_code` macro. That way you don't have to update the code in two places when changes are needed.

Also, syntax highlighting will work more reliably if you specify the language in the opening fence, e.g. \`\`\`js instead of just ```

content/tutorials/quickstart/transfer-nftokens.md
```diff
@@ -38,22 +38,14 @@ You can download the [Quickstart Samples](https://github.com/XRPLF/xrpl-dev-port
 
 
 1. Open `4.transfer-nftokens.html` in a browser.
-2. Get test accounts.
+2. Choose **XLS20-NFT** as your ledger instance.
```
mDuo13, line 41: Same comment as above re "ledger instance" vs "network"

content/tutorials/quickstart/transfer-nftokens.md
```diff
@@ -175,34 +169,59 @@ Connect to the ledger and get the wallet accounts.
...
+  let net = getNet()
+  const client = new xrpl.Client(net)
+  results = 'Connecting to ' + getNet() + '...'
+  document.getElementById('standbyResultField').value = results
+  await client.connect()
+  results += '\nConnected. Creating sell offer...'
+  document.getElementById('standbyResultField').value = results
+```
 
+Compute the Expiration Date, if present. The expiration date represents the number of seconds after the Ripple Epoch that the offer should expire. Subtract the Ripple Epoch from the current system date divided by 1,000, and add the number of days (or fractions of a day) multiplied by 86400 to get the correct expiration date.
```
mDuo13: This conversion is confusing (I didn't know that doing a division operation on a `Date()` object in JS implicitly converts it to milliseconds first) and a lot of this math could be simplified by using functions that are built into xrpl.js (see my comment on the actual code below).

content/tutorials/quickstart/transfer-nftokens.md
```diff
@@ -175,34 +169,59 @@ Connect to the ledger and get the wallet accounts.
...
 
+```
+  //------------------------------------- Prepare Expiration Date
+  var expirationDate = null
+  if (standbyExpirationField.value !="") {
+    var days = document.getElementById('standbyExpirationField').value
+    var secondsInDay = 86400
+    var rippleEpoch = 946684800
+    var currentTimestamp = (new Date() / 1000) - rippleEpoch
+    expirationDate = parseInt(Math.floor(currentTimestamp + (days*secondsInDay)))
```
mDuo13: Rather than figuring out tricky format conversions on their own, we should encourage users to rely on the conversion functions that are built into xrpl.js. For example:

```suggestion
    let d = new Date()
    d.setDate(d.getDate() + parseInt(days))
    var expirationDate = xrpl.isoTimeToRippleTime(d)
```

(The difference between `let` and `var` is a separate thing, which is probably worth discussing at some point, but not necessary to fix at this time)

content/tutorials/quickstart/broker-sale.md
```diff
@@ -0,0 +1,1053 @@
+---
+html: broker-sale.html
+parent: xrpl-quickstart.html
+blurb: Broker a sale between a sell offer and a buy offer.
+labels:
+  - Accounts
+  - Quickstart
+  - Broker
```
mDuo13, line 8: At some point we should add landing pages and colors for these labels (they'll be unclickable gray labels as is). I can push a commit that does that later though.

mDuo13 (approved):
The labels stuff won't work quite as desired here, and there are some areas where I suggested some longer-term or follow-up improvements plus at least one typo. But nothing major that should block the updates from going on.

content/tutorials/quickstart/create-accounts-send-xrp.md
DennisDawson: Will do in the next release.

content/tutorials/quickstart/create-trustline-send-currency.md
DennisDawson: Agreed, but will do in a future release.

content/tutorials/quickstart/create-accounts-send-xrp.md
amarantha-k, line 544:
```suggestion
      Choose your network:  
```

content/tutorials/quickstart/create-trustline-send-currency.md
amarantha-k, line 57:
```suggestion
6. Click **Create Trust line**.
```

content/tutorials/quickstart/transfer-nftokens.md
amarantha-k:
```suggestion
Compute the Expiration Date, if present. The expiration date represents the number of seconds after the Ripple Epoch when the offer should expire. Subtract the Ripple Epoch from the current system date divided by 1,000, and add the number of days (or fractions of a day) multiplied by 86400 to get the correct expiration date.
```

amarantha-k (approved):
Other than some suggestions added in the review, lgtm

content/tutorials/quickstart/transfer-nftokens.md
DennisDawson: Not sure I agree with this one, so I'll let it go rather than risk another edit.

DennisDawson:
Merging, with future improvements noted.

PR #1416, opened 2022-06-06
Update apt-key add instructions

legleux:
`apt-key add` is well deprecated and most instructions specifically say _not_ to do what we recommmend.

This is the new way.

Update a few versions also.

Someone should proof-read / update the Japanese instructions also.


mDuo13 (approved):
Looks good (including Japanese)

PR #1417, opened 2022-06-08
Add 'Normalize Currency Code' code sample

mDuo13:
Add a JavaScript code sample for decoding a variety of hex currency codes into human-readable strings.

### DIFF
--- a/content/_code-samples/normalize-currency-codes/js/normalize-currency-code.js
+++ b/content/_code-samples/normalize-currency-codes/js/normalize-currency-code.js
@@ -1,0 +1,65 @@
+////////////////////////////////////////////////////////////////////////////////
+// Normalize Currency Codes
+// Original authors: Ali (XUMM), nixer89
+// Upstream version: https://github.com/nixer89/xrpl-services-frontend/blob/master/src/app/utils/normalizers.ts
+////////////////////////////////////////////////////////////////////////////////
+// Convert from a string from either the XRP Ledger's "standard" 3-character
+// format, or several possible decodings of the "non-standard" 40-character
+// hexadecimal format into a string for humans to read.
+//
+// This code is intended for use with Node.js.
+////////////////////////////////////////////////////////////////////////////////
+
+function convertDemurrageToUTF8(demurrageCode) {
+
+    let bytes = Buffer.from(demurrageCode, "hex")
+    let code = String.fromCharCode(bytes[1]) + String.fromCharCode(bytes[2]) + String.fromCharCode(bytes[3]);
+    let interest_start = (bytes[4] << 24) + (bytes[5] << 16) + (bytes[6] <<  8) + (bytes[7]);
+    let interest_period = bytes.readDoubleBE(8);
+    const year_seconds = 31536000; // By convention, the XRP Ledger's interest/demurrage rules use a fixed number of seconds per year (31536000), which is not adjusted for leap days or leap seconds
+    let interest_after_year = Math.pow(Math.E, (interest_start+year_seconds - interest_start) / interest_period)
+    let interest = (interest_after_year * 100) - 100;
+
+    return(`${code} (${interest}% pa)`)
+}
+
+function normalizeCurrencyCode(currencyCode, maxLength = 20) {
+    if(!currencyCode) return "";
+
+    if(currencyCode.length === 3 && currencyCode.trim().toLowerCase() !== 'xrp') {
+        // "Standard" currency code
+        return currencyCode.trim()
+    }
+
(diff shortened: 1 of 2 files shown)


content/_code-samples/normalize-currency-codes/js/normalize-currency-code.js
```diff
@@ -0,0 +1,65 @@
...
+    return "";
+}
+
+console.log("Demurrage code (should be 'XAU (-0.5% pa)')...")
+console.log(normalizeCurrencyCode("0158415500000000C1F76FF6ECB0BAC600000000"))
+
+console.log("XLS-16d code (should be 'Purple moon')")
+console.log(normalizeCurrencyCode("0252000B03B6296F507572706C65206D6F6F6E00"))
+
+console.log("Long-alphanumeric code (should be 'XWORKS')...")
```
ckniffen, line 70: Can you add an example of it being sent `xRp` or something similar?

ckniffen (approved):

content/_code-samples/normalize-currency-codes/js/normalize-currency-code.js
```diff
@@ -0,0 +1,65 @@
...
+// Upstream version: https://github.com/nixer89/xrpl-services-frontend/blob/master/src/app/utils/normalizers.ts
+////////////////////////////////////////////////////////////////////////////////
+// Convert from a string from either the XRP Ledger's "standard" 3-character
+// format, or several possible decodings of the "non-standard" 40-character
+// hexadecimal format into a string for humans to read.
+//
+// This code is intended for use with Node.js.
+////////////////////////////////////////////////////////////////////////////////
+
+function convertDemurrageToUTF8(demurrageCode) {
```
JST5000: What does `demurrage` mean? (Maybe there's a more intuitive word for it?)

content/_code-samples/normalize-currency-codes/js/normalize-currency-code.js
```diff
@@ -0,0 +1,65 @@
...
+    let interest_start = (bytes[4] << 24) + (bytes[5] << 16) + (bytes[6] <<  8) + (bytes[7]);
+    let interest_period = bytes.readDoubleBE(8);
+    const year_seconds = 31536000; // By convention, the XRP Ledger's interest/demurrage rules use a fixed number of seconds per year (31536000), which is not adjusted for leap days or leap seconds
+    let interest_after_year = Math.pow(Math.E, (interest_start+year_seconds - interest_start) / interest_period)
+    let interest = (interest_after_year * 100) - 100;
+
+    return(`${code} (${interest}% pa)`)
+}
+
+function normalizeCurrencyCode(currencyCode, maxLength = 20) {
```
JST5000, line 13: Since this sample is about `normalizeCurrencyCode` should we put this at the top? I was surprised when reading to see another function which has a different intent than the filename first.

content/_code-samples/normalize-currency-codes/js/normalize-currency-code.js
```diff
@@ -0,0 +1,65 @@
...
+        }
+        const decodedHex = Buffer.from(hex, 'hex').toString('utf-8').slice(0, maxLength).trim()
+        if (decodedHex.match(/[a-zA-Z0-9]{3,}/) && decodedHex.toLowerCase() !== 'xrp') {
+            // ASCII or UTF-8 encoded alphanumeric code, 3+ characters long
+            return decodedHex
+        }
+    }
+    return "";
+}
+
```
JST5000, line 44: Should we start with a basic case to give an intuition for how they'd normally use it? (Ex. "USD" in hex)

content/_code-samples/normalize-currency-codes/js/normalize-currency-code.js
intelliot: https://en.wikipedia.org/wiki/Demurrage_(currency)
https://xrpl.org/demurrage.html

I'm not aware of any other words for demurrage, but I think it would be fine to add one or both of the above links for anyone reading the code to reference.

content/_code-samples/normalize-currency-codes/js/normalize-currency-code.js
intelliot: Demurrage-charged currencies have the interest rate encoded in their currency code, so parsing that rate and including it in the human-readable string is part of normalization.

That said, I think it would be fine to reorder the functions anyway

content/_code-samples/normalize-currency-codes/js/normalize-currency-code.js
mDuo13: The latter link already is in the code (in the `normalizeCurrencyCode()` function). So moving that function earlier should address this.

PR #1419, opened 2022-06-10
Feat event add hackathon

JakeatRipple:
Moved events that have past from upcoming to past sections. 
Added new hackathon and image in upcoming section


mDuo13 (approved):
LGTM!

PR #1423, opened 2022-06-21
Update amendments.md 

Silkjaer:
Default vote behaviour in amendment voting. Copy pasted "the server uses a default vote which is defined in the source code" from the first paragraph in the Amendment Voting section.

### DIFF
--- a/content/concepts/consensus-network/amendments/amendments.md
+++ b/content/concepts/consensus-network/amendments/amendments.md
@@ -37,7 +37,7 @@ By convention, an amendment's ID should be the SHA-512Half hash of the amendment
 
 ## Amendment Process
 
-Every 256th ledger is called a "flag" ledger. The process of approving an amendment starts in the ledger version immediately before the flag ledger. When `rippled` validator servers send validation messages for that ledger, those servers also submit votes in favor of specific amendments. If a validator does not vote in favor of an amendment, the server uses a default vote which is defined in the source code. ([Fee Voting](fee-voting.html) also occurs around flag ledgers.) [Updated in: rippled 1.8.1][]
+Every 256th ledger is called a "flag" ledger. The process of approving an amendment starts in the ledger version immediately before the flag ledger. When `rippled` validator servers send validation messages for that ledger, those servers also submit votes in favor of specific amendments. If a validator does not vote in favor of an amendment, that is the same as voting against the amendment. However, the `rippled` reference implementation has a default vote for each known amendment, which determines how a validator votes on any amendment for which that validator does not have an explicit configuration. ([Fee Voting](fee-voting.html) also occurs around flag ledgers.) [Updated in: rippled 1.8.1][]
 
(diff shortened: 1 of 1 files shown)


intelliot (approved):

mDuo13 (changes requested):
The original sentence was still correct, and these changes arguably make it incorrect. At a protocol level, things have not changed with regards to how validator votes are communicated and tracked: the lack of a vote in favor is still a vote against. This is separate from how validator votes are _configured_, which has changed. The way (yes/how) validators vote without explicit configuration now varies, but the way (how) a validator communicates a "no" vote is the same: by not sending any "yes" votes.

wojake (approved):

content/concepts/consensus-network/amendments/amendments.md
```diff
@@ -37,7 +37,7 @@ By convention, an amendment's ID should be the SHA-512Half hash of the amendment
 
 ## Amendment Process
 
-Every 256th ledger is called a "flag" ledger. The process of approving an amendment starts in the ledger version immediately before the flag ledger. When `rippled` validator servers send validation messages for that ledger, those servers also submit votes in favor of specific amendments. If a validator does not vote in favor of an amendment, that is the same as voting against the amendment. ([Fee Voting](fee-voting.html) also occurs around flag ledgers.)
+Every 256th ledger is called a "flag" ledger. The process of approving an amendment starts in the ledger version immediately before the flag ledger. When `rippled` validator servers send validation messages for that ledger, those servers also submit votes in favor of specific amendments. If a validator does not vote in favor of an amendment, the server uses a default vote which is defined in the source code. ([Fee Voting](fee-voting.html) also occurs around flag ledgers.) [Updated in: rippled 1.8.1][]
```
intelliot:
```suggestion
Every 256th ledger is called a "flag" ledger. The process of approving an amendment starts in the ledger version immediately before the flag ledger. When `rippled` validator servers send validation messages for that ledger, those servers also submit votes in favor of specific amendments. If a validator does not vote in favor of an amendment, that is the same as voting against the amendment. However, the `rippled` reference implementation has a default vote for each known amendment, which determines how a validator votes on any amendment for which that validator does not have an explicit configuration. ([Fee Voting](fee-voting.html) also occurs around flag ledgers.) [Updated in: rippled 1.8.1][]
```

intelliot (dismissed):
@mDuo13 brought up a good point. I have suggested alternate wording which I believe to be accurate.

content/concepts/consensus-network/amendments/amendments.md
wojake: I like this one better @Silkjaer

Silkjaer: That's also fine for me, I just didn't like the somewhat opposite statements in the documentation.

mDuo13 (approved):

Issue #1430, opened 2022-07-07
add to inline documenting ticket can be used


shortthefomo:
https://github.com/XRPLF/xrpl-dev-portal/blob/1564931ece24085577dabeb93787c79bd2b592c7/content/_code-samples/secure-signing/js/signPayment.js#L8

one also has the ability to use a ticket here in place of the sequence number

shortthefomo:
fix was merged

PR #1431, opened 2022-07-07
add comments for tickets issue #1430

shortthefomo:
light bit of documentation for good place to use tickets.

### DIFF
--- a/content/_code-samples/secure-signing/js/signPayment.js
+++ b/content/_code-samples/secure-signing/js/signPayment.js
@@ -4,7 +4,8 @@ const xrpl = require('xrpl')
 // Load seed value from an environment variable:
 const my_wallet = xrpl.Wallet.fromSeed(process.env['MY_SEED'])
 
-// For offline signing, you need to know your address's next Sequence number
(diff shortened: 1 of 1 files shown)


content/_code-samples/secure-signing/js/signPayment.js
```diff
@@ -4,7 +4,8 @@ const xrpl = require('xrpl')
 // Load seed value from an environment variable:
 const my_wallet = xrpl.Wallet.fromSeed(process.env['MY_SEED'])
 
-// For offline signing, you need to know your address's next Sequence number
+// For offline signing, you need to know your address's next Sequence number alternatively one could use a ticket 
+// in place of the sequence number, this is useful when you need multiple signatures.
```
intelliot:
```suggestion
// For offline signing, you need to know your address's next Sequence number.
// Alternatively, you could use a Ticket in place of the Sequence number.
// This is useful when you need multiple signatures and may want to process transactions out-of-order.
// For details, see: https://xrpl.org/tickets.html
```

intelliot (changes requested):
Just a suggestion

mDuo13 (approved):

intelliot (approved):

PR #1442, opened 2022-07-28
Update pages explaining how to run rippled

intelliot:
Update docs per [recommendations](https://discord.com/channels/886050993802985492/886053414289682522/1001732617135206551) from @nixer89

### DIFF
--- a/content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-stock-server.md
+++ b/content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-stock-server.md
@@ -24,7 +24,9 @@ To participate in the [consensus process](consensus.html) as a validator, [run r
 
 ## Install and run `rippled`
 
-For more information, see [Install `rippled`](install-rippled.html).
+The default package installation installs a stock server with a small amount of transaction history. For installation steps, see [Install `rippled`](install-rippled.html).
+
+After installation, you can adjust how much history your server stores at a time. For steps on how to do this, see [Configure Online Deletion](configure-online-deletion.html).
 
 ## Troubleshooting
 
--- a/content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-stock-server.md
+++ b/content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-stock-server.md
@@ -1,16 +1,16 @@
 ---
-html: run-rippled-as-a-wallet-server.html
+html: run-rippled-as-a-stock-server.html
 parent: configure-rippled.html
 blurb: A multipurpose configuration for anyone integrating XRP.
 labels:
   - Core Server
 ---
-# Run rippled as a Wallet Server
+# Run rippled as a Stock Server
 
-A wallet server is a multipurpose configuration for `rippled`. With a wallet server, you can submit transactions to the XRP Ledger, access ledger history, and use the latest [tools](software-ecosystem.html) to integrate with XRP.
+A stock server is a multipurpose configuration for `rippled`. With a stock server, you can submit transactions to the XRP Ledger, access ledger history, and use the latest [tools](software-ecosystem.html) to integrate with XRP. It is also a server that you can use to connect a wallet with the XRPL.
 
 
-A wallet server does all of the following:
+A stock server does all of the following:
 
 - Connects to a [network of peers](consensus-network.html)
 
--- a/content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-stock-server.md
+++ b/content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-stock-server.md
@@ -19,7 +19,7 @@ A stock server does all of the following:
 - Maintains a local copy of the complete shared global [ledger](ledgers.html)
 
 
-In order to participate in the [consensus process](consensus.html) as a validator, [run rippled as a validator](run-rippled-as-a-validator.html) instead.
+To participate in the [consensus process](consensus.html) as a validator, [run rippled as a validator](run-rippled-as-a-validator.html) instead.
 
 
 ## Install and run `rippled`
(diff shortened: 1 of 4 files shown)


nixer89:
PR looks good so far.

But I really have a hard time with the term "Wallet Server".
What is a "Wallet Server"?
From a newcomer's perspective it could sound like there is a wallet in the server? Or that the server "runs" your wallet?

I think we should change that term.

Some suggestions could be:
- "stock node"
- "submission node"
- "standard node"
- "non-validating node"

What do you think? Any other ideas?

intelliot:
"stock server" is a term that is already used on XRPL.org, and I'd prefer not to introduce a new term unless we have a very good reason to do so.

intelliot:
@mDuo13 Could we rename "Wallet Server" to "Stock Server"?

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-stock-server.md
```diff
@@ -1,39 +1,30 @@
 ---
-html: run-rippled-as-a-wallet-server.html
+html: run-rippled-as-a-stock-server.html
```
mDuo13, line 2: Whenever we change a filename, we need to add an entry in the `dactyl-config.yml` to redirect from the old filename. [Here's an example](https://github.com/XRPLF/xrpl-dev-portal/blob/47ab235afb202cb9e8ceb263603ff3d032ec199e/dactyl-config.yml#L221-L229). (the `html` of this entry is the old filename, and the `redirect_url` is the new filename)

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-validator.md
```diff
@@ -20,33 +20,37 @@ A [`rippled` server](xrpl-servers.html) running in [validator mode](rippled-serv
...
+    A good validator's votes match the outcome of the consensus process as often as possible. To do otherwise could indicate that your validator's software is outdated, buggy, or intentionally biased. Always run the [latest `rippled` release](https://github.com/XRPLF/rippled/tree/release) without modifications. [Watch `rippled` releases](https://github.com/XRPLF/rippled/releases) and subscribe to the [Google Group](https://groups.google.com/g/ripple-server) to be notified of new releases.
 
 - **Issuing timely votes**
 
     A good validator's votes arrive quickly and not after a consensus round has already finished. To keep your votes timely, make sure your validator meets the recommended [system requirements](system-requirements.html), which include a fast internet connection.
 
+Due to the need for consistent performance, it is not recommended to submit new transactions to, or query data from, a validator. A validator should be dedicated to participating in consensus.
```
mDuo13: This is debatable. The point of the "wallet server" configuration was to lower the barriers for businesses running validators, since having an entire separate machine for it is a big expense and if you're only doing light API queries (submitting occasional transactions and checking the balance and transactions of your own account) it's reasonable to expect "good enough" validation performance from a server in that configuration. (If it's powerful enough to handle your queries, it's probably powerful enough to do validation too.)

In an ideal world every validator would be perfect, but in the real world I think it's a fair and pragmatic approach for the network to have a larger population of validators run by different entities that have an interest in the XRP Ledger. Rather than putting validators on a pedestal, let's say these are goals and take a softer stance on servers that do both. Also, indent 4 spaces so it's a continuation of the previous bullet point.
 
```suggestion
    It is possible to submit new transactions and query data using a validator, but heavy loads of API queries may make the validator less reliable at keeping up with consensus. If your API needs are light enough, then you can use a server for both purposes. Ideally, a validator should be dedicated to participating in consensus.
```

mDuo13 (changes requested):

nixer89:
The build job still throws an error. Can this be resolved?

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-stock-server.md
```diff
@@ -1,39 +1,30 @@
...
 
 - Connects to a [network of peers](consensus-network.html)
 
 - Relays cryptographically signed [transactions](transaction-basics.html)
 
 - Maintains a local copy of the complete shared global [ledger](ledgers.html)
 
-- Participates in the [consensus process](consensus.html) as a validator
 
+In order to participate in the [consensus process](consensus.html) as a validator, [run rippled as a validator](run-rippled-as-a-validator.html) instead.
```
mDuo13: Per our style guidelines: you can almost always write "to" instead of "in order to", like here:

```suggestion
To participate in the [consensus process](consensus.html) as a validator, [run rippled as a validator](run-rippled-as-a-validator.html) instead.
```

content/tutorials/manage-the-rippled-server/configuration/run-rippled-as-a-stock-server.md
```diff
@@ -1,39 +1,30 @@
...
-
-For more information, see [Enable validation on your `rippled` server](run-rippled-as-a-validator.html#3-enable-validation-on-your-rippled-server).
+## Install and run `rippled`
 
-**Warning:** Validators should not be accessible to the public. Do not allow public WebSocket access to your wallet server or any other form of public access.
-
-## 3. Provide domain verification
-
-For more information, see [Provide domain verification](run-rippled-as-a-validator.html#6-provide-domain-verification).
+For more information, see [Install `rippled`](install-rippled.html).
```
mDuo13:
```suggestion
The default package installation installs a stock server with a small amount of transaction history. For installation steps, see [Install `rippled`](install-rippled.html).

After installation, you can adjust how much history your server stores at a time. For steps on how to do this, see [Configure Online Deletion](configure-online-deletion.html).
```

mDuo13 (approved):
I pushed changes that fix the build. Also, I left some optional suggestions to improve the text of the pages.

nixer89 (approved):
looks good!

PR #1456, opened 2022-08-05
New version for Education Platform release

DennisDawson:
This version has a CSS to make the forms look a little better.


mDuo13 (approved):
I think it would be nice to move away from storing this in the repo as a zip file in the future, since it makes it hard to review and compare the diff over time, but for now this is fine.

PR #1457, opened 2022-08-05
Add download section for logos

JakeatRipple:


template/page-community.html.jinja
```diff
@@ -233,6 +233,33 @@
...
+      </div>
+      <p class="mb-3 py-4">{% trans %}Just like the technology itself, XRPL assets are open source and available for anyone to use. Check out this helpful framework for using XRPL visuals. {% endtrans %}</p>      
+      <div class="col pr-2 d-lg-none d-block mb-4 pb-3 mb-lg-3">
+        <div class=" pr-1 mr-3">
+          <img src="./assets/img/community/community-design-assets.png" class="w-100">
+        </div>
+      </div>
+     
+      <div >
+        <a class="btn btn-primary btn-arrow" target="_blank" href=" https://github.com/XRPLF/xrpl-dev-portal/tree/master/content/XRPL_Logo_Kit.zip">{% trans %}Download the PDF and Assets{% endtrans %}</a>
```
mDuo13: Tip: if you use `{{target.github_forkurl}}` and `{{target.github_branch}}` this should work on the preview and will also be portable in case the repo name changes. (For local builds it won't work unless you pass these in via `--vars` or make temporary changes to the `dactyl-config.yml`.)

Also, remove the leading space, and if you change `/tree/` to `/raw/` it'll take you straight to the download instead of to [a page like this one](https://github.com/JakeatRipple/xrpl-dev-portal/blob/feat-community-visual%E2%80%94guide-download/content/XRPL_Logo_Kit.zip) where the users have to look around more for a second download button.

```suggestion
        <a class="btn btn-primary btn-arrow" target="_blank" href="{{target.github_forkurl}}/raw/{{target.github_branch}}/content/XRPL_Logo_Kit.zip">{% trans %}Download the PDF and Assets{% endtrans %}</a> 
```

mDuo13 (changes requested):
Left one suggestion for the URL. Otherwise, this looks good.

mDuo13 (approved):

PR #1459, opened 2022-08-05
Update top banner for Apex 2022

JakeatRipple:


styles/_top-banner.scss
```diff
@@ -1,8 +1,7 @@
 .top-banner {
     height: 46px;
-    background: url(../img/backgrounds/bg-apex-banner.svg);
+    background: url(../img/backgrounds/bg-apex-banner-2.gif);
```
mDuo13: Why are the new banner backgrounds in GIF format? Wouldn't SVG be preferable for this type of image? We _must_ have access to the source file for this graphic, right? Even if it's only available in raster form, PNG is more compressible with color options that are as good or better than GIF for still images (and these aren't animated).

mDuo13 (approved):
Looks good, though if we can switch the GIFs out for higher-quality assets first I would prefer that.

styles/_top-banner.scss
JakeatRipple: Agreed I normally export as svgs.  these were changed after I had the figma file and just sent as gif.  
I can check if they have svgs or the source file for these.

JakeatRipple: Found the files in figma. Updated to pngs for now as I was having issues saving as svg.  The bottom left checkered pattern would not show the purple bg.  Would always show white and black.

PR #1468, opened 2022-08-16
Clio server_info: Update some response field descriptions

amarantha-k:


content/references/http-websocket-apis/public-api-methods/clio-methods/server_info-clio.md
```diff
@@ -482,6 +482,7 @@ The `info` object may have some arrangement of the following fields:
 | `Field`                             | Type            | Description          |
 |:------------------------------------|:----------------|:---------------------|
 | `complete_ledgers`                  | String          | Range expression indicating the sequence numbers of the ledger versions the local `rippled` has in its database. This may be a disjoint sequence such as `24900901-24900984,24901116-24901158`. If the server does not have any complete ledgers (for example, it recently started syncing with the network), this is the string `empty`. |
+| `counters`                          | Object          | This is present only if the client connects to the `Clio` server over `localhost`.
 | `rpc`                               | Object array    | Stats on each RPC handled by the Clip server since startup. |
```
cjcobb23: `rpc` and `subscriptions` are nested inside `counters`, so also only appear if the client connects over localhost

cjcobb23 (changes requested):
Do you think we should have two example responses? One for when the client connects over localhost, and one for when they don't?

amarantha-k:
Incorporated feedback. PTAL

amarantha-k:
Rebased on master to pick up changes to fix broken links.

content/references/http-websocket-apis/public-api-methods/clio-methods/server_info-clio.md
```diff
@@ -475,20 +477,104 @@ An example of a successful response:
...
+
 The response follows the [standard format][], with a successful result containing an `info` object as its only field.
 
 The `info` object may have some arrangement of the following fields:
 
 | `Field`                             | Type            | Description          |
 |:------------------------------------|:----------------|:---------------------|
 | `complete_ledgers`                  | String          | Range expression indicating the sequence numbers of the ledger versions the local `rippled` has in its database. This may be a disjoint sequence such as `24900901-24900984,24901116-24901158`. If the server does not have any complete ledgers (for example, it recently started syncing with the network), this is the string `empty`. |
-| `rpc`                               | Object array    | Stats on each RPC handled by the Clip server since startup. |
+| `counters`                          | Object          | This is present only if the client connects to the `Clio` server over `localhost`.
```
mDuo13: The convention I've been using is `_(May be omitted)_` on fields that are sometimes not included in a response (along with context about when the field will/won't be there in the description, like you have here).

Also, this description doesn't say what the `counters` field represents; it probably should, right?

content/references/http-websocket-apis/public-api-methods/clio-methods/server_info-clio.md
```diff
@@ -475,20 +477,104 @@ An example of a successful response:
...
+| `rpc`                               | Object array    | Stats on each RPC handled by the Clip server since startup. Since this is nested within the `counters` object, this is also present only if the client connects to the `Clio` server over `localhost`. |
```
mDuo13: If it's nested within the `counters` array, then by the convention used elsewhere in this table, the Field column should be `counters.rpc`)

mDuo13: Also, based on the example above, `rpc` is _not_ an "Object array" but is simply an "Object". (The word "array" implies an ordered list of unnamed children, using square brackets, but this is a map of named values which may or may not be ordered meaningfully.)

I would describe this field as "A map of API methods names to stats about those methods." I try not to use the term "RPC" in like this in the docs since people tend to associate that term with JSON-RPC specifically even though in situations like this it can refer to the method being called through JSON-RPC _or_ WebSocket.

content/references/http-websocket-apis/public-api-methods/clio-methods/server_info-clio.md
```diff
@@ -475,20 +477,104 @@ An example of a successful response:
...
 | `rpc.*.started`                     | Number          | Number of RPCs of this type that the Clio server has started processing since startup. |
 | `rpc.*.finished`                    | Number          | Number of RPCs of this type that the Clio server has finished processing since startup. |
 | `rpc.*.errored`                     | Number          | Number of RPCs of this type that have resulted in some sort of error since startup.  |
 | `rpc.*.forwarded`                   | Number          | Number of RPCs of this type that the Clio server has forwarded to a `rippled` P2P server since startup. |
 | `rpc.*.duration_us`                 | Number          | The total number of microseconds spent processing RPCs of this type since startup. |
```
mDuo13, line 576: I find that notation in the style of `rpc.*.whatever` gets hard to read or mentally parse, so the convention I've been using in the reference docs is to break these out into a separate table below, with the table description saying something like, "The `rpc` object is a map of API method names to API Stat Objects (see below)." (Better yet—use a link instead of "see below") and then below I have a sub-heading for "API Stats Object" with a table of the fields that (may) appear in each one. See [account_nfts](https://xrpl.org/account_nfts.html#response-format) for a practical example.

mDuo13 (changes requested):
Suggested some changes for consistency with how we've been formatting API objects elsewhere.

content/references/http-websocket-apis/public-api-methods/clio-methods/server_info-clio.md
cjcobb23: Change `Clip` to `Clio`

content/references/http-websocket-apis/public-api-methods/clio-methods/server_info-clio.md
cjcobb23: For description, we can say something like `Stats on API calls handled since server startup`

content/references/http-websocket-apis/public-api-methods/clio-methods/server_info-clio.md
amarantha-k: Noted. Given that the information is correct and updates in this PR need to be published sooner than later, I suggest addressing this change in a separate PR.

cjcobb23 (approved):
LGTM

content/references/http-websocket-apis/public-api-methods/clio-methods/server_info-clio.md
mDuo13: Sounds good. Don't forget to make a ticket to track the follow-up work so it doesn't get lost!

mDuo13 (approved):

PR #1476, opened 2022-08-24
new dev tools page 

jonathanlei:

PR #1482, opened 2022-08-29
Feat - Developer Funding

JakeatRipple:


dactyl-config.yml
```diff
@@ -4248,6 +4248,14 @@ pages:
...
             - en
 
+    -   name: Developer Funding
+        html: developer-funding.html
+        parent: contribute.html
+        template: page-funding.html.jinja
+        sidebar: disabled
+        targets:
+            - en
+
```
mDuo13: Add the untranslated page to the Japanese target just so it's not a missing link there.

```suggestion
            - en
            - ja

```

template/page-funding.html.jinja
```diff
@@ -0,0 +1,342 @@
...
+                    </div>
+                    <div class="px-lg-3 pb-3 pt-5">
+                        <img id="funding-04">
+                        <div class=" pt-3">
+                            <h6 class="mb-3">Funding Levels</h6>
+                            <p class="">Prize money and awards</p>
+                        </div>
+                    </div>
+                </div>
+                <!-- end col 2 -->
```
mDuo13: This should probably be a for loop, right? That should also make it easier to fix some nitpicks with the markup, like:

- For the `<p class="">`, omit the empty attribute
- Close the image tags XHTML style. Also, a change that'll matter for Redocly: technically the src attribute isn't supposed to be omitted so you can provide an empty data URL to satisfy browsers that are picky about that ``

Same goes for the other funding lists below.

template/page-funding.html.jinja
```diff
@@ -0,0 +1,342 @@
...
+                            <p class="">$10,000 - $200,000</p>
+                        </div>
+                    </div>
+                </div>
+                <!-- end col 2 -->
+            </div>
+        </div>
+
+        <div class="d-lg-none order-3 mt-4 pt-3">
+            <a class="btn btn-primary btn-arrow btn-arrow-out" target="_blank" href="https://xrplgrants.org/">{% trans %}Visit XRPL Grants{% endtrans %}</a>
```
mDuo13: The `btn-arrow` and `btn-arrow-out` classes are clashing and resulting in weird jaggedy movement on the hover animation. Removing `btn-arrow` here fixes it, making the arrow behave the same as the "Learn more" for the Creator Fund link below:

```suggestion
            <a class="btn btn-primary btn-arrow-out" target="_blank" href="https://xrplgrants.org/">{% trans %}Visit XRPL Grants{% endtrans %}</a>
```

mDuo13 (changes requested):
We'll need to coordinate with #1476 to make sure whoever merges second doesn't clobber the (S)CSS updates, but otherwise this looks mostly good. I noticed one little bug in the animation and a couple opportunities for cleaning up the code.

template/page-funding.html.jinja
JakeatRipple: Yes I think some loop could work but have not figured out an easy way to do it.  The staggered on desktop changes order from the mobile. 

I can look into it further but for the time being I fixed those other issues so it wont hold up the PR.

mDuo13 (approved):
Looks like the build failure is because the link checker doesn't understand the new `data:,` URIs, which, go figure. You make one tool happy and it makes the other tool upset. They still display correctly to end-users, which is what really matters. :+1:  to this PR.

PR #1484, opened 2022-08-31
Add screenshots to dev tools

mDuo13:
Builds on #1476 by adding screenshots of the dev tools to better set expectations of what they are.

For future reference, here's the process I used for taking screenshots:

1. Navigate to a representative page of the tool
2. Full-screen the browser window. These were taken on a 1920×1080 resolution screen, although that's not a strict requirement.
3. Take a screen capture and save as PNG
4. Resize the image to 540×304. I chose this resolution to be almost exactly 2x the width the images are scaled to in the cards on a desktop with no pixel scaling applied—so that they should look OK at up to 2x pixel scaling. If your screenshot's original resolution isn't 1920x1080, you may need to crop it to achieve the same aspect ratio as the existing images.
5. Use compresspng.com (you could also use another tool) to compress the size of the image file. Since the details aren't as important as the general impression, it's best to sacrifice some detail for more images to load faster.


mDuo13:
For the updated screenshots, I wrote a quick script to take the screenshots, scale, and optimize them all at once. (Uses ImageMagick and pngcrush, which I have installed at the system level.) In case this might be useful for someone else in the future, these are the settings I found to work best:

```bash screencrush.sh
#!/bin/sh
activeWinLine=$(xprop -root | grep "_NET_ACTIVE_WINDOW(WINDOW)")
activeWinId=${activeWinLine:40}
mkdir -p /tmp/screencrush/
cd /tmp/screencrush
fname="$(date +%F_%H%M%S_%N).png"
import -window "$activeWinId" $fname
magick $fname -kmeans 32 -resize 540x304^ -gravity center -extent 540x304 -type palette $fname
pngcrush -reduce -brute -ow $fname
```

(Note: for the faucet I separately and manually pixelized the "secret" section just as a good practice)

jonathanlei (approved):

PR #1494, opened 2022-09-03
Code sample for JS tx-serialization

TusharPardhe:


TusharPardhe:
This PR is to understand the process of bounty program, as per my understanding we had to add a code sample using libraries or adding the functionality to the library itself. 

If I misunderstood please use this PR as sample so everyone else can further understand the process.

Bounty [DGE-94](https://github.com/XRPLBounties/Proposals/blob/main/bounties/0%20JavaScript%20code%20samples.md) Milestone 2

content/_code-samples/tx-serialization/js/main.js
```diff
@@ -0,0 +1,45 @@
...
+        const complete_payload = await client.autofill(payload);
+
+        // convert the JSON tx to serialized using ripple-binary-codec
+        const decoded_payload = ripple_binary_codec.encode(complete_payload);
+
+        // push binary tx into output array
+        outputs.push(decoded_payload);
+
+        // write the output into binary_outputs.json file.
+        fs.writeFileSync("./binary_outputs.json", JSON.stringify(outputs, null, "\t"));
```
mDuo13: Why write to file instead of just printing the result via `console.log()`?

content/_code-samples/tx-serialization/js/binary_outputs.json
```diff
@@ -0,0 +1 @@
+[]
```
mDuo13: It doesn't make sense to me to commit an output file to the repository like this.

content/_code-samples/tx-serialization/js/main.js
```diff
@@ -0,0 +1,45 @@
...
+
+        // connect to a public server
+        const client = new Client("wss://xrplcluster.com/");
+        await client.connect();
+
+        // autofill all the missing fields
+        const complete_payload = await client.autofill(payload);
+
+        // convert the JSON tx to serialized using ripple-binary-codec
+        const decoded_payload = ripple_binary_codec.encode(complete_payload);
```
mDuo13: This is aside the point for this code sample, but you can actually use [`xrpl.encode(payload)`](https://js.xrpl.org/functions/encode.html) to perform this without needing to separately import `ripple-binary-codec`. Turns out, xrpl.js also imports the binary codec and exports most of its methods directly.

mDuo13 (changes requested):
I think there has been a bit of a misunderstanding in this case.

The `tx-serialization` code sample should have the same scope as [the Python version](https://github.com/XRPLF/xrpl-dev-portal/tree/master/content/_code-samples/tx-serialization/py). In this case, that means demonstrating the actual underlying code of serialization, not just how to use a library to perform serialization.

content/_code-samples/tx-serialization/js/main.js
TusharPardhe: Good point, I will change it right away.

content/_code-samples/tx-serialization/js/binary_outputs.json
TusharPardhe: removing this file

content/_code-samples/tx-serialization/js/main.js
TusharPardhe: Updated the code, please review.

JST5000:
Here's the updated link to the bounty referred to in this PR: https://github.com/XRPLBounties/Proposals/blob/main/bounties/0031%20JavaScript%20Code%20Samples.md

content/_code-samples/tx-serialization/js/main.js
```diff
@@ -0,0 +1,36 @@
+const { Client, encode } = require("xrpl");
+
+const payload = require("./json_input.json");
+
+const main = async () => {
+    try {
+        console.log(`Converting your transaction...\n`);
+
+        // connect to a public server
```
JST5000, line 9: Style note - Please capitalize the first letter of comments
```suggestion
        // Connect to a public server
```

content/_code-samples/tx-serialization/js/main.js
```diff
@@ -0,0 +1,36 @@
+const { Client, encode } = require("xrpl");
+
+const payload = require("./json_input.json");
```
JST5000, line 3: It's worth bringing in the transaction formation within the snippet because ideally this code should run on it's own. Someone should be able to copy + paste it and start playing with it without any other setup.

content/_code-samples/tx-serialization/js/main.js
```diff
@@ -0,0 +1,36 @@
...
+
+const main = async () => {
+    try {
+        console.log(`Converting your transaction...\n`);
+
+        // connect to a public server
+        const client = new Client("wss://xrplcluster.com/");
```
JST5000, line 10: Instead of pointing to a mainnet node, it'd be better to point at a test network so that people who use this code snippet can play around with the values themselves.

mDuo13:
The problem here is that this code is not at all the equivalent of the Python code, which demonstrates the actual binary operations—the nuts and bolts of the serialization process. This code merely demonstrates how to use xrpl.js to serialize a transaction, which may be useful in its own way but is beside the point of the Python code sample.

Put another way, the Python code sample is for teaching how to create a client library, whereas this code sample only teaches how to use a client library that already exists.

JST5000 (changes requested):
For this to qualify, it needs to have feature parity with the Python implementation here: https://github.com/XRPLF/xrpl-dev-portal/tree/master/content/_code-samples/tx-serialization/py

One of the main things required then is going into the actual serialization logic in order to illustrate that to the reader.

(Ah, sorry I thought the original concerns were already addressed when diving into the code, sorry for a duplicated concern)

PR #1496, opened 2022-09-03
Added key pair sample code

TusharPardhe:
XRPL bounty sample for [DGE-94](https://github.com/XRPLBounties/Proposals/blob/main/bounties/0%20JavaScript%20code%20samples.md) Milestone 1

----

The goal of this bounty is to support JavaScript for all of the [code samples](https://xrpl.org/code-samples.html). Milestone 1 is [Cryptographic Key Derivation](https://github.com/XRPLF/xrpl-dev-portal/tree/master/content/_code-samples/key-derivation), previously available only in Python.


wojake:
Seems too high level for a sample code. I think you'd need to use a crypto lib to derive the keypairs from a seed in a low level manner.

TusharPardhe:
> 

Can pull the implementation from xrpl.js or go more deeper but first want to know whether we need to be precise with the examples using ripple libs or have to write them from scratch.

wojake:
Using the implementation used on xrpl.js would be fine (i guess)

TusharPardhe:
> Using the implementation used on xrpl.js would be fine (i guess)

As suggested, updated the code to derive ed25519 key pairs.

mDuo13 (changes requested):
I think we want this to have the same scope as the [Python key derivation code sample](https://github.com/XRPLF/xrpl-dev-portal/blob/master/content/_code-samples/key-derivation/py/key_derivation.py). That means:

- No using Ripple or XRPLF libraries for seed derivation. It's OK to use them for base58 encoding since that's not the point of this sample. (Hint: all 16-byte UInts are valid seeds for Ed25519 keys, so you just need to generate a random number using a cryptographically-secure source of randomness.)
- Should cover both ed25519 and secp256k1 derivation. That's important because secp256k1 derivation is significantly more complicated.

This code sample accurately handles the easiest part (using an off-the-shelf crypto library to derive a Ed25519 key pair from seed) but doesn't cover some of the more challenging aspects of the code it's supposed to port, so this is incomplete thus far.

TusharPardhe:
> 

Thanks for the info, that's exactly what I was looking for. I will update the code plus will suggest to make bounties more descriptive as we move ahead.

TusharPardhe:
> I think we want this to have the same scope as the [Python key derivation code sample](https://github.com/XRPLF/xrpl-dev-portal/blob/master/content/_code-samples/key-derivation/py/key_derivation.py). That means:
> 
>     * No using Ripple or XRPLF libraries for seed derivation. It's OK to use them for base58 encoding since that's not the point of this sample. (Hint: all 16-byte UInts are valid seeds for Ed25519 keys, so you just need to generate a random number using a cryptographically-secure source of randomness.)
> 
>     * Should cover both ed25519 and secp256k1 derivation. That's important because secp256k1 derivation is significantly more complicated.
> 
> 
> This code sample accurately handles the easiest part (using an off-the-shelf crypto library to derive a Ed25519 key pair from seed) but doesn't cover some of the more challenging aspects of the code it's supposed to port, so this is incomplete thus far.



> > 
> 
> Thanks for the info, that's exactly what I was looking for. I will update the code plus will suggest to make bounties more descriptive as we move ahead.

I have updated the code as per your comments. @mDuo13 Please review.

JST5000:
Here's the updated link to the bounty referred to in this PR: 
https://github.com/XRPLBounties/Proposals/blob/main/bounties/0031%20JavaScript%20Code%20Samples.md

TusharPardhe:
@JST5000 @mDuo13 @mvadari Is this bounty complete? If not can you go through the current code, I think its now somewhat equivalent to the python code.

JST5000:
> @JST5000 @mDuo13 @mvadari Is this bounty complete? If not can you go through the current code, I think its now somewhat equivalent to the python code.

I'll check today. Sorry for the delay - didn't get a notification for this message for some reason :/

JST5000 (approved):
LGTM

jonathanlei (approved):

TusharPardhe:
Thanks @jonathanlei and @JST5000. What do I have to do next to get the PR merged?

JST5000:
Hey @TusharPardhe we have decided to **accept** your work for the bounty, but ended up merging a different submission (#1507). 

I'm going to close this PR, **but please submit to claim the bounty - you will receive the full amount** for your the unique contribution

Here's the form to claim a bounty: [https://xrpl.typeform.com/xrpl-bounties](https://xrpl.typeform.com/xrpl-bounties?utm_source=xxxxx&utm_medium=xxxxx&utm_campaign=xxxxx&utm_term=xxxxx&utm_content=xxxxx)

TusharPardhe:
> 

Ahh I would suggest updating the bounty page so we don't have similar PRs (It'll get difficult for the reviewers as well). Also, thank you for considering my code as well its always a good experience to learn more about XRPL.

Looking forward for more contributions!

PR #1498, opened 2022-09-03
Added key derivation js example based on XRPL.JS

roodtim:
hope its not too late!

Included a few tests as well.


roodtim:
hope its not too late!

intelliot:
@roodtim this looks like a good start. Can you add some additional comments and documentation (explanation)?

What did you learn from doing this exercise?

Can you add a few original tests? From what I saw, it looks like these are copied from the existing implementation in the `ripple-keypairs` library.

JST5000:
Here's a link to the bounty this PR is referring to: https://github.com/XRPLBounties/Proposals/blob/main/bounties/0031%20JavaScript%20Code%20Samples.md

JST5000:
Just following up on this :) 

For the bounty, the goal is to have feature parity with this implementation in python here: https://github.com/XRPLF/xrpl-dev-portal/tree/master/content/_code-samples/key-derivation/py

At a high level, the hope is that someone could look at the code, and get a better sense for what's going on in this tutorial: https://xrpl.org/cryptographic-keys.html#key-derivation

So, if you can comments to make it clear what each step in the process is doing and why, that'd be helpful to future readers :)

JST5000 (changes requested):
(See above comment)

JST5000:
Hey @roodtim I'm closing this PR as we have accepted a different submission for this bounty (https://github.com/XRPLF/xrpl-dev-portal/pull/1507).

PR #1514, opened 2022-09-13
DGE-94-Cryptographic-Key-Derivation

interc0der:
Bounty pull request for 0031 - DGE-94-Cryptographic-Key-Derivation

### DIFF
--- a/content/_code-samples/key-derivation/js/derive.js
+++ b/content/_code-samples/key-derivation/js/derive.js
@@ -1,0 +1,51 @@
+'use strict'
+// Dependencies for Node.js.
+const api = require('ripple-keypairs')
+
+const entropy= [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16]
+
+// Derive XRPL wallet using ed25519  --------------------
+const seed_ed25519 = api.generateSeed({entropy:entropy, algorithm:'ed25519'});
+const keypair_ed25519 = api.deriveKeypair(seed_ed25519);
+const address_ed25519 = api.deriveAddress(keypair_ed25519.publicKey);
+const accountObject_ed25519  = {
+    'accountType': 'familySeed',
+    'address': address_ed25519,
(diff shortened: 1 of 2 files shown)


jonathanlei:
Good start of implementation using existing xrpl-js tools, could you add some tests?

JST5000:
For context this is referring to the xrpl.js code snippets bounty: https://github.com/XRPLBounties/Proposals/blob/main/bounties/0031%20JavaScript%20Code%20Samples.md

content/_code-samples/key-derivation/js/derive.js
```diff
@@ -0,0 +1,51 @@
+'use strict'
+// Dependencies for Node.js.
+const api = require('ripple-keypairs')
+
+const entropy= [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16]
```
JST5000, line 5: nit:
```suggestion
const entropy = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16]
```

JST5000:
For this bounty the goal is to have feature parity with these snippets so they can be used interchangeably depending on what language the new reader is most comfortable with: https://github.com/XRPLF/xrpl-dev-portal/tree/master/content/_code-samples/key-derivation/py

And specifically, the snippets are supposed to give a concrete way to understand these high-ish level concepts: https://xrpl.org/cryptographic-keys.html#key-derivation

So instead of using the client library to abstract away the details of the encode/decode, it's actually important to step through it and add comments explaining the intent. (The python one has some of those comments, and the high level concept also may have some useful things to point out in the code)

JST5000 (changes requested):
See above comment

JST5000:
Hey @interc0der I'm closing this PR as we have accepted a different submission for this bounty (https://github.com/XRPLF/xrpl-dev-portal/pull/1507).

PR #1519, opened 2022-09-19
Add Batch Minting

DennisDawson:


content/tutorials/quickstart/batch-minting.md
```diff
@@ -0,0 +1,494 @@
...
+
+**Note:** Running this command throws an error in the JavaScript console because the `getAccountsFromSeeds` function in `ripplex1-send-xrp.js` looks for the operational seed field, which is not included in this form. You can ignore the error (or fix it in your own implementation).
+
+## Batch Mint NFTokens
+
+This example lets you mint multiple NFTokens for a single unique item. The NFToken might represent "prints" of an original artwork, tickets to an event, or another limited set of unique items. 
+
+To batch mint a non-fungible token objects:
+
+1. Set the **Flags** field. For testing purposes, we recommend setting the value to _8_. This sets the _tsTransferable_ flag, meaning that the NFToken object can be transferred to another account. Otherwise, the NFToken object can only be transferred back to the issuing account. See [NFToken Mint](https://xrpl.org/nftokenmint.html#:~:text=Example%20NFTokenMint%20JSON-,NFTokenMint%20Fields,-NFTokenMint%20Flags) for information about all of the available flags for minting NFTokens.
```
mDuo13: Use a relative link here rather than a fully-qualified URL so that it doesn't get marked as an "external" link. Also, that `#:~:text=...` anchor doesn't seem to do anything, so I removed it.

```suggestion
1. Set the **Flags** field. For testing purposes, we recommend setting the value to _8_. This sets the _tsTransferable_ flag, meaning that the NFToken object can be transferred to another account. Otherwise, the NFToken object can only be transferred back to the issuing account. See [NFTokenMint](nftokenmint.html) for information about all of the available flags for minting NFTokens.
```

content/tutorials/quickstart/batch-minting.md
```diff
@@ -0,0 +1,494 @@
...
+  results += "\n\nNFTs:\n"
+  let nfts = await client.request({
+    method: "account_nfts",
+    account: standby_wallet.classicAddress,
+    limit: 400
+  })
+  results += JSON.stringify(nfts,null,2)
+```
+
+If the list of NFTokens exceeds your limit, the result will include a `marker` field that you can use as a parameter for the next `account_nfts` request. The `marker` indicates where the next batch of records starts. While the `marker` field is not null, continue to request another batch of NFToken records.
```
mDuo13:
```suggestion
If the list of `NFTokens` exceeds your limit, the result includes a `marker` field that you can use as a parameter for the next `account_nfts` request. The `marker` indicates where the next batch of records starts. While the `marker` field is present, continue to request another batch of NFToken records.
```

content/tutorials/quickstart/batch-minting.md
```diff
@@ -0,0 +1,494 @@
...
+
+```javascript
+  while (nfts.result.marker != null)
```
mDuo13:
```suggestion
  while (nfts.result.marker)
```

The expression as written misleadingly implies that the API returns the actual JSON value `null` in the marker field when there are no additional pages. I don't think this is the case. If it's like any of the other methods, the API omits the `nfts.results.marker` field entirely when there are no additional pages. I suspect that the code as written only works (if it even does!) because JavaScript implicitly converts the value `undefined` to something that it considers equivalent to `null` in this operation.

But I don't have an account with 401+ NFTs on hand to test with, so I'm not sure.

content/tutorials/quickstart/batch-minting.md
```diff
@@ -0,0 +1,494 @@
...
+    <script src='https://unpkg.com/xrpl@2.2.3'></script>
+    <script src='ripplex1-send-xrp.js'></script>
+    <script src='ripplex3-mint-nfts.js'></script>
+    <script src='ripplex7-batch-minting.js'></script>
+    <script>
+      if (typeof module !== "undefined") {
+        const xrpl = require('xrpl')
+      }
+     
+    </script>
```
mDuo13, line 352: I think this can be omitted entirely for this example code.

The `.js` files I use as examples in some other places do this sort of thing so that they can either be included in an HTML page, or run directly with Node.js; it doesn't make sense to include this code inline in an HTML file.

mDuo13 (approved):
Looks good, though I did suggest a couple little improvements.

content/tutorials/quickstart/batch-minting.md
DennisDawson: Will catch this next time around.

Issue #1183, opened 2021-10-04
Double-check and fix ordering of ledger object field tables


mDuo13:
The [Ledger Object Types](https://xrpl.org/ledger-object-types.html) documentation has parameter tables for each object type, but the ordering of the fields in the table is not consistent.

We should update the tables for all the ledger object types to list the fields in a consistent order. The order we have been using most often is to list required fields in alphabetical order, followed by optional fields in alphabetical order. I've been thinking maybe it would make more sense to put all the fields in alphabetical order (regardless of optionality) since users who are looking up a field in the table after seeing it in an object may not know if the field is required or optional, and having them separated may make it easier to miss the field.

I think the formats returned by the rippled APIs typically put the fields in (case-sensitive) alphabetical order, so that dovetails reasonably well with putting all fields in alphabetical order overall.

PR #1524, opened 2022-09-21
Put ledger objects into alphabetical order and minor improvements

LimpidCrypto:
## Summary
This PR just adds some minor improvements such as fix a misspelling, fixes some nits and a false typing.
It also addresses the issue #1183 and puts the ledger object fields in alphabetical order, regardless of wether they are optional or not.

## Tasks
- [x] I also like that the NFToken... objects have extra columns to tell if a field is optional. I find it way clearer than defining it in the fields description. Should I add an extra column to all ledger objects?
- [x] If an object has the field `Flags` it is still possible the object has no flags. We should have a consistent description for these cases, including an explanation that the value is always `0`.


mDuo13:
I like what you've done here so far. I agree it would be good to have a consistent description for the case where no flags are defined, and I'm on board with adding a column to all object types to indicate whether a field is required or "optional". (I find the term "optional" to be a bit weird in some of these cases, because it implies that the user can omit it, but the actual meaning is that the object may or may not have it.) I guess it's fine if the column title is "Required" and the column contents are "Yes" or "No"

wojake (approved):
I like it this way @mDuo13

LimpidCrypto:
> I like it this way @mDuo13 

I personally think the '(Option)' text sometimes gets lost between all the field descriptions. I would find a column 'Required', as @mDuo13 suggested, between types and description way clearer. But maybe that's just me.

LimpidCrypto:
Since this is my first PR I could use some context why the Link Checker failed. A broken link in public API method `account_nfts`, but I didn't changed anything there. Thank you :)

content/references/protocol-reference/ledger-data/ledger-object-types/accountroot.md
```diff
@@ -36,29 +36,29 @@ The `AccountRoot` object type describes a single [account](accounts.html), its s
...
+| `TicketCount`                 | Number    | UInt32            | No        | How many [Tickets](tickets.html) this account owns in the ledger. This is updated automatically to ensure that the account stays within the hard limit of 250 Tickets at a time. This field is omitted if the account has zero Tickets. _(Added by the [TicketBatch amendment][].)_ |
+| `TickSize`                    | Number    | UInt8             | No        | How many significant digits to use for exchange rates of Offers involving currencies issued by this address. Valid values are `3` to `15`, inclusive. _(Added by the [TickSize amendment][].)_ |
+| `TransferRate`                | Number    | UInt32            | No        | A [transfer fee](transfer-fees.html) to charge other users for sending currency issued by this account to each other. |
+| `WalletLocator`               | String    | Hash256           | No        | **DEPRECATED**. Do not use. |
+| `WalletSize`                  | Number    | UInt32            | No        | **DEPRECATED**. Do not use. |
```
mDuo13, line 61: While you're here, can you update these descriptions based on the updated descriptions for the corresponding AccountSet fields in #1517?

Some context: I believe it's currently [impossible to set `WalletSize`](https://github.com/ripple/rippled/issues/3816) but the source allows it as an AccountRoot field. There miiiiight be legacy objects with this field defined?

WalletLocator is OK to use, and has no inherent meaning. It's just an arbitrary 256-bit field you can set to whatever you want.

content/references/protocol-reference/ledger-data/ledger-object-types/nftokenoffer.md
```diff
@@ -27,179 +27,28 @@ The unique ID (`NFTokenOfferID`) of the `NFTokenOffer` object is the result of t
...
-   </td>
-   <td><code>0x00000001</code>
-   </td>
-   <td>If set, the offer is a sell offer. Otherwise, the offer is a buy offer.
-   </td>
-  </tr>
-</table>
+|Flag Name|Flag Value|Description|
+|---|---|---|
+|lsfSellNFToken|0x00000001|If set, the offer is a sell offer. Otherwise, the offer is a buy offer.|
```
mDuo13: I suggest this minor tweak to match our style (aside: the style guide in this repo still needs to be updated, sorry).

```suggestion
| `lsfSellNFToken `| `0x00000001` | If enabled, the offer is a sell offer. Otherwise, the offer is a buy offer. |
```

content/references/protocol-reference/ledger-data/ledger-object-types/nftokenpage.md
```diff
@@ -175,56 +95,11 @@ Each `NFTokenPage` costs an incremental reserve to the owner account. This speci
...
-   </td>
-   <td>0.03125 XRP
-   </td>
-  </tr>
-</table>
+| Incremental Reserve | 1 NFToken | 8 NFTokens | 16 NFTokens | 32 NFTokens | 64 NFTokens |
+|:--------------------|:----------|:-----------|:------------|:------------|:------------|
+| 5 XRP               | 5 XRP     | 0.625 XRP  | 0.3125 XRP  | 0.15625 XRP | 0.07812 XRP |
+| 2 XRP               | 2 XRP     | 0.25 XRP   | 0.125 XRP   | 0.0625 XRP  | 0.03125 XRP |
+| 1 XRP               | 1 XRP     | 0.125 XRP  | 0.0625 XRP  | 0.03125 XRP | 0.01562 XRP |
```
mDuo13, line 102: You may want to take a look at #1523 which also modified this section. But, basically, I think you can remove the "64 NFTokens" column.

mDuo13 (approved):
Thanks for doing this. I noted a couple things to keep an eye out regarding other recent PRs, but I'm happy to merge this regardless. And we should probably do so sooner rather than later so that you don't have to continue keeping it up to date with other PRs.

PR #1526, opened 2022-09-23
NFToken: fix basis points numbers

mDuo13:
Basis points are 1/10,000, meaning one basis point is "a percent of a percent". The XLS-20 spec originally had fees in basis points and at some point that changed to ⅒ basis points, which caused some inconsistency/confusion in the docs. This fixes couple of NFT examples (in both English and Japanese) that had invalid percentages listed. _(Updated.)_


mDuo13:
Actually, I think the standard was changed to 1/10 of a basis point (units of 1/100,000 or 0.001%) at some point, so this might still be wrong. (If it is, the Japanese version needs to be updated too.) Need to double-check...

DennisDawson (changes requested):
The actual value is 1/10 of a basis point, or .001%.

Line 71 should read:
"The `TransferFee` value specifies the percentage fee, in units of 1/1,000, charged by the issuer for secondary sales of the token. Valid values for this field are between 0 and 50,000, inclusive. A value of 1 is equivalent to 1/10 of a basis point or 0.001%, allowing transfer rates between 0% and 50%."

Line 76 should read:
"This value sets the transfer fee to 314, or .0314%"

mDuo13:
I believe line 76 should read "314, or 0.314%" not .0314, because `0.0314` would be in 1/100ths of basis points.

DennisDawson (approved):

DennisDawson:
Yes, .314% is correct.

PR #1527, opened 2022-09-23
Xrp

ghost:
Xrp


ghost (commented):
Files

mDuo13:
Please provide more context because as it stands, this PR looks like spam.

I don't think there is a need to merge the changes from master into the `add_batch_minting` working branch.

PR #1537, opened 2022-10-03
generate faucet account client side

jonathanlei:
Context: 
currently [faucet account](https://xrpl.org/xrp-testnet-faucet.html) is generated through faucet backend and sent to client side. This refactor will generate the faucet account client side and send to faucet API.

### DIFF
--- a/assets/js/test-net.js
+++ b/assets/js/test-net.js
@@ -64,6 +64,7 @@ function rippleTestNetCredentials(url, altnet_name) {
         test_wallet.address).fadeIn('fast')
       secret.hide().html('<h3>Secret</h3> ' +
       test_wallet.seed).fadeIn('fast')
+      // TODO: currently the faucet api doesn't return balance unless the account is generated server side, need to make upates when faucet repo is updated. 
       balance.hide().html('<h3>Balance</h3> ' +
         Number(data.amount).toLocaleString('en') + ' XRP').fadeIn('fast')
       sequence.html('<h3>Sequence</h3> <img class="throbber" src="assets/img/xrp-loader-96.png"> Waiting...').fadeIn('fast')
--- a/assets/js/test-net.js
+++ b/assets/js/test-net.js
@@ -1,90 +1,138 @@
 async function wait_for_seq(network_url, address) {
-  const api = new xrpl.Client(network_url)
-  await api.connect()
+  const api = new xrpl.Client(network_url);
+  await api.connect();
   let response;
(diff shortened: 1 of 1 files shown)


wojake (approved):

ckniffen (approved):

assets/js/test-net.js
```diff
@@ -37,29 +44,35 @@ function rippleTestNetCredentials(url, altnet_name) {
...
       credentials.hide().html('<h2>Your '+altnet_name+' Credentials</h2>').fadeIn('fast')
       address.hide().html('<h3>Address</h3> ' +
-        data.account.address).fadeIn('fast')
+        test_wallet.address).fadeIn('fast')
       secret.hide().html('<h3>Secret</h3> ' +
-        data.account.secret).fadeIn('fast')
+      test_wallet.seed).fadeIn('fast')
       balance.hide().html('<h3>Balance</h3> ' +
-        Number(data.balance).toLocaleString('en') + ' XRP').fadeIn('fast')
+        Number(data.amount).toLocaleString('en') + ' XRP').fadeIn('fast')
```
mDuo13, line 69: I believe the `amount` is how much the Faucet sent, whereas `balance` is how much they have now; if you're adding funds to an existing account, your actual "balance" is of course the sum of how much you had before plus how much you received.

Currently it doesn't matter because we are still funding fresh accounts, but if we build the extension that I have in mind to let users save their testnet credentials in browser storage for reuse, it would be inaccurate to call the "amount" and "balance" the same. The accurate "Balance" in that case, which the user probably would like to know, is of course, their previous balance _plus_ the amount the faucet sent.

For now, could you just leave a comment for later reminding us to adjust this if we allow wallet reuse? Also, if you wanted to merge the changes from #1535 into this PR I wouldn't object.

mDuo13 (approved):

assets/js/test-net.js
jonathanlei: yeah so currently the blocker is that the faucet code doesn't send balance info unless it's creating an account, the balance is updated through `wait_for_seq,` I just added that as a temporary balance until the  ` wait_for_seq` returns a result. but I will make a note.

jonathanlei: and in terms of adding https://github.com/XRPLF/xrpl-dev-portal/pull/1535 into the PR, I think more testing once the faucet server PR is merged is needed so I prefer to keep them separate.

PR #1545, opened 2022-10-17
Feat learning portal banner

JakeatRipple:


template/component-top-nav.html.jinja
```diff
@@ -1,15 +1,15 @@
-{# <div class="top-banner fixed-top"> 
+<div class="top-banner fixed-top"> 
   <!-- mobile -->
-  <div class="justify-content-center d-md-none">
+  <!-- <div class="justify-content-center d-md-none">
     <span><p class="mb-0">{% trans %}Registration is live for Apex: The XRPL{% endtrans %}</p></span>
-    <span><p class="mb-0"> Developer Summit <a href="https://www.apexdevsummit.com/" target="_blank" class="btn btn-outline-secondary">Register Now</a></p></span>
-  </div>
+    <span><p class="mb-0"> Developer Summit <a href="https://www.apexdevsummit.com/" target="_blank" class="btn btn-outline-secondary">Start Learning</a></p></span>
+  </div> -->
```
mDuo13: Rather than commenting out the old banner, can we fully delete it?

mDuo13 (approved):
Looks good, though I suggested one bit of cleanup. (Also, <s>I haven't yet confirmed it actually builds</s>, but it looks like it will.)

EDIT: confirmed it builds and looks good. :+1:

template/component-top-nav.html.jinja
JakeatRipple: Removed.

JakeatRipple:
@mDuo13   I merged the other branch.  Then pulled those changes into this one and rebuilt the css file.  bumped version to 10 for this so that will included other changes so far.

PR #1549, opened 2022-10-18
Add Get NFToken List example

DennisDawson:


content/tutorials/quickstart/get-nftoken-list.md
```diff
@@ -0,0 +1,337 @@
...
+```
+
+Connect to the XLS-20Devnet server.
+
+```javascript
+  const client = new xrpl.Client('wss://xls20-sandbox.rippletest.net:51233')
+	await client.connect()
+```
+
+Check to see if the **Marker** field is populated with a 64 character value. If so, include the marker value in the `ledger_data` request. Set the limit of records to be returned to 200 transactions.
```
mDuo13, line 64: Markers are an explicitly undefined format; checking that they're a 64-character value may not be future proof. This works for now but it might get broken by "non-breaking" changes in other releases or server versions.

content/tutorials/quickstart/get-nftoken-list.md
```diff
@@ -0,0 +1,337 @@
...
+
+```javascript
+  if (markerField.value.length == 64) {
```
mDuo13, line 67: I learned today that this is a niche JavaScript feature called _named access_ where the element with the ID (`markerField` here) is implicitly assigned to a global variable matching its name. Apparently there are [some reasons not to do this](https://www.tjvantoll.com/2012/07/19/dom-element-references-as-global-variables/). It's also not that widely known, so others might also be confused by how you're referencing this variable that has never been assigned. 

And, of course, this only works in browser JS with a specific HTML file, so if you want to do something similar in Node.js you'll have to adapt this code and it's not immediately obvious that porting is necessary—if you did a `document.getElementById('markerField')` first, it would at least be clear that this is a browser-specific thing.

Not strictly a blocker, but I think ideally we'd revise the sample code not to do this.

content/tutorials/quickstart/get-nftoken-list.md
```diff
@@ -0,0 +1,337 @@
...
+		nftoken_data = await client.request({
+			command: "ledger_data",
+			id: "NFToken data request",
+			limit: 200,
+			marker: markerField.value
+		})
```
mDuo13, line 73: I notice this doesn't specify a ledger index / hash. I'm not 100% sure how `ledger_data` handles its markers, but I am worried this might mean that the ledger you're iterating over might shift while you're in the process of iterating it.

Probably the more secure way of doing this is specifying `ledger_index: "validated"` in the case where there's no marker, and specifying `ledger_index: nftoken_data.result.ledger_index` in the case where there _is_ a marker. (That way, you default to "whatever the latest validated ledger is" for the initial call, and "the ledger index you started with" for each follow-up.) I think technically with the way this script is architected you also have to do something extra to make sure that the response from the _previous_  `ledger_data` call (well, at least the `result.ledger_index` value) is available when you do the next call.

content/tutorials/quickstart/get-nftoken-list.md
```diff
@@ -0,0 +1,337 @@
...
+
+```javascript
+	jason =  JSON.stringify(nftoken_data.result, null, 2)
+```
+
+Parse the state tree from the JSON object.
+
+```javascript
+	stateTree = JSON.parse(jason).state
+```
```
mDuo13, line 99: This looks to me like you are taking a JSON object, converting it to a string, and then back into a JSON object. Am I missing something here, or is this a completely unnecessary convert→unconvert step?

content/tutorials/quickstart/get-nftoken-list.md
```diff
@@ -0,0 +1,337 @@
...
+
+## Usage
+
+You can download the [Quickstart Samples](https://github.com/XRPLF/xrpl-dev-portal/tree/master/content/_code-samples/quickstart/js/quickstart.zip){.github-code-download} archive to try the sample in your own browser.
+
+### Get NFTokens
+
+All transactions on the XRP Ledger are available for public review. You do not need an account or credentials to access information about minting and trading NFTokens.
+
+To get a list of current `NFTokens` on the XRP Ledger, click **Get NFTokens**. The button retrieves 200 transactions (the maximum batch allowance), then filters to show only the information from `NFToken` pages. The row marker for the response is stored in the **Marker** field. Each new search starts at the marker to return a unique set of records for the next 200 transactions. Click **Get NFTokens** repeatedly to retrieve all of the `NFTokens` on the XRP Ledger.
```
mDuo13, line 22: I'd add a note here that actually retrieving _all_ the NFTokens is likely to a a _lot_ of calls and a very long time on Mainnet because it needs to iterate over the entire ledger contents (including non-NFT data), which is multiple gigabytes. And doing it too fast might get you rate-limited, so you might need to run your own server to do this in a timely fashion.

content/tutorials/quickstart/get-nftoken-list.md
```diff
@@ -0,0 +1,337 @@
...
+		if (ledger_entry_type == "NFTokenPage") {
+			nfTokens = stateTree[i].NFTokens
+			for (j=0; j < nfTokens.length; j++) {
+				results += JSON.stringify(nfTokens[j])
+			}
+		}
+	}
+```
+
+Disconnect from the XLS-20Devnet.
```
mDuo13, line 115: The official name is "NFT-Devnet", not "XLS-20Devnet", but I suggest saying "network" instead so that you don't have to update this again when you update the doc later. (NFT-Devnet is probably going to go away at some point after NFTs become enabled on Mainnet & Testnet.)

content/tutorials/quickstart/get-nftoken-list.md
```diff
@@ -0,0 +1,337 @@
...
+  client.disconnect()
+```
+
+Report the filtered `NFToken` objects in the resultField.
+
+```javascript
+  resultField.value = results
+```
+
+Save the marker value in the **Marker** field.
```
mDuo13, line 127: I suggest being a little more explicit where this is storing it, because "field" can be ambiguous (at first read, I thought it was referring to a field of a JSON object or something).

```suggestion
Save the marker value in the **Marker** field of the HTML form.
```

content/tutorials/quickstart/get-nftoken-list.md
```diff
@@ -0,0 +1,337 @@
...
+Declare variables used in the function.
+
+```javascript
+async function getNFTokens () {
+	let jason
+	let ledger_entry_type
+	let nfTokens
+	let nftoken_data
+	let results = ""
+	let stateTree
```
mDuo13, line 54: This feels like a very "C/C++" coding style, not modern JavaScript. Some of these variables (like `ledger_entry_type`) can be defined as `const` more locally when they're first assigned; only a few (like `nftoken_data`) need to be defined in advance so that they have the appropriate scope.

Also, please don't use tab characters. I notice that the zipped `ripplex8-all-nftokens.js` file mixes space and tab characters, which results in inconsistent indentation depending on how wide the user's text editor decides a tab character should be. It looks like this code was written with the assumption that a tab is two spaces wide, but one of my text editors is showing them as 4 spaces wide and the GitHub preview is showing them as 8 spaces—both of those mean that the parts that are meant to be indented the same amount don't show that way.

mDuo13 (changes requested):
This works, but I have a number of concerns regarding the code sample. Not all of them may require changes, but I'd like to follow up with the subject-matter expert before publishing as-is.

PR #1558, opened 2022-10-28
Create airgapped-wallet.py

wojake:
Create an airgapped wallet code sample in Python, more details in proposed README.md & bounty details below.

Bounty Details: https://github.com/XRPLBounties/Proposals/blob/main/bounties/Code%20Sample%20For%20Airgapped%20Wallet.md


content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
```diff
@@ -0,0 +1,203 @@
...
+    if OS == "Windows":
+        # If it's Windows, use this path:
+        File = PureWindowsPath(str(usr) + '/Wallet')
+        Path_ = str(PureWindowsPath(str(usr)))
+    if OS == "Linux":
+        # If it's Linux, use this path:
+        File = PurePath(str(usr) + '/Wallet')
+        Path_ = str(PurePath(str(usr)))
+        
+    # If the Wallet's folder doesn't exist, continue on
```
mDuo13: I think this is supposed to say "already exists"?

```suggestion
    # If the Wallet's folder already exists, continue on
```

mDuo13 (commented):
I don't have time to look at this thoroughly just yet, but someone from Ripple's DGE team will. I did notice one apparent typo though.

jonathanlei (commented):
Looks good to me. Could you add some tests if applicable?

wojake:
Ye I didn't think of that, I'll add in a unit test file. Thanks.

content/_code-samples/airgapped-wallet/py/README.md
```diff
@@ -0,0 +1,111 @@
...
+- `relay-transaction.py` - This code could be stored in any online machine, no credentials is stored on this code other than a signed transaction which would be sent to an XRPL node for it to be validated on the ledger.
+
+Preferably, `airgapped-wallet.py` should be on a Linux machine while `relay-transaction.py` could be on any operating system.
+
+# Security Practices
+Strongly note that an airgapped system's security is not determined by its code alone but the security practices that are being followed by an operator.
+
+There are channels that can be maliciously used by outside parties to infiltrate an airgapped system and steal sensitive information.
+
+There are other ways malware could interact across airgapped networks, but they all involve an infected USB drive or a similar device introducing malware onto the airgapped machine. (They could also involve a person physically accessing the computer, compromising it and installing malware or modifying its hardware.)
```
JST5000: nit: No need for parens here since you end with a period anyway. 
```suggestion
There are other ways malware could interact across airgapped networks, but they all involve an infected USB drive or a similar device introducing malware onto the airgapped machine. They could also involve a person physically accessing the computer, compromising it and installing malware or modifying its hardware.
```

content/_code-samples/airgapped-wallet/py/README.md
```diff
@@ -0,0 +1,111 @@
...
+  <img src="https://user-images.githubusercontent.com/87929946/197970678-2a1b7f7e-d91e-424e-915e-5ba7d34689cc.png" width=75% height=75%>
+</p>
+
+# Setup
+- Machine 1 - An airgapped computer (during setup, it must be connected to the internet to download the files)
+- Machine 2 - A normal computer connected to the internet
+- Phone - A normal phone with a working camera to scan a QR
+
+## Machine 1 Setup
+Since this machine will be airgapped, it is best to use Linux as the Operating System.
```
JST5000, line 57: I'm not clear on why that is - is it because Linux has better stability without updates? Better in-built security?

content/_code-samples/airgapped-wallet/py/README.md
```diff
@@ -0,0 +1,111 @@
...
+1. Install Python 3.8:
+
+**Linux Command Line**:
+```
+sudo apt-get update
+sudo apt-get install python3.8 python3-pip
+```
+**Website**: https://www.python.org/downloads/source/
+
+2. Clone all the files under the [`airgapped-wallet`](https://github.com/XRPLF/xrpl-dev-portal/tree/master/content/_code-samples/airgapped-wallet/py) directory
```
JST5000, line 68: Is there an easily copy + pastable command you could add here for that?

content/_code-samples/airgapped-wallet/py/relay-transaction.py
```diff
@@ -0,0 +1,47 @@
...
+          f"\n         Transaction Hash: {tx_hash}"
+          f"\n  Transaction Destination: {tx_destination}"
+          f"\n           Transacted XRP: {tx_xrp_amount}"
+          f"\n              Wallet Used: {tx_account}"
+         )
+
+
+if __name__ == '__main__':
+    tx = "ENTER TX BLOB HERE"
+    send_transaction(tx)
```
JST5000: Would it be better to have this wait for command-line input instead of requiring the user to edit the code directly? (It would only swap the order of steps 4 and 5 in the instructions)

content/_code-samples/airgapped-wallet/py/README.md
```diff
@@ -0,0 +1,111 @@
...
+## Phone Setup
+The phone requires a working camera that is able to scan a QR code and an internet connection for it to be able to transmit the signed transaction blob to Machine 2.
+
+Once you have signed a transaction in the airgapped machine, a QR code will be generated which will contain the signed transaction blob. Example:
+
+<img src="https://user-images.githubusercontent.com/87929946/196018292-f210a9f2-c5f8-412e-98c1-361a72286378.png" width=20% height=20%>
+
+Scan the QR code using the phone and transmit it to Machine 2, which will then be sending it to a rippled node.
+
+You could use Discord, Whatsapp or even e-mail the signed transaction blob to yourself in order for you to receive the signed transaction blob in Machine 2.
```
JST5000: Slight rephrasing suggestion
```suggestion
You can send a message to yourself using Discord, WhatsApp or even e-mail, then open up the message using Machine 2 to receive the signed transaction blob.
```

content/_code-samples/airgapped-wallet/py/airgapped-wallet-TEST.py
```diff
@@ -0,0 +1,228 @@
...
+        address = keypairs.derive_classic_address(pub)
+        private = keypairs.derive_classic_address(priv)
+
+        return address, private, seed
+    if silent:
+        seed = keypairs.generate_seed()
+        pub, priv = keypairs.derive_keypair(seed)
+        address = keypairs.derive_classic_address(pub)
+        private = keypairs.derive_classic_address(priv)
+        return address, private, seed
```
JST5000: Slight simplification
```suggestion
    else:
        seed = keypairs.generate_seed()
        pub, priv = keypairs.derive_keypair(seed)
        address = keypairs.derive_classic_address(pub)
        private = keypairs.derive_classic_address(priv)
    
    return address, private, seed
```

content/_code-samples/airgapped-wallet/py/airgapped-wallet-TEST.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def sign_transaction(_xrp_amount, _destination, _ledger_seq, _wallet_seq, password):
```
JST5000: I often see `_var_name` used to denote a variable which is assigned, but never supposed to be used. (Or if it's a function name, to indicate a private function). In this case I think using that convention is a bit confusing because these variables are used in the function.

content/_code-samples/airgapped-wallet/py/airgapped-wallet-TEST.py
```diff
@@ -0,0 +1,228 @@
...
+    print("1. Retrieving encrypted private key and salt...")
+    with open(get_path("/WalletTEST/private.txt"), "r") as f:
+        _seed = f.read()
+        _seed = bytes.fromhex(_seed)
+
+    with open(get_path("/WalletTEST/salt.txt"), "rb") as f:
+        salt = f.read()
+
+    print("2. Initializing key...")
+    # Line 49-58: initialize key
```
JST5000: This comment seems out of date, and the print explains it anyway so i think you can just delete it.
```suggestion
```

content/_code-samples/airgapped-wallet/py/README.md
wojake: I don't think so, I tried using git to only clone the files under the `airgapped-wallet` directory but git could only clone the entire repo (IIRC).

content/_code-samples/airgapped-wallet/py/README.md
wojake: As compared to Windows, Linux is less prone to malware since a huge number of viruses are developed specifically for Windows, most viruses can't harm Linux in this case.

content/_code-samples/airgapped-wallet/py/relay-transaction.py
wojake: It doesn't work for input(). Unless we create another file (.json) for the tx blob and load it into the script. This is the best solution that I can find.

content/_code-samples/airgapped-wallet/py/airgapped-wallet-TEST.py
wojake: Right, changes have been made.

content/_code-samples/airgapped-wallet/py/airgapped-wallet-TEST.py
```diff
@@ -0,0 +1,227 @@
...
+        last_ledger_sequence=validated_seq + 100,
+        # +100 to catch up with the ledger when we transmit the signed tx blob to Machine 2
+        sequence=_wallet.sequence,
+        fee="10"
+    )
+
+    print("6. Signing transaction...")
+    my_tx_payment_signed = safe_sign_transaction(transaction=my_tx_payment, wallet=_wallet)
+
+    img = qrcode.make(my_tx_payment_signed.to_dict())
```
JST5000, line 85: nit: the img generation should probably be under the below print statement since it's part of that QR code logic

content/_code-samples/airgapped-wallet/py/airgapped-wallet-TEST.py
```diff
@@ -0,0 +1,227 @@
...
+    print("END RESULT: Successful")
+
+
+def get_path(file):
+    """
+    Get path (filesystem management)
+    """
+
+    global File_
+    # Checks what OS is being us
```
JST5000:
```suggestion
    # Checks what OS is being used
```

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
```diff
@@ -0,0 +1,203 @@
...
+def create_wallet():
+    """
+    Generates a keypair
+    """
+
+    seed = keypairs.generate_seed()
+    pub, priv = keypairs.derive_keypair(seed)
+
+    address = keypairs.derive_classic_address(pub)
+    private = keypairs.derive_classic_address(priv)
```
JST5000: Is `derive_classic_address(priv)` the right function here? (I thought `priv` was already the private key?)

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
```diff
@@ -0,0 +1,203 @@
...
+    )
+
+    key = base64.urlsafe_b64encode(kdf.derive(bytes(password.encode())))
+    crypt = Fernet(key)
+    
+    # Decrypts the wallet's private key
+    _seed = crypt.decrypt(_seed)
+    
+    # Initialize XRPL wallet using decrypted private key
+    _wallet = wallet.Wallet(seed=_seed.decode(), sequence=0)
```
JST5000, line 61: This sequence number seems like it could easily fall out of date (It works for the first transaction, but next time we run this code how do we update the sequence?)

JST5000: (Or is the thought that since we're regenerating wallets it doesn't matter?)

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
```diff
@@ -0,0 +1,203 @@
...
+    
+    # Construct Payment transaction
+    my_tx_payment = Payment(
+        account=_wallet.classic_address,
+        amount=xrp_to_drops(xrp=_xrp_amount),
+        destination=_destination,
+        last_ledger_sequence=validated_seq + 100, # +100 to catch up with the ledger when we transmit the signed tx blob to Machine 2
+        sequence=_wallet.sequence,
+        fee="10"
+    )
```
JST5000, line 75: Relatedly, how does the user initially fund the wallet? (Should that be a step in the tutorial?)

content/_code-samples/airgapped-wallet/py/relay-transaction.py
```diff
@@ -0,0 +1,47 @@
...
+    my_tx_signed = Payment.from_dict(transaction_dict)
+
+    tx = send_reliable_submission(transaction=my_tx_signed, client=client)
+
+    tx_hash = tx.result['hash']
+    tx_destination = tx.result['Destination']
+    tx_xrp_amount = int(tx.result['Amount']) / 1000000
+    tx_account = tx.result['Account']
+
+    print(f"\n           XRP-L Explorer: https://livenet.xrpl.org/transactions/{tx_hash}"
```
JST5000: nit:
```suggestion
    print(f"\n           XRPL Explorer: https://livenet.xrpl.org/transactions/{tx_hash}"
```

content/_code-samples/airgapped-wallet/py/relay-transaction.py
```diff
@@ -0,0 +1,47 @@
...
+          f"\n         Transaction Hash: {tx_hash}"
+          f"\n  Transaction Destination: {tx_destination}"
+          f"\n           Transacted XRP: {tx_xrp_amount}"
+          f"\n              Wallet Used: {tx_account}"
+         )
+
+
+if __name__ == '__main__':
+    tx = "ENTER TX BLOB HERE"
```
JST5000: If no tx blob is entered, it should probably print something to the user e.g. "Please set `tx` to the blob you get from scanning the offline-wallet's QR code to try sending that transaction!"

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
wojake: You're right, my mistake

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
wojake: They'd have the create the wallet and fund it themselves, I'll add this step in the tutorial.

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
wojake: `_wallet.sequence = _wallet_seq`: The user has to input in the account's sequence manually to sign a transaction.

content/_code-samples/airgapped-wallet/py/README.md
JST5000: Maybe better phrasing would be `Download` instead of `Clone` then? (Because otherwise I feel obligated to google it and then I'm in a rabbit hole where Git doesn't do what I want)

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
JST5000: Why is that done on a separate line? (I wouldn't expect to look elsewhere for the initialization since the `sequence` param is named explicitly)

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
JST5000: I don't quite see this step in the tutorial yet - Marking conversation as unresolved in the meantime (I read a couple times, but if I just missed it please let me know)

content/_code-samples/airgapped-wallet/py/README.md
```diff
@@ -0,0 +1,113 @@
...
+2. Clone all the files under the [`airgapped-wallet`](https://github.com/XRPLF/xrpl-dev-portal/tree/master/content/_code-samples/airgapped-wallet/py) directory
+
+3. Import all the modules required by running:
+```
+pip install -r requirements.txt
+```
+
+4. Airgap the machine by following the security practices written [here](#security-practices).
+
+5. Run `airgapped-machine.py`
```
JST5000: There is no `airgapped-machine.py` - should this be `airgapped-wallet.py`?

content/_code-samples/airgapped-wallet/py/airgapped-wallet-TEST.py
```diff
@@ -0,0 +1,225 @@
...
+    global File_
+    # Checks what OS is being used
+    OS = platform.system()
+    usr = Path.home()
+
+    # Get PATH format based on the OS
+    if OS == "Windows":
+        File_ = PureWindowsPath(str(usr) + file)
+    if OS == "Linux":
+        File_ = PurePath(str(usr) + file)
```
JST5000:
```suggestion
    # Get PATH format based on the OS
    if OS == "Windows":
        File_ = PureWindowsPath(str(usr) + file)
    else: # Assuming Linux-style file format
        File_ = PurePath(str(usr) + file)
```

My macbook is running `Darwin` so couldn't run the code. I think using else is good enough usually.

JST5000: (This has ripple effects down below since there are other checks for Linux which would need to be updated as well)

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
```diff
@@ -0,0 +1,200 @@
...
+            if ask == 3:
+                with open(get_path("/Wallet/public.txt"), "r") as f:
+                    print(f"\n  Wallet Address: {f.read()}")
+
+                image = Image.open(get_path("/Wallet/public.png"))
+                image.show()
+
+    else:
+        # If the Wallet's folder does not exist, create one and store wallet data (encrypted private key, encrypted seed, account address) 
+        pub, priv, seed = create_wallet()
```
JST5000: `create_wallet` only returns 2 things, did you mean to also return `priv` from `create_wallet`? (Or omit it from this unpacking?)

content/_code-samples/airgapped-wallet/py/README.md
```diff
@@ -0,0 +1,113 @@
...
+**Website**: https://www.python.org/downloads/source/
+
+2. Clone all the files under the [`airgapped-wallet`](https://github.com/XRPLF/xrpl-dev-portal/tree/master/content/_code-samples/airgapped-wallet/py) directory
+
+3. Import all the modules required by running:
+```
+pip install -r requirements.txt
+```
+
+4. Edit line 47 @ `relay-transaction.py` and insert the signed transaction blob
```
JST5000: I'm trying to follow the steps, but I'm confused where we get a signed transaction blob - seems like a couple steps are missing, but I'm not sure what they are ("Maybe grab the wallet generated via the QR code and use it in Xumm to sign an accountset transaction or something"?)

JST5000: Or is there supposed to be a QR code I can scan with the transaction on it? When running the steps I just get one that gives me an address as far as I can tell

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
```diff
@@ -0,0 +1,200 @@
...
+        
+        os.makedirs(File)
+        
+        img = qrcode.make(pub)
+        img.save(get_path("/Wallet/public.png"))
+        openimg = Image.open(get_path("/Wallet/public.png"))
+        
+        openimg.show()
+
+        password = str(input("\n        Enter Password: "))
```
JST5000, line 185: Just running the code I'm not sure where I can find that - Do I need to run the 'Create Wallet' step first - then fund that, then enter that secret here?

JST5000: (Entering dummy values results in a crash because `private.txt` does not exist)

JST5000:
```
python airgapped-wallet.py
/Users/jacksonmills/Wallet

 1. Transact XRP
 2. Generate an XRP wallet (read only)
 3. Showcase XRP Wallet Address (QR Code)

 Enter Index: 2


       XRP WALLET CREDENTIALS
  Wallet Address: rpRaYChQJFNwcPFvQ7hgE52Bn8pRJxGQW2
            Seed: sEdSuKDGQbhgxL98UpZrrdDDNepT57D

 1. Transact XRP
 2. Generate an XRP wallet (read only)
 3. Showcase XRP Wallet Address (QR Code)

 Enter Index: 1
             Enter Password: sEdSuKDGQbhgxL98UpZrrdDDNepT57D

           Enter XRP To Send: 1

        Enter Destination: rpRaYChQJFNwcPFvQ7hgE52Bn8pRJxGQW2

    Enter Wallet Sequence: 3

    Enter Ledger Sequence: 20
Traceback (most recent call last):
  File "/Users/jacksonmills/Documents/wojake-xrpl-dev-portal/xrpl-dev-portal/content/_code-samples/airgapped-wallet/py/airgapped-wallet.py", line 201, in <module>
    main()
  File "/Users/jacksonmills/Documents/wojake-xrpl-dev-portal/xrpl-dev-portal/content/_code-samples/airgapped-wallet/py/airgapped-wallet.py", line 137, in main
    sign_transaction(_xrp_amount=amount,
  File "/Users/jacksonmills/Documents/wojake-xrpl-dev-portal/xrpl-dev-portal/content/_code-samples/airgapped-wallet/py/airgapped-wallet.py", line 39, in sign_transaction
    with open(get_path("/Wallet/private.txt"), "r") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/jacksonmills/Wallet/private.txt'
```

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
```diff
@@ -0,0 +1,200 @@
...
+                            "\n 3. Showcase XRP Wallet Address (QR Code)"
+                            "\n\n Enter Index: "
+                            ))
+
+            if ask == 1:
+                password = str(input("             Enter Password: "))
+                amount = float(input("\n           Enter XRP To Send: "))
+                destination = input("\n        Enter Destination: ")
+                wallet_sequence = int(input("\n    Enter Wallet Sequence: "))
+                ledger_sequence = int(input("\n    Enter Ledger Sequence: "))
```
JST5000: Getting the Ledger Sequence is non-trivial and should have a hint about how to get it I think. (My wording is probably not the best, but this is an example of what that could look like)
```suggestion
                ledger_sequence = int(input("Look up the latest ledger index on livenet.xrpl.org and put it here\n    Enter Ledger Sequence: "))
```

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
```diff
@@ -0,0 +1,200 @@
...
+        # If it's Windows, use this path:
+        File = PureWindowsPath(str(usr) + '/Wallet')
+        Path_ = str(PureWindowsPath(str(usr)))
+    if OS == "Linux":
+        # If it's Linux, use this path:
+        File = PurePath(str(usr) + '/Wallet')
+        Path_ = str(PurePath(str(usr)))
+        
+    # If the Wallet's folder already exists, continue on
+    if os.path.exists(File):
```
JST5000: Instead of just checking the folder exists - it would probably be better to check if the individual files exist during the setup because I accidentally didn't complete all the steps and had a tough time debugging why the sample wasn't working for me. 

(On Mac the QR code opens up an image and that takes you away from the terminal screen - so it's very easy to miss the "Please enter a password:" prompt - which then means I have a `Wallet` folder, but no password file which makes the sample not function properly next time I run it because it assumes I have a `private.txt` file)

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
JST5000: Ah I understand now - This happened because I didn't complete the initial setup properly - I was alt-tabbed by the QR code appearing and I guess I didn't finish setting a password somehow. But there was no error which showed that until much later.

JST5000: (I have a Wallet folder and a public.png, but no password file)

content/_code-samples/airgapped-wallet/py/README.md
wojake: Right.

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
wojake: image

content/_code-samples/airgapped-wallet/py/README.md
wojake: Yup, my mistake.

content/_code-samples/airgapped-wallet/py/airgapped-wallet-TEST.py
wojake: I see, thank you.

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
wojake: Ye I think that's a good addition, since this'll be used in a dev environment: I'll refer the testnet.

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
wojake: Changes will be made.

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
wojake: Good catch, this is outdated code. `created_wallet()` returned 3 variables and now it's only 2, changes will be made.

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
wojake: You need to run `airgapped-wallet.py` first to set the wallet up. Set your password and scan the QR to obtain the address, then you'd fund the account via a faucet.

I can see where/how this can go wrong, fixing. Good catch (yet again!)

content/_code-samples/airgapped-wallet/py/README.md
wojake: There's no use of XUMM in this code sample, after you've signed a transaction blob on the airgapped machine:
A QR code will pop up and you'd need to scan it, the content is the signed transaction blob.

>When running the steps I just get one that gives me an address as far as I can tell

When you first run `airgapped-wallet.py`, it just sets up the files & wallet credentials. *This ^* QR code contains the wallet's address. You need to re-run `airgapped-wallet.py` to be able to sign transactions and such.

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
wojake: Gotcha, I'll add in another file check (`/Wallet/public.txt`) since it's the last file created during setup.

content/_code-samples/airgapped-wallet/py/airgapped-wallet-TEST.py
JST5000: There are still 3 places to make this change - One in `airgapped-wallet-TEST.py` and two in `airgapped-wallet.py` (Searching for 'OS == "Linux"' should reveal them all)

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
```diff
@@ -0,0 +1,218 @@
...
+        with open(get_path("/Wallet/private.txt"), "w") as f:
+            f.write(priv.hex())
+
+        with open(get_path("/Wallet/public.txt"), "w") as f:
+            f.write(pub)
+
+        openimg = Image.open(get_path("/Wallet/public.png"))
+        openimg.show()
+
+        print("Re-run this script!")
```
JST5000:
```suggestion
        print("Finished generating an account.")
        print("Please scan the QR code on your phone and use https://test.bithomp.com/faucet/ to fund the account.\n After that, re-run this script to use this account and credentials!")
```

JST5000: Adding a status message to let the user know that running it for the first time did accomplish something which will make the 2nd run do something different.

JST5000: (If we want them to re-run it though, should the code just jump into the main loop logic directly after completing the first step?)

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
```diff
@@ -0,0 +1,218 @@
...
+        File = PurePath(str(usr) + '/Wallet')
+        Path_ = str(PurePath(str(usr)))
+
+    # If the Wallet's folder already exists, continue on
+    if os.path.exists(File) and os.path.exists(get_path("/Wallet/public.txt")):
+        while True:
+            ask = int(input("\n 1. Transact XRP"
+                            "\n 2. Generate an XRP wallet (read only)"
+                            "\n 3. Showcase XRP Wallet Address (QR Code)"
+                            "\n\n Enter Index: "
```
JST5000, line 127: Allow for graceful exit (Part 1/2)

```suggestion
                            "\n 4. Exit"
                            "\n\n Enter Index: "
```

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
```diff
@@ -0,0 +1,218 @@
...
+            if ask == 2:
+                _pub, _seed = create_wallet()
+
+            if ask == 3:
+                with open(get_path("/Wallet/public.txt"), "r") as f:
+                    print(f"\n  Wallet Address: {f.read()}")
+
+                image = Image.open(get_path("/Wallet/public.png"))
+                image.show()
+
```
JST5000, line 158: Allow for graceful exit (Part 2/2)
```suggestion
if ask == 4:
       break
```

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
JST5000: Also adding in instructions to fund the account using the bithomp faucet site since that's an important step we need the user to do.

content/_code-samples/airgapped-wallet/py/relay-transaction.py
```diff
@@ -0,0 +1,50 @@
...
+    my_tx_signed = Payment.from_dict(transaction_dict)
+
+    tx = send_reliable_submission(transaction=my_tx_signed, client=client)
+
+    tx_hash = tx.result['hash']
+    tx_destination = tx.result['Destination']
+    tx_xrp_amount = int(tx.result['Amount']) / 1000000
+    tx_account = tx.result['Account']
+
+    print(f"\n           XRPL Explorer: https://livenet.xrpl.org/transactions/{tx_hash}"
```
JST5000: By default we're connecting to testnet - so this should point to that as well. It may be worth adding a variable which can dynamically choose the right network if you change where `client` is pointing to above :)
```suggestion
    print(f"\n           XRPL Explorer: https://testnet.xrpl.org/transactions/{tx_hash}"
```

content/_code-samples/airgapped-wallet/py/README.md
```diff
@@ -0,0 +1,113 @@
...
+3. Import all the modules required by running:
+```
+pip install -r requirements.txt
+```
+
+4. Airgap the machine by following the security practices written [here](#security-practices).
+
+5. Run `airgapped-wallet.py`
+
+6. Scan the QR code and fund the account using the [testnet faucet](https://test.bithomp.com/faucet/)
```
JST5000, line 79:
```suggestion
6. Scan the QR code and fund the account using the [testnet faucet](https://test.bithomp.com/faucet/) 

7. Re-run the script and input '1' to generate a new transaction by following the instructions.

8. Use your phone to scan the QR code, then to send the signed transaction to Machine 2 for submission
```

content/_code-samples/airgapped-wallet/py/README.md
```diff
@@ -0,0 +1,113 @@
...
+3. Import all the modules required by running:
+```
+pip install -r requirements.txt
+```
+
+4. Edit line 47 @ `relay-transaction.py` and insert the signed transaction blob
+
+5. Run `relay-transaction.py`
+
+## Phone Setup
```
JST5000, line 108: I'd recommend removing this section, and just having the instructions be part of Machine 1 setup (Ex. of how that could be done in the above recommendation)

JST5000: That way the tutorial has a linear flow from top to bottom with steps to do.

content/_code-samples/airgapped-wallet/py/README.md
```diff
@@ -0,0 +1,113 @@
...
+**Website**: https://www.python.org/downloads/source/
+
+2. Clone all the files under the [`airgapped-wallet`](https://github.com/XRPLF/xrpl-dev-portal/tree/master/content/_code-samples/airgapped-wallet/py) directory
+
+3. Import all the modules required by running:
+```
+pip install -r requirements.txt
+```
+
+4. Edit line 47 @ `relay-transaction.py` and insert the signed transaction blob
```
JST5000: Calling back so people know where to look if they missed a step.

```suggestion
4. Edit line 47 @ `relay-transaction.py` and insert the signed transaction blob from scanning the QR code Machine 1 generated.
```

content/_code-samples/airgapped-wallet/py/airgapped-wallet.py
```diff
@@ -0,0 +1,218 @@
...
+                            "\n 2. Generate an XRP wallet (read only)"
+                            "\n 3. Showcase XRP Wallet Address (QR Code)"
+                            "\n\n Enter Index: "
+                            ))
+
+            if ask == 1:
+                password = str(input("             Enter Password: "))
+                amount = float(input("\n           Enter XRP To Send: "))
+                destination = input("\n        Enter Destination: ")
+                wallet_sequence = int(input("\n    Enter Wallet Sequence: "))
```
JST5000: Similar to the latest ledger sequence, I realized when going through the steps that this sequence requires looking up the value.
```suggestion
                wallet_sequence = int(input("Look up the 'Next Sequence' for the public account using test.bithomp.com and enter it below!"
                "\n    Enter Wallet Sequence: "))
```

JST5000:
I was able to mostly follow the steps and get it all working :) (The comments above are the missing pieces in the process I had to mentally add, so just trying to make them explicit)

content/_code-samples/airgapped-wallet/py/relay-transaction.py
wojake: Good catch, I'll change that!

content/_code-samples/airgapped-wallet/py/README.md
wojake: I have to disagree here, I'm sure that there will definitely be users who aren't familiar with this.

content/_code-samples/airgapped-wallet/py/relay-transaction.py
```diff
@@ -0,0 +1,50 @@
...
+          f"\n              Wallet Used: {tx_account}"
+         )
+
+
+if __name__ == '__main__':
+    tx_blob = "ENTER TX BLOB HERE"
+    if tx_blob is "ENTER TX BLOB HERE":
+        print("Set tx to 'tx_blob' received from scanning the QR code generated by the airgapped wallet")
+    
+    send_transaction(tx_blob)
```
JST5000: Two suggestions: 
1. is -> ==
2. Don't attempt to use the dummy text with `send_transaction`
```suggestion
    tx_blob = "ENTER TX BLOB HERE"
    if tx_blob == "ENTER TX BLOB HERE":
        print("Set tx to 'tx_blob' received from scanning the QR code generated by the airgapped wallet")
    else:
        send_transaction(tx_blob)
```

Otherwise it crashes saying: 
```
 SyntaxWarning: "is" with a literal. Did you mean "=="?
  if tx_blob is "ENTER TX BLOB HERE":
Set tx to 'tx_blob' received from scanning the QR code generated by the airgapped wallet

   ---   Connected to Node
Traceback (most recent call last):
  File "/Users/jacksonmills/Documents/wojake-xrpl-dev-portal/xrpl-dev-portal/content/_code-samples/airgapped-wallet/py/relay-transaction.py", line 50, in <module>
    send_transaction(tx_blob)
  File "/Users/jacksonmills/Documents/wojake-xrpl-dev-portal/xrpl-dev-portal/content/_code-samples/airgapped-wallet/py/relay-transaction.py", line 28, in send_transaction
    my_tx_signed = Payment.from_dict(transaction_dict)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/xrpl/models/transactions/transaction.py", line 367, in from_dict
    return cast(Transaction, super(Transaction, cls).from_dict(value))
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/xrpl/models/base_model.py", line 92, in from_dict
    raise XRPLModelException(
xrpl.models.exceptions.XRPLModelException: E not a valid parameter for Payment```

File: content/_code-samples/airgapped-wallet/py/airgapped-wallet.py [hunk 33]
```diff
@@ -0,0 +1,226 @@
...
+                            "\n 2. Generate an XRP wallet (read only)"
+                            "\n 3. Showcase XRP Wallet Address (QR Code)"
+                            "\n 4. Exit"
+                            "\n\n Enter Index: "
+                            ))
+
+            if ask == 1:
+                password = str(input("             Enter Password: "))
+                amount = float(input("\n           Enter XRP To Send: "))
+                destination = input("\n        Enter Destination: ")
```
JST5000: Getting a valid destination requires leaving the program and looking things up - for a test though just using the faucet account is good enough so I think this prompt simplifies trying it out.
```suggestion
                destination = input("If you just want to try it out, you can use the faucet account rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe"
                "\n        Enter Destination: ")
```

JST5000:
I asked my teammate for a second review tomorrow :)

JST5000 (approved):
LGTM

PR #1565, opened 2022-11-02
Update and rename content/concepts/consensus-network/amendments/amend…

thanhvu1997:
…ments.md to Nội dung/các khái niệm/mạng lưới đồng thuận/sửa đổi/amendments.md


mDuo13:
This PR looks like it has been opened in error since that's a syntax error.

PR #1575, opened 2022-11-07
fix memos format

jonathanlei:
Fixing memos format, to pass in the correct memos data param name. 
before: 
Screen Shot 2022-11-06 at 5 46 12 PM

after (local):
Screen Shot 2022-11-06 at 5 46 29 PM


ckniffen (approved):

JST5000 (approved):
LGTM

PR #1581, opened 2022-11-14
docsearch3 migration

jonathanlei:
migrate to docsearch3 for better search results - https://docsearch.algolia.com/docs/DocSearch-v3


mDuo13 (changes requested):
The modal is underneath the fixed header in the current preview. This can be fixed with a new CSS rule, e.g.

```css
.DocSearch-Modal {
  top: 85px;
}
```

Also, the center search on the Docs landing page needs to be updated, too. In the preview you can see that it doesn't work at all. The 404 page also has a center search bar you may want to update (but the 404 page doesn't work as intended in the preview link—never has—so you'll probably have to test that locally)

jonathanlei:
> The modal is underneath the fixed header in the current preview. This can be fixed with a new CSS rule, e.g.
> 
> ```css
> .DocSearch-Modal {
>   top: 85px;
> }
> ```
> 
> Also, the center search on the Docs landing page needs to be updated, too. In the preview you can see that it doesn't work at all. The 404 page also has a center search bar you may want to update (but the 404 page doesn't work as intended in the preview link—never has—so you'll probably have to test that locally)

adjusted the spacing and migrated the center search as well (looks still kind of off, need redesign),404 page doesn't work locally either for me

amarantha-k:
@jonathanlei The center search on https://xrplf.github.io/xrpl-dev-portal/pr-preview/docsearch3-migration/docs.html isn't working - don't see any search results. 

Also, is it possible to move the Search button on the top nav to the right? Something like this:
Screen Shot 2022-11-15 at 2 52 17 PM

jonathanlei:
> @jonathanlei The center search on https://xrplf.github.io/xrpl-dev-portal/pr-preview/docsearch3-migration/docs.html isn't working - don't see any search results.
> 
> Also, is it possible to move the Search button on the top nav to the right? Something like this: Screen Shot 2022-11-15 at 2 52 17 PM

don't know why the preview version isn't working, it's working fine on my local build, will investigate. 
In terms of styling, are we getting redesigns? or moving to the right would be fine for now. 
@amarantha-k

jonathanlei:
> > @jonathanlei The center search on https://xrplf.github.io/xrpl-dev-portal/pr-preview/docsearch3-migration/docs.html isn't working - don't see any search results.
> > Also, is it possible to move the Search button on the top nav to the right? Something like this: Screen Shot 2022-11-15 at 2 52 17 PM
> 
> don't know why the preview version isn't working, it's working fine on my local build, will investigate. In terms of styling, are we getting redesigns? or moving to the right would be fine for now. @amarantha-k

now it's working - I think preview takes a while to update from previous builds due to caching. also have the top search aligned to the right

template/page-docs.html.jinja
```diff
@@ -129,12 +129,13 @@
...
-indexName: 'xrpl',
-inputSelector: '#docsearchbox',
-algoliaOptions: { 'facetFilters': ["lang:{{target.lang}}"] },
-debug: false
-});
+  apiKey: '3431349deec23b0bc3dcd3424beb9a6e',
+  indexName: 'xrpl',
+  container: '#centersearchboxcontainer',
+  appId: 'R39QY3MZC7',
+  algoliaOptions: { 'facetFilters': ["lang:en"] },
```
mDuo13: Looks better now, but I noticed that one of my first searches took me to a page that's in the Japanese target (but presented in English because it's not translated yet). I'm not sure if that's a bug on Algolia's side or if there's something wrong with how this option is given. Maybe it's because this is only for the `#centersearchboxcontainer` and not the `#topsearchboxcontainer`?

Also, from the looks of it, this will always search English even if the user is on the Japanese version of the site, which is not ideal. The syntax in the old diff (using `{{target.lang}}`) should still work here.

amarantha-k:
> 

Thanks - the preview looks good with the top search aligned to the right. 
I'm checking with our designer for any concerns; will post updates here.

amarantha-k:
Tagging @bugsbunnies for design input.

bugsbunnies:
Thanks! A few suggestions (some may depend on how flexible the UI component is):

- Search bar in dark mode should adapt to dark mode colors like it does currently, can use current colors for search icon, filler text and search bar bg

- Increase search bar width to 700px wide on the top nav so it's more visible.

image

- When a user clicks on a search bar and it becomes active, can we show results below the search bar as an expanded container instead of in a separate pop up that smokes screens the rest of the page? (Search results container width should match the width of the search box) 

- Can we not show the search bar in body for the Docs landing page? I know that's what exists currently on the live site but I'm not sure why it needs to repeat again when we have a global search bar on the top nav that does the same thing. If we absolutely have to show it, please make the width of the search bar also 700px in the body so it's consistent with top nav.

template/page-docs.html.jinja
jonathanlei: What are you searching? when searching EVM it was giving me Japanese at first too but now the results are in English. Also the preview with "/ja" doesn't really work - can't tell if it's working or not.

jonathanlei:
> Thanks! A few suggestions (some may depend on how flexible the UI component is):
> 
> * Search bar in dark mode should adapt to dark mode colors like it does currently, can use current colors for search icon, filler text and search bar bg
> * Increase search bar width to 700px wide on the top nav so it's more visible.
> 
> image
> 
> * When a user clicks on a search bar and it becomes active, can we show results below the search bar as an expanded container instead of in a separate pop up that smokes screens the rest of the page? (Search results container width should match the width of the search box)
> * Can we not show the search bar in body for the Docs landing page? I know that's what exists currently on the live site but I'm not sure why it needs to repeat again when we have a global search bar on the top nav that does the same thing. If we absolutely have to show it, please make the width of the search bar also 700px in the body so it's consistent with top nav.

1. Trying to edit the default styles - will continue work on that next week 
2. We can get rid of it - but decision up to @mDuo13 @amarantha-k

amarantha-k:
> > Thanks! A few suggestions (some may depend on how flexible the UI component is):
> > 
> > * Search bar in dark mode should adapt to dark mode colors like it does currently, can use current colors for search icon, filler text and search bar bg
> > * Increase search bar width to 700px wide on the top nav so it's more visible.
> > 
> > image
> > 
> > * When a user clicks on a search bar and it becomes active, can we show results below the search bar as an expanded container instead of in a separate pop up that smokes screens the rest of the page? (Search results container width should match the width of the search box)
> > * Can we not show the search bar in body for the Docs landing page? I know that's what exists currently on the live site but I'm not sure why it needs to repeat again when we have a global search bar on the top nav that does the same thing. If we absolutely have to show it, please make the width of the search bar also 700px in the body so it's consistent with top nav.
> 
> 1. Trying to edit the default styles - will continue work on that next week
> 2. We can get rid of it - but decision up to @mDuo13 @amarantha-k

+1 from me to remove the Search bar in the body for the Docs landing page.

bugsbunnies:
Thanks @jonathanlei! 

- I'm seeing in a few different browser sizes that the search dropdown isn't aligned with the search bar, can we make sure they're aligned? 

image

- Also is there a way to search with the top nav and only show results on the expanded view? Right now it feels strange to activate the search bar and then another search bar pops up. I understand there might be some restrictions with algolia but just wanted to see if we can make it as smooth as possible. 

- The mobile view is a bit broken right now because when the search is active, the top nav covers the search area like so 

Screenshot 2022-11-22 at 10 05 56 AM

jonathanlei:
> Thanks @jonathanlei!
> 
> * I'm seeing in a few different browser sizes that the search dropdown isn't aligned with the search bar, can we make sure they're aligned?
> 
> image
> 
> * Also is there a way to search with the top nav and only show results on the expanded view? Right now it feels strange to activate the search bar and then another search bar pops up. I understand there might be some restrictions with algolia but just wanted to see if we can make it as smooth as possible.
> * The mobile view is a bit broken right now because when the search is active, the top nav covers the search area like so
> 
> Screenshot 2022-11-22 at 10 05 56 AM

All fixed now, thought I had the aligning right locally but preview is fixed. I basically have the Search Modal on top of the search bar when clicking search - let me know what you guys think! @bugsbunnies @mDuo13 @amarantha-k

bugsbunnies:
> > Thanks @jonathanlei!
> > 
> > * I'm seeing in a few different browser sizes that the search dropdown isn't aligned with the search bar, can we make sure they're aligned?
> > 
> > image
> > 
> > * Also is there a way to search with the top nav and only show results on the expanded view? Right now it feels strange to activate the search bar and then another search bar pops up. I understand there might be some restrictions with algolia but just wanted to see if we can make it as smooth as possible.
> > * The mobile view is a bit broken right now because when the search is active, the top nav covers the search area like so
> > 
> > Screenshot 2022-11-22 at 10 05 56 AM
> 
> All fixed now, thought I had the aligning right locally but preview is fixed. I basically have the Search Modal on top of the search bar when clicking search - let me know what you guys think! @bugsbunnies @mDuo13 @amarantha-k

Thanks @jonathanlei that's a good way around it. Minor nit pick - if we're to cover the search bar, let's cover it completely because a small part of it is still showing on top, otherwise all good!

image

amarantha-k (approved):
The updated preview looks good! Thanks for incorporating the design input, @jonathanlei.

mDuo13 (approved):
Looks great! However, please hold off on merging for now because I want to get the AMM references out today as well and they are going to have merge conflicts because both update the CSS file.

jonathanlei:
> Looks great! However, please hold off on merging for now because I want to get the AMM references out today as well and they are going to have merge conflicts because both update the CSS file.

feel free to merge it when AMM references are done

PR #1582, opened 2022-11-14
Update XCode link

### LINKED ISSUE #1529: XCode link incorrect
Link located on this page: https://xrpl.org/build-run-rippled-macos.html

Currently is: https://developer.apple.com/download/

Should be: https://developer.apple.com/xcode/

mDuo13:
Fix #1529


oeggert (approved):
Link goes to the correct page.

PR #1583, opened 2022-11-14
Update network_id stanza note.

### LINKED ISSUE #1250: Document network_id
A lot of testnet validations/proposals leak onto the mainnet.

network_id is described in rippled-example.cfg (in the rippled repo) but we should also have mention of it on xrpl.org.

We should add it as a step to https://xrpl.org/connect-your-rippled-to-the-xrp-test-net.html

Testnet nodes should include the following in their config:
```
[network_id]
testnet
```

For sidechains, test networks, and other custom networks, a random 32-bit number should be used. Per OverlayImpl.cpp, only `main`, `testnet`, and `devnet` are valid string options; otherwise it needs to be a uint32, and ideally it's a random number to prevent potential collisions.

[In the future, operators may not need to set this. It could be in the "validators_file" and there should be multiple options: main, testnet, devnet, custom. Or, it should be specified in a VL.]

oeggert:
1. Remove wording of network_id being "optional". Network should be set to prevent wasting resources between servers on different networks.
2. Add note about assigning an unsigned 32-bit integer for sidechains and custom networks.

### DIFF
--- a/content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
+++ b/content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
@@ -82,7 +82,9 @@ Edit your `rippled.cfg` file.
 
     <!-- MULTICODE_BLOCK_END -->
 
-    **Note:** Use a random integer from 0 through 4,294,967,295 for sidechains and custom networks.
+    For sidechains and custom networks, everyone who connects to the network should use a value unique to that network. When creating a new network, choose a network ID at random from the integers 11 to 4,294,967,295.
+
+    **Note:** This setting helps your server find peers who are on the same network, but it is not a hard control on what network your server follows. The UNL / trusted validator settings (in the next step) are what actually define what network the server follows.
 
 ## 2. Set your trusted validator list.
 
--- a/content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
+++ b/content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
@@ -82,7 +82,7 @@ Edit your `rippled.cfg` file.
 
     <!-- MULTICODE_BLOCK_END -->
 
-    **Note:** This setting is optional, and does not strictly define which network your server follows, but it helps servers find peers who are following the same network.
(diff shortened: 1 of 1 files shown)


content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
```diff
@@ -82,7 +82,7 @@ Edit your `rippled.cfg` file.
 
     <!-- MULTICODE_BLOCK_END -->
 
-    **Note:** This setting is optional, and does not strictly define which network your server follows, but it helps servers find peers who are following the same network.
+    **Note:** Use a random integer from 0 through 4,294,967,295 for sidechains and custom networks.
```
mDuo13: I'm afraid people may take this to mean that they should personally roll a random integer when doing this setup. But really, everyone on whatever network it is should be using the same value. Also, they should not use 0, 1, or 2 since those are already allocated to mainnet, testnet, and devnet respectively. (Probably, we should have an XLS that defines how to allocate / claim network IDs so that people don't use the same one, but at least if they're choosing truly at random the chances are very small. Coming up with that standard is certainly out of scope for this PR.) As a precaution I'm going to say new networks should not choose a network ID of 10 or less.

Also, it _is_ important to note that the setting does not strictly define which network your server follows. The UNL/validator settings are what _actually_ defines what network you're following and the `network_id` is just a formality to make peering more efficient. If people misunderstand that, it could lead to bad stuff including, worst case, people losing money. 

Lastly, I guess it's OK to omit this detail, but for posterity: the field is optional in the sense that the server will start up and can operate successfully without it, which may be relevant to know when debugging. (I'm not sure what happens if you make a typo in the field name, but if it's ignored then your server will use a value of 0 which implies Mainnet.)

```suggestion
    For sidechains and custom networks, everyone who connects to the network should use a value unique to that network. When creating a new network, choose a network ID at random from the integers 11 to 4,294,967,295.

    **Note:** This setting helps your server find peers who are on the same network, but it is not a hard control on what network your server follows. The UNL / trusted validator settings (in the next step) are what actually define what network the server follows.
```

mDuo13 (changes requested):
This happens to have some important implications, so I'm being a little extra picky about this change. Sorry for the wall of text!

mDuo13 (approved):

PR #1593, opened 2022-11-18
Fix public RPC URL

amarantha-k:
Change https://rpc.evm-sidechain.xrpl.org/ to https://rpc-evm-sidechain.xrpl.org/


DennisDawson (approved):
Looks good to me.

GrowYourCrypto:
Thank you for this! Could I propose that someone with access updates the docs in GitBook? Ref https://xrpl.org/connect-metamask-to-xrpl-evm-sidechain.html for new devs to seamlessly proceed without needing to dig for the correct RPC.

oeggert:
Good catch. I'll update the doc.
#1657

PR #1598, opened 2022-11-22
Add code samples for markers and pagination.

### LINKED ISSUE #1554: Code sample showing correct pagination techniques
There are a couple tricky little details about [pagination](https://xrpl.org/markers-and-pagination.html) so it would be helpful to have a code sample and maybe a brief tutorial demonstrating the right way to use markers to iterate over a set of results.

Challenges to cover:

- If you do a query without specifying the ledger, and perform several calls over a period of time, the ledger your query searches can update ("shift under you"), causing your full results to be a mix of results from different ledger versions. The preferred pattern for handling this case is to start with querying a ledger by keyword (for example, `"validated"`) in the first request, then save the `ledger_hash` or `ledger_index` from the first response with a marker and using that value (in the appropriate request field) alongside the marker in all subsequent requests.
- The type and contents of the `marker` field are "officially undefined" so that it is _not_ a breaking change if the API changes the inner details of how a marker is formatted. It's not a good idea to, for example, check the length of a string marker or the inner fields of an object marker. It should be sufficient to check for presence or absence of the object. In statically-typed languages like Java, you might also need to know the data-type of the marker (it varies by API method), but that's all.

Suggestions:

- Code sample (preferably multiple languages) showing correct pagination to iterate over a request that needs multiple calls, like [account_tx](https://xrpl.org/account_tx.html) or [ledger_data](https://xrpl.org/ledger_data.html).
    - ledger_data could be a good case, since iterating over an entire ledger is something that various use cases come up with (for example, to take a snapshot of balances for an airdrop or something, or if you want to export a ledger's contents to another type of database for analysis). You do have to look out for the fact that ledgers are especially large, though, so you might run out of memory and crash, or get rate-limited from making too many calls, if you do this the naive way
- Brief tutorial linked from the Markers and Pagination page using the code.
    - You can use `{{ include_code("file", start_with="some string", end_with="other string" }}` to embed code samples directly from a complete source file. This makes it easier to maintain the code without needing to update the excerpts that are in the doc every time you make a minor change. See example usage in [this tutorial](https://raw.githubusercontent.com/XRPLF/xrpl-dev-portal/master/content/tutorials/use-tokens/trade-in-the-decentralized-exchange.md).

oeggert:
Add Python and Javascript code samples to highlight using markers to iterate over multiple calls.


heatherleeann (approved):

heatherleeann (approved):

mDuo13:
The build was failing with this error:

```
Building page: <Page 'Markers and Pagination' (markers-and-pagination.html)>
  File "/usr/lib/python3.10/site-packages/dactyl/page.py", line 218, in preprocess
    md = self.pp_template.render(**context, **self.filter_exports())
  File "/usr/lib/python3.10/site-packages/jinja2/environment.py", line 1291, in render
    self.environment.handle_exception()
  File "/usr/lib/python3.10/site-packages/jinja2/environment.py", line 925, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "content/references/http-websocket-apis/api-conventions/markers-and-pagination.md", line 11, in top-level template code
  File "/usr/lib/python3.10/site-packages/jinja2/utils.py", line 84, in from_obj
    if hasattr(obj, "jinja_pass_arg"):
Preprocessor error in page <Page 'Markers and Pagination' (markers-and-pagination.html)>: 'include_code' is undefined.
```

The one-line fix would be to add the `include_code` filter on the Markers and Pagination page as well; but I've instead added a slightly bigger fix to make that filter available by default for all pages. It wasn't strictly, but for cleanliness, I also removed the lines to enable it on a page-by-page basis from all the pages that were doing so.

content/_code-samples/markers-and-pagination/js/pagination-with-markers.js
```diff
@@ -0,0 +1,42 @@
...
+async function main() {
+
+  // Create a client and connect to the network.
+  const client = new xrpl.Client("wss://xrplcluster.com/")
+  await client.connect()
+
+  // Specify a ledger to query for info.
+  let ledger = await client.request({
+    "command": "ledger_data",
+    "ledger_index": 500000,
```
mDuo13: This is not quite the best practice that we want to showcase in this code sample.

There are three shortcut strings you can use when [specifying ledgers](https://xrpl.org/basic-data-types.html#specifying-ledgers), most notably `"validated"` meaning the latest validated ledger version. Typically when you want to query the ledger, you want to get the latest validated data, so this shortcut string is important.

However, it's possible that a new ledger version can get validated while you're iterating over multiple pages with a marker—the more pages, the more likely this is to happen. If you continue to specify `"validated"` your queries might start paging through the _new_ ledger instead, which might mean you get inconsistent data, or something is skipped or repeated because of the differences between the underlying ledgers. You might also get an error saying that your marker is invalid.

So, the best practice to avoid this problem is to use `"ledger_index": "validated"` on the first request (the one with no marker). Then, the response tells you what ledger version it used, so you can save the `ledger_index` or `ledger_hash` from that response (alongside the marker) and specify that same index or hash on each subsequent request.

content/_code-samples/markers-and-pagination/js/pagination-with-markers.js
```diff
@@ -0,0 +1,42 @@
...
+  // Create a function to run on each API call.
+  function code(){
+    console.log(ledger["result"])
+  }
+ 
+  // Execute function at least once before checking for markers.
+  do {
+    code()
+    
+    if (ledger["result"]["marker"] == null) {
```
mDuo13: As a best practice, avoid `==` in JavaScript which performs implicit data type conversion. I think this code will work as written, but if you switch to `===` then it needs to be `undefined` rather than `null` (since `null` is an actual JSON value that can be specified explicitly, but the API does not do that, it just omits the field).

```suggestion
    if (ledger["result"]["marker"] === undefined) {
```

content/_code-samples/markers-and-pagination/js/pagination-with-markers.js
```diff
@@ -0,0 +1,42 @@
...
+    })
+    ledger = ledger_marker
+
+  } while (true)
+
+  // Disconnect when done. If you omit this, Node.js won't end the process.
+  client.disconnect()
+}
+
+main()
```
mDuo13: Please end each file with a newline

content/_code-samples/markers-and-pagination/py/pagination-with-markers.py
```diff
@@ -0,0 +1,23 @@
+from xrpl.clients import JsonRpcClient
+from xrpl.models.requests import LedgerData
+
+# Create a client to connect to the network.
+client = JsonRpcClient("https://xrplcluster.com/")
+
+# Specify a ledger to query for info.
+ledger = LedgerData(ledger_index=500000)
```
mDuo13: Same as above, the best practice we want to showcase is looking up by `"validated"` the first time and then specifying the ledger_index from the response in each follow-up request.

mDuo13 (changes requested):
This is a good start but the code samples don't _quite_ show the best practice we want them to. I think you'll be able to get there with the information I've supplied below, but feel free to ask me if you need more info.

content/_code-samples/markers-and-pagination/js/pagination-with-markers.js
```diff
@@ -0,0 +1,42 @@
...
+  await client.connect()
+
+  // Specify a ledger to query for info.
+  let ledger = await client.request({
+    "command": "ledger_data",
+    "ledger_index": 500000,
+  })
+
+  // Create a function to run on each API call.
+  function code(){
```
jonathanlei: This could use a better name like `printLedgerResult` for better readability

content/_code-samples/markers-and-pagination/py/pagination-with-markers.py
```diff
@@ -0,0 +1,23 @@
...
+ledger_data = client.request(ledger).result
+
+# Create a function to run on each API call.
+def code():
```
jonathanlei: same thing here

content/_code-samples/markers-and-pagination/js/pagination-with-markers.js
oeggert: My mistake. I meant to do this, but for the sake of seeing the code run from start to finish I used a lower index. Updated code to use most recent validated ledger.

mDuo13 (approved):

PR #1601, opened 2022-11-23
PR build improvements

mDuo13:
- Update `checkout`, `setup-python` actions to latest versions
- Use `head` instead of `cat` so very long link reports don't fail the build with an 'Argument list too long' error

### DIFF
--- a/.github/workflows/link-checker-pr.yml
+++ b/.github/workflows/link-checker-pr.yml
@@ -12,10 +12,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: "3.9"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/link-checker-pr.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       run: |
         sed -i '1,/---------------------------------------/d' linkreport.txt
         echo 'LINKREPORT<<EOF' >> $GITHUB_ENV
-        cat linkreport.txt >> $GITHUB_ENV
+        head -c 65300 linkreport.txt >> $GITHUB_ENV
```
maria-robobug, line 46: Nice! Not used the `head` command before, just out of curiosity, how come we set **65300** specifically here?

maria-robobug (approved):
👍 Just had a question, more for my benefit than anything else! But LGTM.

amarantha-k (approved):
lgtm

.github/workflows/link-checker-pr.yml
mDuo13: Apparently the source of the failure is that the maximum length of a GitHub comment (and/or the `$GITHUB_ENV` variable??) is 65536 (that's 2<sup>16</sup>) characters long. This somewhat arbitrary cutoff of 65300 gives enough headroom for the little bit of extra text the job adds to the comment aside from the link report itself. I based it on [this similar issue](https://github.com/robburger/terraform-pr-commenter/issues/6#issuecomment-826966670).

PR #1612, opened 2022-11-30
Fix getAccount in Quickstart

DennisDawson:


Tomyleexrp:
``

oeggert (approved):
Some of the code needs to be tagged for the syntax highlighting to show up in broker-sale, create-trustline-send-currency, mint-and-burn-nftoken, transfer-nftokens. Otherwise looks good to me.

PR #1613, opened 2022-12-01
Try fixing Japanese search mixup 

jonathanlei:
Currently Japanese page results show up with English titles when searching in English, trying Algolia options to fix it.


mDuo13:
I'm seeing a whole lot of "No results" in the staging env with this fix. Will try locally and see if that works differently.

mDuo13:
Looks like searching from Japanese is still broken?

mDuo13:
Merged as 8a34b1e6078598fa783c5efe16435ff4fa88362e.

PR #1614, opened 2022-12-01
Use DNS names instead of old invalid IPs

legleux:
Might as well use these since they might be less likely to change than the IPs (which are all wrong.)
Once again, someone might want to double check Japanese.


mDuo13 (approved):
Looks good (including Japanese)

PR #1617, opened 2022-12-02
Update unfreeze_token.py

ObiajuluM:
minimal changes

### DIFF
--- a/content/_code-samples/freeze/py/unfreeze_token.py
+++ b/content/_code-samples/freeze/py/unfreeze_token.py
@@ -9,7 +9,7 @@ client = JsonRpcClient("https://s.altnet.rippletest.net:51234") # connect to tes
 token_name = "USD"
 
 # Amount a trustline can handle
-value = "0"
+value = "100"
 print("Generating two test wallets...")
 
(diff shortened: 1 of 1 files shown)


mDuo13:
Why? Per [the tutorial step for this](https://xrpl.org/freeze-a-trust-line.html#7-optional-send-trustset-transaction-to-end-the-freeze):

> The amount of currency you trust this counterparty to issue to you, as a quoted number. As an issuer, this is typically "0".

ObiajuluM:
Ahh, I see. But doesn't this mean that I'll have to set the value parameter in freeze token to '0'? [here](https://xrpl.org/freeze-a-trust-line.html#4-send-trustset-transaction-to-freeze-the-trust-line) 
Screenshot_20221207-035524

mDuo13:
Yes, that's a good point.

ObiajuluM:
If I set the value in either the freeze or the freeze transactions to '0', it will require the other party to reset the trustline to its previous value. It will be stressful.

mDuo13:
I think the core issue is better addressed in a separate, more comprehensive PR. Closing this one since it's stale.

PR #1619, opened 2022-12-02
AMM: Fix trading fee decimal comparison

mDuo13:
"units of 1/100,000" is equivalent to "⅒ basis points" which is correct. This 10,000 number is a carry over from when it was basis points, not ⅒ thereof.

(The AMMCreate page already has it right; the object definition fixed here was a typo.)


maria-robobug (approved):
👍

amarantha-k (approved):

PR #1636, opened 2022-12-13
Fix airgapped wallet tests

mDuo13:
While testing #1609 I found a couple things in the airgapped wallet tests that were failing, apparently since some changes weren't made to match other updates.


jonathanlei (approved):

JST5000 (approved):
LGTM

Issue #1628, opened 2022-12-09
Change NFToken to NFT (and similar) throughout documentation


DennisDawson:
While we used NFToken to be consistent with the API early on, NFT is probably the better choice for SEO considerations.

mDuo13:
Don't forget to update the [term in the contributing guidelines](https://github.com/XRPLF/xrpl-dev-portal/blob/master/CONTRIBUTING.md#terminology) with the new preferred usage as part of this task.

DennisDawson:
Done.

PR #1637, opened 2022-12-14
Clarify NFTokenTaxon

mDuo13:
Lathan in Mattermost called out that this line is confusing. I think this rewrite should clear up the confusion. (Note, I'm using "NFT" to get a head start on #1628.)

The note on the maximum value is not necessary because that's the actual hard max of a UInt32 type anyway.


shortthefomo:
Someone pointed out that xls-24d also has implications here... https://twitter.com/xrpledgernft/status/1603118134688350212?s=46&t=GLBN-Q4Pgf9D41p2FmlRFg

Not sure how that vs this existing field are supposed to inter operate.

mDuo13:
@lathanbritz That's good to know about but i don't think this doc page is the right place to go into XLS-24d details

oeggert (approved):
Looks good to me.

PR #1640, opened 2022-12-16
feat: Osano for cookie management

justinr1234:


mDuo13:
Maybe should wait for the updated privacy policy before merging this?

justinr1234:
> Maybe should wait for the updated privacy policy before merging this?

I think it can go without as it’s turned off in the backend.

intelliot (approved):
I guess this is called Osano (not Osana as the PR title currently says)

justinr1234:
> I guess this is called Osano (not Osana as the PR title currently says)

Fixed, thanks for catching!

PR #1641, opened 2022-12-17
AMM-Devnet connection info

mDuo13:
- Add info about connecting to AMM-Devnet in a few places
- Redo the faucet page to have a radio selector for network rather than having multiple buttons. (With 4+ test networks the number of buttons was getting out of hand)
- Update the "Connect Your rippled to a Test Network" tutorial in a couple extra ways:
    -  Make sure every step with tabbed code samples has tabs for all networks (this helps with the thing where it switches all tabs at the same time for you)
    - Delete the section on looking up the network's ledger index using `--rpc_ip`. The changes from #1614 don't actually work, but the IP addresses there were old. It's possible to look up IP addresses first (using `dig` or `getent` for example) but that's a lot more complex. Telling people to use the WebSocket tool seems like an option that'll scale better. (I thought about inlining some HTML that would load xrpl.js and just tell people the latest ledger index on every network, but that might be a bit unnecessarily wasteful and complex.)
- Rename `test-net.js` to `faucet.js` partly because I think it's a better name and partly because it means browser's won't load the old file from cache. If you get the cached version of the JS with the updated HTML, the "Generate Credentials" button doesn't do anything unless you refresh, but you don't see an error, so this avoids that situation.


content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
```diff
@@ -80,6 +85,11 @@ Edit your `rippled.cfg` file.
         [network_id]
         20
 
+    *AMM-Devnet*
+
+        [network_id]
+        25
```
intelliot, line 91: tangent / just curious: how was this number chosen?

content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
```diff
@@ -151,61 +169,72 @@ Edit your `validators.txt` file. This file is located in the same folder as your
...
+$ sudo systemctl restart rippled
 ```
 
-*Devnet*
+## 5. Verify that your server syncs.
 
-```sh
-rippled --rpc_ip s.devnet.rippletest.net:51234 server_info | grep seq
-```
+It takes about 5 to 15 minutes to sync to the network after a restart. After your server is synced, the [server_info method][] shows the a `validated_ledger` object based on the network you are connected to.
```
intelliot:
```suggestion
It typically takes about 5 to 15 minutes to sync to the network after a restart. After your server is synced, the [server_info method][] shows a `validated_ledger` object based on the network you are connected to.
```

content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
```diff
@@ -151,61 +169,72 @@ Edit your `validators.txt` file. This file is located in the same folder as your
...
 
+To verify that your `rippled` is connected to the right network, compare the results from your server to [a public server][public servers] on the Testnet or Devnet. The `seq` field of the `validated_ledger` object should be the same on both servers (possibly off by one or two, if it changed as you were checking).
```
intelliot:
```suggestion
To confirm that your `rippled` is connected to the right network, compare the results from your server to [a public server][public servers] on the Testnet or Devnet. The `seq` field of the `validated_ledger` object should be the same on both servers (possibly off by one or two, if it changed as you were checking).
```

`verify` seems too strong a word, since we're just looking at the `seq` number. If two new ledgers are launched at about the same time, or if there is a fork/divergence in a network, then this test could mislead.

content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
```diff
@@ -151,61 +169,72 @@ Edit your `validators.txt` file. This file is located in the same folder as your
...
 
 ```sh
-rippled --rpc_ip s1.ripple.com:51234 server_info | grep seq
+rippled server_info | grep seq
 ```
 
-
-<!-- MULTICODE_BLOCK_END -->
-
+You can use [server_info in the WebSocket Tool](websocket-api-tool.html#server_info) to look up the latest ledger index on the intended network.
```
intelliot:
```suggestion
You can use [server_info in the WebSocket Tool](websocket-api-tool.html#server_info) to look up the latest ledger index (`seq`) on the intended network.
```

intelliot (approved):

content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
mDuo13: I have no idea. I think the NFT-Devnet used 20 because of XLS-20d, but the AMM proposal is XLS-30d so that doesn't apply here.

content/tutorials/manage-the-rippled-server/configuration/connect-your-rippled-to-the-xrp-test-net.md
mDuo13: I think "typically" is excessive here since it already has "about".

```suggestion
It takes about 5 to 15 minutes to sync to the network after a restart. After your server is synced, the [server_info method][] shows a `validated_ledger` object based on the network you are connected to.
```

PR #1658, opened 2022-12-23
Remove XRP charts page

oeggert:
Ignore this PR; page shouldn't be removed yet.


oeggert:
Deprecation was recent; keep page a while longer before removing.

PR #1673, opened 2023-01-12
Added tx-serialization JavaScript examples analog to python version

AlexanderBuzz:
I've added the [Transaction Serialization](https://github.com/XRPLBounties/Proposals/blob/main/closed_bounties/0031%20JavaScript%20Code%20Samples.md) Milestone 2 from the JavaScript code samples bounty proposal. It's inspired by the existing python version and delivers the same output.


content/_code-samples/tx-serialization/js/test-cases/README.md
```diff
@@ -0,0 +1,52 @@
+# Transaction Serialization Test Cases
+
+This folder contains several transactions in their JSON and binary forms, which
+you can use to verify the behavior of transaction serialization code.
+
+For example (starting from the `tx-serialization/` dir above this one):
+
+```bash
+$ python3 serialize.py -f test-cases/tx2.json | \
```
jonathanlei: Please change this to JS commands

jonathanlei (changes requested):
other than the test file readme everything looks good to me, will run it again locally once you change the tests. Thanks!

content/_code-samples/tx-serialization/js/test-cases/README.md
AlexanderBuzz: Ok, it's done, I've overlooked this one

jonathanlei (approved):
LGTM, test ran successfully. waiting on a double check from @mDuo13. btw, please don't remove reviews requested by us, as per bounty submission needs 2+ reviews to be merged. thanks!

content/_code-samples/tx-serialization/js/test-cases/README.md
```diff
@@ -0,0 +1,52 @@
+# Transaction Serialization Test Cases
+
+This folder contains several transactions in their JSON and binary forms, which
+you can use to verify the behavior of transaction serialization code.
```
JST5000: nit:
```suggestion
you can use to verify the behavior of the transaction serialization code.
```

content/_code-samples/tx-serialization/js/helpers.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
```
JST5000: Since these helpers are only used in one file, I'd recommend just including them in the `index.js` file.

content/_code-samples/tx-serialization/js/index.js
```diff
@@ -0,0 +1,687 @@
...
+                logger(fieldName + ": " + fieldBytes)
+                fieldsAsBytes.push(fieldBytes)
+            }
+        }
+
+        return fieldsAsBytes.join('')
+    }
+}
+
+// Startup stuff begin
```
JST5000: I'd recommend moving the class definition into a separate file, then importing it here with a comment explaining the serializer details are in the imported file. 

That way you can break up the large file into running implementation vs serializer implementation.

content/_code-samples/tx-serialization/js/index.js
```diff
@@ -0,0 +1,687 @@
...
+     * @param fieldName
+     * @param fieldValue
+     * @returns {string}
+     */
+    fieldToBytes(fieldName, fieldValue) {
+        const fieldType = this.definitions["FIELDS"][fieldName]["type"]
+        logger("Serializing field " + fieldName + " of type " + fieldType)
+
+        const idPrefix = this.fieldId(fieldName)
+        logger("IdPrefix is: " + idPrefix)
```
JST5000: nit:
```suggestion
        logger("ID Prefix is: " + idPrefix)
```

JST5000:
As part of this PR I'd recommend changing the JS example on https://xrpl.org/serialization.html#serialization-format to point to these code snippets instead of the `ripple-binary-codec` example.

content/_code-samples/tx-serialization/js/index.js
```diff
@@ -0,0 +1,687 @@
...
+    objectToBytes(object) {
+        const childOrder = Object.keys(object).sort(sortFuncCanonical.bind(this))
+
+        let fieldsAsBytes = [];
+
+        for (const fieldName of childOrder) {
+            if (this.definitions["FIELDS"][fieldName]["isSerialized"]) {
+                const fieldValue = object[fieldName]
+                const fieldBytes = this.fieldToBytes(fieldName, fieldValue)
+                logger(fieldName + ": " + fieldBytes)
```
JST5000: I'd recommend changing this to skip the `ID Prefix` values when logging in verbose mode if possible because currently looks like the ID Prefix should appear twice, when in reality it only appears once. 

Ex. (Current)
```
Serializing field OfferSequence of type UInt32
IdPrefix is: 2019
OfferSequence: 2019001abed7
```

vs

Recommended:
```
Serializing field OfferSequence of type UInt32
IdPrefix is: 2019
OfferSequence: 001abed7
```

(The number of bytes can be a variable amount of bytes (2-4 is common) - so it may require another helper, or getting more information in the return type of `this.fieldToBytes`)

JST5000:
Works as expected on my computer :) - Just a couple comments to improve readability for new people using it. Please leave a comment that mentions my name (ex. @JST5000 ) after you've made revisions/responded to comments and I'll take another look!

content/_code-samples/tx-serialization/js/index.js
```diff
@@ -0,0 +1,687 @@
...
+
+    /**
+     *     Helper function; serializes a hash value from a hexadecimal string
+     *     of any length.
+     *
+     * @param contents
+     * @returns {string}
+     */
+    hashToBytes(contents) {
+        return Buffer.from(contents).toString("hex")
```
ckniffen: Not a big deal but there is an edge case where if the contents is empty it should be zero bytes.

See https://github.com/XRPLF/xrpl.js/pull/2156/files

AlexanderBuzz:
Good, I will commit the suggestions and then restructure the whole thing as suggested

content/_code-samples/tx-serialization/js/index.js
AlexanderBuzz: In now handle the logging in fieldToBytes function, output is as expected

content/_code-samples/tx-serialization/js/index.js
AlexanderBuzz: @ckniffen Hi Caleb, just for clarification: The expected behavior is that an empty hash should NOT return a byte string of zeros the appropriate length but an empty string so that a field can be unset - is this correct? "Zero bytes" is a homonym here :)

content/_code-samples/tx-serialization/js/index.js
AlexanderBuzz: @JST5000 Done, output now corresponds to what you recommended

content/_code-samples/tx-serialization/js/index.js
AlexanderBuzz: So, if I get this right, an empty input should return an zero-byte padded string of hash-type-length, correct?

AlexanderBuzz:
Somehow I removed @mDuo13 from review...

JST5000:
> Somehow I removed @mDuo13 from review...

I think you can only request one or two extra reviewers if you're not a maintainer of the library. Not sure exactly where that setting is, but I think the intent of that feature is to avoid someone spamming all maintainers whenever they comment.

JST5000:
Two more recommendations:
1. I'd recommend capitalizing the serialized sub-components as you log them to match the eventual output: 
Ex. 
`Fee: 400000000000000a` -> `Fee: 400000000000000A`

2. I'd recommend printing out the transaction you're serializing at the top, then show the serialized one below, with a couple lines explaining what's happening. (This will affect your tests)

The reason I'd recommend that is currently if someone just does `node index.js` it's quite hard to tell what's happening, because it just shows the serialized output without any context. 

But if you add a couple print statements for context they can figure out what happened from just the console logs. 

Ex. 

```
Serializing this transaction: 
{
  "Account": "rMBzp8CgpE441cp5PVyA9rpVV7oT8hP3ys",
  "Expiration": 595640108,
  "Fee": "10",
  "Flags": 524288,
  "OfferSequence": 1752791,
  "Sequence": 1752792,
  "SigningPubKey": "03EE83BB432547885C219634A1BC407A9DB0474145D69737D09CCDC63E1DEE7FE3",
  "TakerGets": "15000000000",
  "TakerPays": {
    "currency": "USD",
    "issuer": "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B",
    "value": "7072.8"
  },
  "TransactionType": "OfferCreate",
  "TxnSignature": "30440220143759437C04F7B61F012563AFE90D8DAFC46E86035E1D965A9CED282C97D4CE02204CFD241E86F17E011298FC1A39B63386C74306A5DE047E213B0F29EFA4571C2C",
  "hash": "73734B611DDA23D3F5F62E20A173B78AB8406AC5015094DA53F53D39B9EDB06C"
}

Serialized Transaction: 120007220008000024001ABED82A2380BF2C2019001ABED764D55920AC9391400000000000000000000000000055534400000000000A20B3C85F482532A9578DBB3950B85CA06594D165400000037E11D60068400000000000000A732103EE83BB432547885C219634A1BC407A9DB0474145D69737D09CCDC63E1DEE7FE3744630440220143759437C04F7B61F012563AFE90D8DAFC46E86035E1D965A9CED282C97D4CE02204CFD241E86F17E011298FC1A39B63386C74306A5DE047E213B0F29EFA4571C2C8114DD76483FACDEE26E60D8A586BB58D09F27045C46
```

content/_code-samples/tx-serialization/js/index.js
ckniffen: Yeah.  The only case this needs to happen is for clearing an Email Hash.  So not the most important thing to demo.  Might be something to check in your PHP library.

AlexanderBuzz: > Two more recommendations:
> 
> 1. I'd recommend capitalizing the serialized sub-components as you log them to match the eventual output:
>    Ex.
>    `Fee: 400000000000000a` -> `Fee: 400000000000000A`
> 2. I'd recommend printing out the transaction you're serializing at the top, then show the serialized one below, with a couple lines explaining what's happening. (This will affect your tests)
> 
> The reason I'd recommend that is currently if someone just does `node index.js` it's quite hard to tell what's happening, because it just shows the serialized output without any context.
> 
> But if you add a couple print statements for context they can figure out what happened from just the console logs.
> 
> Ex.
> 
> ```
> Serializing this transaction: 
> {
>   "Account": "rMBzp8CgpE441cp5PVyA9rpVV7oT8hP3ys",
>   "Expiration": 595640108,
>   "Fee": "10",
>   "Flags": 524288,
>   "OfferSequence": 1752791,
>   "Sequence": 1752792,
>   "SigningPubKey": "03EE83BB432547885C219634A1BC407A9DB0474145D69737D09CCDC63E1DEE7FE3",
>   "TakerGets": "15000000000",
>   "TakerPays": {
>     "currency": "USD",
>     "issuer": "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B",
>     "value": "7072.8"
>   },
>   "TransactionType": "OfferCreate",
>   "TxnSignature": "30440220143759437C04F7B61F012563AFE90D8DAFC46E86035E1D965A9CED282C97D4CE02204CFD241E86F17E011298FC1A39B63386C74306A5DE047E213B0F29EFA4571C2C",
>   "hash": "73734B611DDA23D3F5F62E20A173B78AB8406AC5015094DA53F53D39B9EDB06C"
> }
> 
> Serialized Transaction: 120007220008000024001ABED82A2380BF2C2019001ABED764D55920AC9391400000000000000000000000000055534400000000000A20B3C85F482532A9578DBB3950B85CA06594D165400000037E11D60068400000000000000A732103EE83BB432547885C219634A1BC407A9DB0474145D69737D09CCDC63E1DEE7FE3744630440220143759437C04F7B61F012563AFE90D8DAFC46E86035E1D965A9CED282C97D4CE02204CFD241E86F17E011298FC1A39B63386C74306A5DE047E213B0F29EFA4571C2C8114DD76483FACDEE26E60D8A586BB58D09F27045C46
> ```

No problem, I just stuck to the py version output 1:1. So I will improve that one so it looks good and reads well, and if it gets approved perhaps I should do a quick update/PR for the py version also to have it reflect it.

ckniffen (approved):

AlexanderBuzz:
Output now looks nice and clean. @ckniffen the edge case has been added an can be tested with `node index.js -v -j "{\"EmailHash\":\"000000000000000000000000000000\"}"`

content/_code-samples/tx-serialization/js/test-cases/README.md
```diff
@@ -0,0 +1,52 @@
+# Transaction Serialization Test Cases
+
+This folder contains several transactions in their JSON and binary forms, which
+you can use to verify the behavior of the transaction serialization code.
+
+For example (starting from the `tx-serialization/js/` dir above this one):
+
+```bash
+$ node index.js -f test-cases/tx2.json | \
+  diff - test-cases/tx2-binary.txt
```
JST5000, line 10: Unfortunately this instruction is out of date now because the pretty print. Could you update the expected result to match the pretty print version? (Maybe renaming the files to `tx2-expected.txt` would better convey what they contain if you do)

JST5000:
The pretty print looks awesome :) 🥇

content/_code-samples/tx-serialization/js/test-cases/README.md
AlexanderBuzz: I did it by adding a --raw CLI parameter for such cases, it's reflected in the README.md

PR #1675, opened 2023-01-13
Add nft_page option

DennisDawson:
Add `ledger_entry - NFT Page` option to the WebSocket tool page.


DennisDawson:
Addresses https://github.com/orgs/XRPLF/projects/4/views/1?pane=issue&itemId=16202152

assets/js/apitool-methods-ws.js
```diff
@@ -211,6 +211,17 @@ Request('ledger_entry - DirectoryNode', {
   }
 })
 
+Request('ledger_entry - NFT Page', {
+  description: "Returns an NFT Page object in its raw ledger format.",
+  link: "ledger_entry.html#get-offer-object",
```
mDuo13: This link goes to the wrong example. I think you can leave out the `link:` attribute entirely, but it would be better to extend the [ledger_entry docs](https://xrpl.org/ledger_entry.html) with the format for this method and then link the new section.

assets/js/apitool-methods-ws.js
```diff
@@ -211,6 +211,17 @@ Request('ledger_entry - DirectoryNode', {
...
+  body: {
+    "id": "example_get_accountroot",
```
mDuo13: The ID is an arbitrary string, but a more appropriate one for this example would be:

```suggestion
    "id": "example_get_nft_page",
```

assets/js/apitool-methods-ws.js
```diff
@@ -211,6 +211,17 @@ Request('ledger_entry - DirectoryNode', {
...
+    "command": "ledger_entry",
+    "nft_page": "255DD86DDF59D778081A06D02701E9B2C9F4F01DFFFFFFFFFFFFFFFFFFFFFFFF",
```
mDuo13, line 240: Confirmed that this ID does in fact refer to an NFTokenPage object on Mainnet. Not sure whose it is, so hopefully it sticks around, but it's great that the example works out of the box. :+1:

mDuo13 (dismissed):

assets/js/apitool-methods-ws.js
amarantha-k: Agree - would be good to update ledger_entry docs with the format for this method.

assets/js/apitool-methods-ws.js
```diff
@@ -211,6 +211,17 @@ Request('ledger_entry - DirectoryNode', {
   }
 })
 
+Request('ledger_entry - NFT Page', {
+  description: "Returns an NFT Page object in its raw ledger format.",
+  link: "ledger_entry.html#get_nft_page",
```
mDuo13: The link checker doesn't check the WS tool page because it's not smart enough to run all the JavaScript first, but I think this link is broken. Based on the changes to `ledger_entry.md`, the correct link is:

```suggestion
  link: "ledger_entry.html#get-nft-page",
```

mDuo13:
That build failure is weird. It doesn't seem to be coming from one of the files changed in this PR and it wasn't happening on the earlier build. I wonder if it's a momentary glitch of some kind.

mDuo13 (approved):

PR #1685, opened 2023-01-20
Python/JS code sample for Memo

wojake:
Validate and send a [Memo](https://xrpl.org/transaction-common-fields.html#memos-field).

This is work for a XRPL Bounty @ [XRPLBounties/Proposals](https://github.com/XRPLBounties/Proposals/blob/main/open_bounties/0079%20Python%20Code%20Samples%202.md) for Milestone 7 and [XRPLBounties/Proposals](https://github.com/XRPLBounties/Proposals/blob/main/open_bounties/0077%20JavaScript%20Code%20Samples%202.md) for Milestone 10

### DIFF
--- a/content/_code-samples/send-a-memo/js/send-a-memo.js
+++ b/content/_code-samples/send-a-memo/js/send-a-memo.js
@@ -41,6 +41,7 @@ if (typeof module !== "undefined") {
     })
 
     const signed = wallet.sign(prepared)  
+    print("Submitting a payment transaction with our memo field...")
     const submit_result = await client.submitAndWait(signed.tx_blob)
 xrpl.convertHexToString
     const tx_MemoData = xrpl.convertHexToString(string=submit_result.result.Memos[0].Memo.MemoData);
--- a/content/_code-samples/send-a-memo/js/send-a-memo.js
+++ b/content/_code-samples/send-a-memo/js/send-a-memo.js
@@ -23,7 +23,8 @@ if (typeof module !== "undefined") {
     // Enter memo data to insert into a transaction
     const MemoData = xrpl.convertStringToHex(string="Example Memo - 123 -=+");
     const MemoType = xrpl.convertStringToHex(string="Text");
-    const MemoFormat = xrpl.convertStringToHex(string="text/csv");
+    // MemoFormat values: # MemoFormat values: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
+    const MemoFormat = xrpl.convertStringToHex(string="text/plain");
 
     // Send AccountSet transaction 
     const prepared = await client.autofill({
--- a/content/_code-samples/send-a-memo/js/send-a-memo.js
+++ b/content/_code-samples/send-a-memo/js/send-a-memo.js
@@ -19,9 +19,9 @@ if (typeof module !== "undefined") {
     console.log("   Seed: ", wallet.seed)
 
     // Enter memo data to insert into a transaction
-    const MemoData = Buffer.from("TEXT 123 -+=", 'utf8').toString('hex')
-    const MemoType = Buffer.from("Text", 'utf8').toString('hex')
-    const MemoFormat = Buffer.from("text/csv", 'utf8').toString('hex')
+    const MemoData = xrpl.convertStringToHex(string="TEXT 123 -=+");
+    const MemoType = xrpl.convertStringToHex(string="Text");
+    const MemoFormat = xrpl.convertStringToHex(string="text/csv");
 
     // Send AccountSet transaction 
     const prepared = await client.autofill({
@@ -40,10 +40,10 @@ if (typeof module !== "undefined") {
 
     const signed = wallet.sign(prepared)  
     const submit_result = await client.submitAndWait(signed.tx_blob)
-
-    const tx_MemoData = Buffer.from(submit_result.result.Memos[0].Memo.MemoData, 'hex').toString('utf8');
-    const tx_MemoFormat = Buffer.from(submit_result.result.Memos[0].Memo.MemoFormat, 'hex').toString('utf8');
-    const tx_MemoType = Buffer.from(submit_result.result.Memos[0].Memo.MemoType, 'hex').toString('utf8');
+xrpl.convertHexToString
+    const tx_MemoData = xrpl.convertHexToString(string=submit_result.result.Memos[0].Memo.MemoData);
+    const tx_MemoFormat = xrpl.convertHexToString(string=submit_result.result.Memos[0].Memo.MemoFormat);
+    const tx_MemoType = xrpl.convertHexToString(string=submit_result.result.Memos[0].Memo.MemoType);
 
     console.log(`\n Encoded Transaction MEMO: ${JSON.stringify({"MemoType": MemoType, "MemoData": MemoData, "MemoFormat": MemoFormat})}`)
     console.log(` Decoded Transaction MEMO: ${JSON.stringify({"MemoType": tx_MemoType, "MemoData": tx_MemoData, "MemoFormat": tx_MemoFormat})}`);
--- a/content/_code-samples/send-a-memo/js/send-a-memo.js
+++ b/content/_code-samples/send-a-memo/js/send-a-memo.js
@@ -1,0 +1,50 @@
+if (typeof module !== "undefined") {
+    // Use var here because const/let are block-scoped to the if statement.
+    var xrpl = require('xrpl')
+  }
+  // Validate and send a Memo
+  // https://xrpl.org/transaction-common-fields.html#memos-field
+  
+  async function main() {
+    // Connect to a testnet node
+    console.log("Connecting to Testnet...")
+    const client = new xrpl.Client('wss://s.altnet.rippletest.net:51233')
+    await client.connect()
+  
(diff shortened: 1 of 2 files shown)


content/_code-samples/send-a-memo/js/send-a-memo.js
```diff
@@ -0,0 +1,53 @@
...
+            }
+        }]
+    })
+
+    const signed = wallet.sign(prepared)  
+    const submit_result = await client.submitAndWait(signed.tx_blob)
+
+    console.log(`\n Transaction MEMO: ${JSON.stringify({"MemoType": MemoType, "MemoData": MemoData, "MemoFormat": MemoFormat})}`)
+    console.log("\n Transaction hash:", signed.hash)
+    console.log("    Submit result:", submit_result.result.meta.TransactionResult)
```
ckniffen, line 55: Can you add a log that shows that the memo was encoded properly by showing the memo decoded derived from `submit_result`?

content/_code-samples/send-a-memo/js/send-a-memo.js
```diff
@@ -0,0 +1,58 @@
...
+    console.log("Requesting addresses from the Testnet faucet...")
+    const { wallet, balance } = await client.fundWallet()
+    
+    console.log("Account: ", wallet.address)
+    console.log("   Seed: ", wallet.seed)
+
+    // Enter memo data to insert into a transaction
+    const MemoData = Buffer.from("TEXT 123 -+=", 'utf8').toString('hex')
+    const MemoType = Buffer.from("Text", 'utf8').toString('hex')
+    const MemoFormat = Buffer.from("text/csv", 'utf8').toString('hex')
```
khancode: There's a utility function for doing string to hex conversions in xrpl.js called `convertStringToHex()` - [source](https://github.com/XRPLF/xrpl.js/blob/76b73e16a97e1a371261b462ee1a24f1c01dbb0c/packages/xrpl/src/utils/stringConversion.ts#L8-L10)

khancode (changes requested):

content/_code-samples/send-a-memo/js/send-a-memo.js
wojake: Thanks!

content/_code-samples/send-a-memo/js/send-a-memo.js
```diff
@@ -0,0 +1,58 @@
+if (typeof module !== "undefined") {
+    // Use var here because const/let are block-scoped to the if statement.
+    var xrpl = require('xrpl')
+  }
```
JST5000: nit: Spacing to line up with the if-statement.
```suggestion
}
```

content/_code-samples/send-a-memo/js/send-a-memo.js
```diff
@@ -0,0 +1,58 @@
...
+  // https://xrpl.org/transaction-common-fields.html#memos-field
+  
+  async function main() {
+    // Connect to a testnet node
+    console.log("Connecting to Testnet...")
+    const client = new xrpl.Client('wss://s.altnet.rippletest.net:51233')
+    await client.connect()
+  
+    // Get account credentials from the Testnet Faucet
+    console.log("Requesting addresses from the Testnet faucet...")
```
JST5000:
```suggestion
    console.log("Requesting an account from the Testnet faucet...")
```

content/_code-samples/send-a-memo/js/send-a-memo.js
```diff
@@ -0,0 +1,58 @@
+if (typeof module !== "undefined") {
+    // Use var here because const/let are block-scoped to the if statement.
+    var xrpl = require('xrpl')
+  }
+  // Validate and send a Memo
```
JST5000:
```suggestion
  // Encode and send a Memo. 
  // In general, Memo's can be added to any transaction, and are sometimes used to
  // share extra information needed for a 3rd party integration.
```

content/_code-samples/send-a-memo/js/send-a-memo.js
```diff
@@ -0,0 +1,58 @@
...
+  
+    // Get account credentials from the Testnet Faucet
+    console.log("Requesting addresses from the Testnet faucet...")
+    const { wallet, balance } = await client.fundWallet()
+    
+    console.log("Account: ", wallet.address)
+    console.log("   Seed: ", wallet.seed)
+
+    // Enter memo data to insert into a transaction
+    const MemoData = xrpl.convertStringToHex(string="TEXT 123 -=+");
```
JST5000: I'd recommend avoiding the word `TEXT` in the actual memo text because it may be confused with the keyword `Text` used below in the `MemoType` object. (Either because they think that you need it, or causing people to think that the below `Text` is also arbitrary)
```suggestion
    const MemoData = xrpl.convertStringToHex(string="Example Memo - 123 -=+");
```

JST5000:
Not sure if there's a good spot to put this information, but Memos are only viewable if you ask a node that "remembers" the ledger where the transaction was submitted. So, full-history nodes will always remember, and for recent transactions most validators will be able to look up memos with the `tx` command, but they don't get stored in the blockchain ledger itself the way `LedgerObjects` are for example. (So, you may have to query full history nodes if you want to do something like search for transactions with a certain memo)

content/_code-samples/send-a-memo/js/send-a-memo.js
```diff
@@ -0,0 +1,58 @@
...
+      "Memos": [{
+            "Memo":{
+                "MemoType": MemoType,
+                "MemoData": MemoData,
+                "MemoFormat": MemoFormat
+            }
+        }]
+    })
+
+    const signed = wallet.sign(prepared)
```
JST5000, line 44:
```suggestion
    const signed = wallet.sign(prepared)  
    print("Submitting a payment transaction with our memo field...")
```

content/_code-samples/send-a-memo/py/send-a-memo.py
```diff
@@ -0,0 +1,61 @@
...
+# This code sample validates and sends a transaction with a Memo attached
+# A funded account on the testnet is provided for testing purposes
+# https://xrpl.org/transaction-common-fields.html#memos-field
+
+# Connect to a testnet node
+JSON_RPC_URL = "https://s.altnet.rippletest.net:51234/"
+client = JsonRpcClient(JSON_RPC_URL)
+
+# Get account credentials from the Testnet Faucet
+print("Requesting account from the Testnet faucet...")
```
JST5000:
```suggestion
print("Requesting an account from the Testnet faucet...")
```

content/_code-samples/send-a-memo/py/send-a-memo.py
```diff
@@ -0,0 +1,61 @@
...
+test_wallet = generate_faucet_wallet(client=client)
+myAddr = test_wallet.classic_address
+
+# Example: TEXT 321 -+=
+memo_data = input("Enter memo data to insert into a transaction: ")
+# Example: text
+memo_type = input("Enter memo type to insert into a transaction: ")
```
JST5000: This field doesn't make as much sense to allow the user to input. (I'm at least not sure what options make sense to put in there) - While it does technically allow arbitrary input, I'd recommend either setting this for the user to `Text` to keep it simple, or to add a print statement explaining what this field is used for - Ex. 
print("Memo Type is used to describe the type of data you're sharing. Common types include 'Text', '...', and '...' although in practice you can put anything in there, as it's only used by your decoding code.")

content/_code-samples/send-a-memo/py/send-a-memo.py
```diff
@@ -0,0 +1,61 @@
...
+        Memo(
+            memo_type=memo_type,
+            memo_data=memo_data,
+            memo_format=memo_format
+        ),
+    ],
+)
+
+# Sign the transaction
+payment_tx_signed = safe_sign_and_autofill_transaction(payment_tx, wallet=test_wallet, client=client)
```
JST5000, line 45:
```suggestion
print("Submitting a payment transaction with our memo...")
payment_tx_signed = safe_sign_and_autofill_transaction(payment_tx, wallet=test_wallet, client=client)
```

content/_code-samples/send-a-memo/py/send-a-memo.py
```diff
@@ -0,0 +1,61 @@
...
+myAddr = test_wallet.classic_address
+
+# Example: TEXT 321 -+=
+memo_data = input("Enter memo data to insert into a transaction: ")
+# Example: text
+memo_type = input("Enter memo type to insert into a transaction: ")
+
+memo_data = memo_data.encode('utf-8').hex()
+memo_type = memo_type.encode('utf-8').hex()
+memo_format = "text/csv".encode('utf-8').hex()
```
JST5000: Might be worth linking to where these `MemoFormat` values can be found. Applies to both files. (I think this is one site you could link? - https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)

content/_code-samples/send-a-memo/py/send-a-memo.py
wojake: Ye I agree, I'll set it to use a variable as the Memo

ckniffen (approved):

JST5000 (approved):

PR #1692, opened 2023-01-20
Add note re download zip

DennisDawson:
Added notes to the first three files to emphasize that the customer needs to download the Zip file in order to run the samples. Addresses https://github.com/XRPLF/xrpl-dev-portal/issues/1443.


mDuo13:
This is just an idea, but one other thing you could do is make the download link a button so it's more prominent. Dactyl should turn any link whose description ends in > into a button, so you could do like this:

```
[Download Quickstart Sample Code >](https://github.com/XRPLF/xrpl-dev-portal/tree/master/content/_code-samples/quickstart/js/quickstart.zip)

intelliot (approved):

Issue #1651, opened 2022-12-20
Update Ubuntu build instructions


mDuo13:
Just recently the steps needed to [build rippled on Ubuntu](https://xrpl.org/build-run-rippled-ubuntu.html) changed (starting with dev versions of 1.10.0) to use Conan to manage packages. This means changes to the build instructions.

Upstream PR: https://github.com/XRPLF/rippled/pull/4367
Related: #1650, #1648.

Add any new dependencies and modify the steps as necessary to build on supported (current) versions of Ubuntu.

## Aside: future of build instructions

It has been previously discussed, and might still be a good idea, to remove the separate build instructions on the xrpl website and only link [the instructions in the source repo](https://github.com/XRPLF/rippled/tree/develop/Builds), since the method of building is specific to the release version. The repo also has build instructions for more platforms. We would like to first make sure the build instructions in the repo are clear and easy enough to follow, though, which might mean making PRs against the `rippled` repo to update them there.

There's also the consideration that the website has Japanese-translated build instructions, which aren't kept _perfectly_ up-to-date with the English but are probably better than nothing for native Japanese speakers; so we might want to port those to the upstream repo as well, but the team may not want to be responsible for those since they it would probably require someone fluent in Japanese to keep things updated as well.

thejohnfreeman:
I prefer to link to the new build instructions in the root directory of the project, in the conventionally named file [`BUILD.md`](https://github.com/XRPLF/rippled/blob/develop/BUILD.md).

I would like to collaborate on fixing up the build instructions in the repository:

- I believe that, under the directory `Builds/`, the directories `VisualStudio2019`, `linux`, and `macos` can be removed. In fact, I believe they contain currently incorrect documentation, on top of being elaborate and confusing. I suspect that `Test.py` and `build_all.sh` are unused and can be removed.
- We should move and update documentation on the project options from `Builds/linux/README.md` to `BUILD.md`.
- I'd like a review from technical writers on the structure of `BUILD.md`. We're trying to serve a few different audiences:
  - Maintainers who are responsible for keeping these instructions up-to-date. It's better to have them all in one file, instead of just-slightly-different-but-mostly-identical instructions in separate files and directories for each platform.
  - Experienced developers who just need to reference the commands, and maybe copy them.
  - New developers who need a walk through, including an introduction to Conan and CMake.

[...]

PR #1695, opened 2023-01-23
Remove outdated build instructions

thejohnfreeman:
Addresses #1651. The first commit in this PR removes all build instructions, indiscriminately, without adding any links to the build instructions in the rippled project. It was suggested that I submit this anyway to start the discussion.


intelliot:
These instructions are obsolete; it won't hurt anyone to remove them. They'll always be available in the commit history, regardless (in addition to archive.org presumably).

mDuo13:
These instructions are still accurate for building the latest stable release (1.9.4) from source, aren't they? I'm hesitant to replace the build instructions completely until we have a stable release based on Conan.

Also, you need to update `dactyl-config.yml`, or the build will fail (as it currently is) due to missing files. We prefer to leave redirects in place rather than taking URLs down entirely.

oeggert:
I agree that we should hold off on deleting the pages from xrplf if the instructions are valid. I can help update BUILD.md from the linked issue and then set up redirects to it once that goes out with a stable release.

oeggert:
Closing this PR. I've got another open that removes the old instructions and adds redirects. #1789

PR #1697, opened 2023-01-24
JS code sample to list account tokens & objects

wojake:
Show tokens and other objects owned by an account.

This is work for a XRPL Bounty @ [XRPLBounties/Proposals](https://github.com/XRPLBounties/Proposals/blob/main/open_bounties/0077%20JavaScript%20Code%20Samples%202.md) for Milestone 4

### DIFF
--- a/content/_code-samples/issue-a-token/js/list-account-tokens.js
+++ b/content/_code-samples/issue-a-token/js/list-account-tokens.js
@@ -22,7 +22,7 @@ if (typeof module !== "undefined") {
         ledger_index: "validated",
     })
 
-    const response_lines = await client.request({
+    const responseLines = await client.request({
         command: "account_lines",
         account: account,
         ledger_index: "validated",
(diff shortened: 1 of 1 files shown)


pdp2121:
Looks good to me! Just some minor nits:
- In your example account object there's one with `undefined` amount. Would be great if that case can be handle.
- I might consider putting `LedgerEntryType` in front of other fields for better view.

wojake:
@pdp2121:
Seems like the accounts' history have been wiped out but their ledger objects are still present on the ledger, I can't really dictate the problem with the `undefined` amount but I've put the `LedgerEntryType` row in front of other fields

content/_code-samples/issue-a-token/js/list-account-tokens.js
```diff
@@ -0,0 +1,59 @@
...
+    // Testnet example: r9CEVt4Cmcjt68ME6GKyhf2DyEGo2rG8AW
+    account = "r9CEVt4Cmcjt68ME6GKyhf2DyEGo2rG8AW"
+
+    // Connect to a testnet node
+    console.log("Connecting to Testnet...")
+    const client = new xrpl.Client('wss://s.altnet.rippletest.net:51233/')
+    await client.connect()
+
+    // Query the ledger for an account's objects and its trustlines...
+    const response_objects = await client.request({
```
khancode:
```suggestion
    const responseObjects = await client.request({
```
The most recommended variable naming convention for JavaScript is camelCase. Can you refactor all variable names to this convention?

content/_code-samples/issue-a-token/js/list-account-tokens.js
```diff
@@ -0,0 +1,59 @@
...
+        "command": "account_objects",
+        "account": account,
+        "ledger_index": "validated",
+    })
+
+    const response_lines = await client.request({
+        "command": "account_lines",
+        "account": account,
+        "ledger_index": "validated",
```
khancode:
```suggestion
        command: "account_lines",
        account: account,
        ledger_index: "validated",
```
Strings aren't required for transaction keys. Can you refactor it everywhere else?

content/_code-samples/issue-a-token/js/list-account-tokens.js
```diff
@@ -0,0 +1,59 @@
...
+    })
+
+    // We'll only display the key data of RippleState ledger objects, other objects are simply shown as their ObjectID and their LedgerEntryType
+    console.log(`\nAccount ${account}'s Ledger Objects:`)
+    for (var i = 0; i < response_objects.result.account_objects.length; i++) {
```
khancode:
```suggestion
    for (let i = 0; i < response_objects.result.account_objects.length; i++) {
```
Use `let`.

content/_code-samples/issue-a-token/js/list-account-tokens.js
```diff
@@ -0,0 +1,59 @@
...
+            console.log(` - Destination: ${response_objects.result.account_objects[i].Destination}`)
+            console.log(` - Amount: ${response_objects.result.account_objects[i].Amount} drops`)
+        } else {
+            console.log(`\n${i+1}. Index (ObjectID/keylet): ${response_objects.result.account_objects[i].index}`)
+            console.log(` - LedgerEntryType: ${response_objects.result.account_objects[i].LedgerEntryType} `)
+        }
+    }
+
+    console.log(`\nAccount ${account}'s Trust lines:`)
+    for (var i = 0 ; i < response_lines.result.lines.length; i++) {
```
khancode:
```suggestion
    for (let i = 0 ; i < response_lines.result.lines.length; i++) {
```
ditto

content/_code-samples/issue-a-token/js/list-account-tokens.js
```diff
@@ -0,0 +1,59 @@
...
+    // Use var here because const/let are block-scoped to the if statement.
+    var xrpl = require('xrpl')
+  }
+  // Show tokens and other objects owned by an account.
+  // https://xrpl.org/account_objects.html#account_objects
+  // https://xrpl.org/account_lines.html#account_lines
+
+  async function main() {
+    // Testnet example: r9CEVt4Cmcjt68ME6GKyhf2DyEGo2rG8AW
+    account = "r9CEVt4Cmcjt68ME6GKyhf2DyEGo2rG8AW"
```
khancode:
```suggestion
if (typeof module !== "undefined") {
    // Use var here because const/let are block-scoped to the if statement.
    var xrpl = require('xrpl')
  }
  // Show tokens and other objects owned by an account.
  // https://xrpl.org/account_objects.html#account_objects
  // https://xrpl.org/account_lines.html#account_lines

  async function main() {
    // Testnet example: r9CEVt4Cmcjt68ME6GKyhf2DyEGo2rG8AW
    account = "r9CEVt4Cmcjt68ME6GKyhf2DyEGo2rG8AW"
```
The indentation looks off. Can you correct for the entire file?

content/_code-samples/issue-a-token/js/list-account-tokens.js
```diff
@@ -0,0 +1,59 @@
+if (typeof module !== "undefined") {
+    // Use var here because const/let are block-scoped to the if statement.
+    var xrpl = require('xrpl')
+  }
```
khancode: I also see a mix of single & double quote string literals being used. Can you stick to using one?

khancode (changes requested):

content/_code-samples/issue-a-token/js/list-account-tokens.js
wojake: Thanks, just noticed that

content/_code-samples/issue-a-token/js/list-account-tokens.js
wojake: The indentation looks and seems to be correct on my computer

khancode: Here's the indentation fixed - https://pastebin.com/E7tqsqha

jonathanlei (approved):

content/_code-samples/issue-a-token/js/list-account-tokens.js
wojake: Thanks for pointing that out, I missed it (again)!

khancode (approved):
LGTM

PR #1701, opened 2023-01-24
Escrows History - JS Bounty

TusharPardhe:
Bounty 6

### DIFF
--- a/content/_code-samples/escrow/js/get-escrows.js
+++ b/content/_code-samples/escrow/js/get-escrows.js
@@ -15,7 +15,7 @@ const main = async () => {
     let account_objects = [];
     
     // Loop through all account objects until marker is undefined
-    while(currMarker !== 'undefined'){
+    do {
 
         const payload = {
             "command": "account_objects",
@@ -30,14 +30,13 @@ const main = async () => {
 
         let { result: { account_objects : escrows, marker } } = await client.request(payload);
 
-        if (typeof marker === 'undefined' || marker === null || marker === currMarker) {
-            currMarker = 'undefined';
-        } else {
-            currMarker =  marker;
-        };
+        if(marker === currMarker){
+            break;
+        }
 
         account_objects.push(...escrows);
-    };
+        currMarker =  marker;
+    } while (typeof currMarker !== 'undefined')
 
     if (source.length > 0) {
         account_objects = account_objects.filter((escrow) =>  escrow.Account === source);
--- a/content/_code-samples/escrow/js/get-escrows.js
+++ b/content/_code-samples/escrow/js/get-escrows.js
@@ -1,57 +1,75 @@
 'use strict'
 if (typeof module !== "undefined") {
-  // Use var here because const/let are block-scoped to the if statement.
-  var xrpl = require('xrpl');
+    // Use var here because const/let are block-scoped to the if statement.
+    var xrpl = require('xrpl');
 };
 
 const address = "rNQhxJsQ5a4newq4acmWsrn8PbrSjLu69p";
-const source = ""; // Optional: filter by source
-const destination = ""; // Optional: filter by destination
+const sourceAddress = ""; // Optional: filter by source address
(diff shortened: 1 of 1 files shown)


content/_code-samples/escrow/js/get-escrows.js
```diff
@@ -0,0 +1,60 @@
+'use strict'
+if (typeof module !== "undefined") {
+  // Use var here because const/let are block-scoped to the if statement.
+  var xrpl = require('xrpl');
+};
+
+const secret = "sEdTPPEeMH6SAgpo6rSj8YW7a9vFfUj";
```
wojake: Users shouldn't need to enter in their secret since we're just querying the ledger for the account's objects (escrow), the best practice for code that only queries the ledger is to only enter in an address instead of a secret.

content/_code-samples/escrow/js/get-escrows.js
```diff
@@ -0,0 +1,60 @@
...
+const source = ""; // Optional: filter by source
+const destination = ""; // Optional: filter by destination
```
wojake: nit: Since the user would be entering in an address that they intend on searching, we should be able to simply filter the escrow objects using `address`

content/_code-samples/escrow/js/get-escrows.js
TusharPardhe: Removed

content/_code-samples/escrow/js/get-escrows.js
TusharPardhe: My intention was that giving an option will be a good, if someone wants to filter using a different source/destination

ckniffen (approved):

content/_code-samples/escrow/js/get-escrows.js
```diff
@@ -0,0 +1,58 @@
...
+    if (destination.length > 0) {
+        account_objects = account_objects.filter((escrow) =>  escrow.Destination === destination);
+    };
+
+    if(account_objects.length === 0) {
+        console.log("No escrows found.");
+        return;
+    };
+
+    console.log(account_objects);
```
JST5000: With the websocket client you have to close the client after you're done by calling `await client.disconnect()` at the end or the code will hang.

content/_code-samples/escrow/js/get-escrows.js
```diff
@@ -0,0 +1,58 @@
...
+            payload.marker = currMarker;
+        };
+
+        let { result: { account_objects : escrows, marker } } = await client.request(payload);
+
+        if (typeof marker === 'undefined' || marker === null || marker === currMarker) {
+            currMarker = 'undefined';
+        } else {
+            currMarker =  marker;
+        };
```
JST5000, line 51: This looping logic can probably be simplified with a `do...while` loop that uses the condition in this if-statement.

TusharPardhe: fixed

content/_code-samples/escrow/js/get-escrows.js
JST5000: It's a bit unclear what these parameters actually should be. It might be useful to give them names that imply their type like `sourceAddress`. 

Also, it'd be good to add a log in case someone doesn't specify either -Ex. Printing "You can filter results to just display those with a specific source address or destination address by modifying `sourceAddress` or `destinationAddress` in this code sample." at the end if neither were specified.

content/_code-samples/escrow/js/get-escrows.js
TusharPardhe: Done

content/_code-samples/escrow/js/get-escrows.js
TusharPardhe: Added

JST5000:
Hey Tushar, we decided to go with WoJake’s implementation for this milestone, so we’re closing this PR. We accepted your work for Milestone 2 in your other Escrow PR.

PR #1705, opened 2023-01-25
Checks - JS Bounty

TusharPardhe:
Bounty 1


TusharPardhe:
Haven't removed old code yet, will remove after review

ckniffen:
> Haven't removed old code yet, will remove after review

Updating the original files will help with the review

content/_code-samples/checks/js/cancel_check.js
```diff
@@ -0,0 +1,36 @@
+const xrpl = require('xrpl');
+
+const secret = "sEdTPPEeMH6SAgpo6rSj8YW7a9vFfUj";
+const checkId = "505E0A4CCBCC971EE07DCD25611A612830CFFA4D50DBC04947EF952D898C94F3"; // Replace with your check ID
```
ckniffen: nit: Having an example checkId prefilled does not make sense to me because the likelihood of it existing is almost 0 because it gets destroyed after one run.  Warn the user if its empty to be more user friendly. 

I understand the original version had it like did it.

content/_code-samples/checks/js/create_check.js
```diff
@@ -0,0 +1,40 @@
...
+            "SendMax": amount,
+            "DestinationTag": 1,
+            "Fee": "12"
+        };
+        // Auto-fill the fields
+        const prepared = await client.autofill(request);
+
+        // Submit the transaction
+        const tx = await client.submitAndWait(prepared, { wallet });
+        console.log(`Transaction: ${JSON.stringify(tx, null, "\t")}`);
```
ckniffen: Output the id for the check.  This will make it easier to plug into future executions of the other scripts.

content/_code-samples/checks/js/get_checks.js
```diff
@@ -0,0 +1,50 @@
+'use strict'
+const xrpl = require('xrpl');
+
+const address = "rEDvX5pbdgC7rYp2eHawTqr5smN7ikMB4r"; // <-- Change this to your account
```
ckniffen: nit: The default address should be for one the seed isnt provided for in the other examples.  That way you dont have to worry about there being no checks the first time you run it.

content/_code-samples/checks/js/get_checks.js
```diff
@@ -0,0 +1,50 @@
...
+            currMarker = 'undefined';
+        } else {
+            currMarker =  marker;
+        };
+
+        account_objects.push(...checks);
+    };
+    
+    // Response: --------------------------------------------------------------
+    console.log("Final transaction result:", JSON.stringify(account_objects, null, "\t"));
```
ckniffen: Update log string.  This is not a transaction.

content/_code-samples/checks/js/cancel_check.js
TusharPardhe: Updated the code as per suggestion

content/_code-samples/checks/js/get_checks.js
```diff
@@ -0,0 +1,50 @@
...
+        };
+
+        if(currMarker !== null) {
+            payload.marker = currMarker;
+        };
+
+        let { result: { account_objects : checks, marker } } = await client.request(payload);
+
+        if (typeof marker === 'undefined' || marker === null || marker === currMarker) {
+            currMarker = 'undefined';
```
ckniffen: Setting `currMarker` to the string `'undefined'` seems odd to me.

TusharPardhe: Changed the logic

content/_code-samples/checks/js/get_checks.js
TusharPardhe: Fixed

content/_code-samples/checks/js/get_checks.js
TusharPardhe: Done

ckniffen (approved):

content/_code-samples/checks/js/create_check.js
TusharPardhe: const response = await client.request({
            "command": "account_objects",
            "account": wallet.address,
            "ledger_index": "validated",
            "type": "check",
        });
        const checkId = response.result.account_objects[0].index;
        console.log(`Check ID: ${JSON.stringify(checkId)}`);

Something like this?

ckniffen: It should be derivable from the transaction.  Itself.  Using account objects i dont think the order is guaranteed.

TusharPardhe: Got it, I have never tried deriving one. Going by the docs I am trying this:

```
const crypto = require('crypto');
const hash = crypto.createHash('sha512');
hash.update(`0x0043${tx.result.Account}${tx.result.Sequence}`);
const hexDigest = hash.digest('hex').toUpperCase().substring(0, 64);
console.log(hexDigest);
```

However, the ID is not matching any of the index for account_objects, can you point me to the right direction.

content/_code-samples/checks/js/cancel_check.js
```diff
@@ -0,0 +1,44 @@
+const xrpl = require('xrpl');
```
JST5000: Could you re-name these files to use `-` instead of `_` to match the other folders' naming convention? (Ex. `cancel_check.js` -> `cancel-check.js`)

content/_code-samples/checks/js/cancel_check.js
```diff
@@ -0,0 +1,44 @@
...
+const main = async ()=> {
+  try {
+      
+    // Connect ----------------------------------------------------------------
+    const client = new xrpl.Client('wss://s.altnet.rippletest.net:51233');
+    await client.connect();
+
+    // Check if the check ID is provided --------------------------------------
+    if (checkId.length === 0) {
+      console.log("Please provide a check ID");
```
JST5000: It's a bit vague on how someone using this snippet could provide the check id.
```suggestion
      console.log("Please edit this snippet to provide a check ID. You can get a check ID by running create_check.js.");
```

JST5000: (Applies to both this and `cash_check.js`)

content/_code-samples/checks/js/create_check.js
```diff
@@ -0,0 +1,40 @@
...
+
```
JST5000: For xrpl.js's clients, you need to add `await client.disconnect()` at the end in order for the code to complete :) (Otherwise it hangs after running through the logic)

content/_code-samples/checks/js/create_check.js
```diff
@@ -0,0 +1,40 @@
...
+        const request = {
+            "TransactionType": "CheckCreate",
+            "Account": wallet.address,
+            "Destination": destination,
+            "SendMax": amount,
+            "DestinationTag": 1,
+            "Fee": "12"
+        };
+        // Auto-fill the fields
+        const prepared = await client.autofill(request);
```
JST5000: With xrpl.js you don't need to do the `prepare` yourself - you can just call `const tx = await client.submitAndWait(request, { wallet })` now.

content/_code-samples/checks/js/create_check.js
```diff
@@ -0,0 +1,40 @@
...
+// Amount of XRP in drops to send
+const amount = "10000000";
+
+const main = async () => {
+    try {
+        // Connect to the XRP Ledger Test Net
+        const client = new xrpl.Client('wss://s.altnet.rippletest.net:51233');
+        await client.connect();
+
+        const wallet = await xrpl.Wallet.fromSeed(secret);
```
JST5000: For this snippet I'd recommend generating a new account using `client.fundWallet()`, then printing out both it's address & secret. The reason being that it'll make using the output of this function to try `cash_check` or `create_check` easier since the secret is available, and it'll avoid the problem where eventually this specified account runs out of money because we never refill it with the faucet.

content/_code-samples/checks/js/create_check.js
ckniffen: You can get the id from the meta.  It is in the created node for 'Check'

content/_code-samples/checks/js/get_checks.js
```diff
@@ -0,0 +1,56 @@
...
+    const client = new xrpl.Client('wss://s.altnet.rippletest.net:51233');
+    await client.connect();
+
+    let currMarker = null;
+    let account_objects = [];
+
+    // Loop through all account objects until marker is undefined --------------
+    let keepLooping = true;
+    
+    while(keepLooping){
```
JST5000: You can probably replace `keepLooping` with a `do...while(marker === undefined)` loop that stops when marker stops existing.

content/_code-samples/checks/js/create_check.js
JST5000: Here's a bit of code which works to do that, not sure why the direct method isn't working.
```
    const checkID = tx.result.meta.AffectedNodes.reduce((prevOutput, node) => {
      if (node?.CreatedNode && node.CreatedNode?.LedgerEntryType == "Check") {
        return node.CreatedNode.LedgerIndex;
      } else {
        return prevOutput;
      }
    }, null);
    if (checkID) {
      console.log("Check ID: ", checkID);
    } else {
      console.log(
        "Unable to find the CheckID from parsing the metadata. Look for the LedgerIndex of the 'Check' object within 'meta'."
      );
    }
```

content/_code-samples/checks/js/get_checks.js
TusharPardhe: I am questioning myself of why I didn't think of that, haha. Done!

content/_code-samples/checks/js/create_check.js
TusharPardhe: done

content/_code-samples/checks/js/create_check.js
TusharPardhe: I usually add it just to be sure, removed.

content/_code-samples/checks/js/create_check.js
TusharPardhe: Oh yes, forgot to write that. Apologies, fixed

content/_code-samples/checks/js/cancel_check.js
TusharPardhe: done

jonathanlei:
please fix link checker error - https://github.com/XRPLF/xrpl-dev-portal/actions/runs/4078204283/jobs/7111028936, then I'll merge

jonathanlei (approved):

TusharPardhe:
@jonathanlei I just checked, [send-a-check.md](https://github.com/XRPLF/xrpl-dev-portal/blob/master/content/tutorials/use-specialized-payment-types/use-checks/send-a-check.md) file needs to be updated i.e. the contents of this webpage will change https://xrpl.org/send-a-check.html. Any guidelines/suggestions that should be followed while updating?

PS: Now the code is in one file rather than separate files.

JST5000:
Hey Tushar, from investigating the Link Checker, we realized that the other files that used to be in this folder are actually part of a tutorial, and so are required for now. We'll go in later and move them so someone looking at the code sample can see the more user-friendly samples, as opposed to the broken up code used for the tutorial. 

For now though, could you re-add the files that were deleted in this PR? (That'll fix the link checker) After that we can merge this PR.

TusharPardhe:
> Hey Tushar, from investigating the Link Checker, we realized that the other files that used to be in this folder are actually part of a tutorial, and so are required for now. We'll go in later and move them so someone looking at the code sample can see the more user-friendly samples, as opposed to the broken up code used for the tutorial.
> 
> For now though, could you re-add the files that were deleted in this PR? (That'll fix the link checker) After that we can merge this PR.



> Hey Tushar, from investigating the Link Checker, we realized that the other files that used to be in this folder are actually part of a tutorial, and so are required for now. We'll go in later and move them so someone looking at the code sample can see the more user-friendly samples, as opposed to the broken up code used for the tutorial.
> 
> For now though, could you re-add the files that were deleted in this PR? (That'll fix the link checker) After that we can merge this PR.

Added the old code as well

PR #1714, opened 2023-01-30
chore: remove top exchanges

justinr1234:


ckniffen (approved):

Silkjaer:
It's not sufficient to remove FTX – Liquid has also halted trading. So I suggest either updating the list to actually reflect CryptoCompare, as it says, or remove the "top exchanges" altogether. I personally think it's better to remove them :)

justinr1234:
> It's not sufficient to remove FTX – Liquid has also halted trading. So I suggest either updating the list to actually reflect CryptoCompare, as it says, or remove the "top exchanges" altogether. I personally think it's better to remove them :)

Do you know where the link is? I couldn’t find where this original list came from.

intelliot (approved):
This is in a section called "Top Exchanges, according to CryptoCompare". In general, the CryptoCompare list has changed significantly, and the list here has not followed it. Consider removing the list from this page entirely, and perhaps linking to some external resource instead.

intelliot:
FWIW, I do not know where the original list is from.

Silkjaer:
Agreed. I think CryptoCompare introduced other measures to rank exchanges as well, so removing it is fine.

justinr1234:
@Silkjaer @intelliot I added a link to coinmarketcap which has a nice list it seems. Please let me know what you think.

image

Silkjaer:
I still think it's better to just remove it. If not, it should be rephrased (it's not XRPL exchanges, and you may not want to only obtain XRP) 👍🏻

justinr1234:
@Silkjaer Updated, let me know

PR #1744, opened 2023-02-16
Add NFT reserves topic

DennisDawson:


content/concepts/tokens/nft-reserve-requirements.md
```diff
@@ -0,0 +1,55 @@
...
+ - Non-fungible Tokens, NFTs
+---
+
+# NFT Reserve Requirements
+
+Minting, holding, and offering NFTs for sale require XRP held in reserve. The reserve charges can add up quickly. Understanding the reserve requirements can help you choose the best approach for your business case.
+
+## Base Reserve
+
+You account must set aside a base reserve, currently 10 XRP. The base reserve XRP amount is subject to change. See [Base Reserve and Owner Reserve](reserves.html#base-reserve-and-owner-reserve).
```
shawnxie999: is it mean to say "Your account" instead of "You account"?

DennisDawson: Thanks for catching that.

oeggert (approved):
Looks good to me.

shawnxie999 (approved):

shashwatmittal92:
Can we also explicitly state the base reserve and other reserve requirements subjected on a user's wallet while transacting with an application? So that'll help the community understand that they need to accommodate the following when facilitating transactions:
- 10 XRP if the user is creating a wallet on their platform or has an inactive wallet
- 2 XRP for the 1st NFT on a new NFToken page
- 2 XRP per offer

shawnxie999:
> Can we also explicitly state the base reserve and other reserve requirements subjected on a user's wallet while transacting with an application? So that'll help the community understand that they need to accommodate the following when facilitating transactions:
> 
> * 10 XRP if the user is creating a wallet on their platform or has an inactive wallet
> * 2 XRP for the 1st NFT on a new NFToken page
> * 2 XRP per offer

Just curious, isn't the reserve requirement already written somewhere in the existing documentations? I think I have stumbled into some of them when I was reading

DennisDawson:
> 

You're right, Shawn. At first, I didn't think the topic was needed, but now that it's written, I think it's helpful to bring _all_ of the reserves and fees together into one topic. I even thought to add a note about having an extra XRP in there to cover all of the trivial transaction fees.

Shashwat has gotten feedback from the community expressing confusion and consternation about reserves, so making it one handy topic will give them a better doc experience.

shawnxie999:
> > 
> 
> You're right, Shawn. At first, I didn't think the topic was needed, but now that it's written, I think it's helpful to bring _all_ of the reserves and fees together into one topic. I even thought to add a note about having an extra XRP in there to cover all of the trivial transaction fees.
> 
> Shashwat has gotten feedback from the community expressing confusion and consternation about reserves, so making it one handy topic will give them a better doc experience.

If that's the case, wouldn't it be helpful if we also have one page that exclusively shows all the different ledger object types and their respective reserve requirements? I think currently we need to kind scramble the doc to learn about the reserves in each scenario

EDIT: If that's what the community find it to be confusing, then it might be preferable to also show everything in one place

DennisDawson:
> > > 
> > 
> > 
> > You're right, Shawn. At first, I didn't think the topic was needed, but now that it's written, I think it's helpful to bring _all_ of the reserves and fees together into one topic. I even thought to add a note about having an extra XRP in there to cover all of the trivial transaction fees.
> > Shashwat has gotten feedback from the community expressing confusion and consternation about reserves, so making it one handy topic will give them a better doc experience.
> 
> If that's the case, wouldn't it be helpful if we also have one page that exclusively shows all the different ledger object types and their respective reserve requirements? I think currently we need to kind scramble the doc to learn about the reserves in each scenario

Possibly, though it's beyond the scope of this particular change. I think that NFTs are a little more complex than the other types of reserve, in that the reserves are by NFTokenPage, not NFT, and there's a reserve of 2 XRP per offer. It seems like the others are pretty well covered on the Reserves page. I'll confer with @oeggert to see what he thinks.

PR #1763, opened 2023-02-23
Update AccountDelete tecTOO_SOON description

kennyzlei:
Adds reasoning for `Sequence` number check prior to account deletion, taken from a code comment: https://github.com/XRPLF/rippled/blob/60c276d90bbc7ca6a88379fadf7dfdfae8f673a9/src/ripple/app/tx/impl/DeleteAccount.cpp#L207-L209


ckniffen (approved):

intelliot (approved):
let's allow the docs team to review/merge

PR #1768, opened 2023-02-27
Visual fix in nftokenacceptoffer.md

zgrguric:
This will fix visually "X" being bolder in first column in table due to CSS nth-child(1) and X not rendered as bold in second column.


mDuo13 (approved):
Not a bad idea, but I did you one better and substituted an appropriate emoji for consistency with the other ones

PR #1769, opened 2023-02-27
Update account_nfts.md

zgrguric:
Change  non-existing example account to existing with NFT-s.


mDuo13 (changes requested):
The example response down the page was originally a valid API response to the example request. This change only updates the example request, so they're not longer consistent with each other. If you're going to change one, please update both.

zgrguric:
I understand, I am not fully happy with account selection for sample as it is prone to change. I suggest  putting this on hold.

mDuo13:
Feel free to reopen if you have a better suggestion. I can also revisit this later if I get a chance.

PR #1770, opened 2023-02-27
Update node provider list

AnnaNodes:
Hello,

We are an XRP node provider NOWNodes.io.

Could you update the node provider list?


mDuo13:
This file is no longer used (it's been replaced by a more customized page defined by https://github.com/XRPLF/xrpl-dev-portal/blob/master/template/page-dev-tools.html.jinja ). Also, I don't think the dev tools list is the right place to advertise a paid service like NowNodes.io even if it is quite useful for getting up and running quickly.

AnnaNodes:
Thank you for your answer! 
May you suggest us a place where we can list our service for your community?
We also have a free tariff plan.

PR #1771, opened 2023-02-27
Update status of fixNonFungibleTokensV1_2

amarantha-k:
as v1.10.0 is still not released.


mDuo13 (approved):
I have a draft in progress that addresses some further issues but this is OK for now

PR #1775, opened 2023-02-28
Fix broken links for freeze misconceptions/offers pages

mDuo13:
These were caused somehow by #1741 (even though the link checker didn't catch them there—not sure why not).


oeggert (approved):
Looks good to me.

PR #1784, opened 2023-03-02
Add fixed token supply topic to toc

DennisDawson:


dactyl-config.yml
```diff
@@ -832,6 +832,12 @@ pages:
             - en
             - ja
 
+    # TODO: translate
```
mDuo13, line 835: Handled by #1746

mDuo13:
I think this is no longer necessary because #1746 fixed this (including translating and adding a separate config file entry for the translation)

PR #1787, opened 2023-03-03
Update OfferCreate response code.

oeggert:
Update the `tecKILLED` response code for transactions with the `tfImmediateOrCancel` flag.

Also cleaned up old amendment tags for other error codes.

### DIFF
--- a/content/references/protocol-reference/transactions/transaction-types/offercreate.md
+++ b/content/references/protocol-reference/transactions/transaction-types/offercreate.md
@@ -60,8 +60,8 @@ Transactions of the OfferCreate type support additional values in the [`Flags` f
 | Error Code               | Description                                       |
 |:-------------------------|:--------------------------------------------------|
 | `temINVALID_FLAG`        | Occurs if the transaction specifies both `tfImmediateOrCancel` and `tfFillOrKill`. |
-| `tecEXPIRED`             | Occurs if the transaction specifies an `Expiration` time that has already passed _(Updated by the [DepositPreauth amendment][]: previously, the transaction used the result code `tesSUCCESS` in this case. In either case, the transaction has no effect except to destroy the XRP paid as a [transaction cost](transaction-cost.html).)_ |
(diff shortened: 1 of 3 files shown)


content/references/protocol-reference/transactions/transaction-types/offercreate.md
```diff
@@ -60,8 +60,8 @@ Transactions of the OfferCreate type support additional values in the [`Flags` f
...
-| `tecEXPIRED`             | Occurs if the transaction specifies an `Expiration` time that has already passed _(Updated by the [DepositPreauth amendment][]: previously, the transaction used the result code `tesSUCCESS` in this case. In either case, the transaction has no effect except to destroy the XRP paid as a [transaction cost](transaction-cost.html).)_ |
-| `tecKILLED`              | Occurs if the transaction specifies `tfFillOrKill`, and the full amount cannot be filled. _(Updated by the [fix1578 amendment][]: previously, the transaction used the result code `tesSUCCESS` in this case. In either case, the transaction does not execute any trades.)_ |
+| `tecEXPIRED`             | Occurs if the transaction specifies an `Expiration` time that has already passed. |
+| `tecKILLED`              | Occurs if the transaction specifies `tfFillOrKill`, and the full amount cannot be filled. If the _[ImmediateOfferKilled amendment][]_ is enabled, this result code also occurs when the transaction specifies `tfImmediateOrCancel` and is canceled without moving funds (previously, this would return `tesSUCCESS`). |
```
scottschurr: I might modify the phrasing of this sentence just a bit.  Maybe like this:
"If the _[ImmediateOfferKilled amendment][]_ is enabled, this result code also occurs when an `OfferCreate` transaction specifies `tfImmediateOrCancel` and executes without moving funds (previously, this would return `tesSUCCESS`)."

DennisDawson (approved):
These changes are fine. I think this could use further explanation in the concepts section. I did have to read it several times to understand the scenario.

PR #1789, opened 2023-03-03
Remove outdated build instructions and add redirects to BUILD.md

### LINKED ISSUE #1651: Update Ubuntu build instructions

oeggert:
With the move to Conan in 1.10.0, the build from source instructions are now invalid. Redirecting to the build instructions on the stable branch of `rippled`.


oeggert:
@mDuo13 The redirects aren't automatically redirecting for the English version of pages, but works fine for the Japanese ones. I'm not sure if I'm doing something wrong with the dactyl config?

intelliot:
We may still need instructions/information specific to Reporting Mode

oeggert:
@intelliot I restored the reporting mode instructions and updated the links in that doc to point to BUILD.md; does this PR look ready to go otherwise?

PR #1796, opened 2023-03-06
Include links to in js.xrpl.org to https://xrpl.org/get-started-using-javascript.html

generalburatai:
Are we able to include links to js.xrpl.org in the Docs? That way, devs see the links directly and do not need to go and search for them.


mDuo13:
It looks like a PR has been opened instead of an issue—probably by mistake?

The changes here are only a reverse merge from the master branch into the 1.10.0 release docs branch; that's not necessary and does not address the issue requested in the title and description.

(We can and should link js.xrpl.org from the getting started as appropriate though.)

PR #1786, opened 2023-03-03
comments carbon calculator link, redirects from carbon calculator pag…

AKCodez:
…e to impact page

PR #1806, opened 2023-03-08
DGE791-remove XLS20 refs

DennisDawson:
Remove stray XLS-20 devnet references from the websocket api tools page and from tutorials.


ckniffen:
There are also these.

- https://github.com/XRPLF/xrpl-dev-portal/blob/74ea71082015a4ab1a4c5c0ca83f84b92f7a662b/content/references/http-websocket-apis/public-api-methods/account-methods/account_nfts.md?plain=1#L46
- https://github.com/XRPLF/xrpl-dev-portal/blob/74ea71082015a4ab1a4c5c0ca83f84b92f7a662b/content/references/http-websocket-apis/public-api-methods/path-and-order-book-methods/nft_sell_offers.md?plain=1#L47
- https://github.com/XRPLF/xrpl-dev-portal/blob/74ea71082015a4ab1a4c5c0ca83f84b92f7a662b/content/references/http-websocket-apis/public-api-methods/path-and-order-book-methods/nft_buy_offers.md?plain=1#L46

mDuo13 (commented):
Looks like the broken links are unrelated to this PR (they're caused by #1786). The changes here look good, although @ckniffen is right about those links also needing an update.

(You just need to delete the query parameter—that is, the part of the link starting from the `?` up to but not including the `#`.)

ckniffen (approved):

PR #1816, opened 2023-03-15
I was wrong about the change to the txn fee

DennisDawson:
Returning it to the previous value of $0.0002

### DIFF
--- a/template/page-xrp-overview.html.jinja
+++ b/template/page-xrp-overview.html.jinja
@@ -73,7 +73,7 @@
               </tr>
               <tr>
                 <td>{% trans %}Low-Cost{% endtrans %}</td>
-                <td>{% trans %}0.00001/tx{% endtrans %}</td>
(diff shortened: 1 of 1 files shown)


intelliot (approved):
I'll be the first to admit that the values are confusing. 

We could equally argue that 1,500 and 3 are misleading/inaccurate, and 0.3% is also a wild guess that is sure to be (at least slightly) wrong.

I think the page is fine because it simply communicates that XRP has advantages over Bitcoin; not specific values that are broadly applicable. We could add a footnote that explains that the values are estimates and approximations; or we could add `~` to convey the same idea.

DennisDawson:
Agreed that the numbers are confusing. Before I made the change, I spent some time looking at the average transaction cost in the Explorer (which bounced around from about .001500 to. 005700) and did the math. The answer changes every 5 seconds, of course, but a guestimate of $.0002 is reasonable. If nothing else, at least I understand this better. Thanks for the help.

PR #1819, opened 2023-03-16
Add note for `ledger_data`  API about `type` field

shawnxie999:
The response of `ledger_data`  might be empty, when `type` is specified. Some people had thought something was wrong. So clarification is added to avoid confusion

### DIFF
--- a/content/references/http-websocket-apis/public-api-methods/ledger-methods/ledger_data.md
+++ b/content/references/http-websocket-apis/public-api-methods/ledger-methods/ledger_data.md
@@ -251,7 +251,7 @@ The response follows the [standard format][], with a successful result containin
 | `state`        | Array                               | Array of JSON objects containing data from the ledger's state tree, as defined below. |
 | `marker`       | [Marker][]                          | Server-defined value indicating the response is paginated. Pass this to the next call to resume where this call left off. |
 
-If a `type` field is mentioned in the request, the `state` array might be empty. In such cases, you'll have to use the `marker` from this response to paginate and retrieve further data. Keep repeating this until you get a non-empty `state` array _(if the object specified in `type` exists on the ledger)_.
+If a `type` field is mentioned in the request, the `state` array will be empty if the first set of array objects does not match the type requested. In such cases, you can use the `marker` from this response to paginate and retrieve further data.
 
 The format of each object in the `state` array depends on whether `binary` was set to true or not in the request. Each `state` object may include the following fields:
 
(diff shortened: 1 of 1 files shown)


content/references/http-websocket-apis/public-api-methods/ledger-methods/ledger_data.md
```diff
@@ -251,6 +251,8 @@ The response follows the [standard format][], with a successful result containin
 | `state`        | Array                               | Array of JSON objects containing data from the ledger's state tree, as defined below. |
 | `marker`       | [Marker][]                          | Server-defined value indicating the response is paginated. Pass this to the next call to resume where this call left off. |
 
+If a `type` field is mentioned in the request, the `state` array might be empty. In such cases, you'll have to use the `marker` from this response to paginate and retrieve further data. Keep repeating this until you get a non-empty `state` array _(if the object specified in `type` exists on the ledger)_.
```
kennyzlei: Just reading this explanation, I don't feel like it provides a clear understanding what is going on under the hood vs. just having a usage suggestion

```suggestion
If a `type` field is mentioned in the request, the `state` array may be empty if the first set of array of objects does not match the type requested. In such cases, you may want to use the `marker` from this response to paginate and retrieve further data.
```

DennisDawson: Is it possible that the state array will _not_ be empty? Will it always be empty if the first set of array objects don't match the type? Consider this version:

If a `type` field is mentioned in the request, the `state` array will be empty if the first set of array objects does not match the type requested. In such cases, you can use the `marker` from this response to paginate and retrieve further data.

shawnxie999: @kennyzlei why is it important for whoever is using the API to know what's going on underneath the hood? I thought we would like to hide such implementation detail for the sake of abstraction

kennyzlei: I think upon reading the usage suggestion sentence you have, my first thought is "Why do I need to keep paginating?  Is this a bug and this is a hacky workaround?  Seems weird to me".  If I knew how pagination comes first and then filtering on type, I understand why the response comes out a certain way, usually you assume the opposite as in filter then paginate if necessary

shawnxie999: cool that make sense if exposing a slight implementation detail gives more clarity to the user. It normally would make sense to do filter then paginate, but that can't be done with the ledger. If this is implemented, that would mean some one could try to call the api on a ledger index that doesn't have this object type, which would last hours to search through millions of ledger objects

shawnxie999: @DennisDawson thanks Dennis, updated

intelliot (approved):

shawnxie999:
@DennisDawson could you re-review this to see if its ready to merge? thx

DennisDawson:
@shawnxie999 LGTM.

shawnxie999:
@mDuo13 this is ready to merge

ledhed2222 (approved):

PR #1831, opened 2023-03-20
[WIP] Update dactyl config to remove in-development docs

amarantha-k:

PR #1837, opened 2023-03-22
AMM Error Updates

legleux:
Error type changed
Clarify `terNO_ACCOUNT`

@mDuo13 What should `tecOWNER` be?


mDuo13 (changes requested):
The AMM docs need a fairly thorough update on the error codes to reflect changes in the implementation over the past month or two; the differences are greater than what's touched here. (For example, AMMCreate can now return tecDUPLICATE as well.)

However, most likely we'll be doing those changes in a separate repo because these docs are moving off of XRPL.org until the AMM feature is in a stable release. (See #1831 for that)

PR #1847, opened 2023-03-27
Change server_state.transitions to string in example

JST5000:
When running against rippled locally, `ServerState.transitions` appears to be a string, and the documentation says it should be a string, so this PR updates the example to use a string instead of a number for that field. (I believe it used to be a number, but at some point changed)

Discovered as a result of using a more up-to-date rippled version for integration tests in xrpl.js here: https://github.com/XRPLF/xrpl.js/pull/2223


mDuo13 (approved):
It's mildly concerning that at some point we apparently had a breaking API change to this field's type, but ¯\_(ツ)_/¯ it is what it is now...

PR #1856, opened 2023-03-31
Touch up stablecoin issuer use case

mDuo13:
- [x] Replace the stub stablecoin use case page with the full-featured "Become a Stablecoin Issuer" guide (removing it from where it was previously)
- [x] Clean up and reword the stablecoin issuer guide for latest style guidelines, and link out to relevant information in more piecemeal tutorials/guides


mDuo13:
A note for reviewers: the way GitHub breaks down the overall PR makes it hard to see what changed and what was in the original article, but if you look at the individual commit diffs it'll show you just the things that I changed for this PR.

oeggert (approved):
I think some pruning needs to happen. This use case page is dense (unavoidable because of the technical details involved with issuing stable coins).

I think we can remove a lot of the regulatory info and just link out to relevant sources so we don't have to maintain that information ourselves. The only people even considering issuing stable coins would probably have a lot of that background info anyway.

Looks good overall, though. I don't see any reason to stop this merge. We'll all probably need to align our use case pages at some point and work on them more.

PR #1857, opened 2023-03-31
Events updates

AKCodez:


ckniffen:
Community Call 3 happened yesterday so should be moved out of upcoming.

AKCodez:
> Community Call 3 happened yesterday so should be moved out of upcoming.
Removed Community Call 3

template/page-events.html.jinja
```diff
@@ -247,26 +219,154 @@
...
       "type": "conference-past",
       "link": "https://www.youtube.com/playlist?list=PLJQ55Tj1hIVZgnreb8ODgxJW032M9Z2XZ",
       "location": "Las Vegas, Tallinn",
       "date": "September 29-30, 2021",
-      "image": "event-conference-apex-2021.png"},
+      "image": "event-conference-apex-2021.png",
+      "end_date": "September 30, 2022",
+      },
+
+    { "name": "Hackathon:<br />NFT Launch Party",
```
mDuo13, line 247: Ideally the indentation for these would match the examples above, but it's…fine

mDuo13 (approved):
I think we should also add a "past events" entry for Apex 2022 at some point. Looks good for now though

PR #1858, opened 2023-04-04
p2p payments: edits to Wallets

mDuo13:
I noticed a couple typos ("secury"; "Non-custodial" where you meant "custodial") but in the process of fixing it I ended up rephrasing the whole sentence and adding a lot more color on both categories.


oeggert (approved):
Looks good to me!

PR #1859, opened 2023-04-05
depositauth - explain how to enable

intelliot:
This makes it faster for new readers to see how the setting can be enabled.


mDuo13 (changes requested):
Can you clarify what inspired this change? I feel it interrupts the high-level overview of the feature with low-level details of specific flag names. Also, this information is already in this same page further down, under the heading "Enabling or Disabling Deposit Authorization".

intelliot:
Some readers were not sure how to enable the setting. 

I agree the information is already there: this just makes it easier for newcomers to find. That said, it's not a big deal. Feel free to close without merging

PR #1860, opened 2023-04-05
Fix for Issue1765 that clarifies currency format

### LINKED ISSUE #1765: Unclear currency format
https://xrpl.org/account_lines.html has `balance` and `currency` fields.

It is unclear whether here the `currency` can be `XRP`, and if yes, what format `balance` has in this case: value in XRP drops or floating point as usual for non-XRP currencies.

amarantha-k:
When merged, fixes #1765.


intelliot (approved):

oeggert (approved):
Looks good to me.

PR #1861, opened 2023-04-06
Infrastructure docs re-architecture

maria-robobug:


dactyl-config.yml
```diff
@@ -1618,9 +1618,60 @@ pages:
...
+    -   md: tutorials/interoperability/evm-sidechain-validator-security.md
+        targets:
+            - en
+            - ja
+
+        # TODO: translate
+    -   md: tutorials/interoperability/evm-sidechain-run-a-validator-node.md
+        targets:
+            - en
+            - ja
```
mDuo13, line 1653: Looks like this work might need to be updated to reflect the EVM stuff being moved off the site—you don't want to re-add the files here, I suspect.

mDuo13 (commented):
This largely looks good to me but I think the EVM thing might be a place where we need to re-sync this branch with master.

dactyl-config.yml
```diff
@@ -1612,9 +1612,60 @@ pages:
...
+# Infrastructure --------------------------------------------------------------------
+
+    -   name: XRP Ledger Infrastructure
+        html: xrpl-infrastructure.html
+        parent: docs.html
+        top_nav_grouping: Article Types
+        template: pagetype-category.html.jinja
+        blurb: Learn how to set up, maintain, test, and troubleshoot the XRP Ledger.
+        targets:
+            - en
```
develoQ, line 1664: It seems to me that a Japanese page should be added to target.

```suggestion
            - en
            - ja
```

amarantha-k (approved):
LGTM. We should handle the dactyl changes in a separate PR.

dactyl-config.yml
tequdev, line 1664:
```suggestion
            - en
            - ja
```

PR #1862, opened 2023-04-06
add nft api topic

DennisDawson:


content/references/http-websocket-apis/public-api-methods/clio-methods/nft_history.md
```diff
@@ -0,0 +1,706 @@
...
+                "FinalFields": {
+                  "Amount": "100000000",
+                  "Flags": 0,
+                  "NFTokenID": "00082710A955EBAD4AE6261466FBEEEBF6A59591AC8C831F16E5DA9C00000001",
+                  "NFTokenOfferNode": "0",
+                  "Owner": "rHKjdRD2aHzEojrioJb4tw6p4i7FS1hUkc",
+                  "OwnerNode": "0",
+                  "PreviousTxnID": "B2E36CBAE4C0E329A0FA373DBC853AE871F9EC0371C6299FE0955F689671F015",
+                  "PreviousTxnLgrSeq": 27481778
+                },
```
ledhed2222: wondering whether it makes sense to use an example with the `binary` option set, to save space. You could then show an additional example with the `binary` flag not set but with only one transaction to show what that format looks like.

ledhed2222 (approved):
just one idea for you - but otherwise looks good to me!

content/references/http-websocket-apis/public-api-methods/clio-methods/nft_history.md
DennisDawson: Good suggestion. What I ended up with is a reduced-size JSON example, coupled with a new example with the _binary_ flag set to true. The JSON is still long, but not CVS Receipt long.

PR #1866, opened 2023-04-10
Docs.html page update

JST5000:
This updates the top-level documentation page with a new format. 

Opening this PR as a draft for now in order to allow people to see a build-preview for the work so far to get feedback, and to make it easier for others to combine other pieces of the work into it.


styles/yarn.lock
```diff
@@ -3,1605 +3,1802 @@
 
 
 "@babel/code-frame@^7.0.0":
```
ckniffen, line 5: What was the reason for alll of these version bumps?  Might make sense for this to be in a separate PR.

styles/package.json
```diff
@@ -1,7 +1,9 @@
 {
   "devDependencies": {
-    "bootstrap": "^4.6.0",
-    "node-sass": "^7.0.0"
+    "bootstrap": "^4.6.1",
+    "jquery": "^1.9.1",
```
ckniffen: Should this be a dev dependency? @JST5000

JST5000: I don't think so, no.

template/page-docs.html.jinja
```diff
@@ -1,126 +1,429 @@
 {% extends "base.html.jinja" %}
 
-{% block mainclasses %}landing page-docs-index landing-builtin-bg{% endblock %}
+{% block mainclasses %}landing page-docs-index landing-builtin-bg jackson-identifier {% endblock %}
```
JST5000: I'm not sure why this isn't showing up locally for me - but `jackson-identifier` should probably be removed.

JST5000:
> File: out/ja/serialization.html Link: [STIssue][]

@mDuo13 do you happen to know why this would be the case? (It seems to be affecting both this PR and @pdp2121's PR here: https://github.com/XRPLF/xrpl-dev-portal/pull/1892 so it seems maybe it's an error on main?)

Meta: How do you normally go about debugging the broken link error?

mDuo13:
> > File: out/ja/serialization.html Link: [STIssue][]
> 
> @mDuo13 do you happen to know why this would be the case? (It seems to be affecting both this PR and @pdp2121's PR here: #1892 so it seems maybe it's an error on main?)
> 
> Meta: How do you normally go about debugging the broken link error?

#1894 explains and resolves the STIssue ink error. In cases like that where there's a broken link in master, PRs that were branched off while the issue was active will show the error unless they're rebased or have a reverse-merge, but even if you do neither of those things they're safe to merge to master.

For debugging broken links, if the change doesn't come from or link to one of the files modified in the file—including via includes—then I check to see if it's a case where a broken link snuck into master, like this one. Otherwise, you can look at the HTML source and search (as in, ctrl-f) for the reported text of the link. In some cases, like this one, `[STIssue][]` being reported as a broken link means that the page used Markdown's ["reference link" syntax](https://www.markdownguide.org/basic-syntax#reference-style-links) but the reference wasn't matched—basically, you tried to reference something without a definition—so the bracket syntax shows up directly in the final page text.

template/page-docs.html.jinja
```diff
@@ -1,126 +1,430 @@
...
+          </div>
+        </div>
+      </div>
     </div>
-    {% endfor %}
-  </div>
-</section><!--/.doc-index-->
+  </section><!-- Browse by recommended and Generate Testnet Credentials -->
+  <section class="container-new my-10">
+    <a href="./docs-index.html" class="external-link ml-auto">See full documenation index <i aria-hidden="true" class="fa fa-external-link"></i></a>
```
mDuo13: Typo. Also, translation tags.

```suggestion
    <a href="./docs-index.html" class="external-link ml-auto">{% trans %}See full documentation index{% endtrans %} <i aria-hidden="true" class="fa fa-external-link"></i></a>
```

template/page-home.html.jinja
```diff
@@ -176,8 +176,8 @@
           "href": "https://hooks-testnet.xrpl-labs.com/" },
         { "chip": _("Enabled"),
           "title":  _("Non-Fungible Tokens"),
-          "description":  _("Lower fees, faster transactions, and custom token functionality make the XRPL ideally suited for building an ecosystem for NFTs. Explore <a href='docs.html#docs-hot-topic'>proposed standards</a> for issuing NFTs."),
-          "href": "docs.html#docs-hot-topic"},
+          "description":  _("Lower fees, faster transactions, and custom token functionality make the XRPL ideally suited for building an ecosystem for NFTs. Explore <a href='docs.html'>proposed standards</a> for issuing NFTs."),
+          "href": "docs.html"},
```
mDuo13:
```suggestion
          "href": "non-fungible-tokens.html"},
```

template/page-home.html.jinja
```diff
@@ -176,8 +176,8 @@
           "href": "https://hooks-testnet.xrpl-labs.com/" },
         { "chip": _("Enabled"),
           "title":  _("Non-Fungible Tokens"),
-          "description":  _("Lower fees, faster transactions, and custom token functionality make the XRPL ideally suited for building an ecosystem for NFTs. Explore <a href='docs.html#docs-hot-topic'>proposed standards</a> for issuing NFTs."),
-          "href": "docs.html#docs-hot-topic"},
+          "description":  _("Lower fees, faster transactions, and custom token functionality make the XRPL ideally suited for building an ecosystem for NFTs. Explore <a href='docs.html'>proposed standards</a> for issuing NFTs."),
```
mDuo13: Probably we should at some point change this whole block to be a different upcoming feature rather than highlighting NFTs, but for now, we should not link "proposed standards" for something that's in use now.

```suggestion
          "description":  _("Lower fees, faster transactions, and custom token functionality make the XRPL ideally suited for building an ecosystem for NFTs. Explore the XRP Ledger's NFT capabilities."),
```

template/page-docs.html.jinja
```diff
@@ -1,126 +1,430 @@
...
+  {% macro primaryButton(href, text, isArrowUp) %}
+    {% if isArrowUp %}
+      <a class="btn btn-primary btn-arrow-out" id="{{href}}-button" href="{{href}}">{{ text }}</a>
+    {% else %}
+      <a class="btn btn-primary btn-arrow" id="{{href}}-button" href="{{href}}">{{ text }}</a>
+    {% endif %}
+  {% endmacro %}
+
+  {% macro flatCard(href, title, description, linkText, imgClass) %}
+    <div class="card video-image flat-card">
```
mDuo13: As we discussed separately, the `video-image` class should not be used for these

Silkjaer:
This really looks great. Great update to have better entries into relevant sections of the documentation.

styles/_docs.scss
mDuo13, line 1: I'm really nitpicking here, but calling this page `_docs.scss` makes it unclear if it's referring to the actual ("leaf") documentation pages, or the "Docs" landing page. Renaming it to `_docs-landing.scss` or `_docs-home.scss` would be clearer.

styles/_docs.scss
```diff
@@ -0,0 +1,352 @@
...
+  .quickstart-image {
+    content:url(../../img/graphics/getting-started-pages-light.png)
+  }
+
+  .dev-tools-img {
+    content:url(../../img/graphics/dev-tools-light.svg)
+  }
+
+  .dev-tools-link:hover p {
+    color: #000
```
mDuo13: There are a bunch of color name variables defined in `_colors.scss`. Wherever possible, please use those instead of hard-coding color values into the SCSS; that way, if we tweak the color palette, the changes automatically update throughout the styles. (For example, in the past, we changed from "pure black" to "off-black" for some things.)

```suggestion
    color: $black
```

styles/_docs.scss
```diff
@@ -0,0 +1,352 @@
...
+  .quickstart-image {
+    content:url(../../img/graphics/getting-started-pages-dark.svg)
+  }
+
+  .dev-tools-img {
+    content:url(../../img/graphics/dev-tools-dark.png)
+  }
+
+  .dev-tools-link:hover p {
+    color: #fff;
```
mDuo13:
```suggestion
    color: $white;
```

styles/_docs.scss
```diff
@@ -0,0 +1,352 @@
...
+  }
+
+  .flat-card-grid .nav-link:hover {
+    color: #E0E0E1;
```
mDuo13:
```suggestion
    color: $gray-200;
```

styles/_docs.scss
```diff
@@ -0,0 +1,352 @@
...
+      width: 1.5rem;
+      height: 1.5rem;
+    }
+  }
+
+  .langs {
+    
+    h5:hover {
+      text-decoration: underline;
+      text-decoration-color: #9A52FF;
```
mDuo13:
```suggestion
      text-decoration-color: $purple;
```

styles/_docs.scss
```diff
@@ -0,0 +1,352 @@
...
+    margin-left: auto;
+    margin-right: auto;
+  }
+
+  .nav-link {
+    border: none !important;
+
+    &:hover {
+      text-decoration: underline;
+      text-decoration-color: #9A52FF;
```
mDuo13:
```suggestion
      text-decoration-color: $purple;
```

styles/_pages.scss
```diff
@@ -1060,7 +969,8 @@
 
 .page-docs-index {
   &::before {
-    background-image: url(../img/backgrounds/bg-docs.png);
```
mDuo13, line 1063: In cases where this PR removes the last use of a particular image (as best I can tell, `bg-docs.png` is only used this once), please remove the corresponding file so the repo doesn't accumulate unused files.

img/backgrounds/docs-splash.svg
mDuo13, line 1: The way the folder structure is intended to work is that the top-level `img/` folder is for images used in the documentation _contents_, and the `assets/img/` folder is for images used in the templates. (That's not consistently how it's been used, but I would like to try and get closer to the ideal rather than farther.) So, preferably, please move this to `assets/img/backgrounds/` and update the SCSS accordingly.

assets/css/devportal2022-v16.css
mDuo13, line 1: When making changes like this that add new styles, we like to increment the version number to prevent the case where users' browsers load the CSS file from cache, causing the page to display incorrectly. (It's not necessary when only removing styles.)

To do that:
1. Modify the `styles/package.json` file to change the version number in all three builds. For example, change it to `devportal2022-v17.css`
2. Rebuild the CSS (`yarn run build-css` from within `styles/`)
3. Delete the old CSS file (v16) from git and add the new one (v17) (`git rm assets/css/devportal2022-v16.css && git add assets/css/devportal2022-v17.css`)
4. Modify the CSS path in `template/base.html.jinja` on line 55 to match the new version number.

styles/_pages.scss
```diff
@@ -1060,7 +969,8 @@
...
+    background-image: url(../../img/backgrounds/docs-splash.svg);
```
mDuo13, line 972: If you move the file as I recommended above, update this path as well.

```suggestion
    background-image: url(../../assets/img/backgrounds/docs-splash.svg);
```

styles/xrpl.scss
```diff
@@ -69,6 +69,7 @@ $line-height-base: 1.5;
 @import "_video.scss";
 //@import "_top-banner.scss";
 @import "_toml-checker.scss";
+@import "_docs.scss";
```
mDuo13: If you rename this file as I recommended above, don't forget to change it here, too.

```suggestion
@import "_docs-home.scss";
```

template/page-docs.html.jinja
```diff
@@ -1,126 +1,426 @@
...
+          </div>
+        </div>
+      </div>
     </div>
-    {% endfor %}
-  </div>
-</section><!--/.doc-index-->
+  </section><!-- Browse by recommended and Generate Testnet Credentials -->
+  <section class="container-new">
+    <a href="./docs-index.html" class="external-link mr-auto">See full documenation index <i aria-hidden="true" class="fa fa-external-link"></i></a>
```
mDuo13: Just noted this on the design figma, which this matches, but I don't think this should be styled as an external link, since it doesn't actually take you off-site.

template/page-docs.html.jinja
```diff
@@ -1,126 +1,426 @@
...
+      {% endfor %}
     </div>
   </section>
 
-  <section class="container-new mb-50-until-sm" id="software-and-sdks">
-    <div class="d-flex flex-column-reverse w-100">
-      <h2 class="h4">{% trans %}Software and SDKs{% endtrans %}</h2>
-      <h6 class="eyebrow mb-3">{% trans %}Start Developing{% endtrans %}</h6>
+  <section class="container-new ">
+    <h4 class="pb-4">Getting Started</h4>
```
mDuo13: There are a lot of cases here where new user-facing text isn't marked in `{% trans %}` tags to make it translatable. For consistency and to make it easier on @develoQ or anyone else doing translation, we should update these.

```suggestion
    <h4 class="pb-4">{% trans %}Getting Started{% endtrans %}</h4>
```

I haven't tagged all of them but in general it applies to any user-visible text written directly into the template. (For now, I think we're punting on text in variables/macros.)

template/page-docs.html.jinja
```diff
@@ -1,126 +1,426 @@
...
+    <div class="card-grid card-grid-2xN quickstart-card">
+      <div class="col">
+        <a href="./xrpl-quickstart.html" class="card float-up-on-hover">
+          <h5 class="mt-7"> Quickstart to XRP Ledger </h5>
```
mDuo13:
```suggestion
          <h5 class="mt-7">{% trans %}Quickstart to XRP Ledger{% endtrans %}</h5>
```

mDuo13:
```suggestion
          <h5 class="mt-7">{% trans %}Quickstart to XRP Ledger{% endtrans %}</h5>
```

template/page-docs.html.jinja
```diff
@@ -1,126 +1,426 @@
...
+    },
+    {
+      "src": "./assets/img/backgrounds/docs-tokenization@2x.png",
+      "title": "Tokenization",
+      "url": "https://www.youtube.com/embed/Oj4cWOiWf4A?rel=0&amp;showinfo=0&amp;autoplay=1",
+    },
+    ]
+  %}
+  {%
+    set devTools = [
```
mDuo13, line 212: I started writing a long thing about how you could dynamically set this in a more translation-friendly way using `pages|selectattr(...)`, but (a) it's very Dactyl-specific, and (b) the dev tools don't actually have blurbs in the config file where you could the (language-appropriate) one, so you'd have to fix that while you're at it.

So, I guess, nevermind. This is good enough.

mDuo13 (changes requested):
Overall, looks great. I dug a bit deeper into the changes and requested a lot of little nitpicky fixes, but in general it looks good.

Silkjaer (approved):
LGTM!

template/page-docs.html.jinja
```diff
@@ -1,126 +1,426 @@
...
+        alt={{title}}
+      />
+      <h5 class="row">
+        <div class="nav-link">{{ title }}</div>
+      </h5>
+      <p class="row faded-text flat-card-padding">
+        {{ description }}
+      </p>
+      <div class="col align-button-on-bottom">
+        <div class="btn btn-primary btn-arrow-out" id="{{href}}-button">{{ linkText }}</div>
```
mDuo13: These should be horizontal, not diagonal arrows.

```suggestion
        <div class="btn btn-primary btn-arrow" id="{{href}}-button">{{ linkText }}</div>
```

mDuo13 (approved):
One minor thing to fix, otherwise this looks good.

PR #1875, opened 2023-04-19
Contributing: clarify hard-wrapping and code sample line length

mDuo13:
It came up in reviewing a bounty today that a couple details are missing from our in-repo conventions.


oeggert (approved):
Looks good to me.

PR #1863, opened 2023-04-07
Remove in-development docs for sidechains and EVM sidechains

amarantha-k:
When merged, this addresses a part of Issue #1820

PR #1894, opened 2023-04-27
Fix Serialization table & STIssue link

mDuo13:
Fixes two issues with master:

1. #1844 uncommented the table entry for "STIssue" but the corresponding section it linked to was commented out by a separate, concurrent PR, resulting in a broken link in master. I failed to notice this issue because it showed up after a merge commit after I had approved the PR; it came to my attention when the link checker started failing on other PRs.
2. #1863 broke the formatting of the type list table by commenting out a line. For future reference, if you try to comment out a line in the middle of a table, that gets treated as an HTML block and interrupts the table formatting, resulting in something that looks like this: 

Since the "STIssue" type is likely to become useful again when either AMM or Sidechains get proposed as an amendment (wouldn't be surprised if other standards use this type either), I left the type entry for the table commented out, but moved it past the bottom of the table so that it won't interfere with any formatting. We'll want to move it back to its proper place if/when we uncomment it.


ckniffen (approved):

amarantha-k (approved):
Thanks for taking care of these issues and the additional info for future reference. LGTM

PR #1895, opened 2023-05-01
adds apex 2023 event adds meetup

AKCodez:


template/page-events.html.jinja
```diff
@@ -328,6 +328,26 @@
...
+    "link": "https://www.meetup.com/xrp-ledger-espana-madrid-y-barcelona/events/292597878",
+    "location": "Madrid",
+    "date": "April 29, 2023",
+    "image": "Madrid.png",
+    "end_date": "April 29, 2023" 
+    },
+    {
+    "name": "APEX 2023: The XRPL Developer Summit",      
+    "description": "Apex XRPL Developer Summit is the annual event where developers, contributors, and thought leaders come together to learn, build, share, network, and celebrate all things XRP Ledger.",
+    "type": "ama",
```
mDuo13: should be type: conference, shouldn't it?

mDuo13 (approved):
Saw one thing that probably needs to change, otherwise looks ok

template/page-events.html.jinja
AKCodez: So, they added new event types and filters. See attached 
image

AKCodez: Nvm, I overlooked this, you are right it is indeed a conference, nice catch

template/page-events.html.jinja
```diff
@@ -328,6 +328,26 @@
...
     },
+    {
+    "name": "XRPL Community Meetup: Madrid",      
+    "description": "We're excited to host our second Miami meetup for XRP Ledger community members on April 6th from 6-8pm, featuring Marco Neri, Developer Advocate at Ripple, who will join us to give a presentation on the XRP Ledger.",
+    "type": "meetup",
+    "link": "https://www.meetup.com/xrp-ledger-espana-madrid-y-barcelona/events/292597878",
+    "location": "Madrid",
+    "date": "April 29, 2023",
+    "image": "Madrid.png",
+    "end_date": "April 29, 2023"
```
sgramkumar, line 339: The dates seem to be having some discrepancy?

> our second Miami meetup for XRP Ledger community members on **April 6th**

>    "location": "Madrid",
    "date": "**April 29, 2023**",

PR #1896, opened 2023-05-01
Tutorials page

pdp2121:
Implement a new design for tutorial page:

Screenshot 2023-05-01 at 5 27 25 PM

Preview link:
https://xrplf.github.io/xrpl-dev-portal/pr-preview/tutorial-page/tutorials.html


Silkjaer (commented):
Looks good, and in line with the Docs update! The interoperability link should be removed though, however the redirect is fine to stay in place in case someone has made direct links from elsewhere.

styles/_tutorials.scss
```diff
@@ -0,0 +1,169 @@
+/* Tutorials ---------------------------------------------------------------- */
+.tutorial-content {
+  #beginner-cards {
+    .card:nth-child(1) .card-footer {
+      background-image: url(../img/cards/4col-magenta-3.svg);
+    }
+  }
+
+  #sdk-cards {
+    .card:nth-child(1) .card-footer {
```
ckniffen: Curious about why different sections have different repeating colors.  I wonder if all cards should use the same color order.  Would make it easier if more sections get added.

ckniffen:
The stylesheet version will need to be bumped like was described in this comment. https://github.com/XRPLF/xrpl-dev-portal/pull/1866#discussion_r1191611959

pdp2121:
> The stylesheet version will need to be bumped like was described in this comment. [#1866 (comment)](https://github.com/XRPLF/xrpl-dev-portal/pull/1866#discussion_r1191611959)

@ckniffen Makes sense to me. Since yours is 17 should mine be 18?

styles/_tutorials.scss
bugsbunnies: > Curious about why different sections have different repeating colors. I wonder if all cards should use the same color order. Would make it easier if more sections get added.

There isn't any specific reason other than having random variety of colors here just makes the page less boring. We don't have that many cards at the moment and having the same color start each section seems monotonous. Since each card is a different thing and we're not mapping each other to a specific topic, might as well keep it fluid.

pdp2121:
> Looks good, and in line with the Docs update! The interoperability link should be removed though, however the redirect is fine to stay in place in case someone has made direct links from elsewhere.

Thanks @Silkjaer , I have removed the interoperability link (still keeping redirect in config).

styles/_tutorials.scss
ckniffen: I was thinking maybe a mixin would make sense where we give it an offset of which color to start a section with and then the order goes the same.

styles/_tutorials.scss
```diff
@@ -0,0 +1,166 @@
...
+      background-image: url(../img/cards/3col-green-purple.svg);
+    }
+    .card:nth-child(2) .card-footer {
+      background-image: url(../img/cards/3-col-light-blue-2.svg);
+    }
+  }
+
+  #servers-cards {
+    .card:nth-child(1) .card-footer {
+      background-image: url(../img/cards/4col-orange-yellow.svg);
```
ckniffen: Some use "3-col" named images and others use 4 col for the same set.  That seems a bit odd.

styles/_tutorials.scss
pdp2121: mixin with an offset sounds like a good idea! Will implement it

styles/_tutorials.scss
pdp2121: I don't really know the distinction between 3col and 4col for the svg colors. Just tried to match the design. Any ideas?

pdp2121:
Will rebase to bump css version to 18 once https://github.com/XRPLF/xrpl-dev-portal/pull/1892 is merged

ckniffen (approved):

tequdev (changes requested):
Japanese version should be supported as well.

dactyl-config.yml L1236-L1243

```
    -   name: チュートリアル
        html: tutorials.html
        parent: docs.html
        top_nav_grouping: Article Types
        template: pagetype-tutorials.html.jinja
        blurb: XRP Ledgerのチュートリアルは、XRP Ledgerを学び、使い始めるための手順と、高度なユースケースでLedgerを使用するための手順を説明しています。
        targets:
            - ja
```

pdp2121:
> Japanese version should be supported as well.
> 
> dactyl-config.yml L1236-L1243
> 
> ```
>     -   name: チュートリアル
>         html: tutorials.html
>         parent: docs.html
>         top_nav_grouping: Article Types
>         template: pagetype-tutorials.html.jinja
>         blurb: XRP Ledgerのチュートリアルは、XRP Ledgerを学び、使い始めるための手順と、高度なユースケースでLedgerを使用するための手順を説明しています。
>         targets:
>             - ja
> ```

Thanks @develoQ ! I have added to the config file.

intelliot (approved):

tequdev (approved):

PR #1905, opened 2023-05-09
Updates send a memo code sample

sibelly:


content/_code-samples/send-a-memo/js/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "description": "",
+  "main": "send-a-memo.js",
+  "scripts": {
+    "start": "node send-a-memo.js"
+  },
+  "author": "",
+  "license": "ISC",
+    "dependencies": {
+      "xrpl": "^2.0.0"
+    }
```
mDuo13: I don't think this section should be indented.

```suggestion
  "dependencies": {
    "xrpl": "^2.0.0"
  }
```

content/_code-samples/send-a-memo/js/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "js",
+  "version": "1.0.0",
+  "description": "",
+  "main": "send-a-memo.js",
+  "scripts": {
+    "start": "node send-a-memo.js"
+  },
+  "author": "",
+  "license": "ISC",
```
mDuo13: Technically, the dev portal's original contents are offered under the MIT, not ISC license. (I think we may switch to some Creative Commons 4 license which is more appropriate for documentation.)

For very basic code samples like this, I think it's more appropriate to declare them public domain—under many jurisdictions, copying a sample like this would fall under Fair Use anyway. The **CC0-1.0** license declares something to be public domain, so that would remove ambiguity over that.

```suggestion
  "license": "CC0-1.0",
```

mDuo13 (approved):
Good catch. I also notice that the comment on line 26 of the memo has "MemoFormat values:" duplicated at the start of the comment (I suspect also carried over from Python like the `print()` instead of `console.log()`.) Wouldn't complain if you fixed that part up, too. ;)

PR #1907, opened 2023-05-15
Fix inaccuracy in load saved ledger

mDuo13:
The note was about stand-alone mode more generally, but for this specific feature and use case the reasoning is a bit different.


amarantha-k (approved):
LGTM

PR #1913, opened 2023-05-18
Ecosystem

AKCodez:


assets/img/uses/right-arrow.png
mDuo13, line 1: It's only ~2kb but I would love if we could reuse arrows from elsewhere such as the one we use in other buttons, e.g. `assets/img/icons/arrow-right.svg` instead of having separate files for every arrow.

assets/img/uses/modalLogos/Xumm.png
mDuo13, line 1: Aside from being a png instead of an svg, is this different than the one we already have at `assets/img/uses/xumm-wallet.svg`?

assets/js/ecosystem/arrow-animation.json
mDuo13, line 1: This is OK as-is, but I think in the past we've used CSS animations for this type of thing. I suspect that might be smaller (using less bandwidth) in this case, too since the `@keyframes` should be relatively short for a simple up-down bounce animation.

styles/_use-cases.scss
```diff
@@ -30,13 +95,422 @@
...
+  display: none;
+  position: fixed;
+  z-index: 1000;
+  left: 0;
+  top: 0;
+  width: 100%;
+  height: 100%;
+  overflow: auto;
+  background-color: rgba(0, 0, 0, 0.4);
+}
```
mDuo13: A lot of this is redefining / overriding [Bootstrap `modal` styles](https://getbootstrap.com/docs/4.1/components/modal/) that already exist. Can we simplify by using the existing Bootstrap modals?

template/page-uses.html.jinja
```diff
@@ -345,86 +358,153 @@
...
 
-    <ul class="card-grid card-grid-3xN ls-none mt-4 pt-lg-2" id="use-case-card-grid">
+    <ul class="card-grid card-grid-4xN ls-none mt-4 pt-lg-2" id="use-case-card-grid">
       {% set uses = [
-        
           {"id": "infrastructure",
             "title": _("Infrastructure"),
-            "description": _("Build and operate components or systems that help the functionality of the XRP Ledger, such as Nodes, dev tools, storage, security and more.")},
+            "number": 7,
+            "description" : "Build and operate components or systems that help the functionality of the XRP Ledger, such as Nodes, dev tools, storage, security and more."
```
mDuo13: The `_(...)` is there to make the strings translatable. If you remove it like this, the Japanese translation won't display properly.

assets/js/modal.js
```diff
@@ -0,0 +1,235 @@
+$(document).ready(() => {
```
mDuo13, line 1: This seems like the kind of thing that should probably be done once at build-time in the template rather than in-browser on the client side.

mDuo13 (changes requested):
The page _mostly_ looks good, but the implementation is a little sloppy. I'd like to see things slimmed down to use stuff we've already got where possible, and any assets we're no longer using removed.

Also, on some screen sizes, the modals don't work quite as well as we'd like. The NFTs one in particular is big and runs off the edge of the screen on some small / old phones:



Perhaps more importantly, the logos seem to get stretched/distorted on some tablet screen sizes.

assets/img/uses/right-arrow.png
AKCodez: The arrow designs are different, the new modal has a new arrow requested in the design so that is what is being integrated. 2kb should not be a big deal, considering the size of the lotti animations we have all over the website, also I know this a different topic but we should be hosting url's for the lotti animation JSONs instead of downloading them into the codebase, we would save a lot of space in the codebase.

AKCodez: Pushed changes for the mobile modal issues.

assets/img/uses/modalLogos/Xumm.png
AKCodez: Yes the size of the assets differ greatly, tried integrating the one that is already there, did not size well with the other items.

assets/js/ecosystem/arrow-animation.json
AKCodez: When migrating to react, we will have much more freedom in integrating animations, will integrate a lighter animation once we migrate to react

styles/_use-cases.scss
AKCodez: I changed the classname from .modal to .modal-uses to avoid any conflicts with bootstrap's modal styles

template/page-uses.html.jinja
AKCodez: Was not aware, added them back.

assets/js/modal.js
AKCodez: Thank you for your suggestion to move this logic server-side. However, the data for the modal is dynamic and can change based on user interactions on the page. To accommodate this, we need to initialize and update the modal on the client-side. Using $(document).ready() ensures that the DOM is fully loaded before we attach any handlers or manipulate elements, which is crucial for the functionality we're implementing.

template/page-uses.html.jinja
```diff
@@ -284,19 +300,16 @@
...
-  "gaming": _("Gaming"),
-  "security": _("Security"),
-  "payments": _("Payments"),
-  "web_monetization": _("Web Monetization"),
-  "sustainability": _("Sustainability"),
-  "cbdcs": _("CBDCs"),
-  "custody": _("Custody"),
-  "defi": _("DeFi"),
-  "other": _("Other"),
+  "interoperability": "Interoperability",
```
mDuo13: Missing `_(...)` here too.

template/page-uses.html.jinja
```diff
@@ -49,6 +49,14 @@
     "category_id": "custody",
     "category_name": _("Custody"),
     "link": "https://www.bitgo.com/",
+  },
+    {
+    "id": "gatehub",
+    "title": _("Gatehub"),
+    "description": _("Gatehub XRP Ledger Markets is an explorer to track Gatehub's inssuances on the XRP Ledger."),
+    "category_id": "custody",
+    "category_name": "Custody",
```
mDuo13: You missed the `_(...)` on the category names in the update.

mDuo13 (approved):
I would like to see the translation stuff restored in a couple places. Otherwise it's good.

PR #1931, opened 2023-06-05
Update NFT ref info

DennisDawson:
Add entries for new NFT-related fields in transaction metadata. Add `NFTokenCountTracking` and `ValidNFTokenPage` entries to the Invariant Checking topic.

### DIFF
--- a/content/concepts/consensus-network/invariant-checking.md
+++ b/content/concepts/consensus-network/invariant-checking.md
@@ -142,6 +142,7 @@ The XRP Ledger checks all the following invariants on each transaction:
     - A NFTokenBurn transaction cannot change the number of minted NFTs.
     
 ### NFTokenCountTracking
+
 - **Invariant Condition(s):**
     - The page is correctly associated with the owner.
     - The page is correctly ordered between the next and previous links.
--- a/content/concepts/consensus-network/invariant-checking.md
+++ b/content/concepts/consensus-network/invariant-checking.md
@@ -133,10 +133,10 @@ The XRP Ledger checks all the following invariants on each transaction:
 ### ValidNFTokenPage
 
 - **Invariant Condition(s):**
-    - The number of minted or burned tokens can only be changed by `NFTokenMint` or `NFTokenBurn` transactions.
+    - The number of minted or burned NFTs can only be changed by `NFTokenMint` or `NFTokenBurn` transactions.
     - A successful NFTokenMint transaction must increase the number of NFTs.
     - A failed NFTokenMint transaction must not change the number of minted NFTs.
-    - A NFTokenMint transaction must not change the number of burned NFTs.
+    - A NFTokenMint transaction cannot change the number of burned NFTs.
     - A successful NFTokenBurn transaction must increase the number of burned NFTs.
     - A failed NFTokenBurn transaction must not change the number of burned NFTs.
     - A NFTokenBurn transaction cannot change the number of minted NFTs.
--- a/content/concepts/consensus-network/invariant-checking.md
+++ b/content/concepts/consensus-network/invariant-checking.md
@@ -1,7 +1,7 @@
 ---
 html: invariant-checking.html
 parent: consensus-network.html
-blurb: Understand what Invariant Checking is, why it exists, how it works, and what invariant checks are active.
+blurb: Understand what invariant checking is, why it exists, how it works, and what invariant checks are active.
 labels:
   - Blockchain
   - Security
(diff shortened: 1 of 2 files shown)


content/references/protocol-reference/transactions/transaction-metadata.md
```diff
@@ -68,6 +68,15 @@ A `ModifiedNode` object contains the following fields:
 
 **Note:** If the modified ledger object has `PreviousTxnID` and `PreviousTxnLgrSeq` fields, the transaction always updates them with the transaction's own identifying hash and the index of the ledger version that included the transaction, but these fields' new value is not listed in the `FinalFields` of the `ModifiedNode` object, and their previous values are listed at the top level of the `ModifiedNode` object rather than in the nested `PreviousFields` object.
 
+## NFT Fields
+
+An `NFT` transaction response can contain the following fields:
```
mDuo13: Putting `NFT` in code font makes it seem like you're referring to a specific thing with that exact name, but there isn't one. Maybe soften it up, like so:

```suggestion
Transactions involving NFTs can contain the following fields in the metadata:
```

Also, "can contain" is vague. Is it possible to be more specific about when it does and doesn't?

Based on these fields being lower-case, I am guessing that they are "non-canonical" fields that are calculated at the API layer and are not part of the hashed metadata—is that right? And are they returned by P2P `rippled` servers or only by Clio servers? These details are good to know.

content/concepts/consensus-network/invariant-checking.md
```diff
@@ -130,6 +130,25 @@ The XRP Ledger checks all the following invariants on each transaction:
...
+    - The number of minted or burned NFTs can only be changed by `NFTokenMint` or `NFTokenBurn` transactions.
+    - A successful NFTokenMint transaction must increase the number of NFTs.
+    - A failed NFTokenMint transaction must not change the number of minted NFTs.
+    - A NFTokenMint transaction cannot change the number of burned NFTs.
+    - A successful NFTokenBurn transaction must increase the number of burned NFTs.
+    - A failed NFTokenBurn transaction must not change the number of burned NFTs.
+    - A NFTokenBurn transaction cannot change the number of minted NFTs.
+    
+### NFTokenCountTracking
+- **Invariant Condition(s):**
```
mDuo13: Leave a blank line before and after bulleted lists to make sure they parse correctly.

```suggestion
### NFTokenCountTracking

- **Invariant Condition(s):**
```

mDuo13 (changes requested):

mDuo13:
Changes look good although I'm still wondering about the two details I asked about: are these part of the hashed binary metadata or added by the server? And are they Clio-exclusive? (I vaguely recall hearing something about Clio adding more metadata)

DennisDawson:
@mDuo13 , the values are Clio-exclusive, and added by the server, not stored in the metadata. I'll add a note to that effect.

content/references/protocol-reference/transactions/transaction-metadata.md
ckniffen: These fields are in `rippled@1.11`

content/references/protocol-reference/transactions/transaction-metadata.md
```diff
@@ -68,6 +68,15 @@ A `ModifiedNode` object contains the following fields:
 
 **Note:** If the modified ledger object has `PreviousTxnID` and `PreviousTxnLgrSeq` fields, the transaction always updates them with the transaction's own identifying hash and the index of the ledger version that included the transaction, but these fields' new value is not listed in the `FinalFields` of the `ModifiedNode` object, and their previous values are listed at the top level of the `ModifiedNode` object rather than in the nested `PreviousFields` object.
 
+## NFT Fields
+
+Transactions involving NFTs can contain the following fields in the metadata. These values are added by the Clio server at request time and are not stored in the hashed binary  metadata:
```
shawnxie999: extra space  in `binary  metadata`

content/references/protocol-reference/transactions/transaction-metadata.md
```diff
@@ -68,6 +68,15 @@ A `ModifiedNode` object contains the following fields:
...
+
+| Field               | Value                     | Description                |
+|:--------------------|:--------------------------|:---------------------------|
+| `nftoken_id`        | String                    | Shows the `NFTokenID` for the `NFToken` that changed on the ledger as a result of the transaction. See [NFTokenID](nftoken.html#nftokenid) |
```
shawnxie999: I think we should be more specific here. This metadata field is only present if the tx is a `NFTokenMint` or `NFTokenAcceptOffer`, so that users know that they need to catch this field _only_ for these two transactions.

content/references/protocol-reference/transactions/transaction-metadata.md
```diff
@@ -68,6 +68,15 @@ A `ModifiedNode` object contains the following fields:
...
+| `nftoken_ids`       | Array                     | Shows all the `NFTokenIDs` for the `NFTokens` that changed on the ledger as a result of the transaction. |
```
shawnxie999: same reason as above, should specify that this field only applies for `NFTokenCancelOffer`

content/references/protocol-reference/transactions/transaction-metadata.md
shawnxie999: should we be explicit that these new fields are for responses `tx` and `account_tx` APIs only? People may be confused after they submit a transaction and don't see the field.

ckniffen: Will they be present on `nft_history` too?

shawnxie999: @ckniffen no lol. It's only for rippled. i def forgot about it, it was a while ago. But i think i will create a ticket for it

shawnxie999 (approved):

mDuo13 (approved):

PR #1932, opened 2023-06-05
Update ecosystem.svg

MorganBergen:
Formally called ripple-lib, but now called xrpl.js for Javascript removed extraneous Javascript for the reason that .js is already provided

https://xrpl.org/software-ecosystem.html#programming-libraries


mDuo13 (approved):
This is another area where these changes are a step in the right direction, but we really should do _more_, since the Data API is not a good example. Maybe "xApps" is a better example of middleware that's currently in use and supported?

PR #1639, opened 2022-12-16
[DO NOT MERGE] xrpl-py-2.0 changes + bug fixes

connorjchen:
- Fixed bugs like wrong variable names to make tests run
- Changed tests to use xrpl-py-2.0 wallet changes and xrpl-py new alias functions

TODO
- pip install xrpl-py==2.0.0b0 to get new beta changes and test all py files again
- Never tested files in build-a-wallet (need to set up stuff to run wx)
- Never tested key_derivation
- monitor-payments-websocket - read_amount_received no base transaction
- tx-serialization address.py - issues importing local base58
- use-tickets.py - ticket_sequence not valid parameter of AccountSet

PR #1933, opened 2023-06-05
feat: Update xrpl-dev-portal to use xrpl-py 2.0

JST5000:
In preparation for breaking changes to the xrpl-py library, this PR updates the code samples & tutorials in order to follow new best practices. 

This should NOT be merged until xrpl-py 2.0 is released. (This will be set as a draft PR until that point)

**This PR may be most digestible if read on a commit-by-commit basis. Most commits only migrate one of these breaking changes.**

## Breaking Changes

**Updates to function names for signing/submitting:**

* `send_reliable_submission` is now `submit_and_wait`
    * `submit_and_wait` can also handle the `autofill` and `sign` steps automatically (similar to how `submit_and_wait` works with xrpl.js)
* `submit_transaction` is now `submit`
* `safe_sign_and_submit_transaction` is now `sign_and_submit`
     * The `wallet` parameter here has been moved after `client` to be consistent with `submit_and_wait`
* `safe_sign_transaction` is now `sign`
* `safe_sign_and_autofill_transaction` is now `autofill_and_sign`
    * The `wallet` parameter here has been moved after `client` to be consistent with `submit_and_wait`

**`Wallet` Class**

* Removed `sequence` field
* Changed constructor to take in public and private key as required parameters
* Changed class methods to use the Ed25519 cryptographic algorithm by default
* Changed `classic_address` to `address` and made it read-only

`AccountSetFlags`
* `AccountSetAsfFlags` now contains Account Set Flags with the `ASF...` prefix.
* `AccountSetFlags` now is exclusively transaction flags.

**Core Keypairs**

* Changed the `generate_seed` method to take a hex string instead of a UTF-8 string
* Ed25519 key string is formatted with padded zeros if length is less than 32 bytes
* `sign` was made synchronous instead of asynchronous (moving a fee check out).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ out/
 package-lock.json
 yarn-error.log
 /.idea
+.venv/
--- a/content/_code-samples/airgapped-wallet/py/airgapped-wallet-TEST.py
+++ b/content/_code-samples/airgapped-wallet/py/airgapped-wallet-TEST.py
@@ -12,6 +12,7 @@ from xrpl.core import keypairs
 from xrpl.utils import xrp_to_drops
 from xrpl.models.transactions import Payment
 from xrpl.transaction import safe_sign_transaction
+from xrpl.wallet.main import Wallet
 
 
 def create_wallet(silent: False):
@@ -63,7 +64,7 @@ def sign_transaction(xrp_amount, destination, ledger_seq, wallet_seq, password):
     seed = crypt.decrypt(seed)
 
     print("4. Initializing wallet using decrypted private key")
-    _wallet = wallet.Wallet.from_seed(seed=seed.decode())
+    _wallet = Wallet.from_seed(seed=seed.decode())
 
     validated_seq = ledger_seq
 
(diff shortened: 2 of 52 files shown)


mDuo13:
How does this PR relate to #1639? Is this a companion or replacement for that one?

JST5000:
> How does this PR relate to #1639? Is this a companion or replacement for that one?

@mDuo13 This would be a replacement for that PR. I didn't realize we'd created a renamed `autofill_and_submit` - I'll add that into this PR, and then we should close the old one from Connor. (Mostly I want to make sure that we're making all the changes at once - so I'll add in his changes to this one)

JST5000:
I went through and ran every modified snippet to verify they would work with xrpl-py-2.0. All did what was expected (noticed a couple small fixes doing that which were in the last couple commits)

ckniffen (approved):

mDuo13 (approved):
Overall looks very thorough and nicely showcases the improvements in xrpl-py 2.0. I removed one import that was rendered redundant by the changes; otherwise, looks great!

PR #1937, opened 2023-06-07
Apex banner

AKCodez:


styles/_top-banner.scss
```diff
@@ -1,69 +1,66 @@
...
-    background: url(../img/backgrounds/bg-apex-banner.svg);
-    background-position: left;
-    background-size: cover;
-    background-color: #7919FF;
-    // padding: 2px 0;
+  height: 46px;
+  background: url(../img/backgrounds/bg-apex-banner.svg);
+  background-position: left;
+  background-size: cover;
+  background-color: #7919ff;
```
mDuo13: I hate to commit the PR sin of asking you to fix things that were that way before, but as long as you're updating this file, I would appreciate it if you could go through this file and replace all the hard-coded colors with variables using the palette defined in `_colors.scss`. For example:

```suggestion
  background-color: $blue-purple-500;
```

styles/_top-nav.scss
```diff
@@ -1,15 +1,54 @@
+// marquee
+@keyframes marquee-horizontal {
+  0% {
+    transform: translateX(0);
+  }
+  100% {
+    transform: translateX(-826px);
+  }
+}
```
mDuo13: I think you can remove this code now that we're not doing the scrolling thing, right?

template/component-top-nav.html.jinja
```diff
@@ -1,13 +1,6 @@
...
-    <span><p class="mb-0 no-wrap">{% trans %}The XRP Ledger now has Native NFT support.{% endtrans %}</p></span>
-    <span><a href="https://xrpl.org/non-fungible-tokens.html?utm_source=web&utm_medium=web&utm_campaign=pencil_banner" class="btn btn-outline-secondary">Learn More</a></span>
-  </div>
-  <div class="d-md-none d-block justify-content-center">
-    <span><p class="mb-0">{% trans %}The XRP Ledger now has Native{% endtrans %}</p></span>
-    <span><p class="mb-0">{% trans %} NFT support.{% endtrans %}<a href="https://xrpl.org/non-fungible-tokens.html?utm_source=web&utm_medium=web&utm_campaign=pencil_banner" class="btn btn-outline-secondary">Learn More</a></p></span>
-  </div>
-</div>#}
+<a href="http://register.apexdevsummit.com/cfs" target="_blank" class="w-inline-block banner-container">
+  <img id="apex-banner" />
```
mDuo13:
```suggestion
  <img id="apex-banner" alt="Register for APEX Dev Summit 2023 in Amsterdam" />
```

assets/css/devportal2022-v18.css
mDuo13, line 1: I think you need to bump the CSS version here. While you're at it, I think the site has accumulated a few versions of the CSS file—can you delete those, too?

mDuo13 (changes requested):
The banner image needs alt text (suggested some below).

It also looks, uh, very large on my 1440p main monitor. We're talking text larger than the h1 of the page. Even on my 1080p monitor it feels a bit "too" big, personally.



Perhaps it would be best to add an "xl" size for the banner that's used on 1080p screens and up.

That said, the sticky position is a big improvement in terms of ceding space to the relevant contents when you scroll down, and I'm very glad it's not animated anymore.

I noted a couple other "nice-to-have" cleanup tasks, but I won't hold up the PR based on those alone.

mDuo13:
!rebuild

mDuo13 (approved):
Got the approval from design and I'm happy with the new HTML implementation so this is a :+1: from me now.

PR #1941, opened 2023-06-08
Contrib: guidelines for linking by anchor

mDuo13:
Minor style guideline to help avoid problems with broken links when pages are translated (resulting in the h1 anchor being different even though the page is the same).


amarantha-k (approved):
lgtm

PR #1943, opened 2023-06-08
remove interop use cases from IA.v3

oeggert:


mDuo13 (approved):

mDuo13:
I think the failure on the build job is from a race condition due to multiple jobs trying to push to gh-pages in parallel, so this should be safe to merge as-is despite that.

PR #1947, opened 2023-06-10
IAv3: Update Ledger Structure page

mDuo13:
Rewrite "Ledger Structure" for clarity, style, and accuracy, including updated diagrams.

## Context on "Ledger Structure" updates

I tried to keep the wording similar to the rewritten version where it was accurate, but I moved some things around to more clearly distinguish between the "header" fields and the non-header sections, the state data, transaction set, and validated status. I think the previous version was a little confusing because the state data and tx set have hashes in the header to identify/verify them.

The way it was presented with text interspersed with big diagrams wasn't working for me—especially because it was not obvious at a glance whether each picture went with the text above or below it—and I felt it would be better presented as a bulleted list. However, as @DennisDawson pointed out, the graphics were an important part in chunking the information. So, as a compromise, I exported smaller icons to put inside the bulleted list. I also updated the basic ledger diagram.

Changes in the ledger diagram:

- Updated colors and styles, based on some Figma work done by Natalia C. (a designer at Ripple).
- Moved the "Ledger Index" icon into the area with other parts of the header that are included in the hash calculation, because it _is_ one of the parts of the immutable ledger hash.
- Changed the icons for transaction set hash and state data hash, to better represent that these are hashes.
- Flipped the "roll up" arrows so that the diagram reads better left-to-right.
- Changed the general shape to be more of a "block"; the previous "page" design could be misleadingly interpreted as a "file".

I made some slight updates to the `include_svg` Dactyl filter and the SCSS files so that the new (Figma-exported) diagrams still render properly including being theme-responsive and allowing diagrams to float to the left of bullet points rather than always having their own line.

A couple more specific things I fixed in the previous text:

- "hash number" isn't a preferred term; I rewrote this to match our preferred usage.
- Each ledger's state data _does_ include all objects including ones that weren't changed since the previous one. (I think this was intending to explain how the server optimizes space by de-duplicating the storage, but it came across as sounding like you wouldn't have all the state data if you downloaded one whole ledger version—actually, you would. Ultimately the fact that the server de-duplicates ledger contents that don't change from version to version is a detail that only matters when administering a server and understanding data storage and retention requirements; it's not relevant for development, so I don't mention that in this page anymore.)
- I elided some of the nittier-grittier details like `sLCF_NoConsensusTime` since those largely don't matter. They're documented in the ledger header reference page for anyone who really needs to know.
- I added some color around the transaction set, state data, and header fields to further clarify how they're used/useful.
- I deleted the "Tree Format" section since it wasn't providing any information that's useful at this level.
- I added a brief section explaining why ledgers have both a hash and an index and when you might care about one over the other.

## Changes Removed from this PR

~~Move ledgers & consensus files out of the `concepts/xrpl/` folder to appropriately named subfolders~~
~~Redo nav so that "Consensus Protocol" is at the top level, after Servers¹ and before Ledgers~~

(These tasks were done in #1951.)


mDuo13:
The plan of splitting this from #1951 in a way that this can be a "fast-foward" didn't work _quite_ as planned, so I pushed a rebase here that strips the PR down to just the "Ledger Structure" updates.

mDuo13:
NFT-related broken links are from the latest IA.v3 (they were introduced by the reverse-merge to keep it up-to-date with master) and not new to this PR. I rebased this PR on IA.v3 because of potential conflicts in the stylesheet file.

content/concepts/ledgers/ledger-structure.md
```diff
@@ -11,51 +11,55 @@ The consensus protocol takes a previous ledger version as a starting point, form
...
+- {{include_svg("img/ledger-hash-icon.svg", "", classes="floating-diagram")}} The _ledger hash_, which uniquely identifies the ledger's contents. The hash is calculated so that if any detail of the ledger version changes, the hash is completely different, which makes it also like a checksum that shows that none of the data in the ledger has been lost, modified, or corrupted.
+- {{include_svg("img/ledger-parent-icon.svg", "", classes="floating-diagram")}} The _parent ledger hash_. A ledger version is largely defined by the difference from the _parent ledger_ that came before it, so the header also contains the unique hash of its parent ledger.
+- {{include_svg("img/ledger-timestamp-icon.svg", "", classes="floating-diagram")}} The _close time_, the official timestamp when this ledger's contents were finalized. This number is rounded off by a number of seconds, usually 10.
+- {{include_svg("img/ledger-state-data-hash-icon.svg", "", classes="floating-diagram")}} A _state data hash_ which act as a checksum on this ledger's state data.
```
DennisDawson: _state data hash_ which acts...

content/concepts/ledgers/ledger-structure.md
```diff
@@ -11,51 +11,55 @@ The consensus protocol takes a previous ledger version as a starting point, form
...
+- {{include_svg("img/ledger-tx-set-hash-icon.svg", "", classes="floating-diagram")}} A _transaction set hash_ which act as a checksum on this ledger's transaction set data.
```
DennisDawson: ...which acts...

content/concepts/ledgers/ledger-structure.md
```diff
@@ -11,51 +11,55 @@ The consensus protocol takes a previous ledger version as a starting point, form
...
 
-**Tip:** Sometimes, a ledger entry is called a "ledger node". For example, transaction metadata returns a list of `AffectedNodes`. Do not confuse this with a "node" (server) in the peer-to-peer network.
+## Ledger Index or Ledger Hash?
 
-In the case of transactions, the identifying hash is based on the signed transaction instructions, but the contents of the transaction object when you look it up also contain the results and metadata of the transaction, which are not taken into account when generating the hash.
+There are two different ways of identifying a ledger version, its ledger index and its ledger hash. These two fields can both identify a ledger, but they serve different purposes. The ledger index tells you _when_ to place a ledger in history, but not necessarily what its contents are. Two ledgers with the same ledger index might be the same, but they might be from different chains, or they could even be different candidates to be the next ledger on the same chain. Two ledgers with the same ledger hash are always completely identical, but the hash by itself doesn't tell you which one came first.
```
DennisDawson: There are two different ways of identifying a ledger version: its _ledger index_ and its _ledger hash_. 

I find the description hard to parse. It would probably be enough to say the ledger index establishes the position of the ledger in in the chain, while the ledger hash reflects the content of the ledger.

DennisDawson (approved):
Approved with suggestions. I like the new ledger image.

PR #1950, opened 2023-06-13
Reorder topics under new "Networks and Servers" and "Infrastructure" top-nav.

oeggert:
- Rename "XRP Ledger Servers" to "Networks and Servers"
- Amendments moved under "Networks and Servers"
- Parallel Networks moved under "Networks and Servers"
- Move Online Deletion & History Sharding to new "Data Retention" section
- Reorganize "Infrastructure" section
- Reword title for "Testing rippled Functionality in Stand-alone Mode"  to "Testing and Auditing"


mDuo13 (approved):
I think we still want to move some more of the configuration topics into subgroups, but that can happen later. This is great progress.

PR #1951, opened 2023-06-13
IA.v3: Ledger reorg only

mDuo13:
Contains _just_ the commits from #1947 that move files around so those changes can be reviewed separately from the rewrites to "Ledger Structure".

(Note: as long as this and #1947 are **not further rebased nor squashed** then git should allow for smooth sequential merging since the commits have the same IDs.)


amarantha-k (approved):
lgtm

PR #1953, opened 2023-06-13
Add conditional escrows, rename time escrows

DennisDawson:


content/_code-samples/quickstart/js/ripplex1-send-xrp.js
mDuo13, line 1: Should we drop "ripplex" from these filenames?

content/_code-samples/quickstart/js/quickstart.zip
mDuo13, line 1: There are still several places linking to the `quickstart.zip` file (they link to the file as stored in the master branch). Those will eventually become broken links when this is merged to master, although the link checker isn't catching them now because they aren't broken yet. Still, we should probably fix those at the same time as removing the zip.

I did a quick search and turned up 25 results in 15 files on IA.v3 that link to `quickstart.zip`. I'd recommend updating those at the same time as removing the file, maybe in a separate PR from this one with the escrow tutorials.

mDuo13 (changes requested):

DennisDawson:
Fixing these issues in new PRs post-IA.v3.

PR #1955, opened 2023-06-14
Delete xrp.md

DennisDawson:


mDuo13:
I think this PR also needs to remove the file from the dactyl-config, and remove the Japanese version of the file as well?

DennisDawson:
Dumping this PR, combining with #1956

PR #1957, opened 2023-06-15
initial intro rewrite

oeggert:
@DennisDawson 
I noticed that there was a second heading that seems to repeat the intro. I tentatively removed it since the info about minimum reserve requirements doesn't seem like something a person needs to know in a first-look into XRPL.


DennisDawson (approved):
These changes look fine. At some point, I'd like to revisit the intro to make it even more conversational and easily consumable, but I don't have good suggestions at the moment. Thanks for doing this.

PR #1959, opened 2023-06-15
IA.v3: Remove now-unused Production Readiness landing

mDuo13:
@oeggert will be doing some more tutorial reorganization for the new IA, but this gets one easy thing out of the way first.


oeggert (approved):
LGTM

PR #1962, opened 2023-06-17
Bullets corresponds to image in nftoken.md

zgrguric:
Update nftoken.md

### DIFF
--- a/content/references/protocol-reference/data-types/nftoken.md
+++ b/content/references/protocol-reference/data-types/nftoken.md
@@ -30,11 +30,15 @@ Unlike full-fledged [ledger entries](ledger-object-types.html), `NFToken` has no
 
 This composite field uniquely identifies a token, and consists of the following sections.
 
-A. 16 bits that identify flags or settings specific to the NFToken
-B. 16 bits that encode the transfer fee associated with this NFToken, if any
-C. A 160-bit account identifier of the issuer
-D. A 32-bit issuer-specified [`NFTokenTaxon`](https://www.merriam-webster.com/dictionary/taxon)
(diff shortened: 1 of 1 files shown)


content/references/protocol-reference/data-types/nftoken.md
```diff
@@ -30,11 +30,15 @@ Unlike full-fledged [ledger entries](ledger-object-types.html), `NFToken` has no
 
 This composite field uniquely identifies a token, and consists of the following sections.
 
-1. 16 bits that identify flags or settings specific to the NFToken
-2. 16 bits that encode the transfer fee associated with this NFToken, if any
-3. A 160-bit account identifier of the issuer
-4. A 32-bit issuer-specified [`NFTokenTaxon`](https://www.merriam-webster.com/dictionary/taxon)
-5. An (automatically generated) monotonically increasing 32-bit sequence number.
+A) 16 bits that identify flags or settings specific to the NFToken
```
DennisDawson, line 33: I like the suggestion (originally to use A. B. C. etc.). When I tried to build the page in that notation, unfortunately it doesn't render as a list. Fiddling around, I was able to make something like a list by inserting a line feed between each item. Still not perfect, but I agree that having the enumeration match the labels in the illustration makes much more sense. Thanks for taking the time to suggest and submit the change.

DennisDawson (approved):

content/references/protocol-reference/data-types/nftoken.md
zgrguric: Thank you for testing. And you are welcome.

PR #1971, opened 2023-06-23
Update outdated "ledger" key to "ledger_index" key

9oelM:
Relevant API methods:
- [account_info](https://xrpl.org/account_info.html)
- [account_objects](https://xrpl.org/account_objects.html#request-format)
- [account_lines](https://xrpl.org/account_lines.html#account_lines)

These methods all have `ledger_index` as a parameter, but it seems that it used to be `ledger` before. The docs are therefore outdated.

This PR just changes `ledger` to `ledger_index`, making the docs up to date with the latest API parameters.


mDuo13 (approved):
Excellent catch, thanks

PR #1976, opened 2023-06-24
rippled 1.11.0 changes

### LINKED ISSUE #1970: Document the NetworkID field (Version 1.11.0)
https://github.com/XRPLF/rippled/releases/tag/1.11.0
hi team, In the new version, I noticed that the NetworkID field has been added. could you explain in detail how I can use it ?

mDuo13:
- [x] Fix #1970 
- [x] Document https://github.com/XRPLF/rippled/pull/4404
- [x] Document https://github.com/XRPLF/rippled/pull/4459
- [x] Document https://github.com/XRPLF/rippled/pull/4352
- [x] Document https://github.com/XRPLF/rippled/pull/4291
- [x] Document https://github.com/XRPLF/rippled/pull/4406 (the amendment has already been added to the Known Amendments page, but the new field `FirstNFTSequence` needs to be added to the AccountRoot reference)


tequdev:
Is it possible to change the Japanese content (*.ja.md) as well, so that it can remain in English?

I think it's better to have the information that new parts are untranslated than to have the content remain old and not updated.

mDuo13:
> Is it possible to change the Japanese content (*.ja.md) as well, so that it can remain in English?
> 
> I think it's better to have the information that new parts are untranslated than to have the content remain old and not updated.

I agree that it's better, but it's a lot more work to find the appropriate places to update the Japanese docs every time we make a change. I would rather just mark the pages as needing translation updates and prioritize them.

content/references/http-websocket-apis/admin-api-methods/status-and-debugging-methods/feature.md
```diff
@@ -188,7 +188,7 @@ The response follows the [standard format][], with a successful result containin
 | `enabled`   | Boolean | Whether this amendment is currently enabled in the latest ledger. |
 | `name`      | String  | (May be omitted) The human-readable name for this amendment, if known. |
 | `supported` | Boolean | Whether the server knows how to apply this amendment. If this field is set to `false` (the server does not know how to apply this amendment) and `enabled` is set to `true` (this amendment is enabled in the latest ledger), this amendment may cause your server to be [amendment blocked](amendments.html#amendment-blocked-servers). |
-| `vetoed`    | Boolean | Whether the server has been instructed to vote against this amendment. |
+| `vetoed`    | Boolean or String | For most amendments, this is a boolean value indicating whether the server has been instructed to vote against this amendment. For amendments that are marked as obsolete in the code, this is the string `Obsolete` instead. [Updated in: rippled 1.11.0][] |
```
DennisDawson, line 191: This is subjective and nit-picky, but I would italicize _obsolete_ since we're talking about a word as a word. "For amendments that are marked as _obsolete_ in the code...."

content/references/http-websocket-apis/public-api-methods/account-methods/account_info.md
```diff
@@ -61,14 +61,13 @@ The request contains the following parameters:
...
-| `signer_lists` | Boolean                    | _(Optional)_ If `true`, and the [MultiSign amendment][] is enabled, also returns any [SignerList objects](signerlist.html) associated with this account. [New in: rippled 0.31.0][] |
-| `strict`       | Boolean                    | _(Optional)_ If `true`, then the `account` field only accepts a public key or XRP Ledger address. Otherwise, `account` can be a secret or passphrase (not recommended). The default is `false`. |
+| `queue`        | Boolean                    | _(Optional)_ If `true`, return stats about [queued transactions](transaction-queue.html) sent by this account. Can only be used when querying for the data from the current open ledger. Not available from servers in [Reporting Mode][]. |
+| `signer_lists` | Boolean                    | _(Optional)_ If `true`, return any [SignerList objects](signerlist.html) associated with this account. [New in: rippled 0.31.0][] |
 
-The following fields are deprecated and should not be provided: `ident`, `ledger`.
+The following fields are deprecated and should not be provided: `ident`, `ledger`, `strict`.
```
DennisDawson, line 68: I always cringe at _should_, because it suggests there are times you would use them. Perhaps "The following fields are deprecated." or "The following fields are deprecated and ignored."

content/references/protocol-reference/transactions/transaction-common-fields.md
```diff
@@ -127,6 +128,27 @@ Example of a transaction with a Memos field:
 }
 ```
 
+## NetworkID Field
+[New in: rippled 1.11.0][]
+
+The `NetworkID` field is a protection against "cross-chain" transaction replay attacks, preventing the same transaction from being copied over and executing on a [parallel network](parallel-networks.html) that it wasn't intended for. For compatibility with existing chains, the `NetworkID` field **MUST BE OMITTED** for any network with a Network ID of 1024 or less, but **MUST BE INCLUDED** for any network with a Network ID of 1025 or greater. The following table shows the status and values for various known networks:
```
DennisDawson: If this is important enough for bold and all caps, does it warrant a call out as a warning?

content/references/http-websocket-apis/public-api-methods/account-methods/account_info.md
```diff
@@ -206,13 +205,32 @@ The response follows the [standard format][], with the result containing the req
...
+
+| `Field`                | Type    | Description                               |
+|:-----------------------|:--------|:------------------------------------------|
+| `defaultRipple`        | Boolean | If `true`, the account allows [rippling](rippling.html) on its trust lines by default. |
+| `depositAuth`          | Boolean | If `true`, the account is using [Deposit Authorization](depositauth.html) and does not accept any payments from unknown parties. |
+| `disableMasterKey`     | Boolean | If `true`, the account's [master key pair](cryptographic-keys.html) is disabled. |
+| `disallowIncomingCheck` | Boolean | If `true`, the account does not allow others to send [Checks](checks.html) to it. |
+| `disallowIncomingNFTokenOffer` | Boolean | If `true`, the account does not allow others to make [NFT buy or sell offers](non-fungible-token-transfers.html) to it. |
+| `disallowIncomingPayChan` | Boolean | If `true`, the account does not allow others to make [Payment Channels](payment-channels.html) to it. |
+| `disallowIncomingTrustline` | Boolean | If `true`, the account does not allow others to make [trust lines](trust-lines-and-issuing.html) to it. |
```
ximinez: AFAIK, these fields are not returned unless the `featureDisallowIncoming` amendment is enabled.

ximinez (changes requested):

DennisDawson (approved):
Looks good. Some ignorable suggestions just to prove I read the whole thing.

content/references/http-websocket-apis/public-api-methods/account-methods/account_info.md
mDuo13: I think `ident` is ignored, `ledger` will technically still work but it's more ambiguous than using one of `ledger_index`/`ledger_hash`, and passing in a non-default (`true`) value for `strict` works but is now the same as the default. (Previously the default, strict=false, meant that you could provide a secret key or something in the account field, which is a bad idea shortcut that was used by automated tests once upon a time.)

So, while there's no reason when you _should_ use them, a case where you _would_ use `strict=true` is if you were writing code that's backwards compatible with older rippled versions (including forked/branched code running on some test networks/sidechains) but still secure.

intelliot (approved):
Awesome - great updates!

PR #1984, opened 2023-06-28
Fix NFTokenPage PageMin field names in example json

nkramer44:
The example JSON on the [`NFTokenPage`](https://xrpl.org/nftokenpage.html#nftokenpage) page seems to be incorrect/outdated. This PR fixes the following issues:
1) The current JSON uses `PreviousTokenPage` and `PreviousTokenNext` instead of `PreviousPageMin` and `NextPageMin`, respectively. 
2) The current JSON does not wrap each element of `NFTokens` in a JSON field called `NFToken`. 
3) The page IDs used for `PreviousPageMin` and `NextPageMin` are only 240 bits, not 256


DennisDawson (approved):
LGTM - thank you for taking the time to make the changes, Noah!

intelliot (approved):

PR #1986, opened 2023-06-29
restructure tutorials nav and pages

oeggert:


DennisDawson (approved):
As discussed, the changes look good structurally, but we need to ensure that the "Getting Started" topics flow properly into the modular tutorials.
I agree that we should settle on "<Doing thing> Using <Language>."

PR #1991, opened 2023-07-07
Improve install-rippled-on-ubuntu.md

intelliot:
Minor improvements:

- `mv` will rename the file (instead of moving it into a destination directory) if the destination directory does not exist. By providing the `/`, we are being explicit that we want the file to be moved.
  - If the earlier `mkdir` command was skipped or failed, we want `mv`  to report an error, instead of silently renaming the key file to `keyrings`, which would not make sense.
- Supplying a command to `gpg` prevents the warning which was previously shown in the gpg output.

### DIFF
--- a/content/tutorials/manage-the-rippled-server/installation/install-rippled-on-ubuntu.md
+++ b/content/tutorials/manage-the-rippled-server/installation/install-rippled-on-ubuntu.md
@@ -31,23 +31,22 @@ Before you install `rippled`, you must meet the [System Requirements](system-req
 
         sudo mkdir /usr/local/share/keyrings/
         wget -q -O - "https://repos.ripple.com/repos/api/gpg/key/public" | gpg --dearmor > ripple-key.gpg
-        sudo mv ripple-key.gpg /usr/local/share/keyrings
+        sudo mv ripple-key.gpg /usr/local/share/keyrings/
 
 
 4. Check the fingerprint of the newly-added key:
 
-        gpg /usr/local/share/keyrings/ripple-key.gpg
+        gpg --import --import-options show-only /usr/local/share/keyrings/ripple-key.gpg
 
     The output should include an entry for Ripple such as the following:
 
-        gpg: WARNING: no command supplied.  Trying to guess what you mean ...
         pub   rsa3072 2019-02-14 [SC] [expires: 2026-02-17]
             C0010EC205B35A3310DC90DE395F97FFCCAFD9A2
         uid           TechOps Team at Ripple <techops+rippled@ripple.com>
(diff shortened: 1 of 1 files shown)


content/tutorials/manage-the-rippled-server/installation/install-rippled-on-ubuntu.md
```diff
@@ -31,23 +31,22 @@ Before you install `rippled`, you must meet the [System Requirements](system-req
 
         sudo mkdir /usr/local/share/keyrings/
         wget -q -O - "https://repos.ripple.com/repos/api/gpg/key/public" | gpg --dearmor > ripple-key.gpg
-        sudo mv ripple-key.gpg /usr/local/share/keyrings
+        sudo mv ripple-key.gpg /usr/local/share/keyrings/
```
legleux, line 34: Hmm, as long as we're sudoing... let's cut out a step and just put it there in the first place? (I know I wrote this :expressionless:) 
```
sudo install -m 0755 -d /etc/apt/keyrings

wget -q -O - "https://repos.ripple.com/repos/api/gpg/key/public" | sudo gpg --dearmor -o /etc/apt/keyrings/ripple-key.gpg

echo \
  "deb [arch="$(dpkg --print-architecture)" signed-by=/etc/apt/keyrings/ripple-key.gpg] https://repos.ripple.com/repos/rippled-deb \
  "$(. /etc/os-release && echo "$VERSION_CODENAME")" stable" | \
  sudo tee /etc/apt/sources.list.d/ripple.list > /dev/null
  ```
  Stole this from docker.
  This also makes most of (the second??) step 4 and the hubbub about choosing distro redundant on 53-61.

File: content/tutorials/manage-the-rippled-server/installation/install-rippled-on-ubuntu.md [hunk 2]
```diff
@@ -31,23 +31,22 @@ Before you install `rippled`, you must meet the [System Requirements](system-req
...
 
 
 4. Check the fingerprint of the newly-added key:
 
-        gpg /usr/local/share/keyrings/ripple-key.gpg
+        gpg --import --import-options show-only /usr/local/share/keyrings/ripple-key.gpg
```
legleux, line 39: Good catch on this, this always annoyed me. Seemed wrong, just was always like this.

legleux (dismissed):

mDuo13 (approved):

content/tutorials/manage-the-rippled-server/installation/install-rippled-on-ubuntu.md
intelliot: @legleux can you submit those changes in a new PR?

PR #2003, opened 2023-07-14
fix: explainer videos autoplaying on first load of xrpl.org/docs.html

JST5000:
This fixes an issue where if you went to https://xrpl.org/docs.html after enabling audio from auto-played videos, you'd hear every explainer video at the same time, which can be a rather jarring experience.

The fix was to turn off autoplay on the youtube videos.


ckniffen (approved):

ckniffen:
I would prefer the iframe get added when you click on the link instead of preloading all of them.  This should fix the issue for now though.

ckniffen (approved):

intelliot (approved):

PR #2005, opened 2023-07-17
Update max no. of signers per signer list

wojake:


intelliot:
Can you add a link/reference to https://xrpl.org/known-amendments.html#expandedsignerlist ?

mDuo13 (approved):

PR #2011, opened 2023-07-19
Create custom transactor

oeggert:
Create custom transactor, using `CreateCheck` as an example.

### DIFF
--- a/content/tutorials/use-specialized-payment-types/create-custom-transactors.md
+++ b/content/tutorials/use-specialized-payment-types/create-custom-transactors.md
@@ -7,11 +7,13 @@ labels:
 ---
 # Create Custom Transactors
 
-You can create custom transactors to modify the XRP Ledger. Transactors follow a basic order of operations:
+A _transactor_ is code that processes a transaction and modifies the XRP Ledger. Creating custom transactors enables you to add new functionality to `rippled`.
+
+Transactors follow a basic order of operations:
 
 1. Access a _view_ into a serialized type ledger entry (SLE).
 2. Update, erase, or insert values in the _view_.
-3. Apply the finalized changes to the ledger.
+3. Apply the finalized changes from the _view_ to the ledger.
 
 **Note:** _Views_ are sandboxes into ledgers. Transactors make all necessary error checks and changes in sandboxes, not directly on the ledger. After values are finalized, changes are applied atomically to the ledger.
 
@@ -63,6 +65,19 @@ Initializing the transactor with `ApplyContext` gives it access to:
 
 The `preflight` function checks for errors before accessing the ledger and incurring fees.
 
+- `PreflightContext` doesn't have a view of the ledger.
+- Use bracket notation to retrieve fields from ledgers and transactions:
+        
+        auto const curExpiration = (*sle*)[~sfExpiration];
+        (*sle)[sfBalance] = (*sle)[sfBalance] + reqDelta;
+            
+    **Note:** The `~` symbol returns an optional type.
+
+- You can view ledger and transaction schemas here:
+    - [`LedgerFormats.cpp`](https://github.com/XRPLF/rippled/blob/master/src/ripple/protocol/impl/LedgerFormats.cpp)
+    - [`TxFormats.cpp`](https://github.com/XRPLF/rippled/blob/master/src/ripple/protocol/impl/TxFormats.cpp)
+
+
 ```c++
 CreateCheck::preflight(PreflightContext const& ctx)
 {
@@ -102,7 +117,6 @@ CreateCheck::preflight(PreflightContext const& ctx)
         }
     }
 
-    // Add `~` for optional types.
     if (auto const optExpiry = ctx.tx[~sfExpiration])
     {
         if (*optExpiry == 0)
@@ -116,24 +130,13 @@ CreateCheck::preflight(PreflightContext const& ctx)
 }
 ```
 
-- `PreflightContext` doesn't have a view of the ledger.
-- Use bracket notation to retrieve fields from ledgers and transactions:
-
-        ```
-        auto const curExpiration = (*sle*)[~sfExpiration];
-        (*sle)[sfBalance] = (*sle)[sfBalance] + reqDelta;
-        ```
-    
-    **Note:** The `~` symbol returns an optional type.
-
-- You can view ledger and transaction schemas here:
-    - [`LedgerFormats.cpp`](https://github.com/XRPLF/rippled/blob/master/src/ripple/protocol/impl/LedgerFormats.cpp)
-    - [`TxFormats.cpp`](https://github.com/XRPLF/rippled/blob/master/src/ripple/protocol/impl/TxFormats.cpp)
-
 
 ### 2. Add a `preclaim` function.
 
 The `preclaim` functions checks for errors that require viewing information on the ledger.
+
+- `PreclaimContext` has a read-only view of the ledger.
+- This function incurs fees.
 
 ```c++
 CreateCheck::preclaim(PreclaimContext const& ctx)
@@ -219,9 +222,6 @@ CreateCheck::preclaim(PreclaimContext const& ctx)
 }
 ```
 
-- `PreclaimContext` has a read-only view of the ledger.
-- This function incurs fees.
-
 
 ### 3. Add a `doApply()` function.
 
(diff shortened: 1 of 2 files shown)


wojake:
Thanks for taking the time to write this down, I do think more materials related to core `rippled` development should be written and I'd love to see you pursue more of this!

content/tutorials/use-specialized-payment-types/create-custom-transactors.md
```diff
@@ -0,0 +1,379 @@
+---
+html: create-custom-transactors.html
+parent: use-specialized-payment-types.html
```
mDuo13, line 3: Maybe the parent page should be "Contribute Code" or something in that tree instead?

content/tutorials/use-specialized-payment-types/create-custom-transactors.md
```diff
@@ -0,0 +1,379 @@
...
+blurb: Create custom transactors to interact with the XRP Ledger.
+labels:
+  - XRP
```
mDuo13, line 6: This feels like entirely the wrong label for this. If there aren't any that fit, you can come up with new ones or just have none.

content/tutorials/use-specialized-payment-types/create-custom-transactors.md
```diff
@@ -0,0 +1,379 @@
...
+---
+# Create Custom Transactors
+
+A _transactor_ is code that processes a transaction and modifies the XRP Ledger. Creating custom transactors enables you to add new functionality to `rippled`.
```
mDuo13, line 10: Maybe just add a note for context that actually using this functionality in the shared network requires going through the amendment process? Something to the effect that developing the transactor is only one step in adding new functionality to the protocol.

content/tutorials/use-specialized-payment-types/create-custom-transactors.md
```diff
@@ -0,0 +1,379 @@
...
+- The transaction that triggered the transactor.
+- A view of the SLE.
+- A journal to log errors.
+
+
+## CPP File
+
+### 1. Add a `preflight` function.
+
+The `preflight` function checks for errors before accessing the ledger and incurring fees.
```
mDuo13, line 66: Instead of "incurring fees" I suggest providing a little more context about what you're looking for at this stage and why.

```suggestion
The `preflight` function checks for errors before accessing the ledger. It should check for internal inconsistencies in a transaction, so that invalid and incorrectly specified transactions can be rejected with a minimal amount of work.
```

content/tutorials/use-specialized-payment-types/create-custom-transactors.md
```diff
@@ -0,0 +1,379 @@
...
+- You can view ledger and transaction schemas here:
+    - [`LedgerFormats.cpp`](https://github.com/XRPLF/rippled/blob/master/src/ripple/protocol/impl/LedgerFormats.cpp)
+    - [`TxFormats.cpp`](https://github.com/XRPLF/rippled/blob/master/src/ripple/protocol/impl/TxFormats.cpp)
+
+
+```c++
+CreateCheck::preflight(PreflightContext const& ctx)
+{
+    if (!ctx.rules.enabled(featureChecks))
+        return temDISABLED;
```
mDuo13, line 85: The tutorial should probably mention amendments before this point to clarify the usage of `ctx.rules.enabled` and maybe some mention of result codes. You can link [`TER.cpp`](https://github.com/XRPLF/rippled/blob/master/src/ripple/protocol/impl/TER.cpp) or [`TER.h`](https://github.com/XRPLF/rippled/blob/master/src/ripple/protocol/TER.h) which have the canonical lists of transaction result codes.

content/tutorials/use-specialized-payment-types/create-custom-transactors.md
```diff
@@ -0,0 +1,379 @@
...
+}
+```
+
+
+### 2. Add a `preclaim` function.
+
+The `preclaim` functions checks for errors that require viewing information on the ledger.
+
+- `PreclaimContext` has a read-only view of the ledger.
+- This function incurs fees.
```
mDuo13, line 139: What fees? How? Who is paying? Please clarify. I suspect the intent is that a transaction that gets to this stage generally gets put into a ledger and burns the transaction cost even if it fails (so, the sender of the transaction pays). Of course, that's not strictly true if they used `fail_hard`, but I'm not sure how that comes in.

The biggest distinction/turning point where a transaction _should_ burn the transaction cost if at all possible is when it's relayed to other servers. I believe that if a transaction gets to `preclaim` then it's probably going to be relayed unless using `fail_hard` but I'm not totally sure. 

I also think the class of result code used may have an effect on that behavior, but I'm not confident about the details. But looking at the sample code, it looks like `preclaim` should return either `tesSUCCESS` (if all went well so far) or a `tec` code  (if something is wrong). Maybe you aren't supposed to return other classes of codes (like `ter` or `tef`) in a `preclaim` function?

content/tutorials/use-specialized-payment-types/create-custom-transactors.md
```diff
@@ -0,0 +1,379 @@
...
+    // If we succeeded, the new entry counts against the creator's reserve.
+    adjustOwnerCount(view(), sle, 1, viewJ);
+    return tesSUCCESS;
+}
+```
+
+
+## Additional Functions
+
+
```
mDuo13, line 314: I think you should have at least a line of text here explaining this section.

```suggestion
## Additional Functions

You can add more helper functions to your custom transactor as necessary. There are a few special functions that are relevant in special cases.
```

content/tutorials/use-specialized-payment-types/create-custom-transactors.md
```diff
@@ -0,0 +1,379 @@
...
+### `calculateBaseFee`
+
+The `calculateBaseFee` function calculates the XRP fee to run a transaction. The fee is usually inherited from the base transactor, but transactors such as `EscrowFinish` require additional fees.
```
mDuo13, line 317: This phrasing, I think, hides the intent here. Here's an attempt to make it more obvious:

```suggestion
Most transactions inherit the default [reference transaction cost](transaction-cost.html). However, if your transactor needs to define a non-standard transaction cost, you can replace the transactor's `calculateBaseFee` method with a custom one. For example, the following example shows how `EscrowFinish` transactions charge an additional cost on conditional escrows based on the size of the fulfillment:
```

content/tutorials/use-specialized-payment-types/create-custom-transactors.md
```diff
@@ -0,0 +1,379 @@
...
+```
+
+
+### `makeTxConsequences`
+
+`rippled` uses a [`TxConsequences`](https://github.com/XRPLF/rippled/blob/master/src/ripple/app/tx/applySteps.h#L41-L44) class to describe the outcome to an account when applying a transaction. It tracks the fee, maximum possible XRP spent, and how many sequences are consumed by the transaction. There are three types of consequences:
+
+- **Normal:** The transactor doesn't affect transaction signing and only consumes an XRP fee.
+- **Blocker:** The transactor affects transaction signing, preventing valid transactions from queueing behind it.
+- **Custom:** The transactor needs to do additional work to determination consequences.
```
mDuo13, line 341: You should probably clarify that any transaction that can _spend_ XRP (beyond the basic transaction cost) needs a custom `makeTxConsequences`. The [PaymentChannelCreate transaction's example](https://github.com/XRPLF/rippled/blob/aded4a7a925a5822522e6bbbb4adfd3fec39257b/src/ripple/app/tx/impl/PayChan.cpp#L165-L169) is a nice small one. (It can spend up to the `Amount` field of the transaction).

content/tutorials/use-specialized-payment-types/create-custom-transactors.md
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    return Transactor::calculateBaseFee(view, tx) + extraFee;
+}
+```
+
+
+### `makeTxConsequences`
+
+`rippled` uses a [`TxConsequences`](https://github.com/XRPLF/rippled/blob/master/src/ripple/app/tx/applySteps.h#L41-L44) class to describe the outcome to an account when applying a transaction. It tracks the fee, maximum possible XRP spent, and how many sequences are consumed by the transaction. There are three types of consequences:
```
mDuo13, line 337: Based on what I'm seeing here, I have a feeling the point of this function is to determine when a transaction is [eligible for the queue](https://xrpl.org/transaction-queue.html#queuing-restrictions) and how it affects the queue eligibility of transactions that come after it. I think, relevantly, this needs to be something that can be evaluated using the fields of the transaction itself without further context like the ledger state. So the function should estimate the transaction's max XRP spend, any implications for account sequence numbers and transaction authorization.

FWIW, "sequences" is not a phrasing we use elsewhere in the docs; for consistency, it's probably better to write "sequence numbers" instead.

content/tutorials/use-specialized-payment-types/create-custom-transactors.md
```diff
@@ -0,0 +1,379 @@
...
+    };
+
+    return TxConsequences{ctx.tx, getTxConsequencesCategory(ctx.tx)};
+}
+```
+
+
+## Next Steps
+
+Submit your transactor to the XRPL `Devnet` for others to test. See: [Contribute Code to the XRP Ledger](contribute-code-flow.html).
```
mDuo13, line 379: It's nonsensical to say, "submit the transactor to the XRPL's `Devnet`". You can't just... submit raw code to a network, let alone Devnet which is a public test network Ripple runs using specifically code that's been merged to the `develop` branch of the XRPLF `rippled` repository.

I think a more reasonable set of next steps include re-compiling the server and testing the transaction in stand-alone mode. If you put the transaction behind an amendment, you'll need to [force-enable the feature using the config file](https://xrpl.org/test-amendments.html#test-amendments).

mDuo13 (changes requested):
This is a great start, but it feels like a lot of the necessary background and context is missing. Not all of that can be addressed in this one tutorial, of course, but I've suggested some places where I happen to know some of the relevant context.

oeggert:
Reopening a new PR to merge into IA.v3 at #2034

PR #2017, opened 2023-07-24
Update software ecosystem

DennisDawson:
Add graphics, remove redundant section.


amarantha-k (changes requested):
Looking at the diagram again, request a few updates to make the diagram current:
- In the Programming Libraries level, rename "xrpl.js" to "Client libraries" as we now have other languages too, and also remove the "Ripple API" box.  
- Add Clio to the Middleware API level?

img/ecosystem-middleware.svg
mDuo13, line 1: This diagram still has some obsolete software in it: the Data API has been shut down for a couple years. There are some other "middleware" APIs provided by the community such as xApps, although I don't know if it's a good idea to namedrop any of them in particular in the diagram.

xrpl.js no longer has a "RippleAPI" interface (that was a 1.x thing; the nearest equivalent is the `Client` class but I don't really think that's a detail worth having in this diagram anymore) and it would make sense to showcase that there are multiple client libraries in different languages. Maybe we don't need to namedrop them in particular in this diagram, either, since which ones are in good standing may change over time.

I suggest renaming "Programming Libraries" to "Client Libraries" for consistency with the reference page that lists and provides details on them.

Also, this SVG appears to have been exported with an extra transparent frame around the outer edge, including the text "Middleware" in black-on-transparent, which probably won't go well. The others, too. Probably you need to re-export them from Figma(?) selecting a little bit less stuff or in a different way so as not to get the extra outer edge.

Moving forward, we might want to re-think these diagrams more comprehensively, including finding the right place to put Clio in there.

mDuo13 (changes requested):

content/concepts/introduction/software-ecosystem.md
```diff
@@ -24,6 +25,8 @@ The XRP Ledger is home to a deep, layered ecosystem of software projects powerin
 
 The peer-to-peer network at the heart of the XRP Ledger requires a highly-reliable, efficient server to enforce the rules of consensus and transaction processing. The XRP Ledger Foundation publishes a reference implementation of this server software, called [**`rippled`**](xrpl-servers.html) (pronounced "ripple-dee") in collaboration with Ripple. The server is available under [a permissive open-source license](https://github.com/XRPLF/rippled/blob/develop/LICENSE.md), so anyone can inspect and modify their own instance of the server, and re-publish with few restrictions.
```
amarantha-k:
```suggestion
The peer-to-peer network at the heart of the XRP Ledger requires a highly-reliable, efficient server to enforce the rules of consensus and transaction processing. The XRP Ledger Foundation publishes a reference implementation of this server software, called [**`rippled`**](xrpl-servers.html) (pronounced "ripple-dee"). The server is available under [a permissive open-source license](https://github.com/XRPLF/rippled/blob/develop/LICENSE.md), so anyone can inspect and modify their own instance of the server, and re-publish with few restrictions.
```

content/concepts/introduction/software-ecosystem.md
```diff
@@ -33,6 +36,8 @@ Core servers expose [HTTP / WebSocket APIs](http-websocket-apis.html) for users
 
 Libraries simplify some of the common work of accessing the XRP Ledger, usually through the HTTP / WebSocket APIs. They convert the data into forms that are more familiar and convenient for various programming languages, and include implementations of common operations. Some client libraries are officially maintained by the XRP Ledger Foundation and some are maintained by other entities in the community.
```
amarantha-k:
```suggestion
Libraries simplify some of the common work of accessing the XRP Ledger, usually through the HTTP / WebSocket APIs. They convert the data into forms that are more familiar and convenient for various programming languages, and include implementations of common operations. 
```

content/concepts/introduction/software-ecosystem.md
amarantha-k: Would not highlight Ripple as there are other community members contributing too.

content/concepts/introduction/software-ecosystem.md
amarantha-k: Afaik, the client libraries listed on xrpl.org are all under the XRPLF org in GitHub.

amarantha-k (approved):
Looks good for this iteration once the suggestions are committed. Will need to revisit this diagram during the next IA phase.

img/ecosystem-middleware.svg
mDuo13, line 1: One thing that feels to me like it's missing from this diagram is how the Clio server gets data from the p2p network, which is directly from a p2p rippled server's (limited) gRPC API (also not depicted, for the mostly good reason that only Clio and Reporting Mode are meant to use it). As it stands, it kind of looks like it gets data through "Middlware APIs" to a database, but a Clio server itself in read-write mode (you can configure them in read-only or read-write mode) is the one that populates the relational DB. It might make more sense if you gave them separate relational DBs so "Middleware APIs" and "Clio" don't look like they're intrinsically linked.

Anyway, none of that is strictly required, I'm just throwing out some ideas here in case you're inclined to keep improving this particular diagram.

content/concepts/introduction/software-ecosystem.md
```diff
@@ -8,6 +8,7 @@ labels:
 # Software Ecosystem
 
 The XRP Ledger is home to a deep, layered ecosystem of software projects powering and enabling an Internet of Value. It's impossible to list every project, tool, and business that interacts with the XRP Ledger, so this page only lists a few categories and highlights some central projects that are documented on this website.
+![The XRPL Ecosystem](img/ecosystem-apps-and-services.svg)
```
mDuo13, line 11: Preferably you should leave a blank line before the image, or else Markdown treats it as a hard-wrapped continuation of the previous paragraph.

```suggestion

![The XRPL Ecosystem](img/ecosystem-apps-and-services.svg)
```

img/ecosystem-apps-and-services.svg
mDuo13, line 1: Rather than repeating the entire diagram again, perhaps this one should be cropped to just show the top part? That might also be a bit awkward because the arrows showing what APIs the apps & services consume go all the way to the bottom, but it feels a bit redundant (or like the bigger diagram was repeated by mistake) to have the full picture at the top _and_ the bottom.

mDuo13 (approved):
Changes look good enough, though I noticed a few more things after building locally:

1. The background color of the diagram doesn't match the background color on the site in either light or dark mode. The dark mode body bg color on the site is `#111112` if you want to match that.
2. The diagram is _really big_ on larger screens. I think there should be a way to control the size of the SVG export so that it doesn't automatically get larger than a certain size, but I'm not sure offhand how to do that.

See below screenshot for what these two issues look like to me:

PR #2021, opened 2023-07-27
Statsd debug command to view metrics being sent to grafana

lalet:
Adding command line to view the metrics being sent from rippled server to the rippledemon udp port

### DIFF
--- a/content/tutorials/manage-the-rippled-server/configuration/configure-statsd.md
+++ b/content/tutorials/manage-the-rippled-server/configuration/configure-statsd.md
@@ -49,6 +49,18 @@ To enable StatsD on your `rippled` server, perform the following steps:
 
     You should periodically see messages indicating outbound traffic to the configured address and port of your `rippledmon` instance.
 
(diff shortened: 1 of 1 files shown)


mDuo13:
Thanks for the suggestion. I'm not all that familiar with tcpdump or statsd and I don't have a server I can test it on handy, so I ran this by my colleagues at Ripple:

@thejohnfreeman says:

> The example command will work. requires `sudo`. not sure if it needs to be spelled out. don't think it's that generally useful. The doc says "imported", but for that you'd need to look in Grafana. this shows you what is being exported.

And per Sophia Xie

> That sample output still looks way simple for that tcpdump command, you’d get a bunch of header stuff printed out, still think it should be tried out.

Could you perhaps clarify the use case / scenario where you want to see what's being sent and how this helps in that situation?

mDuo13:
I'm going to close this for now since it's not making progress. If you address the questions above then feel free to reopen it.

PR #2022, opened 2023-07-27
Clarify partial-payments.md

intelliot:
- Clarify that the sender of any Payment can enable the Partial Payment flag (it's not something that recipients control).
- Avoid burying the lede. Remind people to use delivered_amount. It can't be repeated too much.
- Clarify that the transaction cost is separate from the Amount.

### DIFF
--- a/content/concepts/payment-types/partial-payments.md
+++ b/content/concepts/payment-types/partial-payments.md
@@ -8,7 +8,7 @@ labels:
 ---
 # Partial Payments
 
-The sender of any Payment can enable the "Partial Payment" flag and send a payment which delivers less than the `Amount` field indicates. When processing any Payment, use [the `delivered_amount` metadata field](#the-delivered_amount-field), not the `Amount` field. The `delivered_amount` is the amount a payment actually delivered.
+The sender of any Payment can enable the "Partial Payment" flag and send a payment which delivers less than the `Amount` field indicates. When processing any Payment, use the `delivered_amount` metadata field, not the `Amount` field. The `delivered_amount` is the amount a payment actually delivered.
 
 If a Payment does not enable the Partial Payment flag, the `Amount` field of a [Payment transaction][] in the XRP Ledger specifies the amount to deliver after charging for exchange rates and [transfer fees](transfer-fees.html). The Partial Payment flag ([`tfPartialPayment`](payment.html#payment-flags)) allows a payment to succeed by reducing the amount received instead of increasing the amount sent. Partial payments are useful for [returning payments](become-an-xrp-ledger-gateway.html#bouncing-payments) without incurring additional costs to oneself.
 
--- a/content/concepts/payment-types/partial-payments.md
+++ b/content/concepts/payment-types/partial-payments.md
@@ -8,7 +8,7 @@ labels:
 ---
 # Partial Payments
 
-The sender of any Payment can enable the "Partial Payment" flag and send a payment which delivers less than the `Amount` field indicates. When processing any Payment, use the `delivered_amount` metadata field, not the `Amount` field. The `delivered_amount` is the amount a payment actually delivered.
+The sender of any [Payment transaction][] can enable the ["Partial Payment" flag](payment.html#payment-flags) and send a payment which delivers less than the `Amount` field indicates. When processing any Payment, use the `delivered_amount` metadata field, not the `Amount` field. The `delivered_amount` is the amount a payment actually delivered.
 
 If a Payment does not enable the Partial Payment flag, the `Amount` field of a [Payment transaction][] in the XRP Ledger specifies the amount to deliver after charging for exchange rates and [transfer fees](transfer-fees.html). The Partial Payment flag ([`tfPartialPayment`](payment.html#payment-flags)) allows a payment to succeed by reducing the amount received instead of increasing the amount sent. Partial payments are useful for [returning payments](become-an-xrp-ledger-gateway.html#bouncing-payments) without incurring additional costs to oneself.
 
--- a/content/concepts/payment-types/partial-payments.md
+++ b/content/concepts/payment-types/partial-payments.md
@@ -8,9 +8,11 @@ labels:
 ---
 # Partial Payments
 
-In the default case, the `Amount` field of a [Payment transaction][] in the XRP Ledger specifies the exact amount to deliver, after charging for exchange rates and [transfer fees](transfer-fees.html). The "Partial Payment" flag ([`tfPartialPayment`](payment.html#payment-flags)) allows a payment to succeed by reducing the amount received instead of increasing the amount sent. Partial payments are useful for [returning payments](become-an-xrp-ledger-gateway.html#bouncing-payments) without incurring additional costs to oneself.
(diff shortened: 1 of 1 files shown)


content/concepts/payment-types/partial-payments.md
```diff
@@ -8,9 +8,11 @@ labels:
 ---
 # Partial Payments
 
-In the default case, the `Amount` field of a [Payment transaction][] in the XRP Ledger specifies the exact amount to deliver, after charging for exchange rates and [transfer fees](transfer-fees.html). The "Partial Payment" flag ([`tfPartialPayment`](payment.html#payment-flags)) allows a payment to succeed by reducing the amount received instead of increasing the amount sent. Partial payments are useful for [returning payments](become-an-xrp-ledger-gateway.html#bouncing-payments) without incurring additional costs to oneself.
+The sender of any Payment can enable the "Partial Payment" flag and send a payment which delivers less than the `Amount` field indicates. When processing any Payment, use the `delivered_amount` metadata field, not the `Amount` field. The `delivered_amount` is the amount a payment actually delivered.
```
ximinez: I think a couple of links would be useful in this intro.
```
The sender of any [Payment transaction][] can enable the ["Partial Payment" flag](payment.html#payment-flags)
and send a payment which delivers less than the `Amount` field indicates. When processing any Payment, use the
`delivered_amount` metadata field, not the `Amount` field. The `delivered_amount` is the amount a payment
actually delivered.
```

They are redundant, but I think valuable to have right up front. Besides, links are cheap and don't really interfere with readability.

ximinez (changes requested):

content/concepts/payment-types/partial-payments.md
intelliot: added

ximinez (approved):

content/concepts/payment-types/partial-payments.md
```diff
@@ -8,9 +8,11 @@ labels:
 ---
 # Partial Payments
 
-In the default case, the `Amount` field of a [Payment transaction][] in the XRP Ledger specifies the exact amount to deliver, after charging for exchange rates and [transfer fees](transfer-fees.html). The "Partial Payment" flag ([`tfPartialPayment`](payment.html#payment-flags)) allows a payment to succeed by reducing the amount received instead of increasing the amount sent. Partial payments are useful for [returning payments](become-an-xrp-ledger-gateway.html#bouncing-payments) without incurring additional costs to oneself.
+The sender of any [Payment transaction][] can enable the ["Partial Payment" flag](payment.html#payment-flags) and send a payment which delivers less than the `Amount` field indicates. When processing any Payment, use the `delivered_amount` metadata field, not the `Amount` field. The `delivered_amount` is the amount a payment actually delivered.
```
oeggert, line 11:
```suggestion
The sender of a [Payment transaction][] can enable the ["Partial Payment" flag](payment.html#payment-flags) to process a successful payment that delivers less than the `Amount` specified.
```

content/concepts/payment-types/partial-payments.md
```diff
@@ -8,9 +8,11 @@ labels:
...
 
-The amount of XRP used for the [transaction cost](transaction-cost.html) is always deducted from the sender’s account, regardless of the type of transaction.
+If a Payment does not enable the Partial Payment flag, the `Amount` field of a [Payment transaction][] in the XRP Ledger specifies the amount to deliver after charging for exchange rates and [transfer fees](transfer-fees.html). The Partial Payment flag ([`tfPartialPayment`](payment.html#payment-flags)) allows a payment to succeed by reducing the amount received instead of increasing the amount sent. Partial payments are useful for [returning payments](become-an-xrp-ledger-gateway.html#bouncing-payments) without incurring additional costs to oneself.
```
oeggert, line 13:
```suggestion
By default, the `Amount` field specifies the amount to deliver after exchange and transfer fees. Partial payments, however, treat the `Amount` field as a max amount to send, reducing the amount received by the fees incurred. This behavior makes partial payments useful for [returning payments](become-an-xrp-ledger-gateway.html#bouncing-payments) without incurring additional costs to oneself.
```

content/concepts/payment-types/partial-payments.md
```diff
@@ -8,9 +8,11 @@ labels:
...
+
+The amount of XRP used for the [transaction cost](transaction-cost.html) is always deducted from the sender’s account, regardless of the type of transaction. This transaction cost, or fee, is not included in the `Amount`.
```
oeggert, line 15: I'm a little confused about the addition at the end here. The transaction fee isn't included in normal payments, but it is factored into the `Amount` on partial payments, right?

oeggert (commented):
A couple suggestions to generalize what a partial payment is. We go into the details in later sections on this page. Also a quick question about the addition of fees not being included in the `Amount`.

content/concepts/payment-types/partial-payments.md
ximinez: I don't like removing references to `delivered_amount` from the first paragraph. Without that, the first reference to _the correct behavior_ is halfway down the page. (In fact, it would probably be good to have a link to that section from the reference in this first paragraph.)

mDuo13: I also think it's weird to say that the sender "can enable... to process a successful payment." We should be sure not to conflate the execution of a payment that happens on-ledger with the off-ledger handling that happens as a response to a payment. From @oeggert's suggested wording, it's unclear if "process" means the on-ledger or off-ledger actions, although it's (hopefully obviously?) impossible for the sender's actions to directly control what the recipient does when they receive a payment.

content/concepts/payment-types/partial-payments.md
mDuo13: No, it's not. Every transaction burns the exact transaction cost specified if at all possible, and that has no impact on the amounts of XRP sent or received unless the sender is literally running out of XRP (in which case the transaction cost takes precedence)

content/concepts/payment-types/partial-payments.md
mDuo13: I'd like to bring back the word "exact" which was a pretty key word in the original phrasing.

Also, @oeggert, that suggestion isn't quite correct. Even with the Partial Payments flag enabled, the `Amount` field is not the "max amount to send". It's the "max amount to deliver". The `SendMax` field is (always) the "max amount to send". 

There are various complex things that can happen in cross-currency payments or with trust lines that have non-default quality settings, so it's just a little bit over-simplified to say that partial payments reduce the amount received by the fees incurred.

content/concepts/payment-types/partial-payments.md
intelliot: @oeggert for the reasons given by ximinez and mDuo13 above, I think your suggested sentence is less helpful than what I had proposed. Leaving this alone for now.

content/concepts/payment-types/partial-payments.md
intelliot: @mDuo13 if the phrasing proposed in this PR isn't acceptable, can you suggest an alternative?

intelliot:
I have read all of the comments to this point and I don't feel there are any changes that need to be made to this PR. @oeggert @mDuo13 - could you give this some thought and let me know if you feel otherwise?

oeggert (approved):
Approving. It looks like I misunderstood some concepts, so disregard my original suggestions.

mDuo13 (approved):
Approved without changes.

PR #2034, opened 2023-08-02
Custom Transactors doc to IA.v3

oeggert:


mDuo13 (approved):
The revised tutorial looks excellent. My only complaint at this point, and it's not a blocker, is the location.

I built the docs locally and I think I guessed like 3 or 4 times, incorrectly each time, where to find the new tutorial in the docs navigation. Ultimately I had to cheat and look at the `parent:` attribute to figure it out. Deep under the "Networks and Servers" _concept_ does not feel to me like a good fit for this or the "Contribute Code..." tutorial that this is currently a sibling of (but feels like it should be a child of?)

One alternative approach I'd pitch is to take the "Contributing to Documentation" header and make a sibling for "Contribute Code to the XRP Ledger" (maybe even: "Modify the XRP Ledger Code"??). We could group them both into a single parent, or leave them both directly under "Documentation" if that seems like it buries them too deep. While we're at it, there are two different pages named "Contribute Code to the XRP Ledger", one under Docs>Concepts>Network & Servers>Amendments and one under Community.

content/tutorials/core-ledger/create-custom-transactors.md
```diff
@@ -0,0 +1,389 @@
...
+        return temDISABLED;
+
+    NotTEC const ret{preflight1(ctx)};
+    if (!isTesSuccess(ret))
+        return ret;
+
+    if (ctx.tx.getFlags() & tfUniversalMask)
+    {
+        // There are no flags (other than universal) for CreateCheck yet.
+        JLOG(ctx.j.warn()) << "Malformed transaction: Invalid flags set.";
```
mvadari, line 97: Perhaps there should be a note about logging here - to explain that this shows up in the debug log.

content/tutorials/core-ledger/create-custom-transactors.md
```diff
@@ -0,0 +1,389 @@
...
+- The transaction that triggered the transactor.
+- A view of the SLE.
+- A journal to log errors.
+
+
+## CPP File
+
+### 1. Add a `preflight` function.
+
+The `preflight` function checks for errors before accessing the ledger. It should reject invalid and incorrectly formed transactions.
```
mvadari:
```suggestion
The `preflight` function checks for errors in the transaction itself before accessing the ledger. It should reject invalid and incorrectly formed transactions.
```

content/tutorials/core-ledger/create-custom-transactors.md
```diff
@@ -0,0 +1,389 @@
...
+}
+```
+
+
+### 2. Add a `preclaim` function.
+
+The `preclaim` function checks for errors that require viewing information on the current ledger.
+
+- `PreclaimContext` has a read-only view of the ledger.
+- If this step returns a result code of `tesSUCCESS` or any `tec` result, the transaction can be queued and broadcast to peers.
```
mvadari:
```suggestion
- If this step returns a result code of `tesSUCCESS` or any `tec` result, the transaction will be queued and broadcast to peers.
```

content/tutorials/core-ledger/create-custom-transactors.md
```diff
@@ -0,0 +1,389 @@
...
+
+```c++
+CreateCheck::preclaim(PreclaimContext const& ctx)
+{
+    AccountID const dstId{ctx.tx[sfDestination]};
+
+    // Use the `keylet` function to get the key of the SLE. Views have either `read` or `peek` access.
```
mvadari, line 148:
```suggestion
    // Use the `keylet` function to get the key of the SLE. Views have either `read` or `peek` access. `peek` access allows the developer to modify the SLE returned.
```

content/tutorials/core-ledger/create-custom-transactors.md
```diff
@@ -0,0 +1,389 @@
...
+    sleCheck->setFieldU32(sfSequence, seq);
+    sleCheck->setFieldAmount(sfSendMax, ctx_.tx[sfSendMax]);
+    if (auto const srcTag = ctx_.tx[~sfSourceTag])
+        sleCheck->setFieldU32(sfSourceTag, *srcTag);
+    if (auto const dstTag = ctx_.tx[~sfDestinationTag])
+        sleCheck->setFieldU32(sfDestinationTag, *dstTag);
+    if (auto const invoiceId = ctx_.tx[~sfInvoiceID])
+        sleCheck->setFieldH256(sfInvoiceID, *invoiceId);
+    if (auto const expiry = ctx_.tx[~sfExpiration])
+        sleCheck->setFieldU32(sfExpiration, *expiry);
```
mvadari, line 269: This can be modified to be easier to read by using the `[]` operator (not sure how exact you want the code to be to the original version)

PR #2038, opened 2023-08-04
generates dynamic link tags to ensure appropriate href's are generate…

AKCodez:
…d for the site on www. urls

### DIFF
--- a/template/base.html.jinja
+++ b/template/base.html.jinja
@@ -16,11 +16,25 @@
     {% if currentpage.canonical_url %}<link rel="canonical" href="{{currentpage.canonical_url}}" />{% endif %}
     <meta name="robots" content="index, follow" />
     <meta name="docsearch:language" content="{{target.lang}}" />
-    {% for lang in config.languages %}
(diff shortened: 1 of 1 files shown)


intelliot (changes requested):
What's the motivation for this change?

mDuo13:
I don't like replacing static `<link>` tags, which are broadly compatible with many browsers, crawlers, and configurations, with JavaScript to add synthetic ones—there will definitely be crawlers and user configurations (e.g. noscript) that don't handle this correctly

I'm guessing from the branch name the intent is to improve SEO by serving canonical names that match the domain used, but I don't think this achieves that purpose. We should instead update the server and/or DNS configuration to always redirect `https://www.xrpl.org` to `https://xrpl.org`

mDuo13:
Confirmed this was for SEO purposes and the better solution is to fix on the DNS and/or server side (make the `www.` URLs non-canonical and always redirect to drop the prefix). But since we're planning to change hosting this quarter, it doesn't make sense to start on that until after we've moved.

PR #2039, opened 2023-08-04
Remove old, unused XRP article

mDuo13:
The contents of this have mostly been replaced with the new "What is XRP?" article


amarantha-k (approved):
lgtm

PR #2058, opened 2023-08-08
Update to reflect current UNL publishers

amarantha-k:


amarantha-k:
@mDuo13 / @develoQ Could you please update the Japanese version of this page?

mDuo13:
I think #2027 already covers this in both Japanese and English.

mDuo13:
Yeah, just confirmed, #2027 covered this more thoroughly. I just merged that one, so this can be closed.

Issue #1978, opened 2023-06-27
Remove old "New in: rippled X.Y.Z" badges


mDuo13:
The badges indicating when a feature was added are extraneous and sometimes distracting information when they refer to really old versions. You can't keep synced with the blockchain unless you're running a recent version anyway, so these are only useful for a limited time unless people are updating _very outdated_ code or they're trying to do some sort of historical digging, in which case they should expect to rely on release notes anyway.

As a general policy, we should keep badges that are < ~2 years old (probably about 4-8 scheduled releases), which aligns with the timeline for retiring amendments (XLS-11d).

To avoid too many merge conflicts, this should be done on the `IA.v3` branch or on the master branch after merging that one. (It's started in some places already.) We should also update the contributing guidelines to add the above rule/guidance.

intelliot:
In certain cases (changes of historical importance), perhaps the badge could be replaced with a footnote, which is less distracting than a colored badge.

PR #2068, opened 2023-08-11
Update tel codes for rippled 1.11.0

### LINKED ISSUE #2000: Add 3 new `tel` Codes
https://github.com/XRPLF/rippled/pull/4612/commits/807a3c710c0361373282130d26d4df8384d0a8aa

Document three new local (`tel`) error codes introduced in version 1.11.0:

  - There are three new local error codes:
    - `telNETWORK_ID_MAKES_TX_NON_CANONICAL`: a `NetworkID` is present but the chain's network ID is less than 1025. Remove the field from the transaction, and try again.
    - `telREQUIRES_NETWORK_ID`: a `NetworkID` is required, but is not present. Add the field to the transaction, and try again.
    - `telWRONG_NETWORK`: a `NetworkID` is specified, but it is for a different network. Submit the transaction to a different server which is connected to the correct network.

The above descriptions are just ones that I wrote. Feel free to use them or rewrite them.

Page to update: https://xrpl.org/tel-codes.html

mDuo13:
Fixes  #2000.

Also removes some extra spaces in the table, which weren't making it actually easier to read, shortens a couple of the queue-related codes' descriptions, and removes old "new in" badges per #1978.


amarantha-k (approved):
LGTM

PR #2077, opened 2023-08-22
Add tooling to check for file/nav consistency

mDuo13:
- From the repo top, run `tool/check_file_consistency.py` to look for Markdown files that exist in the `content/` directory but aren't used in the documentation.
- New `enforce_filenames` filter prints a warning to console when building, if a file's path and filename don't match expectations based on its place in the nav and top heading.
- Remove the old "Get Started" page (unused in IA.v3) and various code that was associated with presenting the editor on that page


amarantha-k (approved):
lgtm

PR #2079, opened 2023-08-22
[IAv3] Top nav and other fixes

### LINKED ISSUE #2078: [IAv3] Fix CSS bug with tabbed code samples
Some new CSS added to allow for images in lists (in the Ledger Structure page) is causing code tabs to display on separate lines.

This should be a very small SCSS fix  (1-3 lines).

mDuo13:
Several of the problems were caused by the organization of the files being a little messy, so these fixes to the top nav involve actually moving files to the right place.

- Rename some files/folders based on their place in the nav.
- Move a bunch of non-documentation stuff, and docs on contributing code and/or docs to the new "Resources" section, as per discussion with @amarantha-k.
    - ~~Known issue: nav spills into a second row on page widths between 993px-1110px inclusive. To be fixed in a later CSS update, maybe along with making the Resources dropdown multi-column.~~ (Fixed)
    - Consolidated the old "contribute code" with the newer one, bringing in some info about the core server source code—the rest was redundant with the new file.
    - Lots of files are still not quite where convention says they should be, but I left those off for now to keep this PR a manageable scale. The tooling added in #2077 should help to track & fix those other ones later.
- Fix #2078 (Known issue: the CSS version in `master` is a higher number file. That'll have to be resolved along with other CSS things, rebuilding with all SCSS changes, when the `IA.v3` branch is merged.)
- Remove pages from footer if they're not supposed to be in the nav (namely the full documentation index)
- Known Issue: A not-new problem with the dropdown tech is that there's no way to group translated and untranslated top nav groupings together. So if, for example, one of the pages under "Join In" is translated including the `top_nav_grouping` and the other is not, they'll show up as different sections in the dropdown. (For the existing dropdowns, we mostly avoided this by not translating the groupings, which is less than ideal. )


mDuo13:
Preview of the new "Resources" dropdown:

content/resources/contribute-documentation/contribute-documentation.md
```diff
@@ -1,25 +1,24 @@
...
 We are happy to review your pull requests. To make the process as smooth as possible, please read this document and follow the stated guidelines.
 
-Contributions become copyright the XRP Ledger Project and are provided under the MIT [LICENSE](https://github.com/XRPLF/xrpl-dev-portal/blob/master/LICENSE).
-
 
 ## About This Site
 
 The XRPL Dev Portal provides comprehensive documentation of the the XRP Ledger, including sample code and other information for developers to start building.
 
+The official source repository for the site is at <https://github.com/XRPLF/xrpl-dev-portal>. Contributions are copyright their respective contributors, but must be provided under the MIT [LICENSE](https://github.com/XRPLF/xrpl-dev-portal/blob/master/LICENSE).
```
AKCodez, line 19: Contributions are copyright to their

AKCodez, line 19: \```suggestion
   Contributions are copyright to their respective contributors, but must be provided under the MIT [LICENSE](https://github.com/XRPLF/xrpl-dev-portal/blob/master/LICENSE).
   \```

amarantha-k (commented):
Under Resources -> Current Status, would be useful to add the XRP Scan Amendments Dashboard.

amarantha-k (approved):
Not a blocker, but I would reconsider the two-column drop-down under Docs. Given that we have simplified the nav and are adding landing pages, we may be able to keep the Docs drop-down consistent with others.  We can look at adding a tile for Popular Pages on the Docs landing page. 
Otherwise, updates in this PR look good to me.

PR #2082, opened 2023-08-22
adds report pdfs

AKCodez:


mDuo13 (commented):
These both seem to be reports that pertain to Ripple-specific work. Perhaps they would be more appropriate to host on opensource.ripple.com ([repo link](https://github.com/ripple/opensource.ripple.com/)) instead of the official XRPL site?

AKCodez:
> These both seem to be reports that pertain to Ripple-specific work. Perhaps they would be more appropriate to host on opensource.ripple.com ([repo link](https://github.com/ripple/opensource.ripple.com/)) instead of the official XRPL site?

Adding to ripple.com codebase instead. Closing PR

PR #2086, opened 2023-08-23
adds bitfrost & towo to software wallets. adds keystone and dcent to h…

AKCodez:
…ardware wallets


styles/light/_light-theme.scss
```diff
@@ -643,7 +643,20 @@ pre code {
...
+  #wallet-bitfrost {
+    content: url("../img/wallets/lightmode/bitfrost.png");
+  }
+  #wallet-towo {
+    content: url("../img/wallets/lightmode/towo.svg");
+  }
+}
+#wallet-keystone {
+  content: url("../img/wallets/lightmode/keystone.svg");
 }
```
mDuo13: The `#wallet-keystone` section needs to be moved _into_ the `#wallets` section above so that it has the proper priority to override the non-light-mode image.

assets/img/wallets/lightmode/dcent.svg
mDuo13, line 1: This image is not needed because the D'Cent logo is monochrome and can be inverted by the `invertible-img` class.

styles/light/_light-theme.scss
```diff
@@ -643,7 +643,20 @@ pre code {
...
+#wallet-dcent {
+  content: url("../img/wallets/lightmode/dcent.svg");
+}
```
mDuo13: This section does not need to be here if you're using `invertible-img` for d'cent.

template/page-xrp-overview.html.jinja
```diff
@@ -181,6 +187,14 @@
               "id": "wallet-ledger",
               "alt": _("Ledger"),
               "imgclasses": "invertible-img" },
+              { "href": "https://keyst.one/",
+              "id": "wallet-keystone",
+              "alt": _("Keystone"),
+              "imgclasses": "invertible-img"},
```
mDuo13:
```suggestion
              "alt": _("Keystone")},
```

Since you've got a separate logo for Keystone in light mode, and the Keystone logo isn't monochrome, it should not have the `invertible-img` class. As-is, it causes the logo to display with an off-brand orange color in light mode, as in this screenshot:



The fix I noted in `_light-theme.scss` is also necessary along with this one to get the logo to display correctly.

mDuo13 (changes requested):
Needs some minor cleanup, but mostly looks good.

Also, this change will need a CSS version number bump before we deploy it.

mDuo13 (approved):

PR #2087, opened 2023-08-24
trivial: tec-codes - update GitHub link & remove old version mentions

intelliot:
I'm skeptical that it is still useful to refer to very old versions of rippled, > 5 years old and/or prior to 1.0.0


mDuo13:
I believe this is already done in IA.v3.

mDuo13:
Confirmed that removing the old version badges is done in IA.v3. Updating the URLs isn't, but I'll handle that all at once in another PR.

PR #2097, opened 2023-08-29
add Countdown Banner for token2049 for APEX region only 

jonathanlei:
use Osano cookies info to display a countdown banner (XRPL Zone hosted at token2029 in Singapore), only for users in Singapore, Malaysia, Indonesia, Thailand, Vietnam, Hong Kong and Australia.
Screenshot 2023-08-29 at 3 08 31 PM


template/component-top-nav.html.jinja
```diff
@@ -1,15 +1,137 @@
...
+        let minutes = document.getElementById(id + '-minutes')
+        let seconds = document.getElementById(id + '-seconds')
+
+        var timeinterval = setInterval(function () {
+          var time = getTimeRemaining(endtime);
+
+          if (time.total <= 0) {
+            clearInterval(timeinterval);
+          } else {
+            days.innerHTML = '<span class=countdown-highlight>' + pad(time.days, 2) + '</span>';
```
mDuo13, line 83: I'm getting an error message in the console, thrown from this line: `Uncaught TypeError: days is null`.

I think this entire script is not supposed to be running if the user is not in the target region.

mDuo13 (changes requested):

template/component-top-nav.html.jinja
jonathanlei: logic moved inside of the country function

mDuo13 (approved):
Looks good with the fixes I added

PR #2098, opened 2023-08-30
Use delivered_amount in JS browser wallet tutorial (Fix #2075)

### LINKED ISSUE #2075: Browser wallet instructions don't show delivered amount
The [Build a Browser Wallet in JS tutorial](https://xrpl.org/build-a-browser-wallet-in-js.html) demonstrates bad practices regarding the amount field. Wallets should primarily show the `delivered_amount`, not the `Amount` field, because [partial payments can trick people](https://xrpl.org/partial-payments.html#partial-payments-exploit) into thinking they've received a lot of money when, really, they barely got anything.

The [transaction history code](https://github.com/XRPLF/xrpl-dev-portal/blob/master/content/_code-samples/build-a-wallet/js/src/transaction-history/transaction-history.js) and associated tutorial step need to be updated to use `delivered_amount` instead.

Also, [the same file sloppily serializes hex currency codes to text](https://github.com/XRPLF/xrpl-dev-portal/blob/290107224d3318ca8d06aa139cdbaad9037d7bab/content/_code-samples/build-a-wallet/js/src/transaction-history/transaction-history.js#L40). With the way it's coded, currency codes might be able to do script injection (writing a `<script>` to the page!) or trick the user into thinking they've received XRP when actually they received a token with a hex currency code _starting_ with `58525000` ("XRP" followed by a null byte). The [Normalize Currency Codes sample](https://github.com/XRPLF/xrpl-dev-portal/blob/master/content/_code-samples/normalize-currency-codes/js/normalize-currency-code.js) demonstrates best practices including (a) only allowing decoded hex with a limited subset of characters, and (b) not allowing a code that matches "XRP" case-insensitive.

mDuo13:
Also improve currency code parsing to remove possible "fake XRP" exploit (hex currency code starting with ascii that decodes to "XRP") and other possible exploits with ASCII currency codes.


content/_code-samples/build-a-wallet/js/src/transaction-history/transaction-history.js
```diff
@@ -29,22 +29,48 @@ header.innerHTML = `
     <th>Account</th>
     <th>Destination</th>
     <th>Fee (XRP)</th>
-    <th>Amount</th>
+    <th>Amount delivered</th>
```
ckniffen: nit: should be consistent with "Transaction Type".

ckniffen (approved):

PR #2100, opened 2023-09-01
docs(Update client-libraries.md): add link to XRPL SDK for Rust

gmosx:
Adds a link to the XRPL SDK for Rust collection of crates.

### DIFF
--- a/content/references/client-libraries.md
+++ b/content/references/client-libraries.md
@@ -15,5 +15,6 @@ For other programming languages, you can access the XRP Ledger through the [HTTP
 | **JavaScript** / **TypeScript** | `xrpl.js` | [Get Started](get-started-using-javascript.html) |  [API Reference](https://js.xrpl.org/) | [Repo](https://github.com/XRPLF/xrpl.js) |
 | **C++**      | `rippled` Signing Library | [Get Started](https://github.com/ripple/rippled/tree/develop/Builds/linux#signing-library) |  | (Part of [`rippled`](https://github.com/ripple/rippled/)) |
 | **Java** | `xrpl4j` | [Get Started Using Java](get-started-using-java.html) | [API Reference](https://javadoc.io/doc/org.xrpl/)  | [Repo](https://github.com/XRPLF/xrpl4j) |
+| **Rust** | `xrpl_sdk` | | [API Reference](https://docs.rs/xrpl_http_client/latest/xrpl_http_client/)  | [Repo](https://github.com/gmosx/xrpl-sdk-rust) |
 
 **Tip:** To add a client library not listed here, please [suggest changes to this page]({{target.github_forkurl}}/edit/{{target.github_branch}}/content/{{currentpage.md}})!


JST5000:
Hey @gmox,

I asked one of my teammates who is more familiar with Rust to go through and port over [this snippet](https://github.com/XRPLF/xrpl-dev-portal/blob/master/content/_code-samples/send-xrp/js/send-xrp.js) from xrpl.js that sends a payment and checks balances then comment their notes on what was challenging / potentially unexpected about using the given library. Here’s the feedback from that:

## Important Issues
These represent blockers from my perspective.

1. No way to construct a private key from a seed, or to construct a public key from a private key as a first class citizen of the library.
  a. AddressCodec is unimplemented
  b. You could potentially use the libsecp256k1 library or dalek-ed25519, but there are a lot of "gotchas" when it comes to encoding xrpl public keys and deriving addresses.
2. It was difficult to figure out how to get the hash of a signed transaction. You have to call hash::hash with a specific prefix and the tx blob, which means you need to know how to serialize the blob and which prefix to use. Would be much better if sign::sign_transaction returned a new struct with the signed transaction and a hash.
3. Transaction structs should be easily serializable without having to know where to find the serializer, ie there should be a method on transactions like to_blob(&self) -> Vec<u8> or to_blob_hex(&self) -> String
4. There wasn’t submit and wait/reliable submission support
  a. The error you get back when the transaction doesn't exist doesn't contain the ledger index either, so you can't check if last_ledger_sequence has passed, making it virtually impossible to do reliable submission yourself
5. TransactionEntryResponse does not have validated field, so you have to request LedgerIndex::Validated and parse the error if it exists to see if the transaction wasn't found or if another error occurred.

## Quality of Life Changes
These are not blockers from my perspective, but may be helpful to know about 🙂

1. Transaction structs all have a common() and common_mut() method, which you have to call to get any fields. Would definitely be better to have a Common and CommonMut trait with getters/setters for all the common fields so I can just do payment.last_ledger_sequence() instead of payment.common().last_ledger_sequence()
2. There was no easy way to convert between XRP and drops. There was also no way to do math on DropsAmount other than getting the u64. Would be nice to implement the various math traits.
3. All client methods go through the .call method with different types of request structs. Not necessarily wrong, but other client implementations have separate methods for each RPC/WS method.
4. Hash types don't have easy conversions to [u8; ?] or &[u8] or hex String, so I have to call hex::encode(hash) myself

## The Snippet

The code snippet they were porting in JavaScript: [send-xrp.js](https://github.com/XRPLF/xrpl-dev-portal/blob/master/content/_code-samples/send-xrp/js/send-xrp.js) 

The code snippet they ended up writing using xrpl-rust:



```Rust

#[tokio::test]
    async fn main() {
        let private_key = SecretKey::parse_slice(
            hex::decode("FDA6B4A3B6FD59F30EF83DD1E9D71525F0C930BBB84260BFF2966D4F89425015")
                .unwrap().as_slice()
        ).unwrap();

        let public_key = PublicKey::from_secret_key(&private_key);

        let address = AccountId::from_address("rPcXURqzuZGfeBKGPw7kCL8w1EWJdmep8G").unwrap();

        let client = Client::builder()
            .base_url(TESTNET_URL)
            .build();

        let mut payment = PaymentTransaction::new(
            address,
            Amount::Drops(DropsAmount::from_drops(22 * 1_000_000).unwrap()),
            AccountId::from_address("rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe").unwrap(),
        );

        client.prepare_transaction(payment.common_mut()).await.unwrap();
        let max_ledger = payment.common().last_ledger_sequence.unwrap();
        println!("Prepared transaction instructions: {:?}", payment);
        match payment.amount {
            Amount::Issued(_) => {
                panic!("Invalid payment amount type!")
            }
            Amount::Drops(drops) => {
                println!("Transaction cost: {:?} XRP", drops.drops() / 1_000_000)
            }
        }
        println!("Transaction expires after ledger: {:?}", max_ledger);

        sign::sign_transaction(&mut payment, &public_key, &private_key).unwrap();
        let signed_blob = serialize::serialize(&payment).unwrap();
        let hash = hash::hash(HASH_PREFIX_SIGNED_TRANSACTION, signed_blob.as_slice());
        println!("Identifying hash: {:?}", hex::encode_upper(&hash.0));
        println!("Signed blob: {:?}", hex::encode_upper(signed_blob.as_slice()));

        let submit_res = client.call(
            SubmitRequest::new(hex::encode_upper(signed_blob.as_slice()))
        ).await.unwrap();

        println!("Response: {:?}", submit_res["engine_result"].to_string());
        let mut validated = false;
        while !validated {
            let response = client.call(
                TransactionEntryRequest::new(hex::encode(hash.0))
                    .ledger_index(LedgerIndex::Validated)
            )
                .await;

            match response {
                Ok(transaction_response) => {
                    validated = true;
                    println!("Transaction result: {:?}", transaction_response.metadata.transaction_result);
                }
                Err(Error::Api(error_msg)) => {
                    if &error_msg != "transactionNotFound" {
                        panic!("Got error while waiting for transaction validation: {}", error_msg)
                    }
                }
                Err(e) => {
                    panic!("Got error while waiting for transaction validation: {}", e)
                }
            }
        }
    }
```

gmosx:
Thank you, this is *very* valuable feedback. I will see what I can do and revert.

mDuo13:
I'm going to close this PR for now, but by all means do reopen it when the Rust library is more mature (the blocking feedback has been addressed)!

PR #2101, opened 2023-09-01
Fix broken link

amarantha-k:


DennisDawson (approved):
Looks correct. Might change in IA.v4.

PR #2103, opened 2023-09-02
Fix various style/template issues

### LINKED ISSUE #2014: Collapse of code block
Background color of text in Code block is broken when in light mode.

(Confirmed in Chrome and Safari)

スクリーンショット 2023-07-20 16 12 56
スクリーンショット 2023-07-20 16 13 06

### LINKED ISSUE #2096: "Explore SDKs" clickable areas are deceptive/offset
On the Docs landing page: Due to the invisible offset for linking to anchors that our CSS automatically adds to headers, the "clickable area" of the "Java" link overlaps with the Javascript one above it:

This means you can click on the text "JavaScript" and get taken to the Java page.

We don't want to remove the `::before` pseudo-element on all headers, but it's fine to remove it from _these_ headers, since we don't link directly to them. That probably means adding a new (S)CSS rule.

mDuo13:
- Add styles to allow fully invisible headers. These can be used to support anchor links in translated pages, so we can move the link_subs out of the dactyl-config.yml. Example in transaction-common-fields.md
- Upgrade jQuery to 3.7.1
- Add left nav to references landings for consistent navigation
- Add breadcrumbs template (previously used Dactyl built-in) which throws an error if you specify a parent: that doesn't exist. (Previously it would write a blank link; the link checker would report this as an error, but it wasn't obvious what the cause was.)
- Fix #2014 (special case for code blocks nested in lists).
- Fix #2096.
- Remove excessive spacing between top nav and main content, and re-align top of main contents in all 3 columns for docs pages.
- Reorganize _content.scss to be easier to read.


amarantha-k (approved):
lgtm

PR #2110, opened 2023-09-06
Add caveat to tfTransferable

DennisDawson:
Per [TOK](https://ripplelabs.atlassian.net/browse/TOK-368) and https://github.com/XRPLF/rippled/issues/4529, note that without this flag an NFT can be returned to the issuer, but only based on a buy offer from the issuer.


amarantha-k (approved):

amarantha-k:
Note that PR#2103 fixes the broken link.

PR #2111, opened 2023-09-06
Porting Clawback documentation to xrpl.org 

amarantha-k:
Documentation for the new Clawback transaction (https://github.com/XRPLF/rippled/pull/4553)

PR #2116, opened 2023-09-11
per TOK-381, clawback is special case.

DennisDawson:
In a Clawback transaction, the `issuer` is the holder of the token.


mDuo13 (approved):

mDuo13:
Approval contingent on #2111 getting merged first, which should resolve the broken link.

PR #2128, opened 2023-09-13
remove apex banner in preperation for new AMM banner

AKCodez:


mDuo13 (changes requested):
Since the deadline for Token2049 has also passed as of this morning, shouldn't we remove both? (It'll probably also be easier to keep things consistent if there's no banner in any case, rather than some regions getting a banner and some not.)

mDuo13 (approved):
I removed the now-unused styles for the removed banners, and fixed some stuff that was causing the 3 main columns to display unevenly in some cases.

Part of that is to keep the top of the 3 columns aligned, as in the following screenshot:



And part of it is to reduce the amount of CSS we're serving, and the amount of unused styles we're keeping around, where possible.

PR #2132, opened 2023-09-14
reorg Stablecoin UC

DennisDawson:


content/concepts/tokens/robustly-monitoring-for-payments.md
```diff
@@ -0,0 +1,27 @@
+---
+html: robustly-monitoring-for-payments.html
+parent: trust-lines-and-issuing.html
+blurb: Issue your own stablecoin, based on assets of equal value outside of the XRP Ledger.
```
mDuo13: Don't forget to update the blurb to fit the page.

img/uc-stablecoin-flow.png
mDuo13, line 1: This diagram seems to have some non-final content in it still, like "Link to future AMM tutorials". Can you take another look and clean it up for publication?

content/use-cases/tokenization/stablecoin-issuer.md
```diff
@@ -4,504 +4,173 @@ parent: tokenization.html
...
+## Choose the Type of Stablecoin
 
-Tokens that are backed by funds in some outside system are sometimes called _stablecoins_. This includes tokens backed by fiat currency in a bank account, by cryptocurrencies on another blockchain, or other types of assets and forms of value. The term "stablecoin" comes from the idea that the exchange rate between the token and the asset it represents should be "stable" at 1:1 (minus fees).
+The first step is to decide the type of stablecoin you want to create. Your choice of stablecoin type might require additional steps, such as signing on financial or audit partners.
 
-For more information, see [Trust Lines and Issuing](trust-lines-and-issuing.html).
+![Stablecoin](img/uc-stablecoin-stable-coin.png)
 
+There are five common types of stablecoin you can create on the XRPL: fiat-backed, crypto-backed, commodity-backed, financial instrument-backed, and non-collateralized. See [Stablecoin Types](stablecoin-types.html).
```
mDuo13: `stablecoin-types.html` appears to be a broken link. Is there a page missing?

Also, this is stretching the definition of a "stablecoin"; if it's non-collateralized, why/how should its value be stable? By definition, a stablecoin is supposed to be backed 1:1 with outside funds of equal value. Otherwise, it can be a token but not a stablecoin, IMO.

img/uc-stablecoin-stable-coin.png
mDuo13, line 1: The slightly wobbly lines in this diagram are really distracting for me. Is that a stylistic thing, or is there something wrong with the image export?

Also, this file has a 1-pixel transparent lines at the top and bottom; those should probably be cropped out. 

Ideally, this seems like the kind of graphic that should change colors to match the theme, but that can be improved later.

content/use-cases/tokenization/stablecoin-issuer.md
```diff
@@ -4,504 +4,173 @@ parent: tokenization.html
...
 
+Publish standard information about your stablecoin to enable potential traders to verify the coin's stability.  
 
-### Hot and Cold Wallets
 
-{% include '_snippets/issuing-and-operational-addresses-intro.md' %}
+### Asset Nomenclature
 
-Main article: [Issuing and Operational Addresses](account-types.html)
+Choose a 3-character string for your currency code. Per ISO 4217, supranational currency codes begin with the letter _X_ for a currency that is not associated with a country. See [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217#X_currencies_(funds,_precious_metals,_supranationals,_other)).
```
mDuo13, line 65: The word "supranational" is obscure, and this sentence can be equally clear without it.

```suggestion
Choose a 3-character string for your currency code. Per ISO 4217, currencies that are not associated with a country should use currency codes that begin with the letter _X_. See [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217#X_currencies_(funds,_precious_metals,_supranationals,_other)).
```

One misconception I've heard that might also be good to clear up here is that multiple issuers can choose the same currency code. (Some people assume that every issuer has to have their own unique currency code.) If you're issuing a stablecoin backed by a national fiat currency, it's actually preferable to use the official code for that currency, such as EUR.

content/use-cases/tokenization/stablecoin-issuer.md
```diff
@@ -4,504 +4,173 @@ parent: tokenization.html
...
 
 
-## Fees and Revenue Sources
+### xrp-ledger.toml 
 
-A stablecoin issuer can earn revenue in a variety of ways, including:
+You can use the _Currencies_ table in an XRPL TOML file to provide additional information about your stablecoin. This makes the information about your cryptocurrency accessible in an expected place and format, and enhances transparency. See [xrp-ledger.toml File](xrp-ledger-toml.html#currencies).
```
mDuo13: It might be good to add a little more context about where this file comes from/goes. For example:

```suggestion
You can use the _Currencies_ table in an XRPL TOML file on your website to provide additional information about your stablecoin. This makes the information about your cryptocurrency accessible in an expected place and format, and enhances transparency. See [xrp-ledger.toml File](xrp-ledger-toml.html#currencies).
```

content/use-cases/tokenization/stablecoin-issuer.md
```diff
@@ -4,504 +4,173 @@ parent: tokenization.html
...
-If a financial institution suspects that funds may be related to criminal activity, the institution must file a Suspicious Activity Report (SAR) with the appropriate regulatory authority. Failure to report suspicious activity may result in in penalties for the institution.
 
-See also:
+### Freeze a Trust Line
 
-- [Suspicious Activity Reporting Overview (USA FFIEC)](https://bsaaml.ffiec.gov/manual/RegulatoryRequirements/04_ep)
+If you issue tokens in the XRP Ledger, you can enable the _No Freeze_ setting to permanently limit your own ability to use the token freezing features of the XRP Ledger. (As a reminder, this only applies to issued tokens, not XRP.) 
 
-- [FATF Recommendation 16: Reporting of suspicious transactions and compliance](http://www.fatf-gafi.org/publications/fatfrecommendations/documents/fatf-recommendations.html)
+If you do not enable the _No Freeze_ setting, when an account shows suspicious activity or violates your institution's terms of use, you have the option of freezing the trust line while you resolve the issue.
```
mDuo13, line 143: I think it would make more sense to reverse the order of these paragraphs (and rephrase accordingly). For example:

> As a token issuer, when an account shows suspicious activity or violates your terms of use, you have the option of freezing your trust line to that account while you resolve the issue. (As a reminder, this only applies to issued tokens, not XRP.)
>
> To make your tokens more "cash-like" and provide assurances to users that you won't misuse these features, you can enable the _No Freeze_ setting to permanently limit your own ability to use the token freezing features of the XRP Ledger.
```

File: content/use-cases/tokenization/stablecoin-issuer.md [hunk 6]
```diff
@@ -4,504 +4,173 @@ parent: tokenization.html
...
-- The account number of the transmittor, if used,
-- The address of the transmittor,
-- The identity of the transmittor's financial institution,
-- The amount of the transmittal order,
-- The execution date of the transmittal order, and
-- The identity of the recipient's financial institution.
+### Global Freeze
 
-<!-- SPELLING_IGNORE: transmittor -->
+If you see signs of suspicious activity, you can enact a global freeze on your account to prevent users from sending your tokens to each other and trading your token in the decentralized exchange.
```
mDuo13, line 150: "suspicious activity" is vague here; there's all kinds of suspicious activity that doesn't justify a global freeze. Maybe it would be better as, "If you suspect that your own account or systems have been compromised," instead?

content/use-cases/tokenization/stablecoin-issuer.md
```diff
@@ -4,504 +4,173 @@ parent: tokenization.html
...
+See [Clawback](clawback.html).
 
-- In the European Union, EU Funds Transfer Regulation requires that the originator’s bank, the beneficiary’s bank, and any intermediary banks include certain details of the payer and payee in transaction details to detect, investigate, and prevent money laundering and terrorist financing.
+![Clawback](img/uc-stablecoin-clawback.png)
 
-    See also:
+### Burn
 
-    - [EU Regulation (EC) No 1781/2006 description](http://eur-lex.europa.eu/LexUriServ/LexUriServ.do?uri=OJ:L:2006:345:0001:0009:EN:PDF)
+One way to manage the value of a stablecoin is to destroy, or _burn_ stablecoins. When you burn stablecoins, you reduce the number of tokens in circulation. That will typically increase the value of the remaining tokens as supply falls.
```
mDuo13, line 175: This goes back to my comment above about stretching the definition of a stablecoin. The value per unit of a stablecoin is supposed to be fixed with respect to its collateral, so burning stablecoins _must_ mean that there is less collateral, unless the coin has become un-pegged and is no longer a stablecoin.

content/use-cases/tokenization/stablecoin-issuer.md
```diff
@@ -4,504 +4,173 @@ parent: tokenization.html
...
 
-    - [Effective June 26, 2017: Regulation 2015/847 on information accompanying transfers of funds](http://eur-lex.europa.eu/legal-content/EN/ALL/?uri=CELEX%3A32015R0847)
+To destroy stablecoins, you can transfer them to a "black hole" account (an account address where no one knows the secret key).
```
mDuo13, line 177: This is misleading/incorrect. You cannot always transfer tokens to a black hole account (it would have to have created a trust line to your issuer before becoming a black hole). However, you _can_ always send tokens back to the issuing account, which directly reduces the number of tokens outstanding. That's burning the tokens in a more direct and literal sense.

However, there's something off about the bigger picture here. For a non-stablecoin token with a fixed total supply  and a blackholed issuer, it might make sense to talk about "burning" tokens. But for a stablecoin, it's weird; typically, the same process is called "redeeming" the tokens (for their off-ledger collateral value). The way a stablecoin (specifically) is supposed to work, the on-ledger supply fluctuates in response to increases or decreases in collateral (that is, deposits and withdrawals) and tokens are "burned" or "issued" regularly.

content/use-cases/tokenization/stablecoin-issuer.md
```diff
@@ -516,36 +185,22 @@ To submit transactions reliably, follow these guidelines:
 
 For more information, see [Reliable Transaction Submission](reliable-transaction-submission.html).
 
+### List on an XRPL Native DEX
+
+Decentralized exchanges are integral to the decentralized finance ecosystem. Listing your token on well known DEXs operating on XRPL will enhance its visibility and accessibility, thereby attracting more liquidity. Begin by placing sell offers on a suitable DEX, such as [Sologenic](https://sologenic.org/trade).
```
mDuo13, line 198: The wording here is a bit misleading: there is _one, global DEX_ built into the XRP Ledger. Sologenic and others are just _interfaces_ into the same shared system. (Offers placed on GateHub can be seen and matched on Sologenic, and vice-versa.)

Also, token issuers don't need to do anything to list on the DEX. They can, if they want, set up a trading account to place offers for their token. This guide should probably provide the advice that they use a separate account for trading than they do for issuing.

content/use-cases/tokenization/stablecoin-issuer.md
```diff
@@ -4,504 +4,173 @@ parent: tokenization.html
...
-
-To send a Partial Payment, enable the [`tfPartialPayment` flag](payment.html#payment-flags) on the transaction. Set the `Amount` field to the amount you received and omit the `SendMax` field. You should use the `SourceTag` value from the incoming payment as the `DestinationTag` value for the return payment.
-
-To prevent two systems from bouncing payments back and forth indefinitely, you can set a new Source Tag for the outgoing return payment. If you receive an unexpected payment whose Destination Tag matches the Source Tag of a return you sent, then do not bounce it back again.
-
-
-## Reliable Transaction Submission
+### Reliable Transaction Submission
 
 The goal of reliably submitting transactions is to achieve the following two properties in a finite amount of time:
```
mDuo13, line 183: Seeing this in the new context, it's more obvious to me that this summary buries the lede on why you should care about reliable transaction submission. What it probably should say here is something like:

> Best practices for reliable transaction submission ensure that your transactions only execute once, within a finite and predictable amount of time, even if your systems suffer problems such as a power failure or network outage. Without taking proper care, it is possible that transactions could be delayed an indefinite amount of time, you might not realize that a payment to a customer has failed, or you might mistakenly send money again for a transaction that has already been settled.

content/use-cases/tokenization/stablecoin-issuer.md
```diff
@@ -4,504 +4,173 @@ parent: tokenization.html
...
-- When sending into the XRP Ledger, always specify your issuing address as the issuer of the token. Otherwise, you might accidentally use paths that deliver the same currency issued by other addresses.
-- Before sending a payment into the XRP Ledger, double check the cost of the payment. A payment from your operational address to a customer should not cost more than the destination amount plus any transfer fee you have set.
-- Before processing a payment out of the XRP Ledger, make sure you know the customer's identity. This makes it harder for anonymous attackers to scam you. Most anti-money-laundering regulations require this anyway. This is especially important because the users sending money from the XRP Ledger could be different than the ones that initially received the money in the XRP Ledger.
-- Follow the guidelines for [reliable transaction submission](#reliable-transaction-submission) when sending XRP Ledger transactions.
-- [Robustly monitor for incoming payments](#robustly-monitoring-for-payments), and read the correct amount. Don't mistakenly credit someone the full amount if they only sent a [partial payment](partial-payments.html).
```
mDuo13, line 291: It's imperative that the stablecoin issuer use case warn people about partial payments. There have been multiple instances of issuers losing thousands to millions of dollars because of an exploit of their systems caused by misunderstanding partial payments.

I would say it's even appropriate to add an entire section to the doc alerting issuers to the possibility of partial payments and linking to the doc.

content/@i18n/ja/concepts/tokens/decentralized-exchange.md
```diff
@@ -1,5 +1,5 @@
 ---
-html: decentralized-exchange.html
+html: de-index.ja.html
```
mDuo13: This line should not be changed, and html filenames should _never_ have `.ja.html` in them. Also, all the other pages that linked to this page should also not have their `parent:` attributes changed.

content/@i18n/ja/concepts/tokens/freezes.md
```diff
@@ -1,7 +1,7 @@
 ---
 html: freezes.html
 parent: tokens.html
-blurb: 発行者はコンプライアンス目的でトークンの取引を停止できます。
+blurb: 凍結では、コンプライアンス目的で発行済み通貨の取引を停止できます。
```
mDuo13: This change (unintentionally?) reverts the changes done in https://github.com/XRPLF/xrpl-dev-portal/commit/27058687b36cdd2a45bb9af6a052252adcab23f9 . This PR should not change it.

content/concepts/tokens/nfts/index.md
```diff
@@ -0,0 +1,46 @@
+---
+html: nft-index.html
```
mDuo13, line 2: It would break fewer links and make migration easier if this file kept the `non-fungible-tokens.html` output filename.

content/concepts/tokens/index.md
```diff
@@ -0,0 +1,53 @@
+---
+html: tk-index.html
```
mDuo13, line 2: Similar with NFTs, this would reduce migration work and break fewer links if it maintained the `tokens.html` output filename.

mDuo13 (changes requested):
I haven't finished going through the whole thing yet, but I wanted to share the things I've seen so far. Big picture, I like the overall structure of the changes, but I noticed several details that should be cleaned up—to be expected, given the scope of changes here.

img/uc-stablecoin-stable-coin.png
DennisDawson: It's a stylistic thing.

content/use-cases/tokenization/stablecoin-issuer.md
DennisDawson: Changed to _currency token_.

content/use-cases/tokenization/stablecoin-issuer.md
DennisDawson: The term supranational is clear in context, if unfamiliar, and used in the linked material. Added info that currency codes don't have to be unique.

content/use-cases/tokenization/stablecoin-issuer.md
DennisDawson: Added the change.

content/use-cases/tokenization/stablecoin-issuer.md
DennisDawson: Added a partial payments section with link.

content/@i18n/ja/concepts/tokens/freezes.md
DennisDawson: Reverted.

content/@i18n/ja/concepts/tokens/decentralized-exchange.md
DennisDawson: Reverted (local builds won't work).

mDuo13:
Merged as #2296

PR #2136, opened 2023-09-18
Update ledger references

### LINKED ISSUE #1824: Ledger object improvements
There are [`common fields` for ledger objects](https://github.com/XRPLF/rippled/blob/develop/src/ripple/protocol/impl/LedgerFormats.cpp#L30) but we don't communicate it way in the docs. 
- `LedgerEntryType`
- `Flags`
- `index`

For transactions there are also `common fields` and they are documented here: https://xrpl.org/transaction-common-fields.html .
This makes it very clear what fields are always included in the transactions. Only the transaction type specific fields are documented in their page, with a referral to the common fields.
We should have a similar way for the ledger objects. A page for ledger object common fields. Currently `index` has it's own page. We could move it to the common fields to have it all in one place. IMO the `Offer` flags seems to describe ledger flags pretty well.

**Also:**
`NFTokenOffer` has a false value (`0x0074`) assigned to it. It's actually [`0x0037`](https://github.com/XRPLF/rippled/blob/develop/src/ripple/protocol/LedgerFormats.h#L162).

mDuo13:
- Fixes  #1824 
- resolves a lot of inconsistencies in the ledger entry types reference
- Makes progress on replacing "ledger object" (non-preferred term) with "ledger entry" (preferred term)


content/references/protocol-reference/ledger-data/ledger-entry-types/accountroot.md
```diff
@@ -83,31 +82,36 @@ Other than those exceptions, these accounts are like ordinary accounts; the LP T
...
-| `lsfDisallowIncomingTrustline`    | `0x20000000` | 536870912         | `asfDisallowIncomingTrustline`    | This account blocks incoming trust lines. _(Requires the [DisallowIncoming amendment][] :not_enabled:.)_ |
+| `lsfDisallowIncomingCheck`        | `0x08000000` | 134217728         | `asfDisallowIncomingCheck`        | This account blocks incoming Checks. _(Added by the [DisallowIncoming amendment][].)_ |
+| `lsfDisallowIncomingNFTokenOffer` | `0x04000000` | 67108864          | `asfDisallowIncomingNFTokenOffer` | This account blocks incoming NFTokenOffers. _(Added by the [DisallowIncoming amendment][].)_ |
+| `lsfDisallowIncomingPayChan`      | `0x10000000` | 268435456         | `asfDisallowIncomingPayChan`      | This account blocks incoming Payment Channels. _(Added by the [DisallowIncoming amendment][].)_ |
+| `lsfDisallowIncomingTrustline`    | `0x20000000` | 536870912         | `asfDisallowIncomingTrustline`    | This account blocks incoming trust lines. _(Added by the [DisallowIncoming amendment][].)_ |
```
mDuo13, line 98: These needed to be updated since DisallowIncoming got enabled

amarantha-k (approved):
LGTM

PR #2139, opened 2023-09-19
rm creator fund tout

AKCodez:
As requested in [ASANA ticket ](https://app.asana.com/0/1203322122024989/1205522920019894/f) .


mDuo13 (approved):
I don't think linking an Asana ticket is very helpful since most people can't see them (even I can't!) but it's fine to remove this section and it the changes are straightforward.

PR #2154, opened 2023-09-22
Unsupported streams in Clio

### LINKED ISSUE #2142: Document unsupported subscription types (Clio)
In Clio the following streams will return unsupported error:
- peer_status
- consensus
- server

godexsoft:
Fixes #2142


intelliot (approved):
Is this actually "New in: Clio v2.0"? What was the behavior in Clio v1?

PR #2165, opened 2023-09-26
add escrow samples

DennisDawson:
Code to accompany the already published Escrow tutorials.

### DIFF
--- a/content/_code-samples/quickstart/js/8.escrow.html
+++ b/content/_code-samples/quickstart/js/8.escrow.html
@@ -1,0 +1,305 @@
+<html>
+  <head>
+    <title>Time-based Escrow Test Harness</title>
+    <link href='https://fonts.googleapis.com/css?family=Work Sans' rel='stylesheet'>
+    <style>
+       body{font-family: "Work Sans", sans-serif;padding: 20px;background: #fafafa;}
+       h1{font-weight: bold;}
+       input, button {padding: 6px;margin-bottom: 8px;}
+       button{font-weight: bold;font-family: "Work Sans", sans-serif;}
+       td{vertical-align: middle;}
+    </style>
+    <script src='https://unpkg.com/xrpl@2.7.0/build/xrpl-latest-min.js'></script>
+    <script src='ripplex1-send-xrp.js'></script>
+    <script src='ripplex2-send-currency.js'></script>
+    <script src='ripplex8-escrow.js'></script>
+  </head>
+  
+<!-- ************************************************************** -->
+<!-- ********************** The Form ****************************** -->
+<!-- ************************************************************** -->
+
+  <body>
+    <h1>Time-based Escrow Test Harness</h1>
+    <form id="theForm">
+      Choose your ledger instance:  
+      &nbsp;&nbsp;
+      <input type="radio" id="tn" name="server"
+        value="wss://s.altnet.rippletest.net:51233" checked>
+      <label for="tn">Testnet</label>
+      &nbsp;&nbsp;
+      <input type="radio" id="dn" name="server"
+        value="wss://s.devnet.rippletest.net:51233">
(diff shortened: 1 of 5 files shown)


content/_code-samples/quickstart/js/8.escrow.html
```diff
@@ -0,0 +1,305 @@
...
+      <br/>
+      <textarea id="seeds" cols="40" rows= "2"></textarea>
+      <br/><br/>
+      <table>
+        <tr valign="top">
+          <td>
+            <table>
+              <tr valign="top">
+                <td>
+                <td>
```
mDuo13: Seems like a duplicated/unclosed table cell here. Thankfully the effects on the page are minimal (it moves the whole thing a couple pixels over) because browsers are pretty good at working around typos like this.

```suggestion
                <td>
```

content/_code-samples/quickstart/js/8.escrow.html
```diff
@@ -0,0 +1,305 @@
...
+                        </table>
+                        <p align="right">
+                          <textarea id="operationalResultField" cols="80" rows="20" ></textarea>
+                        </p>
+                      </td>
+                      </td>
+                    </tr>
+                    </td>
+                    </tr>
+                  </table>
```
mDuo13: There are more duplicate/extra close tags here. I believe the following fixes it:

```suggestion
                      </td>
                    </tr>
                  </table>
```

content/_code-samples/quickstart/js/8.escrow.html
```diff
@@ -0,0 +1,305 @@
...
+                <td>
+                <td>
+                  <button type="button" onClick="getAccount('standby')">Get New Standby Account</button>
+                  <table>
+                    <tr valign="top">
+                      <td align="right">
+                        Standby Account
+                      </td>
+                      <td>
+                        <input type="text" id="standbyAccountField" size="40"></input>
```
mDuo13, line 52: Technically `<input>` elements aren't supposed to have an end tag. You can omit it entirely, or (my preference) use the XML-like syntax for a tag that closes itself.

```suggestion
                        <input type="text" id="standbyAccountField" size="40" />
```

content/_code-samples/quickstart/js/8.escrow.html
```diff
@@ -0,0 +1,305 @@
+<html>
```
mDuo13: Best practice is to add a DOCTYPE so browsers don't use "quirks mode" to render the document.

```suggestion
<!DOCTYPE html>
<html>
```

content/_code-samples/quickstart/js/ripplex8-escrow.js
```diff
@@ -0,0 +1,240 @@
+// *******************************************************
+// ************* Add Seconds to Current Date *************
+// *******************************************************
+
+function addSeconds(numOfSeconds, date = new Date()) {
```
mDuo13, line 5: Instead of using an `addSeconds` method with a magic number of 946684800, a more semantically clear option would be to use one of xrpl.js's built-in time-conversion methods. If you want to convert numeric seconds, [`unixTimeToRippleTime(...)`](https://js.xrpl.org/functions/unixTimeToRippleTime.html) is the one, or if you want to convert from a `Date` instance, [`isoTimeToRippleTime(...)` method](https://js.xrpl.org/functions/isoTimeToRippleTime.html) does that.

Tip: When it comes time to do the Python equivalent, xrpl-py has [`posix_to_ripple_time(...)`](https://xrpl-py.readthedocs.io/en/stable/source/xrpl.utils.html#xrpl.utils.posix_to_ripple_time) which takes int/float seconds as input, and one for converting from Python's native Datetime class type too.

mDuo13 (approved):
Looks OK. I left some nitpicks you can consider.

PR #2168, opened 2023-09-26
Remove XRPL Bounties from funding page

JST5000:
This program is officially sunset, meaning it is not creating any new bounties (although it is still paying out normally for work that is still in progress). So, this should no longer be included on the xrpl.org funding page.

Note: On my local preview, the Creator Fund pop up that's currently on the site wasn't appearing, and I wasn't sure why since it didn't seem my code impacted it. Not sure where that section is coded on the site.


JST5000:
Before:
image
image


After:
image

mDuo13:
The creator fund was removed by #2139 which has been merged but not yet deployed.

mDuo13 (approved):
:+1: Confirmed that it builds and looks right, and there are no SCSS styles nor images that are orphaned by this change. (The `funding-01` through `funding-04` images and related styles are used several times on this page including by the remaining sections.)

PR #2210, opened 2023-10-10
docs: rename AMMAccount to Account for AMM

ckniffen:
The correct field for an `AMM` ledger object is `Account`.

## Context of Change

A while ago I created a ticket (https://github.com/ripple/explorer/issues/734).  Turns out the code for showing a link to the AMM account page already existed but the property in the meta was not the same as is in the unit tests,

We were checking for `AMMAccount` on an `AMM` ledger object in a transaction's meta data. Upon viewing this [transaction](http://localhost:3001/transactions/DCD3DA3A82954C1A0C5FE21092A9F400B28709752F1AD95676B04A321CD66A75/raw) I saw that the property is `Account`.

Another Example: https://xrpl.org/websocket-api-tool.html?server=wss%3A%2F%2Famm.devnet.rippletest.net%3A51233%2F&req=%7B%22command%22%3A%22ledger_entry%22%2C%22index%22%3A%22526E82EAE65D52387843D532CC1EB3F435E6DA66439C5AA368D19E219C7DED66%22%2C%22ledger_index%22%3A%22validated%22%7D

Recreated from https://github.com/ripple/opensource.ripple.com/pull/63

### DIFF
--- a/content/references/protocol-reference/ledger-data/ledger-entry-types/amm.md
+++ b/content/references/protocol-reference/ledger-data/ledger-entry-types/amm.md
@@ -75,7 +75,7 @@ In addition to the [common fields](ledger-entry-common-fields.html), `{{currentp
 |:-----------------|:--------------------|:------------------|:----------|--------------|
 | `Asset`          | Object              | STIssue           | Yes       | The definition for one of the two assets this AMM holds. In JSON, this is an object with `currency` and `issuer` fields. |
 | `Asset2`         | Object              | STIssue           | Yes       | The definition for the other asset this AMM holds. In JSON, this is an object with `currency` and `issuer` fields. |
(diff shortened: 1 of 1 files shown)


mDuo13 (approved):
I think this changed at some point in development and wasn't caught. Thanks for the update.

mDuo13 (commented):
this is correct, but the example AMM JSON needs to be updated too

tequdev:
Can you make the same changes for Japanese?

intelliot:
> this is correct, but the example AMM JSON needs to be updated too

@ckniffen - would you be able to update this?

mDuo13:
Merged as 4a904ab887bf691c1a0518eb8f7384cff3898ee0, 0f474c2f93565e96f86a0cae8f6f2ac99c69b66f, 4f398503c20f6ff00cbad21e4174858cbed7aad0

PR #2213, opened 2023-10-11
Move introduction files

mDuo13:
Relating to IAv3 and Redocly migration work, this moves introduction files to match their location in the IA, so that their "pretty" URLs under Redocly will also match.

There is more of this work, to be done in batches before the switch.


DennisDawson (approved):
The changes look good to me.

PR #2214, opened 2023-10-12
adds  xls-30 banner

AKCodez:


intelliot (changes requested):
do not merge yet - hold for XRPLF review

Silkjaer (changes requested):
Shouldn't be merged. Voting is irrelevant for 99.9% of visitors and for those that it's relevant will already know. We also shouldn't promote any amendments over others, neither pressure/push validators to vote in favor of anything not fixes.

jscottbranson:
Hello,
Let's not merge this, as it sets a unique precedent given that the vast majority of other amendments have not received a banner when they were available to voting.

mDuo13:
Agreed, the banner should not be used to push for votes on specific amendments.

PR #2215, opened 2023-10-14
Add index md files

mDuo13:
Part of the prep work for migrating to Redocly. This "refactor" should result in basically no visible changes to the actual site. (There are some minor quirks around when blurbs are provided explicitly vs implicitly vs inherited from the parent page, but I've tried to work around them.)

Create simple `.md` files to represent the automatic index pages that previously only existed in the `dactyl-config.yml` file. Due to the order pages are generated, I had to explicitly specify blurbs for some. In a few cases I made the actual content of the page a tiny bit more thorough while I was there.

Also includes the script I used for generating the indexes, but it probably shoudn't be needed anymore.


content/@i18n/ja/concepts/index.md
```diff
@@ -0,0 +1,8 @@
+---
+html: concepts.html
+parent: docs.html
+top_nav_grouping: カテゴリ
+template: pagetype-category.html.jinja
+---
+# コンセプト
+XRP Ledgerの基本的な部分の背景に「何があるか」、「なぜなのか」を学びましょう。
```
amarantha-k, line 9: Need an empty line after the page title.

PR #2220, opened 2023-10-17
Fix typo in payment types table

mDuo13:
It refers to the types of the transaction so it should be `Account` (the sender), not `Address` (not a real field).


intelliot:
> It refers to the types of the transaction so it should be `Account` (the sender), not `Address` (not a real field).

types -> fields ?

intelliot (approved):

PR #2225, opened 2023-10-18
Document CTID in tx method

### LINKED ISSUE #2157: Document ctid field
Support for Compact Transaction Identifiers (CTID) was added in https://github.com/XRPLF/rippled/pull/4418

This affects the API reference docs.

I'm not totally sure which API methods are affected. For sure `tx` responses can include it; and I think  `tx` requests can take a CTID in the `transaction` field? Not sure about the commandline interface.

Other places where I might expect CTIDs to show up, and would need to verify if they're affected:

- `ledger` with expanded transactions
- `account_tx`
- subscription messages for transactions

mDuo13:
Fixes #2157.

I looked into it and the PR _only_ added the field to the request/response formats for the `tx` command, not any of the other places transactions can appear. I think transaction unification in API v2 will cause it to appear in more places, but that's a problem for later.


content/references/http-websocket-apis/public-api-methods/transaction-methods/tx.md
```diff
@@ -464,9 +274,9 @@ This means that a `txnNotFound` on its own is not enough to know the [final outc
 
 To further narrow down the possibilities, you can provide a range of ledgers to search using the `min_ledger` and `max_ledger` fields in the request. If you provide **both** of those fields, the `txnNotFound` response includes the following field:
 
-| `Field`        | Type      | Description                              |
+| Field          | Type      | Description                              |
```
intelliot, line 277: nice catch

intelliot (approved):

tequdev:
The link for `transactions.html#identifying-hash` does not exist and is causing an error.

PR #2235, opened 2023-10-24
[DO NOT MERGE] adds allbridge to interoperability

AKCodez:


mDuo13 (approved):
Generally looks good, but I noticed that the numbers in the circle grid are weird. For example "Interoperability" has two icons but the number is 3; Infrastructure has 7 icons and number 7 but there are 8 items in the filterable list below.

PR #2243, opened 2023-10-30
Migrate the faucet page to Redocly

JST5000:
Migrating this page: https://xrpl.org/xrp-testnet-faucet.html

# Steps to migrate
 - [x] Get the developer url (localhost:4000/xrp-testnet-faucet.html) to display a new file that we define in Redocly (likely using Redocly’s equivalent of dactyl-config.yml)
    - You can find the page at http://localhost:4000/dev-tools/xrp-faucets/ and edit it at dev-tools/xrp-faucets.page.tsx- It is missing a sidebar though, so that needs to be fixed somehow later on.
 - [x] Get the basic HTML elements displaying (migrating a lot of page-xrp-faucets.html.jinja without the reactivity or CSS)
 - [x] Add the logic from faucet.js into the page (requires migrating JQuery to React syntax)
 - [x] Fix the css so that it displays like it used to
 - [x] Make sure that clicking “Faucets” from the page top navigation works
 - [x] Make sure that clicking the faucet link from the docs homepage (https://xrpl.org/docs.html) works 
 

## Visible pieces of the page to migrate
1. Some text to be formatted near the top (title + explanation)
2. A right-hand sidebar that has copyable links for each of the chains
3. A modal dialogue to pick your network
4. A button which when you press it triggers a faucet call
5. A loading icon when the button is clicked that appears underneath (uses class “throbbing” in xrpl-dev-portal)
6. Headers + text to appear under displaying the retrieved data (transformation is implemented in faucet.js currently)


JST5000:
Will take this out of draft when done, wanted to open this as a PR so that work in progress could be used as a reference for others :)

JST5000:
TODO before merging: Double check that the link from the top nav has a nice name for the page. (As opposed to a filename)

JST5000:
Remaining to fix:
- [x] AMM Devnet faucet isn't working in my branch (specifically when trying to use `fundWallet`) - Currently investigating.
- [x] Making the files more readable by breaking out longer lines (feedback from @ckniffen )
- [x] Making top-nav link have a nice header instead of a filename / path
- [ ] Fix build preview (unsure why it's broken, I don't have access to the logs)

JST5000:
TODO before merging:
- [x] Type everything using xrpl for types (only using type imports is fine :) )

content/dev-tools/test.js
```diff
@@ -0,0 +1,17 @@
+import xrpl from "xrpl"
```
ckniffen: This file seems unneeded

content/dev-tools/xrp-faucets.page.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+  const [generatedCredentialsFaucet, setGeneratedCredentialsFaucet] = useState("")
+  const [address, setAddress] = useState("")
+  const [secret, setSecret] = useState("")
+  const [balance, setBalance] = useState("")
+  const [sequence, setSequence] = useState("")
+
+return (<div>
+    {/* TODO: Once xrpl.js 3.0 is released, replace this with a direct xrpl.js import */}
+    <script src="https://unpkg.com/xrpl@2.5.0-beta.0/build/xrpl-latest-min.js" async />
```
ckniffen: This looks to be included in the redocly.yml.  may not need to be her too.

Why are you using version 2.5?

content/dev-tools/xrp-faucets.page.tsx
```diff
@@ -0,0 +1,228 @@
...
+        <button id="generate-creds-button" onClick={
+            () => generateFaucetCredentialsAndUpdateUI(
+              selectedFaucet, 
+              setGeneratedCredentialsFaucet, 
+              setAddress, 
+              setSecret, 
+              setBalance, 
+              setSequence)
+          } className="btn btn-primary mr-2 mb-2">
+            Generate {selectedFaucet.shortName} credentials
```
ckniffen: Translate

content/dev-tools/xrp-faucets.page.tsx
```diff
@@ -0,0 +1,228 @@
...
+        </button>
+      </div>
+    {/* </XRPLGuard> */}
+
+    {generatedCredentialsFaucet && <div id="your-credentials">
+      <h2>{translate("Your")} {generatedCredentialsFaucet} {translate("Credentials")}</h2>
```
ckniffen: Can this be one value with interpolation like i18next allows?

content/docs.page.tsx
```diff
@@ -407,7 +407,7 @@ export default function Docs() {
                       'Connect to the XRP Ledger Testnet network to develop and test your apps built on the XRP Ledger, without risking real money or impacting production XRP Ledger users.'
                     )}
                   </p>
-                  <a className="btn btn-primary btn-arrow" href="xrp-testnet-faucet.html">
+                  <a className="btn btn-primary btn-arrow" href="../dev-tools/xrp-faucets">
```
ckniffen: Make sure redirects are configured so old links still work after the rename

content/dev-tools/xrp-faucets.page.tsx
```diff
@@ -0,0 +1,228 @@
...
+    {secret && <div id="secret"><h3>{translate("Secret")}</h3>{secret}</div>}
+    {(address && !balance) && (<div><br/>
+        <div id="loader" style={{display: (address && !balance) ? "inline" : "none"}}>
+          <img alt="(loading)" className="throbber" src="/img/xrp-loader-96.png" /> {translate("Funding account...")}
+        </div>
+      </div>)}
+    
+    {balance && <div id="balance">
+      <h3>{translate("Balance")}</h3>
+      {(Number(balance) * 0.000001).toLocaleString("en")} {translate("XRP")}
```
ckniffen: Doesn't `xrpl` have a utility for this calculation?

content/dev-tools/xrp-faucets.page.tsx
JST5000: Turns out it is unnecessary :) 

(And it was 2.5 due to some experimenting I was doing I believe, I forget what I was trying to learn previously though hahaha)

content/docs.page.tsx
JST5000: @mDuo13 Said they'd handle that separately :)

content/dev-tools/xrp-faucets.page.tsx
JST5000: They do, but I can't use `xrpl` at the top-level. Doing so dereferences a null value. It can only be used within functions like `onChange` / `onClick` (I believe because if you include it at the top level it tries to run the code at compile/loading the DOM time instead of after the `<script>` is loaded

ckniffen (approved):

content/dev-tools/faucets.json
mDuo13, line 1: Couple things to update in this PR:

1. Add Xahau Testnet. ( #2282 )
2. Drop AMM-Devnet ( #2288 )

Also, please make sure all files end in a newline.

content/@i18n/ja/concepts/networks-and-servers/parallel-networks.md
```diff
@@ -33,7 +33,7 @@ Ripple社は、TestnetとDevnetでメインサーバーを運用しています
 ## 関連項目
 
 - **ツール:**
-  - [XRP Testnet Faucet](xrp-test-net-faucet.html)
+  - [XRP Testnet Faucet](/dev-tools/xrp-faucets)
```
mDuo13: I'd like to handle link migration separately because we're going to have to do mass replacements of other links and I want to minimize the occasions for merge conflicts. How hard would it be to change the PR to exclude all these changes to update links in `.md` files?

content/static/js/xrpl-guard.tsx
```diff
@@ -0,0 +1,86 @@
...
+      {isXRPLLoaded ? (
+        isFunction(children) ? (
+          children()
+        ) : (
+          children
+        )
+      ) : (
+        <div id="loader">
+            <img alt="(loading)" className="throbber" src="/img/xrp-loader-96.png" />{translate("Loading...")}
+        </div>)}
```
mDuo13: In the first pass migration, we made [a reusable component](https://github.com/ripple/xrpldotorg-redocly-migration-poc/blob/master/_components/XRPLoader.tsx) for an XRPL loading widget. I'm not sure how different the syntax is now but I think we should still be able to do something like that.

content/dev-tools/xrp-faucets.page.tsx
mDuo13: Redocly added some kind of onready callback function that might be useful for this sort of use case, though I haven't been able to find the docs for how to use it.

template/page-docs.html.jinja
```diff
@@ -212,7 +212,7 @@
     set devTools = [
     {
       "title": _("Faucets"),
-      "link": "./xrp-testnet-faucet.html",
+      "link": ".//dev-tools/xrp-faucets",
```
mDuo13: I don't think `.//` is the right way to link this. Maybe just the following instead?

```suggestion
      "link": "/dev-tools/xrp-faucets",
```

mDuo13 (changes requested):
It's weird that the faucet page doesn't have a sidebar even though it's listed in the sidebar. I wonder if we need to do something custom with the template to add it explicitly.

Mostly works but I would like to reduce the scope of the PR a little to reduce merge conflicts.

ckniffen:
You should incorporate the changes from https://github.com/XRPLF/xrpl-dev-portal/pull/2282 and removing AMM devnet which will be in https://github.com/XRPLF/xrpl-dev-portal/pull/2288

content/@i18n/ja/concepts/networks-and-servers/parallel-networks.md
JST5000: Very doable :)

content/dev-tools/xrp-faucets.page.tsx
JST5000: Using the xrpl one now :) (Beta version works as a direct import)

content/@i18n/ja/concepts/networks-and-servers/parallel-networks.md
JST5000: Done :)

template/page-docs.html.jinja
JST5000: Removed this change, this was an accidental overzealous search and replace :)

content/static/js/xrpl-guard.tsx
JST5000: Added XRPLoader based on that :)

content/concepts/networks-and-servers/parallel-networks.md
```diff
@@ -19,7 +19,7 @@ To help members of the XRP Ledger community interact with XRP Ledger technology
...
 | AMM-Devnet | [XLS-30d pre-release](https://github.com/gregtatcam/rippled/tree/amm-core-functionality/) | A preview of the [XLS-30d](https://github.com/XRPLF/XRPL-Standards/discussions/78) standard for Automated Market Makers on the XRP Ledger. <!-- SPELLING_IGNORE: 30d --> [Binary builds for testing](https://github.com/legleux/scheduled/releases) are also available. Library support: [xrpl.js 2.6.0-beta.0](https://www.npmjs.com/package/xrpl/v/2.6.0-beta.0), [xrpl-py 1.8.0b0](https://pypi.org/project/xrpl-py/1.8.0b0/) |
 
-Each altnet has its own separate supply of test XRP, which is [given away for free](xrp-testnet-faucet.html) to parties interested in experimenting with the XRP Ledger and developing applications and integrations. Test XRP does not have real-world value and is lost when the network is reset.
+Each altnet has its own separate supply of test XRP, which is [given away for free](xrp-test-net-faucet.html) to parties interested in experimenting with the XRP Ledger and developing applications and integrations. Test XRP does not have real-world value and is lost when the network is reset.
```
mDuo13: Revert this change. (old URL)

content/static/js/xrpl-guard.tsx
mDuo13: Shouldn't you be using it here, then?

template/page-docs.html.jinja
```diff
@@ -212,7 +212,7 @@
     set devTools = [
     {
       "title": _("Faucets"),
-      "link": "./xrp-testnet-faucet.html",
+      "link": "./xrp-test-net-faucet.html",
```
mDuo13: Revert this one

content/_code-samples/get-started/py/prepare-payment.py
```diff
@@ -7,7 +7,7 @@
 
 # Create a wallet using the testnet faucet:
 # https://xrpl.org/xrp-testnet-faucet.html
-from xrpl.wallet import generate_faucet_wallet
+# from xrpl.wallet import generate_faucet_wallet
```
mDuo13: I think there's some mistake here. If you comment out this import, the code sample no longer works. I'm not sure why this PR should be changing a code sample anyway.

template/page-docs.html.jinja
```diff
@@ -92,7 +92,7 @@
       },
       {
         "description": _("XRP Faucet"),
-        "link": "./xrp-testnet-faucet.html",
+        "link": ".xrp-testnet-faucet.html",
```
mDuo13: Revert this change too. (`.xrp-testnet-faucet.html` with the leading `.` and no slash is not right either)

template/page-xrp-faucets.html.jinja
mDuo13, line 1: For now, let's leave the old `.jinja` templates. We still need to be able to run the Dactyl build on this branch for purposes of the link migration script, and it'll fail if a template it references has been deleted.

After the link migration is done, I'll delete all the old templates.

styles/yarn.lock
mDuo13, line 1: I'm not sure if you need to upgrade the styles lockfile or the `.css` file for this PR. If you don't need to modify them in this PR, it's best to leave those files untouched (we're going to have enough merge conflicts on the `devportal2022-v25.css` file as it is; no need to have more).

content/static/components/XRPLoader.tsx
```diff
@@ -0,0 +1,13 @@
...
+    message?: string
+}
+
+export default function XRPLoader(props: XRPLoaderProps) {
+  return (
+  <div id="loader" style={{ display: "inline" }}>
+    <img alt="(loading)" className="throbber" src="/img/xrp-loader-96.png" />
+    {props.message}
+  </div>);
+}
```
mDuo13: Nit: end files in a newline (this and a couple others)

mDuo13 (changes requested):
Looks like there are still some changes in this diff that are unrelated to the task at hand. If you can rebase to squash/drop as much of those as possible, that'll make the merging process easier. (We're going to have to rebase the `redocly-migration` branch onto the latest master after merging these, possibly more than once, so I want to keep each PR as constrained as possible, to minimize complications there.)

content/dev-tools/faucets.json
```diff
@@ -0,0 +1,28 @@
+{
```
pdp2121, line 1: Maybe put this file into a `data` folder or something

JST5000: I think it's useful to keep the data near where it's used in this case since it's super related to the test & there's not many files so it doesn't clutter a reader's ability to tell what all the files are for / doing.

mDuo13 (approved):

PR #2246, opened 2023-11-01
Fix typo in auto-bridging blurb

mDuo13:
Reported on Slack.

Unrelated, I haven't figured out why the sidebar for this page is partly buried under the header. It seems the styles are exactly the same as the one on pages that don't have problems...


amarantha-k (approved):
Thanks for fixing the typo.

intelliot (approved):

PR #2247, opened 2023-11-01
ecosystem updates mismatching modals vs cards

AKCodez:


assets/img/uses/modalLogos/Cryptum.png
mDuo13, line 1: Change filename to all lowercase so we don't have to worry about case mismatches on *nix. (Update the CSS accordingly)

mDuo13 (changes requested):
Generally looks good, but let's not add more title case image files.

assets/img/uses/modalLogos/Cryptum.png
AKCodez: done

mDuo13 (approved):

assets/img/uses/modalLogos/Cryptum.png
mDuo13: Actually, it looks like this is not done? Still showing as `Cryptum.png` in the GitHub preview.

mDuo13 (changes requested):

assets/img/uses/modalLogos/Cryptum.png
mDuo13: This must be some kind of platform issue, because it's still showing in title case here.

I'll try merging and fixing the problem myself, since I'm on Linux and have a case-sensitive filesystem.

PR #2250, opened 2023-11-02
account_tx only returns validated transactions

mDuo13:
Per https://github.com/XRPLF/rippled/pull/4775#discussion_r1380587801 this command only returns transactions from validated ledgers, so this clarifies that detail


intelliot (approved):
good catch

PR #2254, opened 2023-11-08
Migrate the tx-sender page (https://xrpl.org/tx-sender.html)

JST5000:
Migrates the [tx-sender page](https://xrpl.org/tx-sender.html) to the new Redocly / React style. 

Files to migrate:
- `page-tx-sender.html.jinja`
- `tx-sender.js`

TODO
- [x] Change all links that previously went to tx-sender.html to dev-tools/tx-sender
  - [x] Top nav
  - [x] Homepage
  - [x] Ctrl + shift + F for "tx-sender" in any links
- [X] Add this to the sidebar
  - [x] Make sure the list item has a pretty name for the page (not the file location / filename)
- [x] Figure out how this code is dynamic (the js files) and then translate that over in a React-like way
  - [x] Specifically, look at `tx-sender.js`
 
Verify that these components are still part of the page:
- Explanatory text at the top
- Main buttons
- Right sidebar which displays the account status
- Green / Red pop ups on transaction completion / resolution (red if error)
- Loading bars with status updates as the partial payments button gets ready to turn on


JST5000:
This is now ready for review. To help with that here's some context on the PR:
- It migrates the https://xrpl.org/tx-sender.html page which can now be found at http://127.0.0.1:4000/dev-tools/tx-sender (Replace http://127.0.0.1:4000 with the preview URL to access it :) )
  - Because the only link to this currently is in the sidebar, and there is no sidebar, you must enter the URL manually in the preview to see it.
- The general structure of the page is:
  1. A brief intro text explaining the tool
  2. A sidebar on the right which shows when the client is connected along with basic info like `sendingWallet` address and current account balance, along with a list of transactions that get sent (with links to the explorer)
  3. A field to change the destination address (This should give a unique warning if a valid X Address is provided, since testnet X addresses start with a T not an X)
  4. Many buttons with slightly different formats which submit transactions. 
  5. Note: After initializing, all buttons EXCEPT the 2nd to last one will succeed by default. (The 2nd to last one requires you to set up a trustline from the destination account to your `sendingWallet`. This has to be done by your code as there's no button which does this for you)

Thanks for taking the time - this is a bit of a long one to review :)

content/static/vendor/bootstrap-growl.jquery.js
```diff
@@ -22,6 +22,11 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 */
 
+/**
+ * Replaced by react-alert after the Redocly migration. 
+ * Please see tx-sender.page.tsx for an example of react-alerts in action.
+ */
+
 (function() {
```
ckniffen, line 30: Might as well delete this file since you already deleted. page-tx-sender.html.jinja

content/dev-tools/utils.tsx
```diff
@@ -0,0 +1,105 @@
+import * as React from 'react'
+import { type Client, type Wallet, type Transaction, type TransactionMetadata, type TxResponse } from 'xrpl'
+import { clsx } from 'clsx'
+
+
+export const TESTNET_URL = "wss://s.altnet.rippletest.net:51233"
+
+export function isoTimeToRippleTime(isoSeconds: number): number {
```
ckniffen: You can use `rippleTimeToISOTime`

content/dev-tools/DestinationAddressInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <div className="form-group">
+            <label htmlFor="destination_address">
+                {translate("Destination Address")}
+            </label>
+            <input type="text" className={clsx("form-control", 
+                // Defaults to not having "is-valid" / "is-invalid" classes
+                (destinationAddress !== defaultDestinationAddress) && (isValidDestinationAddress ? "is-valid" : "is-invalid"))}
+                id="destination_address" 
+                onChange={(event) => onDestinationAddressChange(event, setDestinationAddress, setIsValidDestinationAddress)}
+                aria-describedby="destination_address_help" defaultValue={destinationAddress} />
```
ckniffen: nit: put defaultValue on new line

content/dev-tools/DestinationAddressInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+            <small id="destination_address_help" className="form-text text-muted">
+                {translate("Send transactions to this XRP Testnet address")}
+            </small>
+        </div>
+        <p className={clsx("devportal-callout caution", !(isValidDestinationAddress && destinationAddress[0] === 'X') && "collapse")}
```
ckniffen: This logic seems wrong.  Wouldn't it not equal 'X' if it is an `r` address?

content/dev-tools/AlertTemplate.tsx
```diff
@@ -0,0 +1,41 @@
...
+  style: any
+  close: any // Callback to close the alert early
+}
+  
+export default function AlertTemplate ({ message, options, style, close }: AlertTemplateProps): React.JSX.Element {
+  return(
+    <div className={clsx("bootstrap-growl alert alert-dismissible", typeToClass(options.type))} style={{ ...alertStyle, ...style }}>
+      <button className="close" data-dismiss="alert" type="button" onClick={close}>
+        <span aria-hidden="true">×</span>
+        <span className="sr-only">Close</span>
```
ckniffen: translate

content/dev-tools/TransactionButton.tsx
```diff
@@ -0,0 +1,156 @@
+import * as React from 'react';
+import { useState } from 'react'
```
ckniffen: you can combine these imports

ckniffen:
Talked with @JST5000 offline about adding a client provider in a future to clean up prop drilling and likely clean up the other tools too.

ckniffen:
Can you move these components to a sub folder?

content/dev-tools/tx-sender.page.tsx
```diff
@@ -0,0 +1,438 @@
...
+    const commonTxButtonParams = {
+        submitConstData,
+        connectionReady,
+        sendingWallet
+    }
+    
+    return (
+        <div className="row">
+            {/* TODO: Once xrpl.js 3.0 is released, replace this with a direct xrpl.js import */}
+            <script src="https://unpkg.com/xrpl@2.5.0-beta.0/build/xrpl-latest-min.js" async />
```
ckniffen: This should be able to be removed now as it is included in the redocly.yml

content/dev-tools/tx-sender.page.tsx
```diff
@@ -0,0 +1,438 @@
...
+                                {
+                                    defaultValue: defaultDropsToSend,
+                                    setInputValue: setDropsToSendForPayment,
+                                    min: 1,
+                                    max: 10000000000,
+                                }}
+                            />
+                            
+                            {/* Partial Payments */}
+                            <TransactionButton
```
ckniffen, line 230: Love the abstraction

content/dev-tools/tx-sender.page.tsx
```diff
@@ -0,0 +1,438 @@
...
+                                    setIsInitEnabled,
+                                    setConnectionReady,
+                                    partialPaymentParams
+                            }}/>
+
+                            <DestinationAddressInput 
+                                {...{defaultDestinationAddress, 
+                                destinationAddress, 
+                                setDestinationAddress, 
+                                isValidDestinationAddress,
```
ckniffen: I might be missing this but where is this used outside of DestinationAddressInput.  This validation could be encapsulate inside of the component and not need to be passed in.

ckniffen (dismissed):

pdp2121 (commented):
Maybe you should put the components like buttons, sidebar, etc. inside a separate folder, and just leave the `tx-sender.page.tsx` file at the root of devtools, to make it easier for future navigation.

content/static/vendor/bootstrap-growl.jquery.js
JST5000: @mDuo13 Asked me not to delete the old jinja files, so I am reverting that change. @mDuo13 Should we keep bootstrap-growl or delete it? (I don't believe it is used in the new pages)

content/dev-tools/DestinationAddressInput.tsx
JST5000: If it's an r address this hides as expected because the 1st character is not X, this error is specifically for valid X addresses which are pointing at mainnet instead of testnet, so won't work with this tool. 

Using a different writing of the logic:
Only display if this is a valid X address pointed at mainnet. 

Screenshot of the error in action:
image

content/dev-tools/TransactionButton.tsx
JST5000: Import * from ____ doesn't seem to work with direct imports (At least I don't know the syntax for it after trying like 5 combinations)

content/dev-tools/tx-sender.page.tsx
JST5000: Good catch! :)

JST5000:
> Maybe you should put the components like buttons, sidebar, etc. inside a separate folder, and just leave the `tx-sender.page.tsx` file at the root of devtools, to make it easier for future navigation.

@pdp2121 Good callout! :)

pdp2121 (approved):

mDuo13 (approved):

mDuo13:
Merged as 06b7e91957e2e19b3c8fe7b27338a36132ffefa6

PR #2255, opened 2023-11-08
AI widget style changes

jonathanlei:
modify the AI widget according to design


styles/package.json
```diff
@@ -1,14 +1,14 @@
...
+    "sass": "^1.69.5"
   },
   "scripts": {
     "build-css": "node-sass --include-path scss xrpl.scss ../assets/css/devportal2022-v25.css --output-style compressed",
     "build-css-out": "node-sass --include-path scss xrpl.scss ../out/assets/css/devportal2022-v25.css --output-style compressed --source-map true",
     "build-css-watch-out": "node-sass --recursive --watch  --include-path scss xrpl.scss ../out/assets/css/devportal2022-v25.css --output-style compressed --source-map true"
   },
   "dependencies": {
-    "sass": "^1.26.10"
+    "node-sass": "^9.0.0"
```
JST5000: Why were these swapped?

template/base.html.jinja
```diff
@@ -57,7 +57,8 @@
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@docsearch/css@3" />
 
     <script src="https://cmp.osano.com/AzyjT6TIZMlgyLyy8/f11f7772-8ed5-4b73-bd17-c0814edcc440/osano.js"></script>
-    <link rel="stylesheet" href="https://www.unpkg.com/@xrpl/ai-css/xrplai.css" />
+    <!-- <link rel="stylesheet" href="https://www.unpkg.com/@xrpl/ai-css/xrplai.css" /> -->
+    <link rel="stylesheet" href="{{currentpage.prefix}}assets/css/xrplai.css" />
```
JST5000, line 61: Why is this link commented? (Should it be reverted to the old link, or should that comment be deleted?)

template/base.html.jinja
```diff
@@ -194,35 +207,37 @@ Answer (including related code snippets if available and markdown links to relev
...
+              // return file.path.replace(/\.[^.]+$/, '');
+              // Change extension to html and remove content/ from the URL
+              if (file.path.endsWith('ja.md')) {
+                // Japanese lives at /ja/<filename>.html
+                return `ja/${file.path.replace(/content\//, '').replace(/\.ja\.md$/, '.html')}`
+              }
+              return file.path.replace(/content\//, '').replace(/\.md$/, '.html')
+            } else {
+              // Otherwise link to a file on the github repo such as .js or README.md
+              return `https://github.com/XRPLF/xrpl-dev-portal/blob/master/${file.path}`;
```
JST5000, line 226: This section's indentation is off which makes it much harder to review because it creates unnecessary diffs. It may seem like a small thing, but it can also mess up the Git blame history and cause future developers to come ask you to explain this code instead of the original writers as this will show up as the last change on this line. I'd recommend skimming through this PR and fixing the indentation of any lines which are actually the same on both sides of the diff.

(If you setup a prettier config I think you can automatically have this formatting stuff fix itself)

styles/package.json
ckniffen: `node-sass` is deprecated.  `sass` should be used instead. So this is going backwards.

template/base.html.jinja
```diff
@@ -232,11 +247,20 @@ Answer (including related code snippets if available and markdown links to relev
...
+          displayedBeforeAnswer: false,
+        }
+      },
     },
   };
 </script>
-<script id="initjs" type="module" src="https://www.unpkg.com/@xrpl/ai-web/dist/init.js"></script>
+<script type="module" src="{{currentpage.prefix}}assets/js/xrplai.js"></script>
+{#
+<script id="initjs" type="module" src="{{currentpage.prefix}}assets/js/xrplai.js"></script> #}
```
JST5000, line 261: Are these necessary? Both seem to be importing the same code and this seems like this might be the same script as near the top of this file as well.

styles/package.json
jonathanlei: reverted the changes

template/base.html.jinja
jonathanlei: waiting for the ai-js library to be updated and deployed on unpkg so I don't have to import those files

template/base.html.jinja
jonathanlei: this would be brought back once the changes on xrplai is deployed

template/base.html.jinja
jonathanlei: thanks for fixing the indentation - I went back and checked things - I commented out the algolia part since it's not used anymore - but might be useful in the future so I didn't delete it

template/base.html.jinja
JST5000: Since this PR is not in draft mode - can you change it to the way it should be merged? (Locally you can have it differently and just not commit in order to test changes)

Otherwise I'm uncomfortable approving the PR because this is not what should be merged

template/base.html.jinja
```diff
@@ -168,19 +169,31 @@ $(document).ready(() => {
 
   /** @type {import('@xrpl/ai-react').MarkpromptOptions} */
   window.markprompt = {
+    // NOTE: This is the development key, DO NOT USE THIS IN PRODCTION
```
JST5000, line 172: Please clean up this comment & related code

template/base.html.jinja
JST5000: You can probably just delete it - if someone needs it they can refer to the git history :)

template/base.html.jinja
```diff
@@ -194,35 +207,37 @@ Answer (including related code snippets if available and markdown links to relev
...
+        },
         getLabel: (reference) => {
           return reference.meta?.leadHeading?.value || reference.file?.title;
         },
       },
       showBranding: false,
       trigger: {
-        floating: true,
+        customElement: true,
+        customElementId: "topsearchboxcontainer",
```
JST5000, line 240: I'd recommend using kebab case for element ids - makes them easier to read. (Please update all other areas that use this id if you do)
```suggestion
        customElementId: "top-search-box-container",
```

template/base.html.jinja
JST5000: Similar to the comment above, please push changes to this branch so it's ready to go as if the other half is already done. (Locally can be set up to allow you to test your code differently). 

I tend to think that PR's should pretend they're already production code, or be in draft mode, to clearly signal that as-written these changes shouldn't be merged. That way no accidents happen where code is fine "except one last change" and then gets merged.

intelliot:
- Justin will handle the feedback from Jackson

amarantha-k:
@justinr1234 Could you update this PR to apply these style changes to the Redocly-based style? Thanks!

mDuo13:
Closing this for now. The re-created version on Redocly will be a separate PR (coming soon™)

Issue #2256, opened 2023-11-09
Improved "Installation" landing page


mDuo13:
Rather than just a bulleted list of page names & descriptions, a better "Installation" landing page could have things grouped by the type of server needed (i.e. a core p2p server, API server), operating system, and install type (packaged binary vs self-built). Graphics or icons (i.e. OS icons) could also help.

PR #2257, opened 2023-11-09
Reorg infrastructure folders

mDuo13:
Another "pre-moving" task leading up to the Redocly migration.

- Rename folders to match intended URL structure and navigation
    - No "rippled" / "Clio" folders under Infra. (Moved Clio install instructions under installation.)
    - Move "Configure Peering" under Configuration
- Move some files out of concepts (history sharding, online deletion) to reflect their current place in the nav
- Rename "Install rippled" index→"Installation" and "Configure rippled"→"Configuration"
- Move the Docker private network and amendment testing steps under "Testing and Auditing"

Intentionally not done in this PR:

- #2256 
- Updating the contents of the pages (i.e. removing old migration instructions)
- Renaming the output filenames (i.e. `install-rippled.html` for "Installation" which includes Clio, or `use-stand-alone-mode.html` for "Testing and Auditing" which includes the private network docs). This would involve migrating links in many existing files and adding a redirect; but we're going to need to add redirects and migrate links differently for the Redocly move; so that work that would need to be redone a month later anyway.


amarantha-k (approved):
Tested a local build and the reorg changes look good.

PR #2259, opened 2023-11-10
Script to migrate various Dactyl/Jinja syntax to Redocly/Markdoc

mDuo13:
Script to convert a bunch of syntax in the Markdown files to be compatible with Redocly. The idea is that we can rebase the `redocly-migration` branch on the latest `master` branch; when we do, we can drop the commit that does the mass find-replace and re-create it using the script, so we don't have to freeze work on `master` while working on the migration.

- [x] tabs to spaces (for consistency and to enable some other replacements)
- [x] `{{ include_code(...) }}` → `{% code-snippet ... /%}`
- [x]  `{% include ... %}` for mds → `{% partial ... %}`
- [x] `{% include ... %}` for non-md → `{% code-snippet ... /%}` (where necessary, uses the "prefix" feature of code-snippet added in Redocly v0.66.0)
- [x] github_fork / github_branch variables → `{% repo-link /%}` component
- [x] `{{currentpage.name}}` variable → `{% code-page-name /%}` component or `{% $frontmatter.seo.title %}` as necessary
- [x] owner / account reserve variables
- [x] category template → child pages component (conversion script works but I'm having issues with the component itself)
- [x]  `include_svg(...)` → `{% inline-svg ... /%} ...` conversion works, with caveats
    - [x] link original diagram. :warning: The URL loads for me if I open it in a new tab, but clicking it doesn't work (possibly the same navigation issue I've been having)
    - [x] use caption from original syntax
    - Note, it also requires the `img/` dir to be moved under `content/` right now, but we probably will move some stuff around differently instead of doing this.
    - Theme-aware diagrams should work with the new system, but the CSS needs to be adjusted a little.
- [x] callouts → admonitions
- [x] Multicode block → `{% tabs %}` component
- [x] indented code blocks → (sometimes-indented) code fences. In a separate script (to be run _after_ the `md_dactyl_to_redocly.py` one).

Usage:

1. Optional: Run link replacement, if desired (that script is from outside of this PR)
2. Optional, temporary: Move `img/` to `content/img/` and fix any pages that were using the old image file paths (there's one in `xrp-overview.page.tsx`, I think)
3. From repo top, run:

        tool/md_dactyl_to_redocly.py

4. From repo top, run:

        node tool/convert-code-blocks.cjs content/


amarantha-k (approved):
Was able to run the script and build a local preview (although with expected issues with links and images). The script looks good to merge.

PR #2262, opened 2023-11-14
Community page

AKCodez:


template/page-community.html.jinja
```diff
@@ -17,386 +17,228 @@
...
-        <h6 class="eyebrow mb-3">{% trans %}XRPL Grants{% endtrans %}</h6>
-      </div>
-      <p class="mb-lg-3 py-lg-4 pt-4 mb-0">{% trans %}The XRPL Grants program funds select open-source projects that solve problems and grow the XRP Ledger community.{% endtrans %}</p>
-      
-      <div class="d-lg-block d-none">
-        <div class="mb-4 pb-3" id="xrplGrantsDark"></div>
-        <a class="btn btn-primary btn-arrow" target="_blank" href="https://xrplgrants.org/">{% trans %}Apply for a Grant{% endtrans %}</a> 
+<section class="carousel">
+    <div class="image-container">
+      <img id="left-image" src="" alt="Left Event Image">
```
mDuo13: This empty `src` attribute is one of the things being reported as a broken link. If the image is being set by CSS, you should omit the `src` attr entirely.

template/page-community.html.jinja
```diff
@@ -17,386 +17,228 @@
...
+      <div class="center-image-wrapper">
+          <img id="center-image" src="" alt="Featured Event Image">
```
mDuo13: Another empty `src` attr.

template/page-community.html.jinja
```diff
@@ -17,386 +17,228 @@
...
+                <span class="icon-location"> </span>
+                <span id="event-location"></span>
+              </div>
+              <div class="flex-align">
+                <span class="icon-date">  </span>
+                <span id="event-date"> </span>
+              </div>
+          </div>
       </div>
+      <img id="right-image" src="" alt="Right Event Image">
```
mDuo13: Third empty `src` attr.

styles/_pages.scss
```diff
@@ -579,12 +617,21 @@
   }
 }
 
+.autoscroll-content {
+  animation: autoscroll 15s linear infinite;
+  white-space: nowrap;
+  overflow: hidden;
+  max-width: 300px;
+  /* adjust this value based on your design requirements */
```
mDuo13: Did you copy-paste this, comment included, from somewhere else? The comment doesn't make sense here.

assets/img/community/card-bg-1.png
mDuo13, line 1: This seems like a graphic that should be an SVG. A few diagonal lines in a transparent field should not be 8kb; the 2 and 3 ones are even worse. I know it may not be much, but it adds up and makes the site load slower, especially for people not on the latest machines and fastest connections.

styles/_contribute.scss
```diff
@@ -0,0 +1,1158 @@
...
+        }
+
+        .td-img {
+            min-width: 60px;
+            /* Remove fixed width for mobile view */
+        }
+    }
+
+    .funding-text {
+        color: var(--black-black-0, #FFF);
```
mDuo13: What is going on with these color names? "black black zero"? We already have `$white`, and others defined in `_colors.scss`.

mDuo13 (changes requested):
A bunch of the new `ic_*` files have uppercase letters in the filenames. As mentioned before, let's try to keep things lowercase to avoid case conflicts across OSes.

Also, as I noted before, the auto-formatter makes it harder to see what substantial changes you made to the scss files because most of the diff is just fiddling with whitespace.

I notice the new community SCSS is also defining its own breakpoints instead of using the Bootstrap ones. That's another thing that'll make maintenance harder in the long run, I think. (Occasionally it's kind of necessary to define custom breakpoints, but these look like just the same ones defined additional times.)

template/page-community.html.jinja
```diff
@@ -407,3 +249,123 @@
...
     </script>
 {% endblock analytics %}
+
+{% block endbody %}
+<script type="application/javascript" src="{{currentpage.prefix}}assets/js/contribute-carousel.js"></script>
+<script type="text/javascript">
+document.addEventListener("DOMContentLoaded", function() {
+const events = [
+  { "name": "New Horizon: Innovate Without Limits: New Horizons Await",      
+    "description": "Join our EVM-compatible chain launch for a chance to win $50,000 in prizes! Unleash your creativity in DeFi and NFTs, with judging criteria focused on novelty, impact, and community engagement.",
```
tequdev: name and description should be translatable

assets/img/community/card-bg-1.png
AKCodez: I replaced them with svg as per your request, but the reason I use png for all images, is I export them as 2x pngs from the figma and the resolution comes out much clean and crisper,for example even in this case the background images now look much grainier oppose to that of the png bg images, I know it takes up a bit more space but I feel as if it is worth it as a cleaner ui will be presented to the user, especially when we are using SSR the extra kb should be irrelevant.

mDuo13: Yeah, I don't know what's going on with these SVGs; like this one is _80kb_ which is 10x larger than the PNG...

AKCodez: Want me to revert it back to the png? @mDuo13 ?

ckniffen: This happened on the grants site originally.  The svg is including the entire shape but then just cropping the viewport.  One way to fix this is to convert it to another file type and then trace it and make it an svg again.

template/page-community.html.jinja
```diff
@@ -407,3 +249,123 @@
     })
     </script>
 {% endblock analytics %}
+
+{% block endbody %}
+<script type="application/javascript" src="{{currentpage.prefix}}assets/js/contribute-carousel.js"></script>
+<script type="text/javascript">
+document.addEventListener("DOMContentLoaded", function() {
+const events = [
+  { "name": _("New Horizon: Innovate Without Limits: New Horizons Await"),
```
mDuo13: The `_(...)` function for translation doesn't work in this context. You can use it in `{% set ... %}` blocks which are handled by the Jinja preprocessor (which performs translation at build time) but you can't just use it in client-side JavaScript. The page is throwing an error in the browser console about `_` being undefined because of this.

One workaround is to wrap the entire events array here in `{{ ... }}` which evaluates it as Jinja syntax first, which will do translation and output a dictionary which happens to be compatible with JS syntax… as long as you consistently quote the keys (i.e. the unquoted key literals on lines 291–324 throw an error).

Another problem confounding this is that you have the event details defined in two places and this one is (currently) not being used, the one in `contribute-carousel.js` is. It's possible to work around that and make it use the translatable one from this file. But you have to do all of the following:

- remove the `events` definition from `contribute-carousel.js`
- use `var` instead of `const`
- move the script tag for `contribute-carousel.js` _below_ the one that defines `events`
- plus fix the things with translatability I mentioned in the previous paragraph.

Since that's complicated, I've pushed a commit ( 1e6c237e409b7755670e9f55034b42f3dad6e0c3 ) that does all of that. But you probably want to look at the updated events list and confirm that it works.

template/page-community.html.jinja
```diff
@@ -17,386 +17,228 @@
...
-        <div class="mb-4 pb-3" id="careersDark-small"></div>
+<section class="bottom-cards-section">
+  <div class="com-card">
+    <img class="top-left-img" />
+    <div class="card-content">
+      <h6 class="card-title">{% trans %}Contribute to Consensus{% endtrans %} </h6>
+      <h6 class="card-subtitle">{% trans %}Run an XRP Ledger network node{% endtrans %} </h6>
+      <p class="card-description">{% trans %}Thank you for your interest in contributing to XRPL.org. {% endtrans %}</p>
+      <div class="card-links">
+        <a class="com-card-link" target="_blank" href="the-rippled-server.html">{% trans %}The Rippled Server{% endtrans %}</a>
```
mDuo13: Two things:

1. This link text fails WCAG contrast guidelines. With foreground #7919FF, background #232325, it has contrast ratio 2.53:1. The guideline AA for "normal text" (this isn't large text) is 4.5:1.
2. The `rippled` server name is written in all lowercase, preferably code font.
```suggestion
        <a class="com-card-link" target="_blank" href="the-rippled-server.html">{% trans %}The <code>rippled</code> Server{% endtrans %}</a>
```

mDuo13 (changes requested):
Still a couple fixes that need to go in before we should merge this.

mDuo13 (approved):

PR #2264, opened 2023-11-14
Add core dump config; update key adding; add Bookworm to supported OSs

legleux:
Add instructions to enable core dumps on Ubuntu.
Update adding Ripple's gpg key.
Add new Debian to supported Linux distros.

### DIFF
--- a/content/infrastructure/installation/install-rippled-on-ubuntu.md
+++ b/content/infrastructure/installation/install-rippled-on-ubuntu.md
@@ -29,14 +29,14 @@ Before you install `rippled`, you must meet the [System Requirements](system-req
 
 3. Add Ripple's package-signing GPG key to your list of trusted keys:
 
-        sudo mkdir /usr/local/share/keyrings/
-        wget -q -O - "https://repos.ripple.com/repos/api/gpg/key/public" | gpg --dearmor > ripple-key.gpg
-        sudo mv ripple-key.gpg /usr/local/share/keyrings/
+        sudo install -m 0755 -d /etc/apt/keyrings && \
+            wget -qO- https://repos.ripple.com/repos/api/gpg/key/public | \
+            sudo gpg --dearmor -o /etc/apt/keyrings/ripple.gpg
 
 
 4. Check the fingerprint of the newly-added key:
 
-        gpg --import --import-options show-only /usr/local/share/keyrings/ripple-key.gpg
+        gpg --show-keys /etc/apt/keyrings/ripple.gpg
 
     The output should include an entry for Ripple such as the following:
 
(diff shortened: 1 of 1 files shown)


content/infrastructure/installation/install-rippled-on-ubuntu.md
```diff
@@ -91,6 +89,28 @@ Before you install `rippled`, you must meet the [System Requirements](system-req
         sudo setcap 'cap_net_bind_service=+ep' /opt/ripple/bin/rippled
 
 
+9. Optional: configure core dumps
+
+    By default Ubuntu is not configured to produce core files useful for debugging crashes.
+    First run:
+    
+            ulimit -c unlimited
```
mDuo13: For consistency and to avoid extraneous spaces in the code block when parsed, this should be indented the same as the other example (4 spaces more than the surrounding paragraph)

```suggestion
        ulimit -c unlimited
```

content/infrastructure/installation/install-rippled-on-ubuntu.md
```diff
@@ -91,6 +89,28 @@ Before you install `rippled`, you must meet the [System Requirements](system-req
...
+    
+    Now run `sudo systemctl edit rippled`. The default editor should open and add
+
+        [Service]
+        LimitCORE=infinity
+
+    This creates `/etc/systemd/system/rippled.service.d/override.conf` saving core dumps while not interfering 
+    with the service file provided by the `rippled` package.
+    Core dumps appearing in `/var/lib/apport/coredump/` can be loaded into `gdb` with
```
mDuo13: We don't hard-wrap paragraph text in this repo. Also, I think this could be a little easier to follow if reworded:

```suggestion
    This creates the file `/etc/systemd/system/rippled.service.d/override.conf` and configures the OS to save core dumps, without changing the service file provided by the `rippled` package. If your server crashes, you can find the core dump in `/var/lib/apport/coredump/`. To load the core dump for inspection, use a command such as the following:
```

content/infrastructure/installation/install-rippled-on-ubuntu.md
```diff
@@ -91,6 +89,28 @@ Before you install `rippled`, you must meet the [System Requirements](system-req
...
+
+        gdb <path_to_rippled> <path_to_core_file>
```
mDuo13: Rather than using `<path_to...>` which may be interpreted as shell syntax, I prefer to use examples that use reasonable values. On Ubuntu, we know the path to the rippled executable, so that's a good start; if you had an example of a full core dump filename that would be even better, but I think this would be an improvement:

```suggestion
        gdb /opt/ripple/bin/rippled /var/lib/apport/coredump/core
```

content/infrastructure/installation/install-rippled-on-ubuntu.md
```diff
@@ -91,6 +89,28 @@ Before you install `rippled`, you must meet the [System Requirements](system-req
...
+
+    provided the `rippled-dbgsym` package has been installed. 
+    You will need permission to read files in the core dump directory.
```
mDuo13: In addition to not hardwrapping lines, let's also not continue sentences from before the code block. Instead:

```suggestion
    **Note:** To debug a core file this way, you must have the `rippled-dbgsym` package installed, and you need permission to read files in the core dump directory.
```

mDuo13 (approved):
Couple small style suggestions, but mostly looks good.

Issue #2266, opened 2023-11-15
Add client library guidelines


mDuo13:
There are many libraries for use with the XRPL at various levels of maturity and completion. We should define a set of guidelines for getting a library listed on xrpl.org so that it's clear why some are accepted and aren't.

Related: #2100 and #2234.

---

Some incomplete thoughts on the matter:

- Scope: It should probably implement serialization, signing, at least basic transaction submission, and some kind of API request interface. It should be possible to submit any transaction that's valid on Mainnet and call any public API method.
    - Native interfaces & models are a nice-to-have, as are conversion functions, i.e. XRP-drops, timestamps, flags, and hex to ASCII, and utilities for stuff like reading metadata, create crypto-conditions, tx submission, etc.
    - There are a lot of gotchas in managing XRPL keys / signing to look out for.
    - We can point to xrpl.js as the gold standard in functionality
- Responsibility: It should not have obvious/major security holes, it should receive regular maintenance (updates at least every, like, 6 months or more?).
    - Naming and stuff should not be misleading, unprofessional, etc
- Preferably the repo should have a widely known open source license like ISC, MIT, GPL, or Apache, and a code of conduct.

PR #2267, opened 2023-11-15
Remove `currentpage.md` variables

mDuo13:
As part of the Redocly migration, I'm removing syntax that isn't compatible with the new portal. In many of these cases, the existing if/else statements weren't working as intended anyway because they hadn't been updated after the files had been renamed.

In the case of the Client Libraries page, I've removed mention of editing the page to contribute client libraries for now; when someone adds guidelines (issue #2266) it would make sense to add a link to those guidelines in that place instead.

P.S. the plan is to merge this to master, then have it be automatically ported over to the redocly-migration branch when we rebase.


amarantha-k (approved):
LGTM

PR #2269, opened 2023-11-16
Brand kit updates

AKCodez:


amarantha-k (commented):
The local build looked good. 
Wondering if the zip files should be under `assets` instead of the `content` folder? @mDuo13 what do you think?

AKCodez:
> The local build looked good. Wondering if the zip files should be under `assets` instead of the `content` folder? @mDuo13 what do you think?

I put it in content as that is where it was being hosted before.

content/XRPL_Brand_Kit.zip
mDuo13, line 1: I don't love that this zip has a `__MACOSX/` folder at the top and `.DS_Store` files in every dir, but it's fine I guess.

mDuo13 (approved):
It's OK for the brand kit to be in `content/` because we're linking people to download it from GitHub via the `github.com/.../raw/...` link. However, if we moved it under `assets/`, then we could serve it directly from xrpl.org at a URL like `https://xrpl.org/assets/XRPL_Brand_Kit.zip`, which _might_ make it easier to track how many times it's been downloaded (using the server access log or something).

content/XRPL_Brand_Kit.zip
amarantha-k: While not a blocker, I'd prefer that we remove these files as they sometimes can cause problems.

amarantha-k (approved):

dactyl-config.yml
```diff
@@ -4945,6 +4945,21 @@ pages:
...
+        blurb: Review guidelines for using XRPL design assets.
+        top_nav_grouping: Development
+        targets:
+            - en
+
+    -   name: XRPLブランドキット
+        html: https://github.com/XRPLF/xrpl-dev-portal/raw/master/content/XRPL_Brand_Kit.zip
+        parent: resources.html
+        blurb: XRPLのデザインアセットのガイドラインを確認しましょう。
+        top_nav_grouping: Development
```
tequdev:
```suggestion
        top_nav_grouping: 開発
```

In `targets: ja`, "開発" is used for “Development"

https://github.com/XRPLF/xrpl-dev-portal/blob/db87dc69635b18c0d16bb8eab680a3b86258599b/dactyl-config.yml#L4788

AKCodez:
> It's OK for the brand kit to be in `content/` because we're linking people to download it from GitHub via the `github.com/.../raw/...` link. However, if we moved it under `assets/`, then we could serve it directly from xrpl.org at a URL like `https://xrpl.org/assets/XRPL_Brand_Kit.zip`, which _might_ make it easier to track how many times it's been downloaded (using the server access log or something).

Moved into assets

PR #2271, opened 2023-11-19
TOML Add x in the principal's details

ihomp:
X handle of the principal, where one can express opinion about voting for amendments and post validator restarts.

### DIFF
--- a/content/references/xrp-ledger-toml.md
+++ b/content/references/xrp-ledger-toml.md
@@ -150,6 +150,7 @@ desc = "Test Net public server cluster"
 [[PRINCIPALS]]
 name = "Rome Reginelli" # Primary spec author
 email = "rome@example.com" # Not my real email address
+x = "Rome" # Not Rome's real X username/handle
 
 [[CURRENCIES]]
 code = "LOL"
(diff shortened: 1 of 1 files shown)


Silkjaer:
I think it's best to keep it vendor neutral, but maybe instead of naming X add a line for alternative way of connecting (which could then be X/Telegram/LinkedIn or whatever preferred)?

jscottbranson:
Currently, the TOML spec clarifies that the file can contain any properly formatted custom field.

I wonder about people having the option of using a social media domain (e.g., LinkedIn.com = "username") as a key and their username as a value, to avoid any vendor lock in.

Alternately, it could be a list of social media URLs, maybe like:

social_1 = "https://website.tld/username"

social_2 = "..."

The benefit of option 2 is that it can provide an ordered list.

I think it's nice to have extra contact info, particularly for getting updates about the server, however, I also feel it's important to avoid pressuring anyone into using a particular platform.

ihomp:
I don't see a pressure here to use a particular platform. 
It's just an example, of how to format data, the example can be extended. 

```
x = "username"
linkedin = "username"
telegram = "username"
wechat = "username" 
kakaotalk = "username"
...platform = "username" 
```

right now some have
`twitter = "@username"`
or 
`twitter = "https://twitter.com/username"`

it would be good to have an example for a format to follow.

jscottbranson:
I definitely see the value in having a consistent spec. Would you be ok editing the original PR to reflect this being across platforms?

ihomp:
@jscottbranson how do I reflect that? - 1) by adding more examples  in the list. or 2) by adding a note that it's not limited to "X" only. Or 3) both

jscottbranson:
Personally, I would find option 3 helpful :+1: .

In addition to another example or two, I think I would benefit from having clarity about how to extend the social media list. For example, highlighting that just the platform name should be used, instead of the domain:

`The X username/handle where this principal can be contacted. For consistency, use site names, not domains, when specifying social media platforms.`

amarantha-k:
In preparation for the toolchain migration to Redocly, the proposed changes have been ported to the `redocly-migration` in #2377.

PR #2276, opened 2023-11-22
migrate Validator Domain Verifier to Redocly

jonathanlei:
migrate https://xrpl.org/validator-domain-verifier.html to Redocly


mDuo13 (changes requested):
Tried a test merge based on the latest redocly-migration branch (running realm 0.71.1) and it largely works. I am seeing some errors in the browser console, though:

1. If the pasted manifest isn't valid hex (i.e. trying with a base64 string returned by rippled's manifest method) it fails with a TypeError instead of gracefully.
```
Uncaught (in promise) TypeError: decodedManifest is undefined
    parseAndVerifyManifest
    ...
```

2. Each time you look up a manifest after the first, it throws a non-unique key error.
```
Warning: Encountered two children with the same key, `manifest-key`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

Aside from that, this looks about ready to go.

mDuo13:
Merged as 49dc66f224538790cb55304b33197bcdf56f0b47

PR #2281, opened 2023-11-25
known-amendments: fix drop size scientific notation

rickyno:
Correction to known-amendments page where scientific notation of a drop size was displayed as `1e-81` when it should be `1e+8` in the `fixReducedOffersV1` description.

### DIFF
--- a/content/resources/known-amendments.md
+++ b/content/resources/known-amendments.md
@@ -780,7 +780,7 @@ In general, offers can be _reduced_ in 3 ways:
 - The offer can be partially consumed after being placed into an order book.
 - The offer can be under-funded, meaning its owner has less funds than it specifies.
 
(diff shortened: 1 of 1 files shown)


mDuo13:
The original text is correct as-is. It's talking about the minimum amount of a token amount, which is 1 × 10¯⁸¹, or 1e-81. Yes, that's 0.0000000000000000000000000000000000000000000000000000000000000000000000000000000001 if you write it out in non-scientific notation. The amount that can be rounded is very small. (See also: [this discussion of token math](https://github.com/XRPLF/rippled/issues/4120).)

"1e+8" would mean 100000000.0, which is definitely wrong.

rickyno:
Oh, sorry. Thanks for the reference.

PR #2294, opened 2023-12-07
Update setregularkey.md

FlorianBouron:
Replacing `master key` by `private key`.
I think we should be clear on private key or seed here.
`master key` doesn't seem to be a standard word here. A master key could be as well a 'password' you use to encrypt data. Here the context is about a private key.
Also to keep consistent with this page within the doc: https://xrpl.org/assign-a-regular-key-pair.html?utm_source=gemwallet.app I would rather use the term `private key`.
Also following this schema: https://xrpl.org/img/cryptographic-keys.svg
If we want to mention the seed, I would rather use seed or rather passphrase.

Let me know if this makes sense :) Happy to comment more about it.

### DIFF
--- a/content/references/protocol/transactions/types/setregularkey.md
+++ b/content/references/protocol/transactions/types/setregularkey.md
@@ -11,7 +11,7 @@ labels:
 
 A `SetRegularKey` transaction assigns, changes, or removes the regular key pair associated with an account.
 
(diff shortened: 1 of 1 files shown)


ckniffen:
I think calling it a private keypair is even more confusing.  That is because it is still not fully private.  Only the private key portion is.

intelliot:
Master key is a standard term used in the context of the XRP Ledger. Please see https://xrpl.org/cryptographic-keys.html#master-key-pair

PR #2295, opened 2023-12-07
fix: font-awesome font loading

ckniffen:
Had to move the font file location so that they are found up after Redocly does its magic.


mDuo13 (approved):
Confirmed that `fa-*` classes work in pages after this change. :tada:

mDuo13 (approved):

PR #2296, opened 2023-12-07
Reorg tokens & stablecoin use case

mDuo13:
Rebased and cleaned-up version of #2132.

Key differences:

- Rebased onto the latest `master` branch which includes some more translations. Squashed a lot of the commits to reduce the back-and-forth changes to some filenames and links.
- In most cases, the HTML filenames have been left as-is from the pre-reorg work. Normally, our convention would be to update these to match the new page names, but with the Redocly migration on the horizon, it didn't make sense to introduce a bunch of additional URLs that'll need to be migrated again very soon.
- Removed the `nfts_by_issuer` API method page, which was incomplete and should be handled in a different PR.
- Changed a couple MD filenames so that their eventual Redocly URLs will be less redundant. For example `tokens/nfts/collections` instead of `tokens/nfts/nft-collections`.
- Removed the "technical details" concept under Stablecoins, which had a bunch of duplicate content, and rolled the remaining recommendations into other pages (primarily "Stablecoin Precautions")

Note to @tequdev: several of the "new" pages in this PR are just split off sections from the previous "Stablecoin Issuer" use case page with minimal changes, so you can probably reuse a lot of the previous translations.


content/concepts/tokens/fungible-tokens/stablecoins/precautions.md
```diff
@@ -0,0 +1,54 @@
...
+## Payments into the XRP Ledger
+
+When sending payments into the XRP Ledger, follow best practices to avoid overpaying on fees and roundabout paths. For details, see [Sending Payments to Customers](sending-payments-to-customers.html).
+
+If you accept payments from in external systems, such as bank deposits or credit/debit payments, be sure to protect yourself against reversible deposits. XRP Ledger payments are irreversible, but many other digital payments are not. Scammers can abuse this to take their fiat money back by canceling a non-XRPL transaction after you've already sent an XRP Ledger payment.
+
+Additionally, follow the [Reliable Transaction Submission guidelines](reliable-transaction-submission.html) to ensure that you know the final outcome of your XRP Ledger transactions even in rare circumstances like power failures or network outages.
+
+
+## Other precautions
```
amarantha-k:
```suggestion
## Other Precautions
```

amarantha-k (approved):
Other than one minor suggestion, this looks good.

PR #2297, opened 2023-12-07
feat: add `react-codemirror` to redocly setup

ckniffen:
Redocly uses a super old fork called `react-codemirror2-react-17` which likely was created at a time when react 17 wasnt supported. That was 3 years ago and the original component supports it now.


pdp2121 (approved):
Nice fix!

ckniffen:
There is a lot more work needed to get this to work properly.

PR #2301, opened 2023-12-11
feat: Use `@uiw/react-codemirror` and `react@18`

ckniffen:
Updating Redocly to latest and leveraging `@uiw/react-codemirror` which it also uses.

Removes jslint and jshint vendor js files as they are no longer needed.

This will allow simple usage of an editor through an abstracted interface:
  - `<JsonEditor value={} />`
  - `<JavascriptEditor value={} />`

code mirror


mDuo13:
Merged as 375e1ace75701c1a458d54d4d4036ae8b1a63768

PR #2303, opened 2023-12-13
adds new events to community and events page

AKCodez:


tool/filter_categorize_dates.py
```diff
@@ -433,17 +433,64 @@
     "image": "paris.png",
     "end_date": "November 28, 2023"
     },
+    # { "name": "XRP Ledger Zone ETHDenver",
```
mDuo13, line 436: I'd prefer not to clutter up files with commented-out draft data.

mDuo13 (approved):
Looks fine, although we might want to revisit the events carousel on the Community page. It's painfully slow to rotate new things in, IMO, and the images for the non-focused events are not useful, so it's basically pointless except for whatever event is the default highlighted one. (I'm not sure why it's still showing the XRPL Community Report Launch Party, which is in the past, as the default one.)

AKCodez:
> Looks fine, although we might want to revisit the events carousel on the Community page. It's painfully slow to rotate new things in, IMO, and the images for the non-focused events are not useful, so it's basically pointless except for whatever event is the default highlighted one. (I'm not sure why it's still showing the XRPL Community Report Launch Party, which is in the past, as the default one.)

I will revisit this logic and refine it once we migrate it to redocly.

PR #2304, opened 2023-12-15
Update AMM discounted trading fee

### LINKED ISSUE #2283: Update AMM's DiscountedFee dynamically
In the latest code, AMM's DiscountedFee is not fixed at 0, but rather varies dynamically depending on the pool's fees.
It would be good to have an explanation of how it is calculated

https://xrpl.org/amm.html#auction-slot-object

https://github.com/xrplf/rippled/blob/4977a5d43ce5e4d159aada628f021b0332708780/src/ripple/app/misc/impl/AMMUtils.cpp#L314-L317

mDuo13:
The discounted trading fee always being 0 was a leftover from the in-development version of the amendment. The version as released in 1.12.0 has a non-zero trading fee based on the AMM's current trading fee.


gregtatcam (approved):
LGTM

tequdev:
this PR will fix https://github.com/XRPLF/xrpl-dev-portal/issues/2283

amarantha-k (approved):

PR #2317, opened 2024-01-03
Bump Redocly/realm to version 0.70.0

mDuo13:
This version includes several fixes we wanted, including for the issues with navigating in dev mode using Firefox. I also included a fix/workaround to prevent it from crashing on certain landing pages (concepts/tutorials/etc) although it doesn't fix the underlying issue with those pages' list of child pages being empty.


JST5000 (approved):
LGTM

PR #2319, opened 2024-01-04
Update server_state docs for Clio

intelliot:
Clio is the primary API server for the XRPL, taking that role away from `rippled` (which will continue to perform transaction processing, consensus, etc.)

Without `"ledger_index": "current"`, API consumers get this error:

```
  "error": "unknownCmd",
  "error_code": 32,
  "error_message": "Unknown method.",
```


DennisDawson (approved):
LGTM

PR #2334, opened 2024-01-11
up the example default delete to 30,000

shortthefomo:
up the example default delete to 30,000 as lower values have shown issues in production


intelliot (approved):
Note, with `advisory_delete=1`, online deletion runs only when prompted with [can_delete](https://xrpl.org/can_delete.html).

shortthefomo:
> Note, with `advisory_delete=1`, online deletion runs only when prompted with [can_delete](https://xrpl.org/can_delete.html).

Correct however most validators will not be keeping history and use the automated process to purge records. The delete seems to be using more resources of the disk than before, pushing this value out simply makes the delete not run very often as the default suggested here it runs every few min.

mDuo13 (approved):

PR #2344, opened 2024-01-23
Update assign-an-authorized-minter-using-javascript.md

MorganBergen:
It would be --nice-- for people with various --types-- of web browsers to give an actual the actual preview of the webpage within the docs instead of a jpg of it, the injection change is merely a design suggestion

which web design is not my fortè

so take as you may


mDuo13:
This is something we on the RippleX Docs team have been working on more comprehensively, and we've been formulating plans to make something like a side-by-side view for tutorials and the interface. However, embedding the entire HTML contents verbatim and in duplicate like this is not quite the solution we're looking for at the moment. So, thanks for the suggestion, but we'll be going about this a somewhat different way. (After the Redocly migration wraps up in the next couple weeks, tutorials will be our next big project.)

PR #2345, opened 2024-01-23
Redocly migration: Remove deprecated stuff

mDuo13:
- Remove old `validation_seed` method page (the method was removed way back in v0.29)
- Remove badge definitions from common links (they're not used anymore anyway with the new `{% badge %}` syntax). This should slightly help startup time for the Redocly server
- Remove badges that are older than ~2 years


amarantha-k (approved):
LGTM

PR #2346, opened 2024-01-23
Migrate dev-blog to Redocly

amarantha-k:
Addresses https://github.com/XRPLF/xrpl-dev-portal/issues/2195

PR #2347, opened 2024-01-24
Add footer and frontmatter for custom pages

### LINKED ISSUE #2330: Custom Redocly pages need frontmatter
Custom pages (anything `*.page.tsx`) need to specify at a minimum the page title in the frontmatter. Otherwise, the browser tab shows a generic "Redocly Portal" title instead of the page name, as in the following screenshot:

Per [Redocly's docs](https://beta-docs.redocly.com/extend/how-to/links-in-frontmatter/#example-of-front-matter-links-in-.page.tsx-page) the way to do this is with `export const frontmatter` as in the following sample code:

```
export const frontmatter = {
  seo: {
    title: 'Tsx page title',
  }
};
```

This can be anywhere in the `.page.tsx` file but by convention it should be near the top, just after any imports.

mDuo13:
- Use the default Redocly footer (we can probably re-style this to look a lot like the current XRPL.org footer)
- Add frontmatter to all the custom pages that are in the footer, which more or less fixes #2330 (the individual dev tools might still need it)

### DIFF
--- a/content/about/history.page.tsx
+++ b/content/about/history.page.tsx
@@ -1,6 +1,13 @@
 import * as React from "react";
 import { useTranslate } from "@portal/hooks";
 
+export const frontmatter = {
+  seo: {
+    title: 'History',
+    description: "The history of the XRP Ledger (XRPL), launched in 2012 as a faster, scalable, more sustainable blockchain.",
+  }
+};
+
 export default function History() {
   const { translate } = useTranslate();
   const [openSections, setOpenSections] = React.useState({});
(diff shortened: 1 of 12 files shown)


amarantha-k (changes requested):
The footer is missing a to Blog under Community. Otherwise looks good.

mDuo13:
I think the blog footer item (line 92 of `redocly.yaml`) should automatically show up when merged into the same branch as the blog updates (#2346).

amarantha-k (approved):
Approving after discussion - the blog link should show up once the files are all merged.

PR #2349, opened 2024-01-24
Redocly migration: add tutorials landing

mDuo13:
Adds a Tutorials landing page similar to the one on the production site now, including card-grid and card components that we can use in other Markdoc files too.

The styling for the cards isn't right yet due to CSS problems, but I hope to fix that separately.


amarantha-k (approved):
Ignoring the CSS for now, updates to add a landing page for tutorials looks good.

PR #2352, opened 2024-01-27
3.0 migration v2

JST5000:
Updates xrpl.org to use xrpl.js 3.0

Should NOT be merged until xrpl.js 3.0.0 has been released as the updated package.json files need that release to function.

Uses the migration guide here: https://github.com/XRPLF/xrpl.js/blob/3.0/MIGRATION.md


JST5000:
Closing in favor of https://github.com/XRPLF/xrpl-dev-portal/pull/2378

PR #2354, opened 2024-01-27
Redocly migration: Fix top nav

mDuo13:
- Modify Navbar component to use the fields that it has access to (arbitrary keys in `top-nav.yaml` are discarded)
- Modify the `top-nav.yaml` to put the usable info in the correct keys, and to update the contents of the Docs dropdown based on the current Docs landing page
    - (The References page is a placeholder, assuming that'll be added in a separate PR)
- Adjust frontmatter of a couple pages to make them display better in the top nav
- Adjust CSS styles to reflect the updated slugs and docs structure
- Fix styling of menus in mobile design


amarantha-k (approved):
lgtm

PR #2359, opened 2024-01-30
Fix broken links in partials and tutorials index

mDuo13:
- Partials using common links need to be included as raw-partials
- xrpl-card href values need to use the production path, not the MD file path.


amarantha-k (approved):
Update for raw-partials looks good.
LGTM

PR #2362, opened 2024-01-31
Fix docs page links

mDuo13:
Some of these may have to be updated again later as pages get moved to better URLs, and unfortunately the link checker doesn't seem to catch stuff in custom React pages/components, but for now they all work. I took a few liberties changing the use cases links to some of the use case documentation that we added since this page was last designed, and changing the "Recommended Pages" to not be as redundant with the dev tools section that's near it already.


amarantha-k (approved):
LGTM

PR #2366, opened 2024-01-31
Add redirect for references.html now that the landing page is available

amarantha-k:


amarantha-k:
This is no longer needed as the redirects were updated after rebasing the redocly-migration.

PR #2378, opened 2024-02-05
Update xrpl.js to 3.0 across xrpl.org

JST5000:
Updates all code snippets to use xrpl.js 3.0. 

Some code snippets were also depending on dependencies which weren't in their package.json, so I updated those to include all necessary dependencies. 

The biggest refactor had to happen in the key-pair derivation as it was using `BN.js` instead of `bignumber.js` and really went into the weeds mirroring the implementation details in `ripple-keypairs` (so had to be refactored to match the new impelmentation details). 

I also cleaned up a vite.config to remove any unnecessary polyfills since xrpl.js 3.0 removes a lot of need for them.

Supersedes: https://github.com/XRPLF/xrpl-dev-portal/pull/2352 (that PR was tested less in-depth, and was pointed at main when it should've been pointed to the redocly-migration branch).


jonathanlei (approved):

_code-samples/key-derivation/js/index.js
```diff
@@ -112,27 +145,25 @@ class Seed {
...
-            .mod(order);
+        const scalar = this._deriveScalar(publicGen.encodeCompressed(), accountIndex)
+        return BigNumber(scalar).plus(privateGen).mod(order);
     }
 
     _deriveScalar(bytes, discrim) {
-        const order = Secp256k1.curve.n;
-        for (let i = 0; i <= 0xffffffff; i++) {
-            // We hash the bytes to find a 256 bit number, looping until we are sure it
+        const order = Secp256k1.curve.n
```
ckeshava, line 153: minor nit: if possible, I'd prefer to see the semi-colons at the end of the lines in this file

_code-samples/key-derivation/js/index.js
```diff
@@ -112,27 +145,25 @@ class Seed {
...
+        for (let i = 0; i <= 0xffff_ffff; i++) {
```
ckeshava, line 154: nice use of underscore for readability!

_code-samples/key-derivation/js/index.js
```diff
@@ -3,16 +3,49 @@
...
 const Secp256k1 = elliptic.ec('secp256k1');
 const hashjs = require("hash.js");
-const Sha512 = require("ripple-keypairs/dist/Sha512")
+const isomorphic = require("@xrplf/isomorphic/sha512")
+const noble = require('@noble/curves/abstract/utils')
 const { codec, encodeAccountPublic, encodeNodePublic } = require("ripple-address-codec");
 
 const XRPL_SEED_PREFIX = 0x21
 
+class Sha512 {
```
ckeshava, line 17: Would it be useful to update the `Sha512` class definition inside the JS client library? `packages/ripple-keypairs/src/Sha512.ts`

Would other projects/developers need similar changes?

ckeshava (approved):

_code-samples/key-derivation/js/index.js
```diff
@@ -141,10 +172,10 @@ class Seed {
         // How often will an (essentially) random number generated by Sha512 be larger than that?
         // There's 2^32 chances (the for loop) to get a number smaller than the order,
         // and it's rare that you'll even get past the first loop iteration.
-        // Note that in TypeScript we actually need the throw, otherwise the function signature would be BN | undefined
+        // Note that in TypeScript we actually need the throw, otherwise the function signature would be bigint | undefined
```
amarantha-k, line 175: Wouldn't this be BigNumber?
```suggestion
        // Note that in TypeScript we actually need the throw, otherwise the function signature would be BigNumber | undefined
```

amarantha-k:
Also,  the package [npm/secp256k1@5.0.0](https://socket.dev/npm/package/secp256k1/overview/5.0.0) is flagged as an alert. Please confirm if this is an acceptable risk, in which case, we can ignore the alert and merge when the PR is approved.

_code-samples/key-derivation/js/index.js
JST5000: I don't think other developers would need those changes - I think that class is mostly an implementation detail of the library which supports the functions we do want to expose :)

JST5000: (This tutorial mostly mirrors the implementation details of xrpl.js to show how it works / why)

JST5000:
@amarntha-k The security notification seems fine to ignore. 

The security risk is pointing out that there is non-javascript code which appears to be bundling settings for operating specific bundling rules, so is not a security risk. (The worry is that the different style of code may be doing something unrelated to the library or malicious in a way that tools designed to parse javascript projects as an example couldn't detect). https://socket.dev/npm/package/secp256k1/files/5.0.0/binding.gyp

PR #2397, opened 2024-02-16
fix conflicting color

AKCodez:


@theme/styles.css
```diff
@@ -159,7 +159,7 @@ ul.nav.navbar-nav {
   --color-geekblue-10: #030852; */
 
   /* Dark mode colors by default */
-  --color-primary-bg: var(--color-gray-9);
+  --color-primary-bg: var(--color-gray-10);
```
AKCodez, line 162: for transparency --color-gray-10 is a reference to color-gray-9 so there will be no change in the color for those vars

amarantha-k (approved):
lgtm

PR #2399, opened 2024-02-17
Fix wallet address output in send-xrp.js example

movitto:
The send-xrp.js example starts with code that imports a secret and outputs the corresponding wallet address to the console. The problem is the expected output (as indicated by the comment) is incorrect, this patch fixes this


oeggert:
I ran the code and the original address in the code comments is correct.

PR #2407, opened 2024-02-23
feat: Osano & XRPL AI for Redocly

justinr1234:


amarantha-k (commented):
A few UI elements need to be updated (please see screenshot): 
- The "X" icon to close the preferences window 
- The slider on the toggle button is off. 
Screenshot 2024-02-22 at 8 12 38 PM

amarantha-k:
Does the AI chatbot need to be updated to respond to queries from a different site? Asking as the only response I see on the preview build is "Sorry, I am not sure how to answer that."

amarantha-k (approved):
Confirmed with @justinr1234 offline that the preview site isn't on the allowed list and the AI chatbot should work as expected on prod site.

PR #2411, opened 2024-02-24
fix released version to 2.1 for last 2 amendments

oeggert:


amarantha-k:
No longer needed as PR#2336 includes this fix.

PR #2415, opened 2024-02-26
Migrate key-value pairs from message.po file to translations.yaml file

amarantha-k:


amarantha-k:
This is an initial commit to migrate the translated text from messages.po to translations.yaml. 
Working on batch conversion to migrate the rest.

tequdev:
In the future, we need a script that can add and remove (comment out) keys from translations.yaml files in other languages from changes to English content that is targeted for i18next.
It would also be very easy to check pages if it is possible to see which pages use the key by commenting it out.

These were done in dactyl_build.

PR #2422, opened 2024-02-27
fix links on code samples page

AKCodez:


amarantha-k (approved):
Thanks for fixing this! LGTM

PR #2427, opened 2024-02-28
Fixes links on community page and translation for navbar items

amarantha-k:


DennisDawson (approved):
Can't independently verify, but these look right.

tequdev (approved):
lgtm as japanese

PR #2447, opened 2024-03-11
fix: point homepage canonical to correct url

AKCodez:


AKCodez:
@amarantha-k Lets tag a redocly dev on this, I think its best we environmentalize this, so for PR previews it is the dev url and for master builds it the live site url.

amarantha-k:
Followup note - no need to add an environment variable for the site url, so no further changes are needed.

amarantha-k (approved):

PR #2453, opened 2024-03-12
fix: remove experimental flag for several account flags

ckniffen:
Remove the experimental flag for `Clawback` and `DisallowIncoming` amendment additions.


amarantha-k (approved):
Thanks for the fix! 
Also pushed a commit to remove these experimental flags from other concept and reference topics.

PR #2482, opened 2024-03-21
Remove graphic that was incorrectly introduced in a previous PR

amarantha-k:


DennisDawson:
That wasn't an accident. Given that the character (who is _not_ me) features prominently in the introduction and use cases, it makes sense to have a greeting at the top of the introduction section. If there's an objection to the fat man in the fedora, we should come up with another image in the same style to anchor the landing.

At the very least, the text that is now in the graphic needs to be pulled out as text, not throwing out the baby with the bath water.

I also notice that the Docs top nav doesn't include the introduction, and I think it should.

amarantha-k:
Since the illustration is not about the XRP Ledger, it should not be the first thing on the Introduction topic.

DennisDawson (approved):
Approved for now to get the update ready for AMM. Need to move the text from the graphic onto the page. I do want to see some sort of graphic in the style of the rest of the introduction to make the experience more cohesive.

Issue #1851, opened 2023-03-28
When there are no offers, nft_sell_offers returns objectNotFound, even for a valid NFT


intelliot:
The documentation should be updated to make it clear that `nft_sell_offers` returns `objectNotFound` when there are no offers, even for a valid NFT.

## Steps to Reproduce

Example:

https://xrpl.org/websocket-api-tool.html?server=wss%3A%2F%2Fs2.ripple.com%2F&req=%7B%22command%22%3A%22nft_sell_offers%22%2C%22nft_id%22%3A%2200090000D0B007439B080E9B05BF62403911301A7B1F0CFAA048C0A200000007%22%2C%22ledger_index%22%3A%22validated%22%7D

With `nft_info`, we can see that the NFT exists.

## Expected Result

Some users have expected an empty list (or empty array), and no error message.

## Actual Result

An error message - `objectNotFound`.

This means "the object wasn't found", where the "object" is the list of buy or sell offers.

## Path Not Taken

Consider these two cases:

1. User asks for offers for an NFT that exists, but which has no offers.
2. User asks for offers of an NFT that does not exist (either because it was never made or because it was burned).

The ideal user experience is that **1** returns an empty array, and **2** returns `object_not_found`. However, that is not feasible on `rippled` because of the way that all of this is stored. So, there is no way to distinguish between the two cases.

If we changed the API to return an empty list, it would also return an empty list for scenario 2. That wouldn't be much better than the current behavior, has the added pain of requiring all users of this API to change how they handle these cases.

See https://github.com/XRPLF/rippled/issues/4339

ripple369:
Thank you for bringing this to my attention. It is possible that the documentation is unclear about the behavior of the nft_sell_offers command when there are no offers for a valid NFT. I will make sure to pass this feedback along to the appropriate parties to review and update the documentation as needed. In the meantime, it is helpful to know that the objectNotFound error can be returned even for a valid NFT if there are no sell offers available.

DennisDawson:
Fixed in PR https://github.com/XRPLF/xrpl-dev-portal/pull/2495

PR #2495, opened 2024-03-26
Add objectNotFound error.

DennisDawson:
Customer was confused when they received an `objectNotFound` error, thinking it referred to the NFToken object rather than the `Offer` object. Added the error with explanation to both `nft_sell_offers` and `nft_buy_offers`. Fixes issue #1851 .


shawnxie999 (approved):
LGTM

mDuo13 (approved):

PR #2499, opened 2024-03-26
Make top nav translatable, with examples

### LINKED ISSUE #2492: Page labels in sidebar cannot be translated
https://github.com/XRPLF/xrpl-dev-portal/blob/ad498f60e1100e01825ba56055d8fc50a462a5be/sidebars.yaml#L85
スクリーンショット 2024-03-26 17 09 14

mDuo13:
Fixes #2492.

Sidebar navigation was already translatable using `labelTranslationKey` and an entry in the `@i18n/[lang]/translations.yaml` file; this adds an example demonstrating that for Tutorials.

The dropdowns in top nav were not translatable; this adds support for translating them the same way as in the sidebar.

### DIFF
--- a/@i18n/ja/translations.yaml
+++ b/@i18n/ja/translations.yaml
@@ -11,6 +11,12 @@ sidebar.docs.tutorials: チュートリアル
 sidebar.docs.references: リファレンス
 sidebars.resources: リソース
 sidebar.resources.codesamples: コードサンプル
+topnav.docs.title: ドキュメント
+topnav.docs.description: XRPL技術に掘り下げて組み込めましょう。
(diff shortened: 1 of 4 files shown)


tequdev (approved):
Can you add `labelTranslationKey`, `groupTranslationKey` to the translatable part in another PR?
Translators will be able to use them to translate.

mDuo13:
> Can you add `labelTranslationKey`, `groupTranslationKey` to the translatable part in another PR? Translators will be able to use them to translate.

Yes, that's a good idea. Could you create a ticket just so it doesn't get forgotten during the AMM/rippled release bustle?

Issue #2501, opened 2024-03-26
Dev Blog migration issues


mDuo13:
- [ ] The old site had landing pages for each year, at https://xrpl.org/blog/2022/ for example, showing posts only from that year. The new site currently shows a 404 at those URLs.
- [ ] Similarly, label landing pages like https://xrpl.org/blog/label/rippled-release-notes.html are currently 404s instead of redirecting to something useful.
- [x] Blog posts from 2014–2019 don't have the appropriate metadata indicating post categories and date (and turning off the GitHub edit button). The metadata for these is in the `dactyl-config.yml` file of Ripple's old, private `dev-blog` repo, and should be moved to the frontmatter of the new blog. (Fixed in #2512)
- [x] The dev blog area of the site has custom navigation that kind of mirrors the top nav of the blog prior to the Redocly migration, but this is pointless. The blog navigation being different from the rest of xrpl.org wasn't an intentional choice, but the result of the blog repo not being updated from the previous design (it used to match). (Fixed in #2512)

mDuo13:
I think the landing page per year can _probably_ be built reusing the blog homepage template and just filtering to the pages for that year.

PR #2503, opened 2024-03-27
Blog: unify nav with rest of site

mDuo13:
As noted in #2501, the top nav for the blog was not intentionally different from the rest of the site but rather had fallen out of date when the main site got updated and the old blog hadn't. When we migrated to Redocly, we re-created the outdated blog top nav, but this was unnecessary.

The dropdowns include links to numerous moved or deprecated pages and are best served by replacing them entirely with the standard site-wide navigation.

Also, I fixed a bug that was causing xrpl.org/ja/blog/ to throw an error instead of showing the list of blog posts.

### DIFF
--- a/@i18n/ja/translations.yaml
+++ b/@i18n/ja/translations.yaml
@@ -273,6 +273,7 @@ Create Payment Channel: Payment Channelを作成
 Send Issued Currency: トークンを送信
 Trust for: トラストラインの設定
 Security: セキュリティ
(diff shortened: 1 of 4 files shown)


amarantha-k:
Screenshot 2024-03-26 at 6 02 46 PM
The filter category (Security) looks odd. Otherwise, the PR looks good.

amarantha-k (approved):
Thanks for the update! Lgtm

PR #2510, opened 2024-03-27
Fix spacing below Apex banner

mDuo13:
- Increase the offset so the top of main area isn't under the top nav
- Add more invisible spacing so anchors aren't under the banner


styles/xrpl.scss
```diff
@@ -71,7 +71,6 @@ $line-height-base: 1.5;
 @import "_feedback.scss";
 @import "_video.scss";
 @import "_contribute.scss";
-// @import "_top-banner.scss";
```
mDuo13, line 74: Removed this redundant import (#2479 added a new one on line 54 instead of uncommenting this one)

justinr1234 (approved):

PR #2515, opened 2024-03-30
Update ledger.md

donovanhide:
s/20/32/-byte


DennisDawson (approved):
This is the correct value, per the description of Hashes ([xrpl.org/docs/references/protocol/data-types/basic-data-types/#hashes](https://xrpl.org/docs/references/protocol/data-types/basic-data-types/#hashes)).

PR #2519, opened 2024-04-01
Fix {{use_network}} appearing in visible text

### LINKED ISSUE #2490: Trade in the DEX tutorial typos
- Steps are numbered 1,2,3, 7, 4, 5, 6 for some reason (fixed by #2511)
- `{{use_network}}` is showing up rather than "Testnet" (fixed in #2519)

mDuo13:
Fix #2490.

This was a Dactyl variable that couldn't be directly translated to a Redocly variable. The fix is to hard-code it to a network name.


DennisDawson (approved):
LGTM. Link works.

PR #2539, opened 2024-04-10
Increase space above headers to reduce overlap

### LINKED ISSUE #2520: Invisible header spacing blocks clicking links in paragraphs above them
Our headers have [CSS to add invisible spacing above them](https://github.com/XRPLF/xrpl-dev-portal/blob/4200ea677119e6c2c051c91b1c8350f88b6d403c/styles/_top-nav.scss#L574-L588) so that navigating to an anchor like [Restrictions on Assets](https://xrpl.org/docs/concepts/tokens/decentralized-exchange/automated-market-makers#restrictions-on-assets) scrolls to have the text of the heading visible instead of having it underneath the top banner. Without it, you get taken to a point in the page where the header is underneath the fixed top banner.

However, in cases where the header's invisible spacing overlaps elements above it, those elements are unclickable because the header's padding takes the click instead. This can affect hyperlinks, buttons, code blocks, etc.

Illustration: the link in the blue area in this screenshot is not clickable:

Screenshot of "Restrictions on Assets" anchor blocking a link in the paragraph above it

Previously the spacing was not big enough to overlap with the text above it, but after #2510 increased the amount of spacing to compensate for the added Apex banner, the spacing now overlaps with the last line of any preceding paragraph.

We've tried various fixes to make sure the invisible padding doesn't absorb mouse clicks and instead passes them through to the underlying elements, but they don't seem to work. Possibly the best fix is to add more _visible_ spacing before headers so that the only thing the spacing overlaps is blank space instead of important elements.

mDuo13:
Fixes #2520


amarantha-k (commented):
While it addresses the issue, the extra space before the headings and especially page titles looks off.

Screenshot 2024-04-11 at 10 22 12 AM

If the issue noted in #2520 is a one-off or not seen in more than a handful of instances, should we reconsider changing the spacing?

amarantha-k (approved):
The update to ensure that the page heading is not impacted looks good. Thanks for incorporating the feedback!

PR #2553, opened 2024-04-17
Add blog about upcoming Testnet+Devnet resets

mDuo13:
Provide some details and info about the testnet/devnet resets.


justinr1234 (approved):

xzhaous (commented):
should we put an rough start time that is lest impact to everyone? some utc time

PR #2562, opened 2024-04-21
AMMBid explain non-trading fee case

zgrguric:
Adds single sentence which addresses case when user bids on Pool which has not Trading fee yet set. For reference this is the transaction on XRPL Mainnet: A3A278DEEF3C5E06126526F948F31436BDDD96E61779E2E7A55090270017BD89

### DIFF
--- a/docs/references/protocol/transactions/types/ammbid.md
+++ b/docs/references/protocol/transactions/types/ammbid.md
@@ -11,7 +11,7 @@ labels:
 
 _(Requires the [AMM amendment][])_
 
-Bid on an [Automated Market Maker](../../../../concepts/tokens/decentralized-exchange/automated-market-makers.md)'s (AMM's) auction slot. If you win, you can trade against the AMM at a discounted fee until you are outbid or 24 hours have passed. If you are outbid before 24 hours have passed, you are refunded part of the cost of your bid based on how much time remains.
+Bid on an [Automated Market Maker](../../../../concepts/tokens/decentralized-exchange/automated-market-makers.md)'s (AMM's) auction slot. If you win, you can trade against the AMM at a discounted fee until you are outbid or 24 hours have passed. If you are outbid before 24 hours have passed, you are refunded part of the cost of your bid based on how much time remains. A Bid on AMM pool without trading fee set may result in sucessful transaction but will have no effect.
 
 You bid using the AMM's LP Tokens; the amount of a winning bid is returned to the AMM, decreasing the outstanding balance of LP Tokens.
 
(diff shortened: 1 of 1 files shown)


oeggert:
That transaction hash shows the bid going through successfully, and the account sending the bid becoming the new auction slot owner.

Also I don't think it's possible to create an amm pool without specifying the trading fee. @mDuo13 Is it possible for a pool to be created without a trading fee?

mDuo13:
It should not be possible to create an AMM without a trading fee. [[Source]](https://github.com/XRPLF/rippled/blob/2d1854f354ff8bb2b5671fd51252c5acd837c433/src/ripple/protocol/impl/TxFormats.cpp#L85-L92) However, it is possible for the trading fee to be 0.

docs/references/protocol/transactions/types/ammbid.md
```diff
@@ -11,7 +11,7 @@ labels:
 
 _(Requires the [AMM amendment][])_
 
-Bid on an [Automated Market Maker](../../../../concepts/tokens/decentralized-exchange/automated-market-makers.md)'s (AMM's) auction slot. If you win, you can trade against the AMM at a discounted fee until you are outbid or 24 hours have passed. If you are outbid before 24 hours have passed, you are refunded part of the cost of your bid based on how much time remains.
+Bid on an [Automated Market Maker](../../../../concepts/tokens/decentralized-exchange/automated-market-makers.md)'s (AMM's) auction slot. If you win, you can trade against the AMM at a discounted fee until you are outbid or 24 hours have passed. If you are outbid before 24 hours have passed, you are refunded part of the cost of your bid based on how much time remains. A Bid on AMM pool without trading fee set may result in sucessful transaction but will have no effect.
```
mDuo13: To my understanding:

- A pool can have a fee of 0 but it cannot have a fee not set.
- If you win the auction slot, you have access to the discounted fee for 24 hours. This is true regardless of whether the existing fee is 0 or a larger number. Of course, 10% of 0 is still 0, so it's not a useful discount during that time.
- The trading fee can change via AMM voting during the time you have the auction slot. So even if the trading fee is 0 when you first get the auction slot, the trading fee can go up later and you'll still be entitled to trading at 10% of the normal fee.

So it's not precisely right to say the transaction will have no effect.

mDuo13 (changes requested):
I think we should either not merge this, or we should rephrase to make it clearer what it means to bid when the fee is 0.

docs/references/protocol/transactions/types/ammbid.md
zgrguric: I suggest rephrasing to:
"A Bid on AMM pool with zero trading fee may result in successful transaction but will have no effect."

mDuo13 (approved):
Better now that I've rephrased it to:

> If the AMM's trading fee is zero, you can still bid, but the auction slot provides no benefit unless the trading fee changes.

PR #2572, opened 2024-05-03
2571 - Update Rippling definition to clarify issuer's role in the process

maria-robobug:
```
Rippling is essential, because it allows users who hold tokens to send those to each other with the issuer as a passive intermediary.
```

Issue: https://github.com/XRPLF/xrpl-dev-portal/issues/2571


maria-robobug:
FYI, compliance seem happy with this.

DennisDawson (approved):
LGTM

PR #2585, opened 2024-05-14
Update Smart Contracts Use Case for style and consistency

DennisDawson:


docs/img/uc-yescrow-holding-lock-and-key.png
amarantha-k, line 1: The crow doesn't seem like the right illustration, but I don't have a good alternative in mind. Would defer this one to UX/design.

amarantha-k (changes requested):
As noted in the comment, I think we need an alternative to the crow for escrow.

docs/img/uc-yescrow-holding-lock-and-key.png
amarantha-k: Tagging @bugsbunnies for input.

bugsbunnies: Awesome illustrations! For the lock part, I would suggest making all the strokes of the lock green as well so both the lock and key stands out in green as highlighted items.

bugsbunnies: Would be good to also match the bird's eyes with the previous bird's eyes for consistency, they look slightly different.

DennisDawson: Fixed, thanks for the feedback. Still need to rethink the crow, though.

DennisDawson: Okay, a more apt metaphor, I reckon. The Escrow is a locker, the condition is a padlock, the fulfillment is the key.

bugsbunnies (approved):

amarantha-k (approved):
Thanks for incorporating the feedback.

Issue #2533, opened 2024-04-08
`PreviousFields` included in `DeletedNode` metadata


mvadari:
Here is a link to the relevant docs page: https://xrpl.org/docs/references/protocol/transactions/metadata/#deletednode-fields

Here is an example of this occurring in a transaction: https://livenet.xrpl.org/transactions/424661CF1FD3675D11EC910CF161979553B6D135F9BD03E6F8D4611D88D27581/raw

If a ledger object is modified before it's deleted, some fields will still show up in `PreviousFields`.

PR #2587, opened 2024-05-14
Add PreviousFields to DeletedNode

DennisDawson:
Per issue #2533, include `PreviousFields` in the list of fields returned for a `DeletedNode`.


docs/references/protocol/transactions/metadata.md
```diff
@@ -226,6 +226,7 @@ A `DeletedNode` object contains the following fields:
 | `LedgerEntryType` | String            | The [type of ledger entry](../ledger-data/ledger-entry-types/index.md) that was deleted.                                                                                                                                                                                             |
 | `LedgerIndex`     | String - [Hash][] | The [ID of this ledger entry](../ledger-data/common-fields.md) in the ledger's [state tree](../../../concepts/ledgers/index.md). **Note:** This is **not the same** as a [ledger index](../data-types/basic-data-types.md#ledger-index), even though the field name is very similar. |
 | `FinalFields`     | Object            | The content fields of the ledger entry immediately before it was deleted. Which fields are present depends on what type of ledger entry was created.                                                                                                                                 |
+| `PreviousFields` | Object             | Selected fields of the ledger entry before it was deleted. Which fields are present depends on what type of ledger entry was created. |
```
mDuo13:
```suggestion
| `PreviousFields` | Object             | _(May be omitted)_ Selected fields of the ledger entry before it was deleted. Which fields are present depends on what type of ledger entry was created. |
```

I think `PreviousFields` only appears in some special cases, and is omitted most of the time. This description also doesn't make it clear what the difference is with `FinalFields`. (I think the difference is that `PreviousFields` provides the intermediate state if the object was first updated, _then_ deleted vs just deleted? Not sure entirely.)

intelliot (approved):
I think there is room for broader improvements. For example, it's awkward to start a statement with the word `Which`, as it makes me expect a question. But that's beyond the scope of this PR.

PR #2592, opened 2024-05-16
add tokenization page

jonathanlei:
add new tokenization page with figma design 
https://www.figma.com/design/yZtBnHIq0Aas9NBZRBOtVH/Use-Cases-Template?t=TH8ONCaBnfi57liM-0


docs/use-cases/tokenizations.page.tsx
mDuo13, line 1: This should be named `tokenization.page.tsx` (no "s")

sidebars.yaml
```diff
@@ -31,11 +31,14 @@
                 - page: docs/use-cases/tokenization/nftoken-marketplace.md
                 - page: docs/use-cases/tokenization/authorized-minter.md
                 - page: docs/use-cases/tokenization/digital-artist.md
+            - page: docs/use-cases/tokenizations.page.tsx
+              label: Tokenizations
```
mDuo13: Isn't this page is meant to replace the `use-cases/tokenization/index.md` landing page (line 24)?

static/img/backgrounds/tokenization-stats.png
mDuo13, line 1: The text for a graphic like this should be rendered using HTML; accessibility and SEO are much worse if the text is in an image. The background could probably be a CSS gradient although it's also OK if the background gradient is an image file.

package.json
```diff
@@ -15,7 +15,7 @@
     "@codemirror/state": "6.3.3",
     "@codemirror/view": "^6.22.2",
     "@lezer/highlight": "^1.2.0",
-    "@redocly/realm": "^0.78.6",
+    "@redocly/realm": "^0.78.8",
```
mDuo13: You can drop the `package.json` and `package-lock.json` changes from this PR; we've already updated to a newer version than this in the `master` branch.

docs/use-cases/tokenizations.page.tsx
```diff
@@ -0,0 +1,310 @@
...
+    title: "XRPL Overview",
+    description:
+      "An introduction to the key benefits and features of the XRP Ledger, a public blockchain driven by the XRPL developer community.",
+  },
+};
+
+const useCases = [
+  {
+    name: "Stablecoin Issuer",
+    url: "",
```
mDuo13: Looks like actual link paths are missing from these entries? The elements they map to look like they're meant to be links but aren't.

Stablecoin Issuer will be this use case:

```suggestion
    url: "/docs/use-cases/tokenization/stablecoin-issuer/",
```

docs/use-cases/tokenizations.page.tsx
```diff
@@ -0,0 +1,310 @@
...
+              <div key={index} className="tokenization-use-case">
+                {useCase.name}
+                <button className="arrow-button" id="rightArrow">
+                  <img
+                    alt="right arrow"
+                    src={require("../../static/img/icons/arrow-right-purple.svg")}
+                  />
+                </button>
+              </div>
+            ))}
```
mDuo13: I believe this element is supposed to be a nav list similar to ones we have on the XRP Overview and Docs landing pages. Of course, the links (as noted in my comment on the `useCases` array up above) are missing, and we should be sure to fill those in.

It would probably be best to create and use a shared React component for all of these. It's especially a little weird here since it's _almost_ but _not quite_ the same in terms of spacing and typesetting, and it would be good to make them consistent.

Side-by-side comparison:

| Previous Example | This Element |
|---|---|
| (cropped screenshot of a list from the Docs landing) | (cropped screenshot of the use cases list from this page) |


Here's the source for those other examples:

- [XRP exchanges list](https://github.com/XRPLF/xrpl-dev-portal/blob/304ed0046ac13da6bc8a1406cb6cfc97eec575f0/about/xrp.page.tsx#L480-L496)
- [Docs landing use cases list](https://github.com/XRPLF/xrpl-dev-portal/blob/304ed0046ac13da6bc8a1406cb6cfc97eec575f0/docs/index.page.tsx#L167-L176) (also "Browse by Recommended Pages" in the same file)

Also, with the way this is coded right now, it re-uses `id="rightArrow"` for every element, which is against the rules of HTML (`id` attributes are supposed to be unique within a page).

Lastly, this is a nitpick, but I find it surprising that "Learn more" would take me to a YouTube video. Maybe it would provide a better user experience to add a little video camera icon as kind of a warning? We can quickly do this with `<i class="fa fa-video-camera"> </i>` which would look like the following:

(screenshot of "Learn more" link prefaced with a video camera icon)

mDuo13 (changes requested):
This is a good start but I noticed a number of issues with the implementation that should be addressed.

docs/use-cases/tokenization/index.page.tsx
```diff
@@ -0,0 +1,389 @@
...
+    stat: "0.4",
+    description: "Network speed (sec)",
+  },
+  {
+    stat: "$0.001",
+    description: "Avg fee per NFT tx",
+  },
+];
+
+// TODO: Fill in url when links are available.
```
mDuo13: I think this TODO should be removed?

docs/use-cases/tokenization/index.page.tsx
```diff
@@ -0,0 +1,389 @@
...
+          id="security-features"
+        >
+          {securityAdvantages.map((advantage) => (
+            <div className="security-card">
+              <div className="card-body p-6">
+                <h4 className="card-title h6">{translate(advantage.title)}</h4>
+                <p>{advantage.description}</p>
+              </div>
+            </div>
+          ))}
```
mDuo13: I think each element generated by the `map` needs a unique `key` attribute for React to stop throwing this error:

> Warning: Each child in a list should have a unique "key" prop.

docs/use-cases/tokenization/index.page.tsx
```diff
@@ -0,0 +1,389 @@
...
+        <div className="d-flex flex-column-reverse">
+          <h4 className="eyebrow mb-16">{translate("Stats")}</h4>
+        </div>
+        <div className="tokenization-stats">
+          {stats.map((statElement) => (
+            <div className="stat-container align-items-center">
+              <div className="stat">{statElement.stat}</div>
+              <div className="desc">{translate(statElement.description)}</div>
+            </div>
+          ))}
```
mDuo13: I think this map is another case where a `key` attribute is necessary.

docs/use-cases/tokenization/index.page.tsx
```diff
@@ -0,0 +1,389 @@
...
+              </h2>
+              <h6 className="eyebrow mb-3 text-large">
+                {translate("Tokenization")}
+              </h6>
+            </div>
+            <p className="mb-10">
+              {translate(
+                "Explore the possibilities of tokenization on the XRP Ledger, revolutionizing ownership, transactions, and value exchange with unparalleled efficiency and security."
+              )}
+            </p>
```
mDuo13, line 283: Another warning the page is throwing right now _might_ have to do with this section.

> Warning: Expected server HTML to contain a matching \<div> in \<p>.
> ```div
> p
> div
> div
> div
> section
> div
> XrplOverview
> ```

styles/_use-cases.scss
```diff
@@ -543,3 +554,349 @@ html.light {
...
+    left: 0;
+    border-radius: inherit;
+    padding: 1px;
+    background: linear-gradient(90deg, #d91aff 26.41%, #1aa4ff 100.32%);
+    mask: 
+      linear-gradient(#fff 0 0) content-box, 
+      linear-gradient(#fff 0 0);
+    mask-composite: exclude;
+    -webkit-mask-composite: xor;
+  }
```
mDuo13: This effect succeeds at creating a linear-gradient border around the card, but at the cost of making it impossible to click or select the text inside the cards. That can be fixed by adding `z-index`:

```suggestion
  .security-card::before {
    content: '';
    position: absolute;
    top: 0;
    right: 0;
    bottom: 0;
    left: 0;
    border-radius: inherit;
    padding: 1px;
    background: linear-gradient(90deg, #d91aff 26.41%, #1aa4ff 100.32%);
    mask: 
      linear-gradient(#fff 0 0) content-box, 
      linear-gradient(#fff 0 0);
    mask-composite: exclude;
    -webkit-mask-composite: xor;
    z-index: -1;
  }
```

docs/use-cases/tokenization/index.page.tsx
```diff
@@ -0,0 +1,389 @@
...
+  },
+  {
+    description: "Digital Artist",
+    link: "/docs/use-cases/tokenization/digital-artist/",
+  },
+];
+
+const securityAdvantages = [
+  {
+    title: "Trustlines & Authorized Trustlines",
```
mDuo13: It's strange to me that these cards aren't links nor do they contain links to read more about the features they tout.

For this one, I suggest linking to `/docs/concepts/tokens/fungible-tokens/`

docs/use-cases/tokenization/index.page.tsx
```diff
@@ -0,0 +1,389 @@
...
+        <div className="card-text">
+          {translate(
+            "Enable specific accounts to hold issued tokens by setting up allowlisting."
+          )}
+        </div>
+      </>
+    ),
+  },
+  {
+    title: "Freeze & Clawbacks",
```
mDuo13, line 56: The two relevant links for this category are:

`/docs/concepts/tokens/fungible-tokens/freezes/`
`/docs/concepts/tokens/fungible-tokens/clawing-back-tokens/`

Not sure how you would link to both though.

docs/use-cases/tokenization/index.page.tsx
```diff
@@ -0,0 +1,389 @@
...
+        <div className="card-text">
+          {translate(
+            "Recover tokens distributed to accounts in error: for example, reclaim funds sent to an account sanctioned for illegal activity."
+          )}
+        </div>
+      </>
+    ),
+  },
+  {
+    title: "Non-transferable Tokens",
```
mDuo13, line 76: I guess this must be referring to NFTs with the transferable flag unset? We don't currently have docs on when/how to create NTTs, but the best related link is:

`/docs/use-cases/tokenization/digital-artist/`

docs/use-cases/tokenization/index.page.tsx
```diff
@@ -0,0 +1,389 @@
...
+    description: (
+      <>
+        {translate(
+          "Native support for nontransferable items such as identity tokens, airline credits, and consumer rewards, honored by all on-chain participants."
+        )}
+      </>
+    ),
+  },
+  {
+    title: "Royalties",
```
mDuo13, line 88: The best page for explaining this, currently, is probably:

`/docs/use-cases/tokenization/digital-artist/`

(This is the same as the one I mentioned for NTTs above)

docs/index.page.tsx
mDuo13, line 1: The auto-formatter you've used on this page makes it very hard to tell what the substantive diff is in this PR. Can you re-submit it _without_ all the whitespace- and quote-related changes?

mDuo13 (changes requested):
This is looking pretty close now, but there were a few more problems I'd like to see addressed including all the error messages in the console. Also, when you're done, please rebase against the latest master to resolve merge conflicts and possibly squash the work into fewer commits.

pdp2121:
@amarantha-k Could you provide the links corresponding with each card in the security advantage section (if we are planning to do so)? I could not find them in the doc

docs/index.page.tsx
```diff
@@ -164,15 +165,7 @@ function UseCasesCard(props: {
...
-      <ul className="nav flex-column">
-        {useCase.subItems.map(item => (
-          <li key={item.link} className="nav-item">
-            <a href={item.link} className="nav-link">
-              {item.description}
-            </a>
-          </li>
-        ))}
-      </ul>
+      <NavList pages={useCase.subItems} />
```
mDuo13, line 168: :tada:

docs/use-cases/tokenization/index.page.tsx
```diff
@@ -0,0 +1,395 @@
...
+};
+
+export default function Tokenization() {
+  const modalRef = useRef(null);
+
+  return (
+    <div className="landing page-tokenization">
+      <div className="position-relative"></div>
+      <div className="position-relative d-none-sm"></div>
+      <section className="container-new mt-16">
```
mDuo13: The top margin here pushes the main context box down excessively in layout sizes medium and below. Suggested fix:

```suggestion
      <section className="container-new mt-lg-16">
```

docs/use-cases/tokenization/index.page.tsx
```diff
@@ -0,0 +1,395 @@
...
+        </div>
+        <div
+          className="security-card-grid nav card-grid"
+          id="security-features"
+        >
+          {securityAdvantages.map((advantage) => (
+            <div className="security-card" key={advantage.id}>
+              <div className="card-body p-6">
+                <h4 className="card-title h6">{translate(advantage.title)}</h4>
+                <p>{advantage.description}</p>
```
mDuo13: This produces invalid markup because the description sections often contain `<div>` tags which are not allowed to nest inside `<p>` tags. Suggested fix:

```suggestion
                <div>{advantage.description}</div>
```

sidebars.yaml
```diff
@@ -21,21 +21,15 @@
...
+          label: Tokenization
           expanded: false
-          items:
-            - page: docs/use-cases/tokenization/stablecoin-issuer.md
-            - page: docs/use-cases/tokenization/nft-mkt-overview.md
-              expanded: false
-              items:
-                - page: docs/use-cases/tokenization/nftoken-marketplace.md
-                - page: docs/use-cases/tokenization/authorized-minter.md
-                - page: docs/use-cases/tokenization/digital-artist.md
```
mDuo13: These child items (lines 26-33) should remain as the children of the (new) Tokenization page.

static/img/backgrounds/tokenization-stats.png
mDuo13, line 1: This file is now unused and should be removed.

docs/use-cases/tokenization/index.page.tsx
```diff
@@ -0,0 +1,395 @@
...
+    url: "https://www.carbonlandtrust.com",
+  },
+  {
+    id: "nautilus",
+    label: "Nautilus wallet",
+    url: "https://github.com/nautls/nautilus-wallet",
+  },
+  {
+    id: "evernode",
+    label: "Evenode",
```
mDuo13: Typo, missing the R.

```suggestion
    label: "Evernode",
```

docs/use-cases/tokenization/index.page.tsx
```diff
@@ -0,0 +1,395 @@
...
+
+  const handleNext = () => {
+    if (currentIndex < projects.length - 3) {
+      setCurrentIndex(currentIndex + 1);
+    }
+  };
+
+  const ProjectCard = ({ project, index }) => (
+    <a
+      className={`col float-up-on-hover ${index % 2 === 0 ? "even" : "odd"}`}
```
mDuo13: To make the cards use light mode colors, add the `card` class here (requires some other SCSS changes but also makes some other SCSS unnecessary)

```suggestion
      className={`col card float-up-on-hover ${index % 2 === 0 ? "even" : "odd"}`}
```

styles/_use-cases.scss
```diff
@@ -543,3 +554,350 @@ html.light {
...
+  .project-cards .col {
+    padding: 32px;
+    background-color: $black;
+    border-bottom-left-radius: 0.5rem;
+    border-bottom-right-radius: 0.5rem;
+    border-top-left-radius: 0.25rem;
+    border-top-right-radius: 0.25rem;
+    text-align: center;
+    position: relative;
+  }
```
mDuo13: The height of the cards can vary depending on whether the project name gets wrapped onto a second line, causing the control arrows below them to move up and down as you're clicking through the carousel. This can be fixed by adding a minimum height to the cards.

If you've added the `card` class to these elements already (see my suggestion on line 204 of `tokenization/index.page.tsx`) then that already sets minimum height on the cards, but it's too large. Also, the rest of these styles become unnecessary because they're part of the `card` class. So instead, all we need to do is set a smaller minimum height:

```suggestion
  .project-cards .card {
    min-height: 240px;
  }
```

styles/light/_light-theme.scss
```diff
@@ -1181,3 +1181,34 @@ main article .card-grid {
     background-color: $gray-200;
   }
 }
+
+.page-tokenization {
+  .project-name {
+    color: $white;
+  }
```
mDuo13: Since I got the go-ahead from design to make light cards here, we don't need this rule anymore, but we do need a new rule to invert the logos instead.

```suggestion
  .project-cards .project-logo {
    filter: invert(100%);
  }
```

mDuo13 (changes requested):
Noticed a couple important typos that should be fixed and suggested a couple ways to improve the styles including making the use cases carousel use light cards (a change to the design that Ellen is on board with)

styles/_use-cases.scss
pdp2121: Screenshot 2024-06-12 at 2 41 09 PM
240px would leave empty spaces on mobile screen with the reduced font. Maybe we should reduce min-height to 200px on small screen?

mDuo13: Less than ~235px causes it to change heights in some cases, based on my testing

mDuo13 (approved):

PR #2597, opened 2024-05-21
Update peer-crawler.md

iJustErikk:
fix description for overlay.active[].port. see src/ripple/overlay/impl/OverlayImpl.cpp:813.

I felt like String (Number) implied that this is always something like `"51235"`, but it can literally be `51235` OR `"51235"`. Not sure why this would be different.


mDuo13:
I'm not sure when/why it would _not_ be a string. I think this bears further investigation before codifying in the docs.

mDuo13:
Looks like this was probably a mistake in the upstream PR ( https://github.com/XRPLF/rippled/pull/1017 ) a long time ago. @seelabs has volunteered to fix it in rippled to always be a Number type, so we can revisit this when that change gets into a rippled release.

That said, this is accurate to how it works today so :+1: to this PR.

mDuo13 (approved):

PR #2603, opened 2024-05-23
public-servers: divide table in two

intelliot:
As this table has grown, I believe Mainnet deserves its own table here.


DennisDawson (approved):
LGTM. Since all of the other networks are "Test" networks at this time, does it make sense to use the heading "Test Networks"? If we add another sort of network in the future, we can consider "Other" as the heading, but it might be clearer to just call them Test networks, if that's what they're all going to be. @amarantha-k

intelliot:
> Since all of the other networks are "Test" networks at this time, does it make sense to use the heading "Test Networks"?

No opinion from me. I think either way is fine.

amarantha-k:
@DennisDawson Yes, it makes sense to call the heading Test networks for now. +1 from me.

amarantha-k (approved):

PR #2604, opened 2024-05-23
amm updates

DennisDawson:
https://github.com/XRPLF/xrpl-dev-portal/issues/2588


mDuo13 (approved):

docs/img/cpt-amm-lp-tokens.png
mDuo13, line 1: This diagram has kind of a lot going on. Maybe it would be better split into two separate diagrams?

DennisDawson: The diagram illustrates the transactions in the paragraph below. In context, the diagram is just fine. I moved the graphic beneath the paragraph, so that they'll get the context before the graphic.

PR #2605, opened 2024-05-23
Earn Passive Income with AMM

oeggert:


docs/tutorials/how-tos/use-tokens/earn-passive-income-as-a-liquidity-provider.md
```diff
@@ -0,0 +1,227 @@
+---
+parent: use-tokens.html
+seo:
+    description: Provide liquidity for an Automated Market Maker (AMM) and earn income from trading fees.
+embed_xrpl_js: true
+status: not_enabled
+filters:
+  - interactive_steps
```
mDuo13: The `embed_xrpl_js` and `filters` frontmatter elements are vestiges of the old Dactyl-based site and can be omitted here. (Same for `parent` actually.)

docs/tutorials/how-tos/use-tokens/earn-passive-income-as-a-liquidity-provider.md
```diff
@@ -0,0 +1,67 @@
+---
+html: earn-passive-income-as-a-liquidity-provider.html
+seo:
+    description: Provide liquidity for an Automated Market Maker (AMM) and earn income from trading fees.
+status: not_enabled
```
mDuo13: The `status: not_enabled` field should be removed since the AMM is enabled on Mainnet now.

docs/tutorials/how-tos/use-tokens/earn-passive-income-as-a-liquidity-provider.md
```diff
@@ -0,0 +1,67 @@
+---
+html: earn-passive-income-as-a-liquidity-provider.html
```
mDuo13: The `html:` field is a legacy frontmatter field that is only used by Dactyl, not Redocly, and can be safely removed.

docs/tutorials/how-tos/use-tokens/earn-passive-income-as-a-liquidity-provider.md
```diff
@@ -0,0 +1,67 @@
...
+seo:
+    description: Provide liquidity for an Automated Market Maker (AMM) and earn income from trading fees.
+status: not_enabled
+labels:
+  - Decentralized Exchange
+  - Tokens
+  - AMM
+steps: ['Connect', 'Generate', 'Acquire tokens', 'Deposit to the AMM', 'Vote on fees', 'Withdraw from the AMM']
```
mDuo13: The `steps` frontmatter field is only necessary if you're using the interactive tutorial code. Since this tutorial is not interactive, you can omit this.

docs/tutorials/how-tos/use-tokens/earn-passive-income-as-a-liquidity-provider.md
```diff
@@ -0,0 +1,67 @@
...
+---
+# Earn Passive Income as a Liquidity Provider
+
+_(Requires the [AMM amendment][])_
```
mDuo13: This link doesn't get parsed unless you also include the common links definition somewhere in the file, i.e. ending with:

```
{% raw-partial file="/docs/_snippets/common-links.md" /%}
```

docs/tutorials/how-tos/use-tokens/earn-passive-income-as-a-liquidity-provider.md
mDuo13, line 1: I think we should move this tutorial into the **Javascript / Trade On-Ledger** category, which I added for the other AMM tutorial that was made in parallel with this one.

docs/tutorials/how-tos/use-tokens/earn-passive-income-as-a-liquidity-provider.md
```diff
@@ -0,0 +1,67 @@
...
+## Steps
+
+### 1. Connect to the network and generate credentials
+
+{% code-snippet file="/_code-samples/add-amm-lp/js/add-amm-lp.js" language="js" from="// Define the network client" before="// Acquire TST tokens" /%}
+
+
+### 2. Acquire TST tokens
+
+{% code-snippet file="/_code-samples/add-amm-lp/js/add-amm-lp.js" language="js" from="// Acquire TST tokens" before="// Deposit assets into AMM" /%}
```
mDuo13: You should probably have at least _some_ prose explaining this part of the code sample. For example:

```suggestion
### 2. Acquire TST tokens

Before you can perform a two-asset deposit, you must first have the assets you want to deposit. The following code uses an [OfferCreate transaction][] to buy TST from the decentralized exchange.

{% code-snippet file="/_code-samples/add-amm-lp/js/add-amm-lp.js" language="js" from="// Acquire TST tokens" before="// Deposit assets into AMM" /%}
```

docs/tutorials/how-tos/use-tokens/earn-passive-income-as-a-liquidity-provider.md
```diff
@@ -0,0 +1,67 @@
...
+
+
+### 4. Vote on trading fees
+
+This step isn't required, but you're encouraged to submit a vote for a fee structure you believe makes sense.
+
+{% code-snippet file="/_code-samples/add-amm-lp/js/add-amm-lp.js" language="js" from="// Vote on fees" before="// Withdraw by redeeming LP tokens" /%}
+
+
+### 5. Withdraw assets from the AMM
```
mDuo13: This tutorial glosses over the most important step in earning passive income, which is: wait for your LP Tokens to accrue value. There may not be any code involved in this step, but I think it's critical to call it out explicitly and to explain where the passive income from the article title comes from: as people incur AMM trading fees or burn LP Tokens on bidding, the amounts in the AMM pools increase relative to LP Tokens, meaning your LPT become worth more than you deposited.

How long should they wait? As long as they want. Days, weeks, years—as long as both assets are actively traded and healthy (i.e. one of them isn't cratering in value relative to the other) then the longer you hold, the more value your AMM deposits should earn. So you would wait until either you need the liquidity for something else, or you're worried that the relative value of the assets may be too unstable for your risk appetite.

It might also be worth noting that it's not guaranteed that you'll get back the same ratio of asset/asset2 that you deposited, because your withdrawal is based on the proportion of the assets that are in the pool now, which shifts based on trading activity. Depending on how the value of the assets shifts relative to one another, this is also how you can take losses on an AMM deposit (in terms of total value withdrawn vs what you put in).

_code-samples/add-amm-lp/js/add-amm-lp.js
```diff
@@ -0,0 +1,151 @@
...
+} else {
+  console.error("Error sending transaction:", response_withdraw)
+}
+
+// Disconnect when done (If you omit this, Node.js won't end the process)
+await client.disconnect()
+
+}
+
+main()
```
mDuo13: Nit: End files in a newline.

```suggestion
main()

```

_code-samples/add-amm-lp/js/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "xrpl": "^3.0.0"
+  }
+}
+
```
mDuo13: Nit: remove trailing spaces
```suggestion

```

package.json
```diff
@@ -26,6 +26,7 @@
     "react": "^18.2.0",
     "react-alert": "^7.0.3",
     "react18-json-view": "^0.2.6",
+    "readline": "^1.3.0",
```
mDuo13: I don't think we should be changing the Redocly-level depdencies for this tutorial, should we?

sidebars.yaml
```diff
@@ -278,6 +278,7 @@
                 - page: docs/tutorials/how-tos/use-tokens/enact-global-freeze.md
                 - page: docs/tutorials/how-tos/use-tokens/freeze-a-trust-line.md
                 - page: docs/tutorials/how-tos/use-tokens/create-an-automated-market-maker.md
+                - page: docs/tutorials/how-tos/use-tokens/earn-passive-income-as-a-liquidity-provider.md
```
mDuo13: If you move the file, don't forget to move the sidebar entry too.

docs/tutorials/how-tos/use-tokens/earn-passive-income-as-a-liquidity-provider.md
```diff
@@ -0,0 +1,67 @@
...
+# Earn Passive Income as a Liquidity Provider
+
+_(Requires the [AMM amendment][])_
+
+Providing liquidity for an [Automated Market Maker (AMM)](../../../concepts/tokens/decentralized-exchange/automated-market-makers.md) can be a great way to earn passive income. By depositing assets into an AMM pool, you'll earn a percentage of the trading fees generated. This tutorial shows how to deposit assets, vote on fees, and withdraw assets.
+
+
+## Prerequisites
+
+- You should be familiar with basic usage of the [xrpl.js client library](https://github.com/XRPLF/xrpl.js/). See [Get Started Using JavaScript](../build-apps/get-started.md) for setup steps.
```
mDuo13: This "Get Started Using Javascript" link is broken based on the relative paths of these files. (It'll also have to change again if you move this file.)

mDuo13 (changes requested):
Noted a couple broken links that should definitely get fixed, and suggested some other less crucial changes.

package.json
oeggert: Odd, not sure how this change even got put in. Reverting.

docs/tutorials/javascript/trade-on-ledger/earn-passive-income-as-a-liquidity-provider.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+### 4. Vote on trading fees
+
+This step is optional, but you're encouraged to submit a vote for a fee structure you believe makes sense.
+
+{% code-snippet file="/_code-samples/add-amm-lp/js/add-amm-lp.js" language="js" from="// Vote on fees" before="// Get LP tokens value" /%}
+
+
+### 5. Check the value of your LP tokens
```
mDuo13: I still think it makes more sense to explicitly call out "the part where you just sit and wait" for an indeterminate amount of time, since the code sample doesn't do that but in a realistic situation you would, so that people know where to break.

```suggestion
### 5. Wait for your LP Tokens to accrue value
```

mDuo13 (approved):
Looks good at this point. Left one minor suggestion which you can apply or not per your preference.

Issue #2600, opened 2024-05-22
Problems with JS code sample for NFTs


mDuo13:
There are several problems with the JavaScript code sample added in #1682:

- It incorrectly has the `.py` file extension.
- There's no README or `package.json` to explain how to install/run it
- The account and NFTs that it's supposed to browse no longer exist following the recent Testnet reset.

The code samples folder in question is here: https://github.com/XRPLF/xrpl-dev-portal/tree/master/_code-samples/non-fungible-token/js

PR #2609, opened 2024-05-29
Fix some issues with NFT code samples

mDuo13:
This is just a quick band-aid the code sample problems noted in #2600.

- Fix the file extension of JS code sample
- Handle error case when buy offers aren't found
- Add a `package.json` file
- Add a brief README for the code samples


_code-samples/non-fungible-token/js/list-nft-pages-and-buy-offers.js
```diff
@@ -47,6 +50,6 @@
     }
     client.disconnect()
     // End main()
-  }
-  
-  main()
+}
+
+main()
```
justinr1234, line 54: If this is running in nodejs I think the program would end with an unresolved promise

intelliot: Thanks, I think I fixed this in c829b7d

intelliot:
I tried to "suggest changes" but github doesn't allow doing that on lines which weren't touched by this PR, so I just pushed my changes as a new commit. Feel free to revert if it looks bad, but I think my edits are all improvements :)

intelliot (approved):
I did `npm i` and `node list-nft-pages-and-buy-offers.js` and it appears to work fine. Seems like there are currently no buy offers for these NFTs though.

PR #2626, opened 2024-06-17
Add a redirect for a variant of the URL

amarantha-k:


mDuo13 (approved):
Confirmed in a local build that the redirect works as intended.

PR #2627, opened 2024-06-17
fix alert banner show/hide logic, hide apex banner 

AKCodez:


mDuo13 (approved):
Approved with the (S)CSS changes I added

PR #2630, opened 2024-06-18
Add new use cases logos and descriptions, fix use cases grid responsiveness

AKCodez:
…nsiveness


static/js/modal.js
mDuo13, line 1: :tada: Love to see removal of unused code and assets!

static/img/uses/Zerpmon.png
mDuo13, line 1: The capital Z in this filename is causing the logo to not display.

styles/_use-cases.scss
```diff
@@ -30,13 +29,48 @@ $custody-logos: "gatehub", "bitgo";
...
+    content: url("../img/uses/zerpmon.png") !important;
+  }
+}
+
 #use_case_companies_list {
   #threezy {
     .biz-logo {
       max-height: 40px;
-      content: url("../img/uses/modallogos/threezy.png");
+      content: url("../img/uses/modallogos/threezy.png") !important;
```
mDuo13: The `!important` doesn't seem to do anything. It's bad practice to use `!important` where you can avoid it.

If it's not needed, can we remove it from here and also the light theme versions of the styles?

mDuo13 (changes requested):

static/img/uses/Zerpmon.png
AKCodez: Are you sure? Its displaying for me in both sections in light and dark mode

styles/light/_light-theme.scss
```diff
@@ -675,8 +679,19 @@ pre code, pre {
...
+  .first-ledger-bot{
+    content: url("../img/uses/lightmode/first-ledger-bot.svg") !important;
+  }
+  .moai-finance{
+    content: url("../img/uses/lightmode/moai-finance.svg") !important;  
+  }
+  .orchestra-finance{
+    max-height: 52px !important;
+    margin: 0 !important;
+    content: url("../img/uses/lightmode/orchestra-finance.svg") !important;
```
mDuo13: Same as the above comment re: `!important` here too

static/img/uses/Zerpmon.png
mDuo13: Yes, it'll look OK in dev mode on Mac or Windows because their filesystems are not case-sensitive, but it won't be found when the host machine is uses a typical Linux filesystem like `ext4`.

mDuo13 (approved):
Fixed the `zerpmon.png` casing myself; rest should be good now. I'll resolve the conflicts with master (redocly 0.86 / css clashes) myself and merge it myself.

mDuo13:
Merged as 0845af6e46dffa1d421fc5e1510ad528746943ce

PR #2633, opened 2024-06-20
add new events

AKCodez:


mDuo13 (approved):
We still need to update the Apex 2024 section of the Events page too but this part looks good as-is.

PR #2635, opened 2024-06-20
Add brief page on Non-transferable Tokens

mDuo13:
While reviewing #2592, I discovered that we don't really have a page describing NTTs or how they work in the XRP Ledger at all, so I created this short page as a quick summary we can link to to explain it to people. I hope it should improve SEO as well.


oeggert (approved):
Looks good to me.

PR #2641, opened 2024-06-24
 adds osana script

AKCodez:


mDuo13 (approved):
This looks OK although it would be nice to have a little more context about this change. It looks like the defaults are fairly privacy-forward at least?

AKCodez:
> This looks OK although it would be nice to have a little more context about this change. It looks like the defaults are fairly privacy-forward at least?

image

PR #2643, opened 2024-06-27
Migration to the new version of Realm

nzicko:
Update Realm to the version `0.86.0`


tequdev:
Rather than using two Hooks to use translate(), how about creating useTranslate as a custom Hook?

```jsx
export const useTranslate = () =>{
  const { useTranslate } = useThemeHooks();
  return useTranslate
}
```
and use this
```js
  import { useTranslate } from './some/thing'
  ...
  const { translate } = useTranslate();
```

@theme/components/Navbar/Navbar.tsx
```diff
@@ -1,12 +1,11 @@
...
-import { useThemeConfig } from "@theme/hooks/useThemeConfig";
-import { LanguagePicker } from "@theme/i18n/LanguagePicker";
-import { useI18n, useTranslate } from "@portal/hooks";
+import { useThemeConfig, useThemeHooks } from "@redocly/theme/core/hooks";
+import { LanguagePicker } from "@redocly/theme/components/LanguagePicker/LanguagePicker";
 import { slugify } from "../../helpers";
-import { Link } from "@portal/Link";
-import { ColorModeSwitcher } from "@theme/components/ColorModeSwitcher/ColorModeSwitcher";
-import { Search } from "@theme/components/Search/Search";
+import { Link } from "@redocly/theme/components/Link/Link";
```
mDuo13, line 6: If we have to change every instance of this, there might be a few new ones that got added after this branch was forked.

mDuo13 (approved):
These changes look good to me although I agree with @tequdev about `useTranslate`.

PR #2656, opened 2024-07-03
Tokenization landing page fixes

mDuo13:
- remove articles focused on price speculation
- fix network speed stat


amarantha-k (approved):
Agree that pricing-related articles shouldn't be featured on this page. Approving the PR without alternatives for the removed cards so the page can be updated asap. Will work on adding more related articles over the next few days.

PR #2664, opened 2024-07-08
Updating path for validator keys

Aaditya-T:
As the validator-keys package no longer works - and it comes in built with rippled, its best to mention the use of validator-keys from rippled itself instead of asking them for other package.

The guide has confused many (including me) as when trying to build validator-keys tool from scratch, it gives multiple errors as its not maintained anymore and people usually get suck at this stage (see many people ask on the dev server, for example:[here](https://discord.com/channels/886050993802985492/886053414289682522/1240929310227959828)

### DIFF
--- a/docs/infrastructure/configuration/server-modes/run-rippled-as-a-validator.md
+++ b/docs/infrastructure/configuration/server-modes/run-rippled-as-a-validator.md
@@ -67,7 +67,7 @@ Enabling validation on your `rippled` server means providing a validator token i
 
 In a secure location **not** on your validator:
 
-1. Generate a validator key pair using the `create_keys` command. You can find it in the rippled package:
+1. Generate a validator key pair using the `validator-keys` tool, which is included in the `rippled` package:
 
     ```
     $ cd /opt/ripple/bin/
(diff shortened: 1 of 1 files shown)


docs/infrastructure/configuration/server-modes/run-rippled-as-a-validator.md
```diff
@@ -67,14 +67,15 @@ Enabling validation on your `rippled` server means providing a validator token i
 
 In a secure location **not** on your validator:
 
-1. Manually build and run the `validator-keys` tool, if you don't already have it installed via a `rippled` RPM.
+1. Generate a validator key pair using the `create_keys` command. You can find it in the rippled package:
```
mDuo13: - Keep `rippled` in monospace font
- Calling it the `create_keys` command makes it seem like that might be the name of the executable on the commandline, so I think it makes more sense to clarify the name of the main executable and let the example show that `create_keys` is the argument you pass to it.

```suggestion
1. Generate a validator key pair using the `validator-keys` tool, which is included in the `rippled` package:
```

mDuo13 (approved):
To be clear, the linked repository itself is not deprecated and is still being updated, although the build instructions may be outdated.

I do think it's an improvement to just suggest people use the version that comes with the RPM/DEB package if they have it.

Approved with small edits to wording/formatting.

PR #2673, opened 2024-07-11
Add UNL concept doc

mDuo13:
This new concept article summarizes the concept of a "UNL", including the "default UNL", and explains the role of validator list publishers in the consensus process.


docs/concepts/consensus-protocol/unl.md
```diff
@@ -0,0 +1,36 @@
+# Unique Node List (UNL)
+
+A _unique node list_ (UNL) is a server's list of validators that it trusts not to collude. Every XRP Ledger server is configured with a UNL, which determines which validation votes it listens to and which votes it throws out during the consensus process. By design, each entry in a UNL should represent an independent entity, which could be a business, a university, another type of organization, or even just an individual hobbyist; by having each entry be a separate entity, no one has more than a minimum share of the responsibility for keeping the network running normally.
```
oeggert, line 3:
```suggestion
A _unique node list_ (UNL) is a server's list of validators that it trusts not to collude. Every XRP Ledger server is configured with a UNL, which determines which validation votes it listens to and which votes it throws out during the consensus process. By design, each entry in a UNL should represent an independent entity, which could be a business, a university, another type of organization, or even just an individual hobbyist; by having each entry be a separate entity, no one has more than a minimum share of the responsibility for keeping the network running normally.
```
My initial thought is that an individual hobbyist has much voting power as an entire university/business? Is it worth addressing this concern here?

docs/concepts/consensus-protocol/unl.md
```diff
@@ -0,0 +1,36 @@
...
+
+Validators are intended to be impartial, and to process every transaction as soon as is possible within the constraints of the technology. Validators must not block or censor some transactions for arbitrary reasons, because that hinders the network from reaching consensus. More importantly, validators should be online and operational as much as possible. However, the XRP Ledger is designed to allow for imperfections, both in the network and in the validators themselves. Even if some validators are offline, misconfigured, buggy, or outright malicious, the network should still be able to make progress if the majority of them are operating normally, and the network will never confirm transactions that contradict the rules or past history of the network unless a supermajority (>80%) agree. Keeping these things in mind, the validators on a UNL should be chosen to minimize the chances that many validators will fail in the same way at the same time, or collude for malicious reasons.
```
oeggert:
```suggestion
Validators are intended to be impartial, and to process every transaction as soon as is possible within the constraints of the technology. Validators must not block or censor some transactions for arbitrary reasons, because that hinders the network from reaching consensus. More importantly, validators should be online and operational as much as possible. However, the XRP Ledger is designed to allow for imperfections, both in the network and in the validators themselves. Even if some validators are offline, misconfigured, buggy, or outright malicious, the network should still be able to make progress if the majority of them are operating normally, and the network will never confirm transactions that contradict the rules or past history of the network unless a supermajority (>80%) agree. Keeping these things in mind, the validators on a UNL are chosen to minimize the chances that many validators will fail in the same way at the same time, or collude for malicious reasons.
```

docs/concepts/consensus-protocol/unl.md
```diff
@@ -0,0 +1,36 @@
...
+
+## UNL Overlap
+
+Each server operator has full control over which validators are in their UNL. However, if two servers operate with totally different UNLs, they are likely to reach different conclusions about when ledgers (and the transactions in them) are validated. This could lead to a _fork_ in the network; when a fork happens, parties on different sides are unable to mutually agree on what has happened and can't transact with one another. To avoid forking, servers in the XRP Ledger should be configured with UNLs that have some amount of overlap with one another.
```
oeggert:
```suggestion
Each server operator has full control over which validators are in their UNL. However, if two servers operate with totally different UNLs, they are likely to reach different conclusions about when ledgers (and the transactions in them) are validated. This could lead to a _fork_ in the network; when a fork happens, parties on different sides are unable to mutually agree on what has happened and can't transact with one another. To avoid forking, servers in the XRP Ledger need to be configured with UNLs that have a high degree of overlap with one another.
```
Suggest we change the language here to be more aggressive, since the next paragraph says we need 90% overlap to not fork.

docs/concepts/consensus-protocol/unl.md
```diff
@@ -0,0 +1,36 @@
...
+
+Recommended validator lists need to be updated over time, to add more quality validators and remove validators that aren't as reliable or are retiring. Typically, a recommended validator list has an expiration time, because the publisher expects to have made an update available by that time. Lists also have a sequence number, such that the highest sequence number is the newest version of that list and supercedes any older versions. Lists can also have an activation date so that servers can coordinate when to switch to the new version, and the updated list has time to propagate to all servers that use it.
+
+Publishers aren't involved in day-to-day validation of new transactions, but they do wield significant power in selecting which validators are widely trusted. Server operators should be cautious in selecting which validator list publishers they trust, because carelessness on the part of the publishers could affect the reliability of the servers using those lists.
```
oeggert, line 25: So people can technically pick any publisher, but really we really just want them using the two widely accepted ones, right? Maybe it's a good idea here to gently push opinion to using these two lists, and elaborate on why these 2 publishers are so trusted in the space?

docs/concepts/consensus-protocol/unl.md
```diff
@@ -0,0 +1,36 @@
...
+## How to Get Your Validator on a Recommended List
+
+Each publisher can define their own criteria for being listed. However, the criteria for having your validator added to the default UNL typically include the following:
+
+- Run a validator with high uptime for at least a year and high agreement with the rest of the network.
+- Set up [domain verification](/docs/references/xrp-ledger-toml/#domain-verification) for your validator.
+- Be a separate and recognizeable entity in the XRP Ledger community—not, for example, an employee of a company that already runs a validator.
+- Run the validator in a separate physical location than most other validators. (So that, for example, an outage at a single data center does not cause many validators to go down.)
+
+Validator list operators may interview candidates for inclusion in the recommended list to confirm that they meet these and other requirements and that they have a commitment to continuing to run the server in the future.
```
oeggert, line 36: Do we have a link to a form or something where a validator can apply to get on the list?

oeggert (changes requested):
Mostly good, but I've got some suggestions for the general tone of UNLs and the ones people should be choosing.

docs/concepts/consensus-protocol/unl.md
mDuo13: Short answer, yes they do, and in practice the hobbyists seem to do a better job because they care more. Can you clarify what concerns you have?

docs/concepts/consensus-protocol/unl.md
mDuo13: Historically, the threshold for not forking has been a lot lower than 90%, which represents a worst-case scenario. But fair enough.

docs/concepts/consensus-protocol/unl.md
mDuo13: I don't really want to editorialize more than this section already has, and right now I don't know of any publishers other than the two default ones anyway. (Coil used to run a list as well, but they went out of business.)

docs/concepts/consensus-protocol/unl.md
mDuo13: No, I think there might've been a form like that in the earlier days [before domain verification worked like it does now] but not anymore. I believe the general policy is, "Don't reach out to us, we'll reach out to you." Validator performance and domain verification are public, so it's likely that the publishers will notice any validator that meets the requirements.

docs/concepts/consensus-protocol/unl.md
oeggert: If I was a newcomer, I'd see that and wonder if the distribution of voting power is fair giving a single person the same vote as a large organization. But if it's not something you've ever really seen anyone mention, then probably not worth even bringing it up and create a problem that doesn't exist.

oeggert (approved):
Approving now after seeing your comments.

xVet:
Well written, short and not too detailed, good run down of the concepts!

docs/concepts/consensus-protocol/unl.md
rippleitinnz: Consensus does not involve 'throwing out' votes.

A Validator only listens to validators on its UNL and can only consider those votes. It can't consider or 'throw out' votes it doesn't know of, and that is from every other validator not on its UNL.

PR #2678, opened 2024-07-15
Add Organization, FAQ and WebPage schema markup

AKCodez:


mDuo13 (commented):
I don't understand what this is for but I really don't like the idea of the FAQ containing duplicate copies of the questions in the frontmatter and the body.

mDuo13:
Can this be refactored in a way that the question and answer text is reused instead of duplicated in the file? Otherwise, this is going to be a maintenance headache.

Also, are we sure that this format makes Redocly serve the appropriate metadata to the search bots? I couldn't test it experimentally nor find any documentation or other indication that it actually works as intended.

amarantha-k:
Closing this as per offline discussion with @AKCodez.

PR #2687, opened 2024-07-18
Migrate to latest version of realm and adjust plugins to match new interface

volodymyr-rutskyi:
The latest version of Realm brings performance improvements but the interface for custom plugins has changed. 

This PR updates Realm to the latest version and updates plugins to use the new interface.


mDuo13:
Moving this work over to #2688.

PR #2691, opened 2024-07-22
Incorporate feedback and info on how to tune nodes to drop proposals …

amarantha-k:
…from untrusted validator nodes


docs/concepts/consensus-protocol/consensus-protections.md
```diff
@@ -66,6 +66,11 @@ Technically, if you run a server, you can configure your own list site or explic
 
 Research is ongoing to design an improved consensus protocol that allows more heterogeneous validator lists. For more information, see the [Consensus Research](consensus-research.md) page.
 
+## Dropping Proposals From Untrusted Validators
+
+For reasons beyond the scope of discussion here, if a server no longer trusts a validator in the network, the server can update the configuration to drop proposals or validations from the untrusted validator instead of relaying (gossiping) them to the network.
```
mDuo13, line 71: I have some reservations about mentioning this with no reasons for/against.

A non-obvious caveat is that it may be bad for the decentralization of the network if operators configure their servers to not relay untrusted proposals. Since your UNL is not necessarily the same as others', enabling these settings may make it harder for some validators' proposals to find their way to the other nodes who _do_ have those validators in their UNL. If only one operator enables this setting then it may not matter, but the higher the percentage who do, the harder it'll be for lesser-known validators to have their voices heard at all.

Of course it's logically less work and less bandwidth to relay less data, but I think untrusted proposals and validations are generally a small portion of overall bandwidth needs—sending the ledger state to syncing peers is probably far more bandwidth on average.

These settings are kind of "selfish" optimizations with a downside for the network, so I think you may only want to turn them on if needed.

MarkusTeufelberger: I don't see any real advantage of not relaying untrusted validations to unknown peers. IMHO this setting might be useful in a setting where you have a "bridgehead/proxy node" that kinda filters traffic from the big bad internet and only forwards well known content to nodes behind that are never directly exposed to other peers.

Such setups are not really described in the documentation so far afaik and are probably also a bit difficult to properly set up and maintain, so in general I think there is not much use to even consider this outside of a theoretical attack where someone tries to spam the P2P layer with extremely large amounts of validation messages. Setting this proactively does more harm than good imho.

amarantha-k:
Thanks for the feedback - agree that it is better to not document this niche setup proactively. Will close the PR without merging.

PR #2692, opened 2024-07-23
rm apex hero, update rio description

AKCodez:


amarantha-k (changes requested):
Need to add Apex 2024 to the list of past events. 
Otherwise, the changes look good.

PR #2694, opened 2024-07-23
Dev Reflections - Filedgr

amarantha-k:


mDuo13 (approved):
The scribbly diagram is a bit hard to read and the language is too marketing-like for my taste, but this is fine from a formatting standpoint.

PR #2697, opened 2024-07-25
adds xrpl builder office hours event to events and community page

AKCodez:


amarantha-k (approved):
Suggestion, not a blocker - we may want to update the image on the card to reflect office hours.  Otherwise, looks good.

AKCodez:
> Suggestion, not a blocker - we may want to update the image on the card to reflect office hours. Otherwise, looks good.

No, info sessions all use the same stock photo :)

PR #2719, opened 2024-08-09
Add Batch faucet

DennisDawson:
Add faucet information for XLS-56d Batch testing.

### DIFF
--- a/resources/dev-tools/faucets.json
+++ b/resources/dev-tools/faucets.json
@@ -36,7 +36,7 @@
             "id": "faucet-select-batch-devnet",
             "wsUrl": "wss://batch.nerdnest.xyz",
             "jsonRpcUrl": "https://batch.rpc.nerdnest.xyz",
-            "faucetUrl": "https://batch.faucet.nerdnest.xyz/accounts",
+            "faucetUrl": "batch.faucet.nerdnest.xyz/accounts",
             "shortName": "Batch-Devnet",
             "desc": "Preview of XLS-56d Batch transactions."
         }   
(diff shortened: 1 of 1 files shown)


resources/dev-tools/faucets.json
```diff
@@ -31,6 +31,14 @@
             "faucetUrl": "mptfaucet.devnet.rippletest.net",
             "shortName": "MPT-Devnet",
             "desc": "Preview of XLS-33d Multi-Purpose Tokens amendment."
-        }
+        },
+        {
+            "id": "faucet-select-batch-devnet",
+            "wsUrl": "wss://batch.nerdnest.xyz",
+            "jsonRpcUrl": "https://batch.rpc.nerdnest.xyz",
+            "faucetUrl": "https://batch.faucet.nerdnest.xyz/accounts",
```
mDuo13: Contrary to what the field name would imply, this doesn't take a full URL but only the hostname part of it. So as it stands, it'll show you an account and seed but the call to the faucet fails with a CORS error because it's trying to access `https://https://batch.faucet.nerdnest.xyz/accounts/accounts` or something like that.

Works OK if you change it to just a hostname like the others.

```suggestion
            "faucetUrl": "batch.faucet.nerdnest.xyz",
```

P.S. I opened issue #2725 to fix the confusing field name.

mDuo13 (changes requested):

resources/dev-tools/faucets.json
DennisDawson: Updated.

mDuo13 (approved):

PR #2724, opened 2024-08-12
adds 2 new events changes name for one other event

AKCodez:


community/events.page.tsx
```diff
@@ -740,7 +740,7 @@ const events = [
     end_date: "September 4, 2024",
   },
   {
-    name: "XRP Community After Hours",
+    name: "XRPL Community After Hours",
```
mDuo13: The Linkup entry is called "XRPL Zone Seoul After Hours". Maybe we should match that?

```suggestion
    name: "XRPL Zone Seoul After Hours",
```

mDuo13 (approved):

PR #2739, opened 2024-08-28
feat: upgrades realm to latest v, adds id to plugin.js to account for break…

AKCodez:
…ing change

Upgraded realm V in order to fix issue with sitemap.xml appending a trailing slash to the end of the URL degrading SEO performance. Tested locally and can now confirm this upgrade fixes that url issue, screenshot attached.
image


amarantha-k (approved):
Spot tested custom plugins on a local build - looks good.

PR #2740, opened 2024-08-28
feat: add new events

AKCodez:


mDuo13 (approved):
These event titles are hard to read over the background image, but that also applies to a couple events that are older than this PR, so we can address that separately.

mDuo13:
Opened #2742 for the contrast issue

PR #2750, opened 2024-09-05
Create Offers modular tutorial

DennisDawson:


_code-samples/quickstart/js/ripplex3a-create-offers.js
```diff
@@ -0,0 +1,276 @@
...
+        '"value": "' + operationalTakerPaysValueField.value + '"}'
+      takerPays = JSON.parse(takerPaysString)
+    }
+    
+      // -------------------------------------------------------- Prepare transaction
+    const prepared = await client.autofill({
+      "TransactionType": "OfferCreate",
+      "Account": operational_wallet.address,
+      "TakerGets": takerGets,
+        "TakerPays": takerPays
```
oeggert:
```suggestion
    const prepared = await client.autofill({
      "TransactionType": "OfferCreate",
      "Account": operational_wallet.address,
      "TakerGets": takerGets,
      "TakerPays": takerPays
```

_code-samples/quickstart/js/ripplex3a-create-offers.js
```diff
@@ -0,0 +1,276 @@
...
+    })   
+    // ------------------------------------------------- Sign prepared instructions
+    const signed = operational_wallet.sign(prepared)
+    results += "\nSubmitting transaction...."
+    // -------------------------------------------------------- Submit signed blob
+    const tx = await client.submitAndWait(signed.tx_blob)
+    
+    results += tx.results + "\n"
```
oeggert: The operational log is showing `undefined` when you submit an offer create transaction. Looks like something is going on here.

docs/tutorials/javascript/send-payments/create-offers.md
```diff
@@ -0,0 +1,1091 @@
...
+
+[![Results of matching offers for XRP and USD](/docs/img/module-create-offer-xrp-for-usd2.png)](/docs/img/module-create-offer-xrp-for-usd2.png)
+
+## Get Offers
+
+Click **Get Offers** to get a list of offers issued by the corresponding account.
+
+[![Get outstanding offers for the Standby (left) account](/docs/img/module-create-offer-get-offers.png)](/docs/img/module-create-offer-get-offers.png)
+
+## Cancel Offer
```
oeggert, line 76: Maybe reorder this so that an existing offer is on the XRPL when you instruct them to Get Offers and Cancel Offers. Something like:

1. Create Offer
2. Get Offer
3. Cancel Offer
4. Create New Offer
5. Create Corresponding Offer

docs/tutorials/javascript/send-payments/create-offers.md
```diff
@@ -0,0 +1,1091 @@
...
+  results  += "\nBalance changes: \n" + tx.result + "\n" +
+    JSON.stringify(xrpl.getBalanceChanges(tx.result.meta), null, 2)
+  operationalResultField.value = results
+
+  operationalBalanceField.value = (await client.getXrpBalance(operational_wallet.address))                 
+  client.disconnect()    
+} // End of oPcancelOffer()
+```
+
+## ripplex3b-name-field-support.js
```
oeggert, line 447: I'm not sure this is the best place to introduce the name field support. This is the only tutorial that supports this functionality; I'd like to see this introduced in the very first quickstart tutorial and refactored into subsequent ones.

oeggert (commented):
I've added some small suggestions. My main concern is about introducing named accounts this far into the modular tutorials, when it should be a foundational piece from the first quickstart tutorial.

docs/tutorials/javascript/send-payments/create-offers.md
DennisDawson: I see what you're saying, but I don't think the reader needs to be hand-held to this degree. We'll see if this is a problem in the field, but once they understand how to create an offer it's better if they're left to experiment, rather than have us spoon feed each step.

docs/tutorials/javascript/send-payments/create-offers.md
DennisDawson: I agree. We need to set a priority for the refactor - if it's sooner than later, we should incorporate Account Name fields starting at Module one. If not, we'll have to go ahead and introduce it here, even if it's not optimal.

oeggert: That's a good point. I'm not sure we'll be able to refactor any time soon, so I'll go ahead and approve.

oeggert (approved):

_code-samples/quickstart/js/ripplex3a-create-offers.js
DennisDawson: Yep, some dev-time kruft that shouldn't have survived this long. Quashed that and a couple of typos. Thanks for pointing that out.

PR #2760, opened 2024-09-10
feat: add new events and images

AKCodez:


DennisDawson (approved):
There are several instances of where two spaces are used rather than one, but it shouldn't impact the output. Consider removing the extra spaces before merge.

PR #2762, opened 2024-09-12
update image for town hall meetup

AKCodez:


amarantha-k (approved):
Recommend adding 5 PM CET in the description. Otherwise, looks good.

PR #2765, opened 2024-09-14
Add rippled v2.2.3 release notes

mDuo13:
- [ ] Add package sums
- [ ] Add commit message

### DIFF
--- a/blog/2024/rippled-2.2.3.md
+++ b/blog/2024/rippled-2.2.3.md
@@ -14,6 +14,8 @@ Version 2.2.3 of `rippled`, the reference server implementation of the XRP Ledge
 
 <!-- BREAK -->
 
+[Sign Up for Future Release Announcements](https://groups.google.com/g/ripple-server)
+
 ## Background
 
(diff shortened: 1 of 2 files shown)


bachase (approved):
👍

blog/2024/rippled-2.2.3.md
```diff
@@ -0,0 +1,104 @@
...
+
+```text
+Free SQLite space for transaction db is less than 512MB. To fix this, rippled
+  must be executed with the vacuum <sqlitetmpdir> parameter before restarting.
+  Note that this activity can take multiple days, depending on database size.
+```
+
+The exact timing of when a server runs out of space can vary based on a few factors. Server operators who encountered a similar problem in 2018 and followed steps to [increase the SQLite transaction database page size issue](../../../docs/infrastructure/troubleshooting/fix-sqlite-tx-db-page-size-issue) may not encounter this problem at all. The `--vacuum` commandline option to `rippled` from that time may work to free up space in the database, but requires extended downtime.
+
+Version 2.2.3 of `rippled` reconfigures the maximum number of SQLite pages so that the issue does not occur.
```
intelliot:
```suggestion
Version 2.2.3 of `rippled` reconfigures the maximum number of SQLite pages so that the database capacity is doubled. With the default page size of 4096, the capacity is roughly 17.5 TB.
```
In 2018, we thought 8.7 TB was enough that the issue would not occur. Now, we think 17.5 TB is enough. But the long-term fix is something else; perhaps changing the architecture to enable sharding, and/or increasing the page size (likely to its max of 65536 bytes) so that the storage device is the limiting factor instead of SQLite.

intelliot (approved):

PR #2766, opened 2024-09-15
Update set-up-multi-signing.md

sosaucily:
Changing the max number of signers to 32, as per the newer documentation: https://xrpl.org/docs/concepts/accounts/multi-signing


intelliot (approved):
good catch

PR #2769, opened 2024-09-17
adds 2 new events

AKCodez:


amarantha-k (approved):
The new events look good. 
Related but non-blocking: 
- The banner on the Events page needs to be updated with an upcoming event. 
- The events on the Community page need to be updated with upcoming events.

AKCodez:
> The new events look good. Related but non-blocking:
> 
> * The banner on the Events page needs to be updated with an upcoming event.
> * The events on the Community page need to be updated with upcoming events.

will bring this up with design & marketing team.

PR #2771, opened 2024-09-18
fix-gemwallet_image

l0vvkey:
fix broken image link for gemwallet


amarantha-k (approved):
Thanks for the fix!

l0vvkey:
The image is still not showing.
Was able to see it locally when I deployed it.
However, the PR I submitted did not include some other changes, see screenshot.
telegram-cloud-photo-size-5-6291768863053955045-y
Should I resubmit with these changes?

PR #2786, opened 2024-09-23
clarify decimal input and hexadecimal api response

### LINKED ISSUE #2785: Add a clarification to the Price Oracles to specify AssetPrice as decimal

oeggert:


docs/references/protocol/transactions/types/oracleset.md
```diff
@@ -57,7 +57,7 @@ Creates a new `Oracle` ledger entry or updates the fields of an existing one, us
 |---------------------|-----------|---------------|-----------|-------------|
 | `BaseAsset`         | String    | Currency      | Yes       | The primary asset in a trading pair. Any valid identifier, such as a stock symbol, bond CUSIP, or currency code is allowed. For example, in the BTC/USD pair, BTC is the base asset; in 912810RR9/BTC, 912810RR9 is the base asset. |
 | `QuoteAsset`        | String    | Currency      | Yes       | The quote asset in a trading pair. The quote asset denotes the price of one unit of the base asset. For example, in the BTC/USD pair, BTC is the base asset; in 912810RR9/BTC, 912810RR9 is the base asset. |
-| `AssetPrice`        | Number    | UInt64        | No        | The asset price after applying the `Scale` precision level. It's not included if the last update transaction didn't include the `BaseAsset`/`QuoteAsset` pair. |
+| `AssetPrice`        | Number    | UInt64        | No        | The asset price after applying the `Scale` precision level. It's not included if the last update transaction didn't include the `BaseAsset`/`QuoteAsset` pair. Must be in decimal format. |
```
gregtatcam: Generally it should be hexadecimal because Json integer value doesn't support a value greater that max uint.

ajitkulk: @oeggert, I spoke to Greg, and here is a suggestion. We could say that it's recommended to always provide the value in hexadecimal. If the XRPL libraries (js, py, etc.) are used to execute this transaction API, they convert a decimal input to hexadecimal values but the API takes hexadecimal. I am clarifying with the DGE team what happens if the value is more than the maximum int 32.

oeggert: What about something like this:

The `rippled` API takes this value in hexadecimal format; client libraries will convert decimal values to hexadecimal, but you'll receive <insert error/transaction failure> if you submit a number larger than the max UInt.

mDuo13: I'm skeptical that passing the value in hexadecimal will even work as intended, if the problem is that number literals are losing precision in being converted to JavaScript's native double (floating point) format.

gregtatcam: It's a string value. No need to convert.

amarantha-k (approved):
lgtm

PR #2798, opened 2024-10-03
Fix legacy checks tutorials

### LINKED ISSUE #1206: Update Checks tutorials
Update the [Use Checks](https://xrpl.org/use-checks.html) tutorials:

- Update ripple-lib 1.x code samples to xrpl.js 2.x
- Add examples in additional programming languages like Python
- Optional: [make some or all of them interactive](https://github.com/XRPLF/xrpl-dev-portal/blob/master/tool/INTERACTIVE_TUTORIALS_README.md)

mDuo13:
- Remove old WebSocket/JSON-RPC/CLI request/response examples, which were causing display issues on the Code Samples page.
- Update sample code from ripple-lib 1.x to xrpl.js 3.x
- Simplify language and sample code significantly
- Use `getBalanceChanges()` to get the results of cashing a check
- Combine two tutorials for looking up checks and add redirects for the old URLs
- Fix the landing page URL (from redundant `use-checks/use-checks` to `use-checks/`), redirect the old URL, and update internal links (in multiple languages)
- Remove Japanese translations of the outdated versions of the tutorials because they were dependent on the removed code samples

This PR fixes #1206.


mDuo13:
The 2 broken links have already been fixed by #2795 but are still showing up here because this was forked from master while it contained the errors. They should be a non-issue for this PR, but I can rebase if necessary.

_code-samples/checks/js/cancel-check.js
```diff
@@ -1,44 +1,56 @@
...
+const xrpl = require('xrpl')
 
-const secret = "sEdTPPEeMH6SAgpo6rSj8YW7a9vFfUj"; // TODO: Replace with your secret
-const checkId = ""; // TODO: Replace with your check ID
+// Define parameters. Edit this snippet with your values before running it.
+const secret = "s████████████████████████████" // Replace with your secret
+const check_id = "" // Replace with your Check ID
 
-const main = async ()=> {
+async function main() {
```
achowdhry-ripple, line 8: curious was this just for readability? Personally usually default to the previous object syntax

_code-samples/checks/js/get-checks.js
```diff
@@ -1,55 +1,57 @@
...
-
-    let currMarker = null;
-    let account_objects = [];
+    const client = new xrpl.Client('wss://s.altnet.rippletest.net:51233')
+    await client.connect()
 
-    // Loop through all account objects until marker is undefined --------------
+    // Loop through account objects until marker is undefined -----------------
+    let current_marker = null
+    let checks_found = []
```
achowdhry-ripple, line 12: are we purposely keeping the boilerplate code duplicated in each file to provide separate examples? Or could we abstract it for groups of samples since the main difference is the payload object

achowdhry-ripple (approved):
LGTM - love the cleanup and concision

_code-samples/checks/js/cancel-check.js
mDuo13: Yeah, maybe it's old fashioned of me but our other examples all use `function name()` or `async function main()` without the arrow syntax so I changed it to be consistent.

_code-samples/checks/js/get-checks.js
mDuo13: Yeah, the code samples are meant to be something you can actually run with minimal modification. I think including all the necessary boilerplate code helps with context, so readers don't have to guess how much additional setup is needed before the code itself can run. We don't know which code sample they'll find first, after all.

I've been writing the actual tutorials to mostly just showcase the sections that are different, which is hopefully the right balance—users see the unique parts in the tutorial, but they can click through to the full sample code to see how it fits into a fully running sample.

mDuo13:
Rebased and fixed merge conflicts.

khancode (commented):
LGTM

khancode (approved):

_code-samples/checks/js/get-checks.js
achowdhry-ripple: makes sense!

PR #2804, opened 2024-10-09
Fix some lingering Redocly migration issues

mDuo13:
- Perform mass find-and-replace on old callout syntax, converting it to Redocly's admonition syntax
    | Before | After |
    |---|---|
    | no callout; caution text is in paragraph | distinct caution callout |

- Fix a couple instances of incorrect admonition syntax (`type="note"` is not valid and results in a missing background for the admonition; it needs to be `type="info"` instead)
    | Before | After |
    |---|---|
    | black icon on transparent box for callout about quickstart samples | blue callout with legible text and icon |

- Remove `{.github-code-download}` syntax. This was formerly used to style the code downloads to be more obvious in the code samples, but there is no equivalent under Redocly so it was just showing up to end-users. Rather than create a special component for this, I simply removed the old syntax.
- Change the code comment color in dark mode to meet WCAG contrast guidelines.
    | Before | After |
    |---|---|
    | hard to read comment in code block | easier to read comment |


amarantha-k (approved):
The syntax updates look good; verified some affected files and they render correctly. Thanks for taking care of these lingering issues! LGTM.

PR #2807, opened 2024-10-11
Airgapped wallet (py): slim down dependencies & update to xrpl-py 3.0.0

mDuo13:
Most of the dependencies listed in the `requirements.txt` were either not actually used or were indirect dependencies (so it's better not to list them and let pip figure out the correct ones based on the explicit dependencies' versions).

I trimmed it down to just the ones that are used, and updated the code to work with xrpl-py 3.0.0, which has breaking changes in `submit_and_wait` (which now uses rippled API v2). In the process I cleared up some discrepancies in the described format of the signed transaction (transaction blob vs dictionary) and a couple small usage things.

Supersedes #2803.


amarantha-k (approved):
Reviewed text changes and they look good.

PR #2819, opened 2024-10-22
Remove references from AI bot

justinr1234:
Workaround for https://github.com/XRPLF/xrpl-dev-portal/issues/2818


mDuo13:
I would much rather re-index the bot than remove the broken source links since that just makes it even harder to fact-check the sometimes incorrect answers it gives...

PR #2783, opened 2024-09-23
chore: upgrade realm version and migrate deprecated config

volodymyr-rutskyi:
* Upgrade `@redocly/realm` to `0.102.0`
* Remove the `theme.` prefix from translation keys (removed in Realm `0.102.0`)
* Removed the deprecated `theme` configuration property from `redocly.yaml` and front matter (backwards compatibility for it will be dropped in `1.0.0` release)
* Renamed `@i18n` directory to `@l10n` and `i18n` config property to `l10n` (the feature was renamed and the backwards compatibility for the `@i18n` directory will be dropped in `1.0.0` release)

PR #2840, opened 2024-11-04
Upgrade realm and migrate deprecated config

amarantha-k:
This PR resolves merge conflicts noted in #2783.


mDuo13 (approved):
Hate huge diffs like this but it seems like the changes are all just the same difference (removing "theme" prefix, changing i18n to l10n) so :+1:

PR #2847, opened 2024-11-10
apiv2 code updates

oeggert:


mDuo13 (changes requested):
We should change the dependencies as well. `xrpl.js` version 4.0.0 is the lowest one that defaults to API v2, but many of these are specifying a lower version. For xrpl-py, it's version 3.0.0 that defaults to API v2.

For folders with a `package.json` (js) or `requirements.txt` (py) we should update that. If there's a `demo.html` file we also need to update its `<script>` tag to pull the correct version.

mDuo13 (approved):

PR #2848, opened 2024-11-11
adds new institutional defi page

AKCodez:


amarantha-k (commented):
A few comments:
- Could you move the page under Tokenization, so the URL would be /docs/use-cases/tokenizaton/ ?
- Could we rename the page to "real-world-assets-tokenization"?

styles/_contribute.scss
```diff
@@ -1041,14 +1041,17 @@
                     /* 150% */
                     margin-top: 15px;
                     margin-bottom: 15px;
+                    a{
+                        color: #9A52FF;
```
mDuo13: This is a color we have defined in `_colors.scss` so let's use the variable instead of having another random constant:

```suggestion
                        color: $blue-purple-400;
```

Calvinjwala (approved):
LGTM!

docs/use-cases/tokenization/real-world-assets.page.tsx
```diff
@@ -0,0 +1,431 @@
+import { useThemeHooks } from "@redocly/theme/core/hooks";
+import { Link } from "@redocly/theme/components/Link/Link";
+
+export const frontmatter = {
+  seo: {
+    title: "RWA Tokenization on XRP Ledger",
```
mDuo13: This title could use some work. "RWA" is jargon that people may not know—even in the finance space, it could mean "risk-weighted assets" for example. And we don't need to emphasize "on XRP Ledger" when this whole _site_ is about the XRP Ledger.

See screenshot for how it seems to stand out:




```suggestion
    title: "Real-World Asset Tokenization",
```

docs/use-cases/tokenization/real-world-assets.page.tsx
```diff
@@ -0,0 +1,431 @@
...
+    alt: "Ripple",
+    aspectRatio: 3.89,
+    link: "https://ripple.com/rwa-tokenization",
+  },
+];
+
+const features = [
+  {
+    title: "Token Issuance Tutorial",
+    link: "https://xrpl.org/docs/tutorials/how-tos/use-tokens/issue-a-fungible-token",
```
mDuo13: Using a fully-specified URL here means that viewers of the translated site will get shunted to the English-language version of the site when they click it.

It's possible to make it work the intended way, but you need to do two things:

1. Use an absolute path here, e.g.

    ```suggestion
        link: "/docs/tutorials/how-tos/use-tokens/issue-a-fungible-token",
    ```
    (NOTE: If you use my commit below, it modifies these lines already so you don't need to use this suggestion separately.)
2. Change the `FeatureItem` definition below to use a `<Link>` element rather than an `onclick` handler.

docs/use-cases/tokenization/real-world-assets.page.tsx
```diff
@@ -0,0 +1,431 @@
...
+  const { useTranslate } = useThemeHooks();
+  const { translate } = useTranslate();
+  return (
+    <li onClick={() => window.open(link, "_blank")} className="feature-item">
+      <div className="feature-item__content">
+        <span className="feature-item__title">{translate(title)}</span>
+        <span className="right-arrow-item"> </span>
+      </div>
+      <div className="feature-item__divider"></div>
+    </li>
```
mDuo13: It's bad practice to use `onclick` when all you want is something that works like a link—it breaks behavior in subtle ways, like middle-clicking won't open it in a new tab and SEO doesn't catch things.

But furthermore, this whole component is re-creating one we already have defined. I've adjusted the shared component and refactored this page to use it here: 6835e174b42f5d4c84ea810670f0185d670b4c5e . Feel free to pull that commit and use it.

mDuo13 (changes requested):
Mostly looks good but I noticed a couple small things I'd like to see fixed before merging.

docs/use-cases/tokenization/real-world-assets.page.tsx
amarantha-k: This was a recommendation for SEO - "RWA Tokenization on XRP Ledger" for the title and "real-world-assets" for the page name.

amarantha-k:
Other than the page title which was recommended for SEO, agree that it would be good to fix the rest of the comments by @mDuo13 before merging this PR.

mDuo13 (approved):

amarantha-k (approved):
Thanks for incorporating the feedback. LGTM!

PR #2851, opened 2024-11-14
add interactive tutorial

oeggert:
Add interactive tutorial to see how offers interact with clob and amm.


oeggert:
Decided against integrating demo, and will create a video instead.

PR #2856, opened 2024-11-19
Rippling redux

DennisDawson:
Streamline the Rippling topic for clarity.


docs/concepts/tokens/fungible-tokens/rippling.md
```diff
@@ -9,92 +7,54 @@ labels:
...
-
-Rippling only occurs along the [paths](paths.md) of a payment. [Direct XRP-to-XRP payments](../../payment-types/direct-xrp-payments.md) do not involve rippling.
-
-For non-issuing accounts, rippling can be undesirable because it lets other users shift obligations between tokens with the same currency code but different issuers. The [No Ripple Flag](#the-no-ripple-flag) disables rippling by default when others open trust lines to your account, unless you enable rippling by default using the [Default Ripple flag](#the-default-ripple-flag).
-
-{% admonition type="warning" name="Caution" %}When you create a trust line, you must explicitly enable the `tfSetNoRipple` flag to block rippling on your side of that trust line.{% /admonition %}
+Rippling occurs when one holder sends issued currency to another holder of the same currency.
```
mDuo13: This summary describes only the most common case, and doesn't account for the rich possibilities of rippling. Here's an attempt at keeping the core insight without leaving out critical details:

```suggestion
Rippling is the _indirect movement_ of funds that occurs for any payment of fungible tokens, except when directly issuing or redeeming tokens. This includes when one holder sends tokens to another holder of the same tokens. Rippling can automatically swap tokens that have the same currency code, even if they have different issuers, to facilitate longer and more complex payments.
```

docs/concepts/tokens/fungible-tokens/rippling.md
```diff
@@ -9,92 +7,54 @@ labels:
...
 
-[{% inline-svg file="/docs/img/noripple-04.svg" /%}](/docs/img/noripple-04.svg "Charlie --($0)-- Institution A --($11)-- Emily --($90)-- Institution B --($12)-- Daniel")
+The funds are transferred through the Issuer, but ultimately the Issuer’s balance doesn’t change. This flow of funds is known as _rippling_. Issuing accounts must allow rippling so that their holders can transfer funds to one another. 
 
-This may surprise Emily, who does not know Charlie or Daniel. Even worse, if Institution A charges her higher fees to withdraw her money than Institution B, this could cost Emily money. The No Ripple flag exists to avoid this scenario. If Emily sets it on both trust lines, then payments cannot ripple through her address using those two trust lines.
+Rippling can involve several accounts as the XRPL server finds a path to transfer the funds. For example, the funds might flow between two issuer accounts that have a trust line between them.
 
-For example:
+[![Holder A transfers currency through Issuers A and B to get to Holder B.](/docs/img/cpt-rippling5.png "Holder A transfers currency through Issuers A and B to get to Holder B.")](/docs/img/cpt-rippling5.png)
```
mDuo13: There's nothing _wrong_ with this scenario, but I think a more important scenario to illustrate is the one where rippling flows through a holder of both issuers' tokens, as shown in the original [diagram #3](https://xrpl.org/assets/noripple-03.892fdd92148fd9e48c8e4e9ecc0a2a73ee532817dd1594c9b7d06134350e2bee.ac57e6ef.svg). Laid out like you have the others, the diagram would be shaped kind of like an "M" (or upside-down-W).

One little detail that the current diagram elides is that Holder A is sending "USD issued by issuer A" and Holder B is receiving "USD issued by issuer B"—technically two different tokens which can have different properties like transfer fees, even though they have the same currency code (rippling can only occur if the currency code is the same). In my experience, this simplification tends to confuse people, who think that all USD issued in the XRPL is the same. So it might be worth adding that detail to the diagram somehow. Conventionally, we've written "USD.issuerA" to represent this type of thing.

docs/concepts/tokens/fungible-tokens/rippling.md
```diff
@@ -9,92 +7,54 @@ labels:
...
-
-
-## The Default Ripple Flag
-
-The **Default Ripple** flag is an account setting that enables rippling on all _incoming_ trust lines by default. Issuers MUST enable this flag for their customers to be able to send tokens to each other.
-
-The Default Ripple setting of your account does not affect trust lines that you create; only trust lines that others open to you. If you change the Default Ripple setting of your account, trust lines that were created before the change keep their existing No Ripple settings. You can use a [TrustSet transaction][] to change the No Ripple setting of a trust line to match your address's new default.
+# The No Ripple Flag
 
+Accounts that do not want to allow rippling can set the No Ripple flag. If you set the No Ripple flag on two of its trust lines, other accounts cannot use those trust lines to ripple payments through your account.
```
mDuo13: - Use "enable" rather than "set" for clarity.
- Pronoun disagreement between "you" and "its"
- This statement may give a false sense of security to people who don't fully grasp the nature of the "both trust lines" aspect of No Ripple.

```suggestion
Accounts that do not want to allow rippling can enable the No Ripple flag. If you enable the No Ripple flag on two of your trust lines, other accounts cannot use that pair of trust lines to ripple payments through your account. However, if you have any other trust lines that do not have the No Ripple flag enabled, payments can still ripple through your account as long as _one_ of the trust lines involved does not have No Ripple enabled.
```

docs/concepts/tokens/fungible-tokens/rippling.md
```diff
@@ -9,92 +7,54 @@ labels:
...
-<!--{# TODO: move these things into their own tutorials #}-->
-
-### Enabling / Disabling No Ripple
-
-The No Ripple flag can only be enabled on a trust line if the address has a positive or zero balance on that trust line. This is so that the feature cannot be abused to default on the obligation the trust line balance represents. (Of course, you can still default by abandoning the address.)
 
-To enable the No Ripple flag, send a [TrustSet transaction][] with the `tfSetNoRipple` flag. You can disable the No Ripple flag (that is, allow rippling) with the `tfClearNoRipple` flag instead.
+The No Ripple flag can only be enabled on a trust line if the address has a positive or zero balance. This is so that the feature cannot be abused to remove the obligation the trust line balance represents.
 
+To enable the No Ripple flag, send a `TrustSet` transaction with the `tfSetNoRipple` flag. To disable the No Ripple flag, send a `TrustSet` transaction with the `tfClearNoRipple` flag.
```
mDuo13: This is a good place to link to the reference documentation—people who want to know more can click through to find more details on how that works.

```suggestion
To enable the No Ripple flag, send a [TrustSet transaction][] with the `tfSetNoRipple` flag. To disable the No Ripple flag, send a TrustSet transaction with the `tfClearNoRipple` flag.
```

docs/concepts/tokens/fungible-tokens/rippling.md
```diff
@@ -9,92 +7,54 @@ labels:
...
-- **Tutorials:**
-    - [Stablecoin Issuer](../../../use-cases/tokenization/stablecoin-issuer.md)
-- **References:**
-    - [account_lines method][]
-    - [account_info method][]
-    - [AccountSet transaction][]
-    - [TrustSet transaction][]
-    - [AccountRoot Flags](../../../references/protocol/ledger-data/ledger-entry-types/accountroot.md#accountroot-flags)
-    - [RippleState (trust line) Flags](../../../references/protocol/ledger-data/ledger-entry-types/ripplestate.md#ripplestate-flags)
+Using the HTTP / WebSocket APIs or your preferred client library, look up trust lines with the `account_lines` method. For each trust line, the `no_ripple` field shows whether the current address has enabled the No Ripple flag on that trust line. The `no_ripple_peer` field shows whether the counterparty has enabled the No Ripple flag.
```
mDuo13, line 64: I strongly disagree with removing the hyperlinks from this paragraph.

docs/concepts/tokens/fungible-tokens/rippling.md
```diff
@@ -9,92 +7,54 @@ labels:
...
 
-Now the above scenario, where Charlie pays Daniel while rippling through Emily's address, is no longer possible.
+# The Default Ripple Flag
 
-### Specifics
+The Default Ripple flag is an account level setting that enables rippling for all incoming trust lines. The default setting of the Default Ripple flag is false. Issuers must enable this flag for their customers to be able to send tokens to one another.
 
-The No Ripple flag makes certain paths invalid, so that they cannot be used to make payments. A path is considered invalid if and only if it enters **and** exits an address node through trust lines where No Ripple has been enabled for that address.
+The Default Ripple setting doesn’t affect trust lines created by your account, only trust lines others open to your account. If you change the setting, trust lines that were created before the change keep their existing No Ripple settings. You can use a TrustSet transaction to change the No Ripple setting of a trust line to match your account’s new default.
```
mDuo13: - Change passive voice back to active voice
- Don't remove useful hyperlinks

```suggestion
The Default Ripple setting doesn’t affect trust lines that you create, only trust lines that others open to your account. If you change the setting, trust lines that were created before the change keep their existing No Ripple settings. You can use a [TrustSet transaction][] to change the No Ripple setting of a trust line to match your account’s new default.
```

docs/concepts/tokens/fungible-tokens/rippling.md
mDuo13, line 1: The use of "issued currency" throughout this document is inconsistent with our [terminology guidelines](https://xrpl.org/resources/contribute-documentation#terminology) so either it should use the term "tokens" or we need to update the style guide to clarify when "issued currency" is an acceptable usage. In my opinion, this is one of those cases that should use "tokens" (or "fungible tokens" as necessary for clarity) because rippling can occur for anything issued on a trust line, regardless of whether it represents a currency (like a stablecoin), tokenized asset like gold, or other type of financial instrument. Lacking any user study data on the point, we should probably poll the team/contributors.

docs/concepts/tokens/fungible-tokens/rippling.md
```diff
@@ -9,92 +7,54 @@ labels:
...
+[![Holder B receives currency through the Issuer.](/docs/img/cpt-rippling4.png "Holder B receives currency through the Issuer.")](/docs/img/cpt-rippling4.png)
 
-[{% inline-svg file="/docs/img/noripple-04.svg" /%}](/docs/img/noripple-04.svg "Charlie --($0)-- Institution A --($11)-- Emily --($90)-- Institution B --($12)-- Daniel")
+The funds are transferred through the Issuer, but ultimately the Issuer’s balance doesn’t change. This flow of funds is known as _rippling_. Issuing accounts must allow rippling so that their holders can transfer funds to one another. 
 
-This may surprise Emily, who does not know Charlie or Daniel. Even worse, if Institution A charges her higher fees to withdraw her money than Institution B, this could cost Emily money. The No Ripple flag exists to avoid this scenario. If Emily sets it on both trust lines, then payments cannot ripple through her address using those two trust lines.
+Rippling can involve several accounts as the XRPL server finds a path to transfer the funds. For example, the funds might flow between two issuer accounts that have a trust line between them.
```
mDuo13: Saying "the XRPL server" here might be interpreted as there being one centralized server that does the calculations. In reality, every XRPL server does the same calculations in parallel when processing the transaction. For simplicity, I like to describe this type of action as being done collectively by "the XRP Ledger".

```suggestion
Rippling can involve several accounts as the XRP Ledger finds a path to transfer the funds. For example, the funds might flow between two issuer accounts that have a trust line between them.
```

mDuo13 (changes requested):

docs/concepts/tokens/fungible-tokens/rippling.md
DennisDawson: Inadvertent, restored.

docs/concepts/tokens/fungible-tokens/rippling.md
DennisDawson: Changed to tokens throughout.

docs/concepts/tokens/fungible-tokens/rippling.md
DennisDawson: Revised, with some shorter sentences.

docs/concepts/tokens/fungible-tokens/rippling.md
```diff
@@ -9,85 +7,67 @@ labels:
...
 
-This may surprise Emily, who does not know Charlie or Daniel. Even worse, if Institution A charges her higher fees to withdraw her money than Institution B, this could cost Emily money. The No Ripple flag exists to avoid this scenario. If Emily sets it on both trust lines, then payments cannot ripple through her address using those two trust lines.
+[![Holder A transfers tokens through Issuers A and B to get to Holder B.](/docs/img/cpt-rippling5.png "Holder A transfers tokens through Issuers A and B to get to Holder B.")](/docs/img/cpt-rippling5.png)
 
-For example:
+Tokens might also flow through intermediary gateway accounts to reach a holder.
 
-[{% inline-svg file="/docs/img/noripple-05.svg" /%}](/docs/img/noripple-05.svg "Charlie --($10)-- Institution A --($1, No Ripple)-- Emily --($100, No Ripple)-- Institution B --($2)-- Daniel")
+[![Holder A transfers tokens through Issuers A and B and Gateway X to get to Holder B.](/docs/img/cpt-rippling6.png "Holder A transfers tokens through Issuers A and B and Gateway X to get to Holder B.")](/docs/img/cpt-rippling6.png)
```
mDuo13: I worry that using the term "gateway" here will confuse people who come across legacy content that uses the term "gateway" differently. From 2013-2016 or so the term "gateway" was widely used in the XRPL community and in Ripple's communications to mean, essentially, "stablecoin issuer."

Perhaps "intermediary" is better here (and in the diagram) to dodge that confusion. The key thing here is that the intermediary is also just an ordinary holder except they allow rippling, not necessarily a special business or anything.

mDuo13 (dismissed):
Looks good with the changes, with one important note on terminology.

docs/concepts/tokens/fungible-tokens/rippling.md
DennisDawson: There's a wider discussion on this - I agree that we need a term for an account that allows rippling but is not an issuer. I want to make it clear that Holders never allow rippling, Exchanges allow rippling, and Issuers allow rippling, then document the basic configurations for each of those accounts. We can pick the term for the intermediary/exchange user after the holidays.

mDuo13 (approved):

PR #2860, opened 2024-11-21
fix(docs): `XChainCreateClaimID` SignatureReward field type

GuillemGarciaDev:
# fix(docs): `XChainCreateClaimID` SignatureReward field type

This PR aims to fix a typo in the [`XChainCreateClaimID`](https://xrpl.org/docs/references/protocol/transactions/types/xchaincreateclaimid) transaction documentation.
The error is found on the `SignatureReward` field. It should be a `Currency Amount` type instead of an `Account`.


mDuo13 (approved):
Confirmed with the source that [`SignatureReward` is an Amount type](https://github.com/XRPLF/rippled/blob/0ec17b6026298dc150099b66a2cecad0fe561d1b/include/xrpl/protocol/detail/sfields.macro#L227).

PR #2865, opened 2024-11-26
Fee Voting revisions

mDuo13:
Rewrite the Fee Voting article:

- Simplify the explanation of the fee compromise to explain that it uses the median. (Props to @mvadari for noticing that the existing algorithm boils down to taking the median vote.)
- Explain some of the rationale for fee settings more explicitly.
- Update the maximum values section to reflect the XRPFees amendment being live.
- Small wording changes for clarity.


ximinez (approved):
I gave this a read-through. Looks good.

mvadari (approved):

PR #2878, opened 2024-12-02
Fix TOML checker and redundant Use Checks landing file

mDuo13:
- Fixes `[warn] Index route used with no sidebar @l10n/ja/docs/tutorials/how-tos/use-specialized-payment-types/use-checks/use-checks.md`. The file in question was an unnecessary duplicate of the `index.md` in the same folder.
- Removes translation hooks from the TOML Checker since they were breaking it. The code needs a bigger refactor for the translation to work, so for now I'm at least making it work in English.


resources/dev-tools/toml-checker/ValidateTomlSteps.tsx
```diff
@@ -204,22 +201,22 @@ async function parseXRPLToml(
...
-                label: translate("NOT FOUND"),
+                label: "NOT FOUND",
                 type: "ERROR"
                 }
             }
         })
 
         addNewLogEntry(setLogEntries, {
-            message: translate("Account not found in TOML file. Domain can not be verified."),
+            message: "Account not found in TOML file. Domain can not be verified.",
```
DennisDawson: perhaps _cannot_?

DennisDawson (approved):
Looks good, just want to be sure you need the emphatic "can not" rather than the more typical "cannot" in one instance.

PR #2888, opened 2024-12-04
Quick fixes

amarantha-k:


docs/references/protocol/binary-format.md
```diff
@@ -155,17 +155,17 @@ All fields in a transaction are sorted in a specific order based first on the fi
 
 ### Type Codes
 
-Each field type has an arbitrary type code, with lower codes sorting first. These codes are defined in [`SField.h`](https://github.com/XRPLF/rippled/blob/master/src/ripple/protocol/SField.h#L57-L74).
+Each field type has an arbitrary type code, with lower codes sorting first. These codes are defined in [`SField.h`](https://github.com/XRPLF/rippled/blob/master/include/xrpl/protocol/SField.h#L60-L98).
 
-For example, [UInt32 has type code 2](https://github.com/XRPLF/rippled/blob/72e6005f562a8f0818bc94803d222ac9345e1e40/src/ripple/protocol/SField.h#L59), so all UInt32 fields come before all [Amount fields, which have type code 6](https://github.com/XRPLF/rippled/blob/72e6005f562a8f0818bc94803d222ac9345e1e40/src/ripple/protocol/SField.h#L63).
+For example, [UInt32 has type code 2](https://github.com/XRPLF/rippled/blob/master/include/xrpl/protocol/SField.h#L67), so all UInt32 fields come before all [Amount fields, which have type code 6](https://github.com/XRPLF/rippled/blob/master/include/xrpl/protocol/SField.h#L71).
```
mDuo13, line 160: I'm a little wary of including line numbers in links to the `master` branch rather than a specific commit, because any change to the previous lines would cause the link to still work, but highlight the wrong lines.

The downside of linking an individual commit is after a release it's still pointing to an outdated version of the codebase. So either way it kind of needs to be updated whenever a new release changes the file. Which is worse, linking the wrong lines of the latest version, or linking the right lines of an outdated version?

Line 158 already links by branch, so it's not a showstopper either way.

In this specific example, it might actually be best to remove the links since the same file is linked in the previous paragraph anyway.

mDuo13 (approved):

PR #2891, opened 2024-12-06
Restore Algolia DocSearch

mDuo13:
The old Algolia DocSearch powered search, which still has an active index, seems to work much better for most typical searches than Redocly's built-in search does. This PR removes the Redocly search interface and re-adds the DocSearch-powered interface in its place. Much of the SCSS for the Algolia-based search was still there, though I had to make a few adjustments to make it work in both light and dark mode in the new interface, including mobile and desktop styles.


justinr1234 (dismissed):

achowdhry-ripple (dismissed):

amarantha-k (dismissed):
Thanks for replacing with the Algolia DocSearch, @mDuo13! The search function returns better results now! 
Nit: The search bar seems be in the middle of the nav. Is it possible to move it to the right, next to the lang picker? 
Screenshot 2024-12-06 at 10 07 41 AM

justinr1234 (approved):

amarantha-k (approved):
LGTM! Thanks

PR #2893, opened 2024-12-06
Improve visual distinction of search on light mode

mDuo13:
Fast follow to #2891.

- In light mode, restore drop shadow to search modal and adjust background color, so that the search box is more visually distinct from page contents.
- Fix indentation of page-specific styles in same file.


amarantha-k (approved):
LGTM

PR #2900, opened 2024-12-10
Update search language filtering

mDuo13:
After re-indexing the site, there were two problems that caused search to return no results:

- DocSearch was no longer configured with `lang` as a filterable attribute. We fixed this in the Algolia interface.
- The language tag on pages changed from `en` to `en-US`. This PR removes the special case that we previously needed to get results for English, which is now causing results not to appear for English.

This PR also fixes a number of typos in the `redirects.yaml` file that were discovered while investigating the search problems.


amarantha-k (approved):
Verified that search is returning valid results. LGTM

PR #2901, opened 2024-12-10
Update broken leagcy links in dev tools

mDuo13:
- Fix several broken links from the dev tools that were using legacy `.html` relative links. These were discovered while debugging search problems.
- Fix styling issue that caused links in transaction sender help text not to be underlined or otherwise recognizably styled as links.


amarantha-k (approved):
LGTM

PR #2907, opened 2024-12-12
Add response headers for CSP (X-Frame-Options) and a noindex for work-in-progress translated pages

amarantha-k:


mDuo13 (approved):
I'm seeing the frame headers in staging, so that's great!


Not sure if/how I can test the ES files. I wasn't able to get them to show up in the preview (which they shouldn't).

redocly.yaml
```diff
@@ -13,6 +13,15 @@ l10n:
       name: 日本語
 redirects:
   $ref: redirects.yaml
+responseHeaders: 
+  '**':
+    - name: X-Frame-Options
+      value: DENY
```
mDuo13, line 19: Actually, based on what I'm seeing on MDN, [`X-Frame-Options` is deprecated](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) and we should be setting things in [`Content-Security-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) instead.

And `frame-ancestors` isn't supposed to be a header, but is one of the directives that can be embedded in the `Content-Security-Policy` header. So we should probably change that.

mDuo13 (changes requested):

redocly.yaml
```diff
@@ -13,6 +13,15 @@ l10n:
...
+    - name: frame-ancestors
+      value: NONE
```
mDuo13: Following up on the above, I think it's probably unnecessary to set `X-Frame-Options` (any browser that doesn't support `Content-Security-Policy` probably can't display the site properly regardless) but not harmful.

As for the frame ancestors, I think this is the right syntax:

```suggestion
    - name: Content-Security-Policy
      value: frame-ancestors 'none'
```

redocly.yaml
amarantha-k: Thanks for looking into this one and adding the suggested change.

mDuo13 (approved):

PR #2922, opened 2024-12-21
Fix gramatical error when referencing Authorized Trust Lines

### LINKED ISSUE #2921: Fix gramatical error when referencing Authorized Trust Lines
https://xrpl.org/docs/concepts/tokens/fungible-tokens/authorized-trust-lines

**This sentence needs to be updated from:**
With a stablecoin on the XRP Ledger and use Authorized Trust Lines, the process of onboarding a new customer might look something like the following
**to:** 
With a stablecoin on the XRP Ledger and the use of Authorized Trust Lines, the process of onboarding a new customer might look something like the following

jonathanirvin:
Fixes #2921


amarantha-k (approved):
Thanks for the update!

PR #2934, opened 2025-01-06
Prevent maintenance issues when adding faucet hosts and fix the loader

### LINKED ISSUE #2725: Faucets page - rename faucetUrl to faucetHost
It's not actually supposed to contain a URL, so that's confusing for maintenance every time we add a new faucet.

Modify https://github.com/XRPLF/xrpl-dev-portal/blob/master/resources/dev-tools/xrp-faucets.page.tsx and https://github.com/XRPLF/xrpl-dev-portal/blob/master/resources/dev-tools/faucets.json to change the `faucetUrl` field to be something more accurate, probably `faucetHost`.

Alternatively, we could modify it to actually take a URL.

jonathanirvin:
1. Prevent maintenance issues when adding new faucet hosts
2. Fix the loader which broke during a past migration

Fixes: #2725

### DIFF
--- a/@theme/components/XRPLoader.tsx
+++ b/@theme/components/XRPLoader.tsx
@@ -6,7 +6,7 @@ export interface XRPLoaderProps {
 
 export default function XRPLoader(props: XRPLoaderProps) {
   return (
-  <div className="loader collapse">
+  <div className="loader">
     <img alt="(loading)" className="throbber" src="/img/xrp-loader-96.png" />
     {props.message}
   </div>);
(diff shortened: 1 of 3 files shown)


mDuo13 (approved):
Confirmed the changes work and the loading spinner is back. Thanks!

Also noticed the batch faucet doesn't work properly (it returns an `invalid_API_version` version error) but that's also true in production and isn't caused by this PR.

mDuo13:
Identified and reported https://github.com/XRPLF/xrpl.js/issues/2872 as the source of the issue with Batch Devnet (and Xahau) faucets.

I believe the MPT Devnet has been decommissioned now that MPTs are merged and available on the regular Devnet.

PR #2935, opened 2025-01-07
Remove MPT-Devnet faucet, which has been shut down; use APIv1 for Faucets

mDuo13:
- MPTs are available on the regular Devnet so the MPT-Devnet (including faucet) has been shut down.
- Xahau network doesn't support APIv2, so explicitly specify APIv1 in the Faucet code. (The other networks' public servers support both API versions and we don't need anything from APIv2 here, so this is purely restoring functionality in this case.)
- Known issue: the Batch faucet still doesn't work; it appears to be broken on the server side.


amarantha-k (approved):
lgtm

PR #2943, opened 2025-01-13
Fix broken links and update redirects

AKCodez:


mDuo13 (approved):
Looks good. OK to merge as long as the link checker passes after the additional edits I made. (Should probably be squashed)

PR #2945, opened 2025-01-15
Update Redocly to Realm v0.115.0 and fix some related issues

mDuo13:
- Fixes an issue introduced by #2934 where the loading animation is always visible in interactive tutorials
- Fix a broken image path in stablecoin settings page in Japanese
- Remove redundant loader component from dev tools page

Supersedes #2892.

Known issue, Redocly build reports this:

```text
[error] [1] File /js/tutorials/how-tos/use-tickets.js does not exist
at ./@l10n/ja/docs/tutorials/how-tos/manage-account-settings/use-tickets.md:1:1
```

~~But as best I can tell all the references to `use-tickets.js` in that file (the Japanese `use-tickets.md`) are absolute paths so it seems like a Redocly bug.~~ The error message was incorrect, but there was a real path issue behind it and I've fixed it.


mDuo13:
@SocketSecurity ignore npm/emoji-regex-xs@1.0.0

amarantha-k (approved):
lgtm

PR #2958, opened 2025-01-23
Devnet Reset Blog Post

oeggert:


blog/2025/devnet-reset.md
```diff
@@ -0,0 +1,57 @@
...
+
+<!--
+{% admonition type="success" name="Update" %}
+The reset has completed successfully. Devnet is online and fully operational.
+{% /admonition %}
+-->
+
+## Background
+
+Ripple operates Clio servers on Devnet to help the community access validated ledger data efficiently. It better handles API calls and stores data in a more space-efficient format than rippled servers, ultimately reducing the load on the network. Clio uses ScyllaDB, which is nearing storage capacity; in order to continue serving the community, Devnet will be reset to clear the Clio databases.
```
amarantha-k:
```suggestion
Ripple operates Clio servers on Devnet to help the community access validated ledger data efficiently. 

Clio handles API calls better and stores data in a more space-efficient format than `rippled` servers, ultimately reducing the load on the network. Clio uses ScyllaDB, which is nearing storage capacity. In order to continue serving the community, Devnet will be reset to clear the Clio databases.
```

amarantha-k (commented):
Added a suggestion

blog/2025/devnet-reset.md
```diff
@@ -0,0 +1,57 @@
+---
+category: 2025
+date: 2025-01-24
+seo:
+    title: Upcoming Devnet Reset Notification
```
mDuo13: We should be able to leave the `title` field off and have it pick the title from the first h1 of the page, I think.

If we can't, we should make sure they match. Right now this doesn't match because it has the extra word "Notification"

blog/2025/devnet-reset.md
```diff
@@ -0,0 +1,57 @@
...
+
+Ripple operates Clio servers on Devnet to help the community access validated ledger data efficiently. It better handles API calls and stores data in a more space-efficient format than rippled servers, ultimately reducing the load on the network. Clio uses ScyllaDB, which is nearing storage capacity; in order to continue serving the community, Devnet will be reset to clear the Clio databases.
+
+## Impact
+
+This reset affects Devnet only. Other networks will continue to operate as usual, including XRPL Mainnet, XRPL Testnet, Xahau, and the Hooks Testnet.
+
+The reset will delete all ledger data in Devnet, including all accounts, transactions, balances, settings, offers, AMMs, escrows, and other data. This means all balances will be reset to zero and the block number will start at one again. No changes are anticipated to services such as Devnet APIs, faucets, Explorers, access rights, and wallet integrations; these services usually manage resets without issues.
+
+Devnet will have its amendment statuses restored to match XRPL Mainnet. Any existing accounts or other data will need new test XRP from the faucet and will need to be re-created.
```
mDuo13: I don't think this is quite the case. Devnet usually has amendments that aren't yet available on Mainnet and I expect it to continue to be the case. (For example, Credentials, AMMClawback, and MPTs are currently available on Devnet and will probably be enabled again on Devnet after the reset.)

blog/2025/devnet-reset.md
```diff
@@ -0,0 +1,57 @@
...
+
+This reset affects Devnet only. Other networks will continue to operate as usual, including XRPL Mainnet, XRPL Testnet, Xahau, and the Hooks Testnet.
+
+The reset will delete all ledger data in Devnet, including all accounts, transactions, balances, settings, offers, AMMs, escrows, and other data. This means all balances will be reset to zero and the block number will start at one again. No changes are anticipated to services such as Devnet APIs, faucets, Explorers, access rights, and wallet integrations; these services usually manage resets without issues.
+
+Devnet will have its amendment statuses restored to match XRPL Mainnet. Any existing accounts or other data will need new test XRP from the faucet and will need to be re-created. 
+
+If code relies on specific addresses, a request to the faucet can fund the same address again. However, any AMMs that are re-created after the reset will generally have different account addresses. As a reminder, it's best not to use the same addresses or key pairs on Mainnet and any developer networks.
+
+The important URLs, Network ID, and validator settings for Devnet will remain the same.
```
mDuo13: I think the validator settings for Devnet are expected to change with the new validator list signing key (if we do that).

mDuo13 (changes requested):
I'm a little concerned that the rationale given makes it sound like Clio is holding the Devnet back. It might give people the wrong idea about Clio if people think it can't even manage the little-used Devnet's level of traffic. I think it's specifically the spec of VMs that we have currently in use for Devnet that are nearing their capacity, not an inherent limit to the Clio server itself.

blog/2025/devnet-reset.md
oeggert: Good catch--removing the mention about amendment matching with mainnet.

blog/2025/devnet-reset.md
oeggert: They're upading t he validator keys--adding instructions for that.

blog/2025/devnet-reset.md
```diff
@@ -0,0 +1,68 @@
...
+{% admonition type="success" name="Update" %}
+The reset has completed successfully. Devnet is online and fully operational.
+{% /admonition %}
+-->
+
+## Background
+
+Ripple operates Clio servers on Devnet to help the community access validated ledger data efficiently. Clio handles API calls better and stores data in a more space-efficient format than `rippled` servers, ultimately reducing the load on the network.
+
+The virtual machines we're running the Clio servers from are nearing storage capacity. In order to continue serving the community, Devnet will be reset to clear the Clio databases.
```
mDuo13: Without a byline, the "we" here is ambiguous. This is the XRPL dev blog, not a Ripple blog, so it shouldn't be assumed that "we" are "Ripple" in this context. It might be better to avoid 1st person entirely.

```suggestion
The virtual machines running the Clio servers are nearing storage capacity. In order to continue serving the community, Devnet will be reset to clear the Clio databases.
```

blog/2025/devnet-reset.md
```diff
@@ -0,0 +1,68 @@
...
+
+If you run a `rippled` server that is connected to Devnet, after the reset you should delete your database data and restart the server. Database files and folders are defined in the config file in the `[database_path]` and `[node_db]` stanzas. If you use the default config, you can run the following commands:
+
+```sh
+rm -r /var/lib/rippled/db/*
+
+systemctl restart rippled.service
+```
+
+You will also have to update the `[validator_list_keys]` keys stanza in your `validators.txt` file:
```
mDuo13: Nit: introduce the code sample with _slightly_ more context:

```suggestion
You will also have to update the `[validator_list_keys]` keys stanza in your `validators.txt` file with a new key, as in the following example:
```

mDuo13 (changes requested):
Looks good with some minor notes

mDuo13 (approved):

PR #2962, opened 2025-01-25
Add Dhali public server to docs

desimmons:
This PR adds Dhali as a paid RPC option to the docs.

### DIFF
--- a/docs/tutorials/public-servers.md
+++ b/docs/tutorials/public-servers.md
@@ -7,16 +7,21 @@ labels:
   - Core Server
 ---
 # Public Servers
-
 If you don't [run your own `rippled` server](../infrastructure/installation/index.md), you can use the following public servers to submit transactions or read data from the ledger.
 
-## Mainnet
-
+## Non-Commercial
 | Operator  | [Network][] | JSON-RPC URL | WebSocket URL | Notes                |
 |:----------|:------------|:-------------|:--------------|:---------------------|
 | XRP Ledger Foundation | **Mainnet** | `https://xrplcluster.com/` <br> `https://xrpl.ws/` [²][] | `wss://xrplcluster.com/` <br>  `wss://xrpl.ws/` [²][] | Full history server cluster with CORS support. |
 | Ripple[¹][]   | **Mainnet** | `https://s1.ripple.com:51234/` | `wss://s1.ripple.com/` | General purpose server cluster |
 | Ripple[¹][]   | **Mainnet** | `https://s2.ripple.com:51234/` | `wss://s2.ripple.com/` | [Full-history server](../concepts/networks-and-servers/ledger-history.md#full-history) cluster |
+
+## Commercial
+| Operator  | [Network][] | JSON-RPC     | Notes                |
+|:----------|:------------|:-------------|:---------------------|
+| XRP Ledger Foundation full history paid API via [Dhali](https://dhali.io/) | **Mainnet** | `https://run.api.dhali.io/199fd80b-1776-4708-b1a1-4b2bb386435d/` | You must [create a paid API key](https://pay.dhali.io/?uuids=199fd80b-1776-4708-b1a1-4b2bb386435d) and embed it in the request's `Payment-Claim` header. |
+| [QuickNode](https://www.quicknode.com/chains/xrpl) | Testnet/Mainnet | N/A | QuickNode provides hosted XRPL RPC mainnet and testnet under their free and paid plans, granting flexible and reliable access to the network.
+
 
 ## Test Networks
 
@@ -30,11 +35,7 @@ If you don't [run your own `rippled` server](../infrastructure/installation/inde
 | Ripple[¹][]   | Sidechain-Devnet | `https://sidechain-net2.devnet.rippletest.net:51234/` | `wss://sidechain-net2.devnet.rippletest.net:51233/` | Sidechain Devnet to test cross-chain bridge features. Devnet serves as the locking chain while this sidechain serves as the issuing chain. |
 | XRPL Labs     | Xahau Testnet | `https://xahau-test.net/` | `wss://xahau-test.net/` | [Hooks-enabled](https://hooks.xrpl.org/) Xahau Testnet |
 
-## RPC Ecosystem Providers
 
-| Operator  | [Networks] | Notes |
-|:----------|:------------|:-------------|:--------------|:---------------------|
-| [QuickNode](https://www.quicknode.com/chains/xrpl) | Testnet/Mainnet | QuickNode provides hosted XRPL RPC mainnet and testnet under their free and paid plans, granting flexible and reliable access to the network. |
 
 
 [Network]: ../concepts/networks-and-servers/parallel-networks.md


mDuo13:
This would probably be better moved to the "RPC Ecosystem Providers" (perhaps "Commercial Providers" would be a better name?) table farther down the list. I'd rather not mix paid options into the table with the free, public servers.

desimmons:
> This would probably be better moved to the "RPC Ecosystem Providers" (perhaps "Commercial Providers" would be a better name?) table farther down the list. I'd rather not mix paid options into the table with the free, public servers.

Thanks for the response @mDuo13. I updated the PR so the servers are partitioned into three sections:
1. Non-commercial
2. Commercial
3. Testnet

Does this seem more appropriate to you?

docs/tutorials/public-servers.md
```diff
@@ -7,17 +7,22 @@ labels:
...
+## Non-Commercial
 | Operator  | [Network][] | JSON-RPC URL | WebSocket URL | Notes                |
 |:----------|:------------|:-------------|:--------------|:---------------------|
 | XRP Ledger Foundation | **Mainnet** | `https://xrplcluster.com/` <br> `https://xrpl.ws/` [²][] | `wss://xrplcluster.com/` <br>  `wss://xrpl.ws/` [²][] | Full history server cluster with CORS support. |
 | Ripple[¹][]   | **Mainnet** | `https://s1.ripple.com:51234/` | `wss://s1.ripple.com/` | General purpose server cluster |
 | Ripple[¹][]   | **Mainnet** | `https://s2.ripple.com:51234/` | `wss://s2.ripple.com/` | [Full-history server](../concepts/networks-and-servers/ledger-history.md#full-history) cluster |
 
+## Commercial (high rate)
+| Operator  | [Network][] | JSON-RPC URL | WebSocket URL | Notes                |
+|:----------|:------------|:-------------|:--------------|:---------------------|
```
mDuo13: I don't think "(high rate)" is necessary for the heading, and the column headers should change since URLs aren't provided in this table.

docs/tutorials/public-servers.md
```diff
@@ -7,17 +7,22 @@ labels:
...
+| XRP Ledger Foundation full history paid API via [Dhali](https://dhali.io/) | **Mainnet** | `https://run.api.dhali.io/199fd80b-1776-4708-b1a1-4b2bb386435d/` | N/A | You must [create a paid API key](https://pay.dhali.io/?uuids=199fd80b-1776-4708-b1a1-4b2bb386435d) and embed it in the request's `Payment-Claim` header. |
```
mDuo13: What is the `uuids` in this link all about? I'm skeptical about putting someone's referral link or something on the site. Also, the `N/A` column should go, to match the other row

mDuo13 (changes requested):

docs/tutorials/public-servers.md
desimmons: I removed "(high rate)" and "URL"

docs/tutorials/public-servers.md
desimmons: I removed the `N/A` column.

Dhali routes API requests to APIs using a uuid at the root of the path: `https://run.api.dhali.io/199fd80b-1776-4708-b1a1-4b2bb386435d/`.

XRPL Cluster's uuid is `199fd80b-1776-4708-b1a1-4b2bb386435d`.

The uuid in the query parameter `uuids` of `pay.dhali.io` is the same XRPL Cluster uuid from above. It allows `pay.dhali.io` to create a payment terminal customised for XRPL Cluster's API. Specifically, it customises:
1. the displayed example,
2. the displayed cost and latency of the API.

It has nothing to do with referral links.

docs/tutorials/public-servers.md
```diff
@@ -7,17 +7,22 @@ labels:
...
 | Ripple[¹][]   | **Mainnet** | `https://s1.ripple.com:51234/` | `wss://s1.ripple.com/` | General purpose server cluster |
 | Ripple[¹][]   | **Mainnet** | `https://s2.ripple.com:51234/` | `wss://s2.ripple.com/` | [Full-history server](../concepts/networks-and-servers/ledger-history.md#full-history) cluster |
 
+## Commercial
+| Operator  | [Network][] | JSON-RPC | Notes                |
+|:----------|:------------|:-------------|:---------------------|
+| XRP Ledger Foundation full history paid API via [Dhali](https://dhali.io/) | **Mainnet** | `https://run.api.dhali.io/199fd80b-1776-4708-b1a1-4b2bb386435d/` | You must [create a paid API key](https://pay.dhali.io/?uuids=199fd80b-1776-4708-b1a1-4b2bb386435d) and embed it in the request's `Payment-Claim` header. |
+| [QuickNode](https://www.quicknode.com/chains/xrpl) | Testnet/Mainnet | QuickNode provides hosted XRPL RPC mainnet and testnet under their free and paid plans, granting flexible and reliable access to the network. |  |
```
mDuo13:
```suggestion
| [QuickNode](https://www.quicknode.com/chains/xrpl) | Testnet/Mainnet | QuickNode provides hosted XRPL RPC mainnet and testnet under their free and paid plans, granting flexible and reliable access to the network. |
```

docs/tutorials/public-servers.md
```diff
@@ -7,17 +7,22 @@ labels:
...
-
+## Non-Commercial
 | Operator  | [Network][] | JSON-RPC URL | WebSocket URL | Notes                |
 |:----------|:------------|:-------------|:--------------|:---------------------|
 | XRP Ledger Foundation | **Mainnet** | `https://xrplcluster.com/` <br> `https://xrpl.ws/` [²][] | `wss://xrplcluster.com/` <br>  `wss://xrpl.ws/` [²][] | Full history server cluster with CORS support. |
 | Ripple[¹][]   | **Mainnet** | `https://s1.ripple.com:51234/` | `wss://s1.ripple.com/` | General purpose server cluster |
 | Ripple[¹][]   | **Mainnet** | `https://s2.ripple.com:51234/` | `wss://s2.ripple.com/` | [Full-history server](../concepts/networks-and-servers/ledger-history.md#full-history) cluster |
 
+## Commercial
+| Operator  | [Network][] | JSON-RPC | Notes                |
```
mDuo13:
```suggestion
| Operator  | [Network][] | Notes                |
```

mDuo13 (commented):
The columns of the table still don't seem quite right but I'm OK with the changes otherwise.

desimmons:
> The columns of the table still don't seem quite right but I'm OK with the changes otherwise.

I've aligned the column edges for the title row (first and second lines of the table). My markdown rendering displays it correctly.  Was that what you were referring to? 

I can do a more heavy handed alignment for the column edges across all the rows, but this arguably makes it more confusing to look at in raw text. The other tables in this page don't seem to take this heavier approach ... I don't think

desimmons:
@mDuo13 I've just squashed one of your suggested commits into a co-authored commit. Regarding the other suggested commit, I've kept the JSON-RPC column in the "Commercial" table because it contains an explicit reference to XRPL Clusters Dhali linked URL. The overall tables look like the attachment now. Let me know if there are any issues; I'll be happy to correct:
Screenshot from 2025-02-01 20-12-57

mDuo13 (approved):

PR #2966, opened 2025-01-28
Fix minor issues with Code Samples

mDuo13:
- Remov a "feature" where the description gets deleted if it matches the title exactly (due to how our plugin finds the text of the README).
    | Before | After |
    |---|---|
    | price oracles description starts with '. ' | price oracles code sample has the correct description |
- Changed the top level of the Code Sampels layout to be a `<main>` element, which allows some styles to apply correctly and also helps the search crawler find the contents of the page.
    - The text now (correctly) does not turn purple when a card is hovered.
- Changed a couple code sample descriptions that had weird casing or typos.

### DIFF
--- a/resources/code-samples.page.tsx
+++ b/resources/code-samples.page.tsx
@@ -29,7 +29,7 @@ export default function CodeSamples() {
   const { codeSamples, langs } = usePageSharedData<any>('code-samples');
 
   return (
-    <div className="landing page-community">
+    <main className="landing page-code-samples">
       <div className="">
         <section className="py-26">
           <div className="col-lg-8 mx-auto text-lg-center">
@@ -126,6 +126,6 @@ export default function CodeSamples() {
           </div>
         </section>
       </div>
-    </div>
+    </main>
   );
 }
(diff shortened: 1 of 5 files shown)


amarantha-k:
The description on the last card needs to be fixed. 
Screenshot 2025-01-29 at 4 18 01 PM

mDuo13:
Yep, that was another instance of the same problem. It turns out, the "feature" of removing text that matches the title was not even needed, so I removed it entirely. Also fixed another typo I spotted along the way.

resources/code-samples.page.tsx
```diff
@@ -29,7 +29,7 @@ export default function CodeSamples() {
   const { codeSamples, langs } = usePageSharedData<any>('code-samples');
 
   return (
-    <div className="landing page-community">
+    <main className="landing page-code-samples">
```
AKCodez, line 32: I cant find any references to this className in the codebase. Maybe we can remove this? Otherwise LGTM

AKCodez (approved):

resources/code-samples.page.tsx
AKCodez: image

mDuo13: It's not used now but it's available if we need it later. Putting it there (on the correct element) is future-proofing.

PR #2967, opened 2025-01-29
Add & use custom Markdoc component for "Try it!" and "Query example transaction" buttons

mDuo13:
Add a new Markdoc component for "Try it!" links, so they show up as a button again, and a related component for "Query example transaction" links. The component automatically provides a translatable label. The "Query example transactions" component uses APIv2 by default. Additionally:

- I updated the existing instances of these to use the new syntax in both English and Japanese.
- I added a couple new instances for methods and transactions that had examples available.
- I documented the new components on the Contribute Documentation page. I made a couple other minor formatting edits to that page and mentioned a couple of our other custom Markdoc components while I was there.
- I added the `manifest` and `version` commands to the WebSocket API tool using the examples from their respective documentation pages. I also changed the `ledger` method in the WS tool to use API v2.
- I switched the order of AccountSet and AccountDelete in the sidebar to make them correctly alphabetical like the other transaction types.


DennisDawson (approved):
LGTM

PR #2970, opened 2025-01-30
1839 - Document ignore_default field for account_lines API method

### LINKED ISSUE #1839: Undocumented `ignore_default` field in `account_lines`
[The doc](https://xrpl.org/account_lines.html#account_lines) does not describe the input field `ignore_default` (bool) that is implemented in rippled and clio.

maria-robobug:
Issue #1839

Adds missing _ignore_default_ request parameter.

### DIFF
--- a/docs/references/http-websocket-apis/public-api-methods/account-methods/account_lines.md
+++ b/docs/references/http-websocket-apis/public-api-methods/account-methods/account_lines.md
@@ -57,7 +57,7 @@ The request accepts the following parameters:
 | `Field`          | Type                 | Description    |
 |:-----------------|:---------------------|:---------------|
 | `account`        | String - [Address][] | Look up trust lines connected to this account. |
-| `ignore_default` | Boolean              | _(Optional)_ If true, suppress the output of incoming trust lines in the default state. Defaults to false. |
+| `ignore_default` | Boolean              | _(Optional)_ If `true`, suppress the output of incoming trust lines in the default state. The default is `false`. |
 | `ledger_hash`    | String               | _(Optional)_ A 20-byte hex string for the ledger version to use. (See [Specifying Ledgers][]) |
 | `ledger_index`   | Number or String     | _(Optional)_ The [ledger index][] of the ledger to use, or a shortcut string to choose a ledger automatically. (See [Specifying Ledgers][]) |
 | `peer`           | String - [Address][] | _(Optional)_ A second account; if provided, filter results to trust lines connecting the two accounts. |
--- a/docs/references/http-websocket-apis/public-api-methods/account-methods/account_lines.md
+++ b/docs/references/http-websocket-apis/public-api-methods/account-methods/account_lines.md
@@ -7,7 +7,8 @@ labels:
   - Tokens
 ---
 # account_lines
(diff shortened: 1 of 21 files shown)


docs/references/http-websocket-apis/public-api-methods/account-methods/account_lines.md
```diff
@@ -53,14 +54,15 @@ rippled account_lines r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59
...
-| `peer`         | String - [Address][] | _(Optional)_ A second account; if provided, filter results to trust lines connecting the two accounts. |
-| `limit`        | Number               | _(Optional)_ Limit the number of trust lines to retrieve. The server may return less than the specified limit, even if there are more pages of results. Must be within the inclusive range 10 to 400.  Positive values outside this range are replaced with the closest valid option. The default is 200. |
-| `marker`       | [Marker][]           | _(Optional)_ Value from a previous paginated response. Resume retrieving data where that response left off. |
+| `Field`          | Type                 | Description    |
+|:-----------------|:---------------------|:---------------|
+| `account`        | String - [Address][] | Look up trust lines connected to this account. |
+| `ignore_default` | Boolean              | _(Optional)_ If true, suppress the output of incoming trust lines in the default state. Defaults to false. |
```
mDuo13: Nit: slight formatting/wording edits.

```suggestion
| `ignore_default` | Boolean              | _(Optional)_ If `true`, suppress the output of incoming trust lines in the default state. The default is `false`. |
```

mDuo13 (dismissed):
I have some questions about this parameter. It is my (strongly held) understanding that a trust line that is _truly_ in its default state is automatically deleted. So this parameter must do something slightly different than that. Perhaps it omits trust lines that have where only the perspective account's settings are in the default state—and probably a zero balance as well?

Otherwise, seems fine.

maria-robobug:
Yeah, I think in this context, "default state" means trust lines where the perspective account has not modified the trust limit and has not applied any custom flags. I don’t think balance is considered in the filtering though.

https://github.com/XRPLF/rippled/blob/develop/src/xrpld/rpc/handlers/AccountLines.cpp#L211-L234

mDuo13:
> Yeah, I think in this context, "default state" means trust lines where the perspective account has not modified the trust limit and has not applied any custom flags. I don’t think balance is considered in the filtering though.
> 
> https://github.com/XRPLF/rippled/blob/develop/src/xrpld/rpc/handlers/AccountLines.cpp#L211-L234

Based on that I think balance _is_ considered because it checks who owes a reserve. The account with a positive (nonzero) balance owes the reserve.

mDuo13 (dismissed):
Should be good after the fixes I applied:

- Clarify that it's only trust lines where the perspective account's side is in the default state.
- Update syntax of parameter tables:
    - "Field" head should not be code
    - Add explicit "Required?" column instead of using italic parentheses in some Description columns
    - Add links to definitions for some of the data types and re-word some of the descriptions, accordingly.
    - Move response examples to included JSON files.

mDuo13:
Rebased to fix merge conflicts.

Also, added a commit that removes an old typo (which had been copy-pasted to many files) regarding the number of bytes in a ledger hash. (The correct number is 32 bytes, not 20.)

amarantha-k (approved):
lgtm

PR #2971, opened 2025-01-30
Add ledgerstatefix txn doc

amarantha-k:


DennisDawson (dismissed):
LGTM.

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
...
+[[Source]](https://github.com/XRPLF/rippled/blob/master/src/xrpld/app/tx/detail/LedgerStateFix.cpp "Source")
+
+
+`LedgerStateFix` is a general purpose transaction used to fix specific issues affecting the XRP ledger. You submit the transaction with the `LedgerFixType` value set to indicate the particular  error state to correct.
+
+## Example {% $frontmatter.seo.title %} JSON
+
+```json
+{
+   "Account" : "<Signer and fee payer>",
```
mDuo13: I would like to have an example with real (or at least, realistic) values for this and other fields in this code block, for consistency.

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
...
+| `LedgerFixType` | uint16 | Required | Currently the only type is _1_, which fixes the NFToken directory for a single account. |
+| `Owner` | STAccount | Optional | Required if `LedgerFixType` == _1_, the account ID that owns the NFToken directory that needs fixing. Need not have any relationship to Account. |
+
+## LedgerStateFix Flags
+
+Transactions of the LedgerStateFix type can support additional values in the `Flags` field. Currently, there are no flags defined. A future `LedgerFixType` might require flag settings.
+
+## Error Cases
+
+Potential errors are those that can occur for all transactions. {% $frontmatter.seo.title %}.
```
mDuo13: Stray variable here.

```suggestion
Potential errors are those that can occur for all transactions.
```

Also I suspect there's at least one error type that doesn't occur for all transactions (if you try to fix something that isn't broken)

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
...
+   "TxnSignature" : "<Signature>",
+}
+```
+
+{% raw-partial file="/docs/_snippets/tx-fields-intro.md" /%}
+
+| Field | Data Type  | Required? | Description |
+|:------|:-----------|:----------|:------------|
+| `TransactionType` | uint16 | Required | Identifies this as a `LedgerStateFix` transaction. |
+| `Account` | STAccount | Required | Identifies the account signing and submitting the transaction as well as paying the Fee. |
```
mDuo13: This is a common field and doesn't need to be listed in this table. Neither does `Fee`, nor `Flags`.

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
...
+- The very last page of the directory contained only one NFToken.
+- The transaction removed the last remaining token from the last page.
+
+When these conditions were met, the last NFToken page was removed and the next-to-last page was left as the final page in the directory.
+
+That would be fine, except the NFToken directory has an expectation that the last page has a specific index. The page with that index was just deleted. When an NFToken is added to the directory, and that token has a high enough value that it doesn't belong on the current last page, then a new last page is created that has no links to the previous page, creating a hole in the middle of the list.
+
+The `fixNFTokenPageLinks` amendment modifies the NFToken page, coalescing code to notice when the very last page of the directory would be removed. In that case, it moves all of the contents of the next lower page into the last page and deletes the next-to-last page. It then fixes up the links.
+
+New invariant checks also validate aspects of the links on pages, so a similar corruption returns a tecINVARIANT_FAILED transaction result. That will prevent this specific type of corruption going forward.
```
mDuo13: Some of these descriptions seem out of place in the transaction reference (it sounds more or less copied from the PR that introduced the invariant checks along with the transaction type)

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
...
+seo:
+    description: Repair corruptions to the XRP ledger.
+labels:
+  - Utilities, troubleshooting
+---
+# LedgerStateFix
+[[Source]](https://github.com/XRPLF/rippled/blob/master/src/xrpld/app/tx/detail/LedgerStateFix.cpp "Source")
+
+
+`LedgerStateFix` is a general purpose transaction used to fix specific issues affecting the XRP ledger. You submit the transaction with the `LedgerFixType` value set to indicate the particular  error state to correct.
```
mDuo13, line 11: Isn't this transaction gated by an amendment? Which one?

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
+---
+seo:
+    description: Repair corruptions to the XRP ledger.
+labels:
+  - Utilities, troubleshooting
```
mDuo13: Multiple labels are supposed to be on separate lines (yaml list). You can't comma-separate them here.

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
...
+   "TransactionType" : "LedgerStateFix",
+   "TxnSignature" : "<Signature>",
+}
+```
+
+{% raw-partial file="/docs/_snippets/tx-fields-intro.md" /%}
+
+| Field | Data Type  | Required? | Description |
+|:------|:-----------|:----------|:------------|
+| `TransactionType` | uint16 | Required | Identifies this as a `LedgerStateFix` transaction. |
```
mDuo13: Nit: casing

```suggestion
| `TransactionType` | UInt16 | Required | Identifies this as a `LedgerStateFix` transaction. |
```

mDuo13 (changes requested):

docs/references/protocol/transactions/types/ledgerstatefix.md
amarantha-k: I couldn't find it, could you help track it down?

docs/references/protocol/transactions/types/ledgerstatefix.md
amarantha-k: Done.

docs/references/protocol/transactions/types/ledgerstatefix.md
oeggert: Added link to amendment.

docs/references/protocol/transactions/types/ledgerstatefix.md
oeggert: I updated the language to try and match a reference doc.

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
...
+seo:
+    description: Repair corruptions to the XRP ledger.
+labels:
+  - Utilities
+  - Troubleshooting
+---
+# LedgerStateFix
+[[Source]](https://github.com/XRPLF/rippled/blob/master/src/xrpld/app/tx/detail/LedgerStateFix.cpp "Source")
+
+_(Added by the [fixNFTokenPageLinks amendment][])_
```
mDuo13, line 13: "Added by..." is the text we use for an amendment that has already been enabled. For amendments that are still open for voting, instead:

```suggestion
_(Requires the [fixNFTokenPageLinks amendment][] {% not-enabled /%})_
```

Also, I have been moving to having this line about the amendment added at the _end_ of the introductory section, to improve SEO because it'll show the actually useful description in search results instead of the (less important) info about what amendment added something.

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
...
+
+| Field | Data Type  | Required? | Description |
+|:------|:-----------|:----------|:------------|
+| `LedgerFixType` | uint16 | Required | Currently the only type is _1_, which fixes the NFToken directory for a single account. |
+| `Owner` | STAccount | Optional | Required if `LedgerFixType` == _1_, the account ID that owns the NFToken directory that needs fixing. Need not have any relationship to Account. |
+
+
+## LedgerFixType
+
+`LedgerStateFix` transactions are targeted solutions for rare and specific issues. They can't be used to fix ledger issues outside of the described types below:
```
mDuo13: Nit: it feels misleading to say "types below" when there's only one. Also, I prefer to word this in the positive rather than negative sense, and I feel the word "corruption" is more accurate and less confusing than "issues" (which is a word that has also been used to refer to a type of issued token).

```suggestion
`LedgerStateFix` transactions are targeted solutions for rare and specific issues. They can only be used to fix a specific type of ledger corruption, described below.
```

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
...
+Corrupt NFT directories resulting from these conditions:
+
+- At least two NFToken pages were in the directory.
+- The next-to-last page was completely full, holding 32 NFTokens.
+- The last page of the directory contained only one NFToken.
+- A transaction removed the last remaining token from the last page, causing the directory to delete the page.
+
+When these conditions were met, the NFToken directory didn't properly update page links, causing holes in the directory when new last pages were created for additional NFTokens.
+
+The [`fixNFTokenPageLinks` amendment][] fixed new instances of this ledger issue by adding invariant checks.
```
mDuo13: Nit: technically, the amendment fixes the code that would create them, _and_ adds an invariant check. An invariant check on its own would fail the transaction with no changes. The amendment changes the normal transaction processing code and also adds an invariant check as an extra layer of protection.

More importantly: Technically, past tense here is wrong because fixNFTokenPageLinks is not enabled. Also, don't use backticks inside a reference-style Markdown link because [there are longstanding bugs with that](https://github.com/Python-Markdown/markdown/issues/495) and it also doesn't match how the reference link is defined.

```suggestion
The [fixNFTokenPageLinks amendment][] {% not-enabled /%} prevents new instances of this type of ledger corruption from happening.
```

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
...
+   "TransactionType" : "LedgerStateFix"
+}
+```
+
+{% raw-partial file="/docs/_snippets/tx-fields-intro.md" /%}
+
+| Field | Data Type  | Required? | Description |
+|:------|:-----------|:----------|:------------|
+| `LedgerFixType` | uint16 | Required | Currently the only type is _1_, which fixes the NFToken directory for a single account. |
+| `Owner` | STAccount | Optional | Required if `LedgerFixType` == _1_, the account ID that owns the NFToken directory that needs fixing. Need not have any relationship to Account. |
```
mDuo13: This table still doesn't match the standard format for transaction reference tables:

- It should have both JSON Type and Internal Type columns
- The contents of the "Required?" column are conventionally written as "Yes" or "No"
- `uint16` should be written as `UInt16`

```suggestion
| Field         | JSON Type | [Internal Type][] | Required? | Description |
|:--------------|:----------|:------------------|:----------|:------------|
| `LedgerFixType` | Number | UInt16 | Yes | The type of fix to apply. See [LedgerFixType](#ledgerfixtype) for possible values. Currently the only type is `1`, which fixes the NFToken directory for a single account. |
| `Owner` | String - [Address][] | Account | No | _(Required if `LedgerFixType` is `1`.)_ The account that owns the NFToken directory to fix. Does not need any relationship to the sender of the transaction. |
```

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
...
+
+When these conditions were met, the NFToken directory didn't properly update page links, causing holes in the directory when new last pages were created for additional NFTokens.
+
+The [`fixNFTokenPageLinks` amendment][] fixed new instances of this ledger issue by adding invariant checks.
+
+
+## LedgerStateFix Flags
+
+Transactions of the `LedgerStateFix` type can support additional values in the `Flags` field. Currently, there are no flags defined. A future `LedgerFixType` might require flag settings.
+
```
mDuo13: This section is unnecessary future-proofing. For consistency with other transaction reference pages, we can simply omit the entire section since there are no flags (currently) defined.

```suggestion
```

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Special Transaction Cost
+
+The `LedgerStateFix` transaction is rare and potentially compute intensive. The minimum fee is the same as the fee for an [`AccountDelete`][] transaction. If the transaction fails with a tec code, the fee is still charged.
```
mDuo13: Instead of obliquely referring to the cost of AccountDelete, I suggest copying some language from that page to explain it with less indirection:

```suggestion
The `LedgerStateFix` transaction is rare and potentially compute intensive, so the transaction must pay a special [transaction cost][] equal to at least the [owner reserve](../../../../concepts/accounts/reserves.md) for one item (currently {% $env.PUBLIC_OWNER_RESERVE %}).

The transaction cost always applies when a transaction is included in a validated ledger, even if the transaction fails. (See [Error Cases](#error-cases).) To greatly reduce the chances of paying the high transaction cost if the transaction fails, [submit the transaction](../../../http-websocket-apis/public-api-methods/transaction-methods/submit.md) with `fail_hard` enabled.
```

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+## Error Cases
+
+Potential errors are those that can occur for all transactions.
```
mDuo13: There are more error cases that are specific to this transaction type. If you look at the [source code link](https://github.com/XRPLF/rippled/blob/master/src/xrpld/app/tx/detail/LedgerStateFix.cpp) for this transactor, you'll see some cases that `return` various other codes. The ones I found are:

```suggestion
Besides errors that can occur for all transactions, {% $frontmatter.seo.title %} transactions can result in the following [transaction result codes](../transaction-results/index.md):

| Error Code                   | Description |
|:-----------------------------|:------------|
| `tecFAILED_PROCESSING`       | The transaction failed to apply the fix. For example, the transaction attempted to repair an NFT directory that was not broken. |
| `tecOBJECT_NOT_FOUND`        | A ledger entry specified in the transaction does not exist. For example, the transaction tried to repair the NFT directory of an account that does not hold any NFTs. |
| `tefINVALID_LEDGER_FIX_TYPE` | The [`LedgerFixType`](#ledgerfixtype) value specified in the transaction is not valid. |
```

docs/references/protocol/transactions/types/ledgerstatefix.md
```diff
@@ -0,0 +1,68 @@
+---
+seo:
+    description: Repair corruptions to the XRP ledger.
+labels:
+  - Utilities
+  - Troubleshooting
+---
```
mDuo13, line 7: This won't do anything _yet_, but I hope to extend Redocly's sidebar so that we can use this frontmatter field to make the "not enabled" flask icon appear in the sidebar again:

```suggestion
---
seo:
    description: Repair corruptions to the XRP ledger.
labels:
  - Utilities
  - Troubleshooting
status: not_enabled
---
```

mDuo13 (changes requested):
I noticed a few more inaccuracies and style inconsistencies that should still be fixed, but this is a lot closer.

docs/references/protocol/transactions/types/ledgerstatefix.md
oeggert: Removed backticks, but keeping present tense since amendement is now live.

mDuo13 (approved):
I'll have to separately look into why the page that lists all transaction types only shows descriptions for a few (MPT and Oracle) transaction types, even though it should be pulling from the frontmatter of all the transaction type pages including this one. But this PR looks good now.

PR #2977, opened 2025-01-31
Upgrade browser wallet (js) deps

mDuo13:
Upgrades xrpl.js and Vite dependencies in the "Build a Browser Wallet in JavaScript" code sample. This was a bit more involved than just bumping version dependencies.

- [x] Changed some usage to match API v2 formats used in xrpl.js 4.x
- [x] Improved error handling of the sample wallet (now it alerts you if you forgot to edit the `.env` file)
- [x] Removed the special configs & dependencies that used to be necessary for using xrpl.js in Vite. (It works out of the box now.)
- [x] Split some longer lines into two to make the code display nicer in the dev site's code blocks.

This PR supersedes #2912 and #2961.

### DIFF
--- a/_code-samples/build-a-browser-wallet/js/index.js
+++ b/_code-samples/build-a-browser-wallet/js/index.js
@@ -36,21 +36,25 @@ txHistoryButton.addEventListener('click', () => {
         });
 
         // Fetch the wallet details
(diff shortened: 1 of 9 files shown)


intelliot (approved):
diff LGTM. i did not run the code

PR #2979, opened 2025-02-03
Improve home page performance

AKCodez:
- Fix Cumulative Layout Shifts (CLS)
- Add lazy loading to images


AKCodez:
Current Site homepage Performance:
image


Vs with improvements:
image

AKCodez:
Need to adjust image size

mDuo13:
I'm not sure how that tool is measuring cumulative layout shift, but it definitely can still shift during loading due to the font loading, depending on your window width. See how it changes how many lines "The Blockchain Built for Business" takes up during page loading, in the following video:

https://github.com/user-attachments/assets/cd57bd61-b991-4899-8e66-ae517fb500bd

I'm less sure of the following, but I suspect that the layout might also change before and after loading the CSS file. A more foolproof way to avoid layout shifts would be to [set the dimensions on the image element itself](https://www.aleksandrhovhannisyan.com/blog/setting-width-and-height-on-images/).

AKCodez:
> I'm not sure how that tool is measuring cumulative layout shift, but it definitely can still shift during loading due to the font loading, depending on your window width. See how it changes how many lines "The Blockchain Built for Business" takes up during page loading, in the following video:
> 
>  cumulative-layout-shift.mp4 
> I'm less sure of the following, but I suspect that the layout might also change before and after loading the CSS file. A more foolproof way to avoid layout shifts would be to [set the dimensions on the image element itself](https://www.aleksandrhovhannisyan.com/blog/setting-width-and-height-on-images/).

The PR does include adding of min-height across 3 different breakpoints to the main hero image that causes the layout shift, (there will still be minor shifts in SOME screen sizes as the image size is dynamic based off windows size but it is a massive improvement none the less)  you can test layout shifts through the performance tab in your chrome dev tools.

mDuo13 (approved):
Agreed it is at least a big improvement over the status quo and it doesn't (seem to) break anything.

PR #2995, opened 2025-02-11
Update Devnet validator list key across docs

intelliot:
These pages still appear in search results and are used by Devnet node operators


amarantha-k (approved):
lgtm

PR #3001, opened 2025-02-18
Apex banner 2025 v2

AKCodez:


amarantha-k (changes requested):
The banner overlaps the breadcrumbs and right-nav on docs pages.

AKCodez:
> The banner overlaps the breadcrumbs and right-nav on docs pages.

Great Catch! Latest commit solves this issue by adding a banner-height variable and utilizing that + the nav height to formulate the proper margin needed for the layout.

amarantha-k (approved):
lgtm

PR #3016, opened 2025-02-25
Update index.md

dimondevceo:
Small typo in index.md

In the Python documentation, it says "Send XRP and issued currency on the XRP Ledger using JavaScript.", but instead should be "Send XRP and issued currency on the XRP Ledger using Python."


amarantha-k (approved):
Thanks for catching and fixing it!

PR #3025, opened 2025-03-05
Update XRPLF UNL migration comms

xVet:
March 10th is upcoming which was set as the migration date. A new migration plan with detailed description and comms campaign seems to be in the works. 

To ensure operators are served with date and urgency neutral comms i updated the blog post. Subject to changes once the plans above are ready to announce.


sappenin (approved):
LGTM

amarantha-k (approved):
Thanks for the updates! LGTM

PR #3026, opened 2025-03-05
add deep freeze topic

DennisDawson:


docs/concepts/tokens/fungible-tokens/deep-freeze.md
oeggert, line 1: I think the conceptual content on this can just be added here: https://xrpl.org/docs/concepts/tokens/fungible-tokens/freezes

Also the ripplestate and trustset flags probably don't belong on a concept topic if they'r ealready documented in their respective reference docs.

oeggert (approved):
Approving for merge for the sake of speed, but I think a separate concept topic is unnecessary since we already have a Freezing Tokens page we can add that info to.

Issue #2957, opened 2025-01-23
Remove `rippled` Reporting Mode


intelliot:
Reporting Mode was deprecated and removed in `rippled` version 2.3.0, released Nov 25, 2024.

- https://github.com/XRPLF/rippled/pull/5092

PR #3028, opened 2025-03-07
specify information about network_ledger field in server_info response

### LINKED ISSUE #1145: server_info missing information about network_ledger
Link to relevant rippled code: https://github.com/ripple/rippled/blob/aee422e8197b9c827587f2a8f009decf1a9f2ac0/src/ripple/app/misc/NetworkOPs.cpp#L2214

It looks like this field exists when the server is waiting for a validated ledger.

ckeshava:
Fix https://github.com/XRPLF/xrpl-dev-portal/issues/1145


mDuo13:
Does this field also exist in `server_state` responses?

ckeshava:
> Does this field also exist in `server_state` responses?

I don't see this field in my empirical observations with the WebSocket tool (or) with my local rippled instance. In the code, this field is set only for `server_info` RPC command. 

However, I haven't done extensive tests.

mDuo13:
Based on my research, the code that emits this field is shared between the `server_state` and `server_info` API methods, so they should both use it. However, I haven't been able to actually get the server to output the field (from either method) in any of my testing, so I'm not sure what actually causes it to occur.

mDuo13:
I got more information—the field is currently only provided when the server is started with the `--net` parameter. (There may also be some surprising behavior when using `--net`, according to @legleux, but that's a different story.) And I was able to confirm that it exists in both `server_state` and `server_info` if it is.

I added this detail, put the field in alphabetical order, and removed some fields that were only returned by reporting mode (see: #2957) while I was at it.

mDuo13:
Closing in favor of #3252.

PR #3033, opened 2025-03-12
add logos

AKCodez:


docs/use-cases/tokenization/real-world-assets.page.tsx
```diff
@@ -56,23 +57,29 @@ const companies = [
...
+    link: "https://ondo.finance/",
   },
   {
     className: "ripple-logo",
     alt: "Ripple",
     aspectRatio: 3.89,
     link: "https://ripple.com/rwa-tokenization",
   },
+  {
+    className: "meld",
```
mDuo13, line 90: What is the logic behind the order of the logos? They seemed to be in kind of a random order before and this changes them to be in a different random order?

AKCodez: Following the same order as ripple.com/rwa-tokenization

mDuo13: What order are those in?

mDuo13 (approved):
Looks good now. I'll want to squash to keep the history cleaner (I think that should clean up the adding and removing of the extraneous FAQ page)

PR #3036, opened 2025-03-19
Fix dead link in rate-limiting.md

lmaisons:
The source link for the 'Rate Per Request' section links to a non-extant file. Update to nearest contemporaneous equivalent.


mDuo13 (approved):
I think the nearest equivalent is actually https://github.com/XRPLF/rippled/blob/master/include/xrpl/resource/Fees.h but this one might be more useful. :+1:

PR #3049, opened 2025-03-25
Update dependencies & fix some bugs in JS code

mDuo13:
- [x] Updates Realm to v0.120.2 which contains a number of fixes including some security issues among dependencies. It was supposed to contain a fix for interactive tutorials, but it turns out that fix was incomplete. Redocly has confirmed they've reproduced the bug and are working on a fix. But for now we can update to satisfy dependabot.
- [x] Updated the "Issue a Fungible Token" tutorial to match the updated format for the Testnet faucet's response format. (The faucet no longer tells you your current balance, just the amount that it sent you.) The "Issue a Fungible Token" interactive tutorial works in local dev mode now if you navigate to it from another page, but (due to the aforementioned Redocly bug) not if you load it directly.
- [x] Updates the Transaction Sender dev tool for the current version of xrpl.js which uses the rippled API v2 by default. Prior to this fix, the tool would fail to send an EscrowFinish transaction because it was looking in the old place for the `Sequence` field.

### DIFF
--- a/resources/dev-tools/tx-sender.page.tsx
+++ b/resources/dev-tools/tx-sender.page.tsx
@@ -77,7 +77,7 @@ async function onClickCreateEscrow(
         }
         setEscrowWidthPercent(0)
 
-        if(escrowCreateResponse.result.Sequence === undefined) {
+        if(escrowCreateResponse.result.tx_json.Sequence === undefined) {
             
             errorNotif(submitConstData.alert, 
                 "Error: Unable to get the sequence number from EscrowCreate, so cannot submit an EscrowFinish transaction.")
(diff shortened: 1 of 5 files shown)


mDuo13:
Rebased on master and added a couple more fixes and updates. Even though the interactive tutorials bug isn't fixed, I believe we should push this sooner to incorporate other updates.

resources/dev-tools/tx-sender.page.tsx
```diff
@@ -77,7 +77,7 @@ async function onClickCreateEscrow(
         }
         setEscrowWidthPercent(0)
 
-        if(escrowCreateResponse.result.Sequence === undefined) {
+        if(escrowCreateResponse.result.tx_json.Sequence === undefined) {
```
intelliot, line 80: relatedly - maybe we can link to https://github.com/xrplf/rippled/blob/4ba9288/API-CHANGELOG.md#modifications-to-json-transaction-element-in-v2 somewhere? This is where we "documented" the change

intelliot (approved):

PR #3053, opened 2025-03-26
Update stablecoin-issuer.md

coralexbadea:


mDuo13 (commented):
This seems like a distinction that doesn't make things clearer. I'm inclined to leave it as-is unless you can provide a stronger case for the change.

PR #3062, opened 2025-04-01
Fix #2858

### LINKED ISSUE #2858: Invalid type code for Binary Format page
The type code of the Paths Field is 18(`0b00010010`), but the binary value of the image represents 19 (`0b00010011`).

スクリーンショット 2024-11-21 13 39 05

mDuo13:
Fixes a typo in a diagram where the binary representation was wrong (should be 18, but the binary said 19)

As a note, `.uxf` files are Umlet source files for diagrams. In this case, one file contains all 4 diagrams, exported in parts as PNG (not as SVG because inlining them and making them theme-aware doesn't cleanly work _inside_ the table they're in).


DennisDawson (approved):
LGTM

PR #3074, opened 2025-04-09
Remove axios dependency

mDuo13:
Replaces #3038.

Instead of bumping the axios dependency, this small refactor removes the dependency on axios entirely in favor of using the browser's built-in `fetch()` API. Since axios was only used for this one type of call in this one place, it was overkill.


DennisDawson (approved):
LGTM

amarantha-k (approved):
lgtm

intelliot (approved):

PR #3082, opened 2025-04-14
Missed change to MPT doc, add error type to clawback.

DennisDawson:
Addresses https://github.com/XRPLF/xrpl-dev-portal/issues/3048


amarantha-k (approved):
lgtm

PR #3088, opened 2025-04-22
Update xrpl JS package version

amarantha-k:


mDuo13 (approved):

amarantha-k:
NOTE: The fs 0.0.1-security release is just a placeholder package on NPM to protect against supply chain attacks. Since Since node.js provide fs natively, we should not need to import one from npm. We'll create a separate PR to update the package.json for the tx-serialization code sample.

PR #3089, opened 2025-04-23
Update/fix deep freeze at the known amendments page

vvysokikh1:
Fixed the name of the amendment and status


mDuo13 (approved):
Good catch.

PR #3091, opened 2025-04-24
Revise payment tutorials, add/remove screenshots

DennisDawson:
refactored code, new CSS, only one active account at a time, but radio button allows switching between two accounts in the active fields.


_code-samples/modular-tutorials/base-module.html
oeggert, line 1: Can we also add a button to fund an existing account with additional XRP?

_code-samples/modular-tutorials/account-support.js
```diff
@@ -0,0 +1,158 @@
...
+
+// *******************************************************
+// ************* Get Account *****************************
+// *******************************************************
+
+async function getAccount() {
+  let net = getNet()
+  const client = new xrpl.Client(net)
+  await client.connect()
+  let results = `\nConnected to ${net}.`
```
oeggert:
```suggestion
  let results += `===Getting Account===\nConnected to ${net}.`
```

This suggestion applies to all instances that add to the result field. I think it's worth keeping a log of all actions that take place, so ensuring all results are `+=` and not overridden with `=`. It'd also like to see a more pronounced headers to identify the type of function that's being called to help differentiate info in the logs. So for an offer create transaction, it starts with something like `===Creating Offer===`  and then has a line break for subsequent metadata.

_code-samples/modular-tutorials/create-offer.js
```diff
@@ -0,0 +1,125 @@
...
+  let offers
+  try {
+    const offers = await client.request({
+      method: "account_offers",
+      account: wallet.address,
+      ledger_index: "validated"
+    })
+    results += JSON.stringify(offers,null,2)
+  } catch (err) {
+      results += err
```
oeggert:
```suggestion
  } catch (err) {
      resultField.value += `${err.message}`
```

This suggestion applies to all functions called by the test harness. Current behavior will just kind of lock up and do nothing if the user adds in a bad field value or is missing something. While it can be a pain to code in proper messaging for all error types, the basic returned error is at least something that gives the user feedback they did a step wrong. I think all functions should include the try, catch block and spit out the error message if there is one.

oeggert (changes requested):
PR looks pretty good, but there are a couple points I have that apply across all functions.

_code-samples/modular-tutorials/account-support.js
```diff
@@ -0,0 +1,158 @@
...
+
+function distributeAccountInfo() {
+  let accountInfo = resultField.value.split("\n")
+  account1name.value = accountInfo[0]
+  account1address.value = accountInfo[1]
+  account1seed.value = accountInfo[2]
+  account2name.value = accountInfo[3]
+  account2address.value = accountInfo[4]
+  account2seed.value = accountInfo[5]
+}
```
maria-robobug, line 123: I'm seeing some weird behaviour in the `send-mpt.html` example when I click "Distribute Account Info" (which calls this function).

Screenshot 2025-04-25 at 12 05 57

It looks like the fields are not populated correctly.

I printed the contents of the `accountInfo` variable in the console to help see what's in it. Seems that we don't have a value for **name**, and **seed** is populated with the wrong value?

docs/tutorials/javascript/send-payments/send-and-cash-checks.md
```diff
@@ -86,20 +90,19 @@ To cash a check you have received:
 1. Enter the **Check ID** (**index** value).
 2. Enter the **Amount** you want to collect, up to the full amount of the check.
 3. Enter the currency code.
-   a. If you cashing a check for XRP, enter _XRP_ in the **Currency** field.
+   a. If you cashing a check for XRP, enter _XRP_ in the **Currency Code** field.
```
maria-robobug: The `a` and `b` points don't seem to render correctly.

Maybe just split like this:

```
    - If you are cashing a check for XRP, enter _XRP_ in the **Currency Code** field.
    - If you are cashing a check for an issued currency token:
	    1. Enter the **Issuer** of the token.
	    2. Enter the **Currency Code** code for the token.
```

docs/tutorials/javascript/send-payments/create-trust-line-send-currency.md
```diff
@@ -6,805 +6,447 @@ seo:
...
 # Create Trust Line and Send Currency Using JavaScript
 
 This example shows how to:
 
-1. Configure accounts to allow transfer of funds to third party accounts.
-2. Set a currency type for transactions.
-3. Create a trust line between the standby account and the operational account.
-4. Send issued currency between accounts.
-5. Display account balances for all currencies.
+1. Create a trust line between the standby account and the operational account.
```
maria-robobug: The UI no longer mentions Standby and Operational accounts, we should update this in the steps.

maria-robobug:
Question about how we reference code snippets, should we not use this format now?

```
{% code-snippet file="/_code-samples/modular-tutorials/send-checks.js" language="js" from="async function sendCheck() {" before="let check_amount = amountField.value" /%}
```

_code-samples/modular-tutorials/account-support.js
```diff
@@ -0,0 +1,158 @@
...
+    command: "gateway_balances",
+    account: wallet.address,
+    ledger_index: "validated",
+  })
+  results += JSON.stringify(balance.result, null, 2)
+  resultField.value = results
+  xrpBalanceField.value = (await client.getXrpBalance(wallet.address))
+  client.disconnect()
+} // End of getTokenBalance()
+
```
oeggert, line 208:
```suggestion
// *******************************************************
// **************** Scroll Results ***********************
// *******************************************************
let results

async function updateResults() {
  resultField.value += results;
  resultField.scrollTop = resultField.scrollHeight;
}

// End of updateResults()
```

This creates a function to help accommodate adding additional logs and scrolling the results log down to the latest. If implemented, later code would look something like this:
```
  // Prepare transaction
  const transaction = {
    "TransactionType": "CredentialCreate",
    "Account": wallet.address,
    "Subject": subject,
    "CredentialType": credentialType
  }
  
  results = `\n\n===Preparing and Sending Transaction===\n\n${JSON.stringify(transaction, null, 2)}`
  updateResults()
  ```

_code-samples/modular-tutorials/account-support.js
DennisDawson: The instructions are pretty clear in the doc. You paste the gathered info, cut and paste the MPT Issuance ID, click Distribute Account Info, then **Get New Account 2** or **Get Account 2 From Seed**. It's a little convenience tool to be able to gather and distribute info - if people choose not to use it, that's okay. But making it robust enough to distribute info in a number of formats is overkill for the tutorials.

_code-samples/modular-tutorials/base-module.html
DennisDawson: I don't know of an easy way to do that using the faucet. If I want to add XRP to an account, I can just create another account and send some XRP. Accounts that run out of XRP should try sending smaller amounts when testing.

_code-samples/modular-tutorials/account-support.js
DennisDawson: I made these changes.

_code-samples/modular-tutorials/create-offer.js
DennisDawson: I made these changes.

_code-samples/modular-tutorials/account-support.js
DennisDawson: This is a bridge too far at the moment. We can discuss this as a future enhancement.

oeggert (dismissed):
Changes look good to me.

maria-robobug (dismissed):

maria-robobug (approved):

PR #3096, opened 2025-04-28
Fix websocket-api-tool

maria-robobug:
The Websocket API tool seems broken on production: https://xrpl.org/resources/dev-tools/websocket-api-tool

Seems related to the `@codemirror/state` - currently version 6.4.1

I've updated the dependency to the latest, `6.5.2`


amarantha-k (approved):
LGTM

PR #3101, opened 2025-04-29
Invalid hot wallet error

maria-robobug:
Fixes issue https://github.com/orgs/XRPLF/projects/4/views/1?pane=issue&itemId=95298179&issue=XRPLF%7Cxrpl-dev-portal%7C2949

I've added an admonition to note that Clio server returns different errors for API v1 and v2 when the `hotwallet` field is invalid.


amarantha-k (approved):
lgtm

PR #3104, opened 2025-05-02
Clean up mpt_holders reference documentation + add Try It

### LINKED ISSUE #3017: mpt_holders documentation needs some cleanup
These issues with the [mpt_holders reference page](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/clio-methods/mpt_holders):

1. It should use `{% tabs %}` components to put the request/response format examples in tabs.
2. The page has some examples of `tx` and `account_objects` responses. I'm not sure why there are responses from other API methods in the reference page for this method. Maybe they're copy-paste errors or maybe there's some additional context missing.
3. The "Type" column for the fields doesn't use title case like our other reference pages do. It also doesn't link to common type definitions (e.g. markers and pagination, string number, ledger index, etc.)
4. Saying the `mpt_amount` is "base 10-encoded" is confusing. It's just a string number. (It's base-10 because that's, like, a normal number as opposed to hexadecimal or something. But making it sound like base58 decoding is probably going to more confusing than helpful.)
5. There is some inconsistent use of code font in instances of "`MPToken` object" or `rippled`.
6. The other Clio methods all have a badge that tells you what Clio version they were added in. This one is probably newer than many of those but it doesn't have a badge. This also makes it easier to miss that it's a Clio-specific method. (It _is_ in the "Clio Methods" category, but I certainly missed that on first read. Maybe that's just because I was coming from #2925 instead of finding the page more organically.)
7. It would be great to add a "Try It" button and example on the WebSocket tool, and in the process double check the request and response formats, but that's not as easy as I would like at the moment since it's a Clio-only method for a feature that's only on Devnet right now. I think there are some Clio Devnet servers somewhere but I don't know the URL. Might be worth investigating.

maria-robobug:
- Cleaned up API reference based on issues raised in https://github.com/orgs/XRPLF/projects/4?pane=issue&itemId=108396542&issue=XRPLF%7Cxrpl-dev-portal%7C3017

- Also added support for `devnet-clio` and `testnet-clio` for the "TryIt" component. This wasn't fully wired up previously it seems.

Preview: https://xrpl-dev-portal--mpt-holders-doc-cleanup.preview.redocly.app/docs/references/http-websocket-apis/public-api-methods/clio-methods/mpt_holders

The "Try it" example defaults to a Devnet example request, as I've created an MPT in that environment. I don't have one for Mainnet, but do let me know if we absolutely require this to be Mainnet.

### DIFF
--- a/@theme/markdoc/components.tsx
+++ b/@theme/markdoc/components.tsx
@@ -145,7 +145,7 @@ export function Badge(props: {
     }
 }
 
-type TryItServer = 's1' | 's2' | 'xrplcluster' | 'testnet' | 'devnet'
+type TryItServer = 's1' | 's2' | 'xrplcluster' | 'testnet' | 'devnet' | 'testnet-clio' | 'devnet-clio'
 
 export function TryIt(props: {
   method: string,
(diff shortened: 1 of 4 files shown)


DennisDawson (approved):
LGTM

mDuo13 (approved):
LGTM.

Having the examples on Devnet is best for now—since MPTs are not even enabled on Mainnet yet—but eventually it would be preferable to switch the examples to use Mainnet so that we don't have to re-create the same data every time there's a Devnet reset. Devnet resets are somewhat irregular but can be as often as multiple times per year, whereas on Mainnet we can generally create sample data once and leave it there forever. So that's a task to note for later, but it can't be done until MPT is enabled and there's no need to do anything until the next Devnet reset.

PR #3106, opened 2025-05-07
Amendments: Bithomp link update

ihomp:
Changing the link from `https://xrplexplorer.com/amendments` to a newer `https://bithomp.com/en/amendments`

Changing the name from the `XRPLExplorer` to `Bithomp`  (re-branded back to the older version)


amarantha-k (approved):
lgtm

PR #3108, opened 2025-05-08
Link MPTs from Payment page

mDuo13:
The "MPT Payments" section of this page is one of the first search results for MPTs. This seems like a place where it's useful to link back to the concept article for MPTs.


amarantha-k (approved):
lgtm 
Additionally, we could also add "MPT" in the title of the conceptual topic.

Topher68chris (approved):

PR #3110, opened 2025-05-09
Update/remove legacy links

mDuo13:
While testing out the updated search, I discovered some places where it was taking you to places with lots of old/dead links to either `ripple.com/build/` or `developers.ripple.com` (the predecessors of `xrpl.org`). Many of those links referred to pages that still exist in a new location or under a new name, so I went on a find-and-replace spree and updated or removed the links as seemed most appropriate on a case-by-case basis.

All of these were in the blog except for one in the Japanese translation of the "List XRP..." use case.

Hopefully this should improve SEO and make it easier to get to what you're looking for if you click through to a link to an old blog (which is entirely possible, given how the search works).


amarantha-k (approved):
lgtm
Something to consider for the future: add an editor's note clarifying that only the links were updated.

Topher68chris (approved):

PR #3123, opened 2025-05-27
[JA] Change freeze translation from kanji to katakana for consistency

t-ube:


mDuo13 (approved):
Seems like a good choice overall and I didn't spot any problems in the diff.

PR #3131, opened 2025-05-29
Allow for variable fees due to a change in the base fee

DennisDawson:
The description was currently specific to 10 drop base fees. While those are still the standard, there are many validators voting in favor of increasing the fee to 200 drops, which would affect the calculated fees. For the moment, it's probably prudent to indicate that the fees are variable, and update the docs when and if the standard base fee goes to 200 drops.


amarantha-k (approved):
lgtm

PR #3151, opened 2025-06-11
chore: update realm to 0.122.1

volodymyr-rutskyi:
Update `@redocly/realm` package to `0.122.1` and `react` dependency to `^19.1.0`


mDuo13 (approved):
Tested a bunch of pages after upgrading (again) to 0.122.3 and upgrading all deps. It seems to work fine. :+1:

PR #3157, opened 2025-06-23
Fix Account Configurator

mDuo13:
- The code samples page was glitching out because it wasn't in a language subfolder. Moved to a js/ folder to fix this.
  | Before | After |
  |---|---|
  | "Before" screenshot w/ broken icons and text overlapping | "After" screenshot with one working JS icon |
- Updated pages that linked to the old location.
- The UI as linked directly in the _code-samples folder was displaying as expected but the JS wasn't loading correctly on the page, so none of the buttons worked. For now, I changed the Broker an NFT Sale page to link the ZIP instead.


amarantha-k (approved):
lgtm

Issue #3057, opened 2025-03-28
MPToken/STCurrency for binary format page


tequdev:
https://xrpl.org/docs/references/protocol/binary-format

DennisDawson:
@tequdev , I'm sorry, I don't follow. What about MPToken and STCurrency need to be documented on that page? I don't see any obvious peer objects.

tequdev:
STCurrency should be added to the [Type List](https://xrpl.org/docs/references/protocol/binary-format#type-list), and MPToken should be mentioned in the [Amount Fields section](https://xrpl.org/docs/references/protocol/binary-format#amount-fields).

mDuo13:
Related: #2859

DennisDawson:
This is part of a larger set of tasks and will be prioritized as part of that group.

mDuo13:
Fixed in #3163

PR #3163, opened 2025-06-25
Update binary format documentation & Python sample code

### LINKED ISSUE #2859: Type list needs to be updated on the Binary Format page
UInt64, UInt96, UInt192, UInt384, UInt512, Vector256, Currency are not listed

https://xrpl.org/docs/references/protocol/binary-format#type-list

スクリーンショット 2024-11-21 16 45 42

mDuo13:
Fixes #2859 and #3057.

Document and implement sample code for "new" types including UInt64, Issue, Currency, Number, MPT amounts, and UInt192. Change the official name of "Hash256" etc. to "UInt256" etc. (The legacy names still exist in some APIs, though, such as `get_counts` and `server_definitions`.)

For the changes to the transaction reference to consistently use the proper type names, see PR #3164.


docs/references/protocol/binary-format.md
```diff
@@ -218,23 +226,40 @@ AccountIDs that appear as stand-alone fields (such as `Account` and `Destination
...
+    XRP is serialized as a 64-bit unsigned integer (big-endian order), except that the most significant bit is always `0`, the second-most-significant bit is `1` to indicate that it is positive, and the third-most-significant bit is `0` to indicate that it is not an MPT. The remaining 61 bits represent the quantity of XRP. Since the maximum amount of XRP (10<sup>17</sup> drops) only requires 57 bits, you can calculate XRP serialized format by taking standard 64-bit unsigned integer and performing a bitwise-OR with `0x4000000000000000`.
 
-- **Tokens**
+- **Fungible Tokens**
 
-    Tokens consist of three segments in order:
+    Fungible tokens (also called "IOUs") consist of three segments in order:
 
-    1. 64 bits indicating the amount in the [token amount format](#token-amount-format). The first bit is `1` to indicate that this is not XRP.
+    1. 64 bits indicating the amount in the [token amount format](#token-amount-format). The first bit is `1` to indicate that this is not XRP nor an MPT.
```
oeggert:
```suggestion
    1. 64 bits indicating the amount in the [token amount format](#token-amount-format). The first bit is `1` to indicate that this is an IOU.
```
Is there a reason we have to say this in the negative of what it's not? Like there may be a future scenarior where `1` can represent an IOU and something else, while still not being an MPT or XRP?

docs/references/protocol/binary-format.md
```diff
@@ -218,23 +226,40 @@ AccountIDs that appear as stand-alone fields (such as `Account` and `Destination
...
+    4. 160 bits indicating the [AccountID][] of the MPT's issuer.
+
+You can tell which of the three sub-types an amount is based on the first and third most significant bits: 
+
+    - If the first bit is a `1`, it's a fungible token (IOU).
+    - If the first bit and third bit are both `0`, it's XRP.
+    - If the first bit is a `0` and the third bit is a `1`, it's an MPT.
+
+{% admonition type="warning" title="Caution" %}
+Not all types of amount are valid in all places. Some fields can only represent XRP, or XRP and fungible tokens but not MPTs. These limitations are defined by the individual transactions and ledger entries.
```
oeggert, line 260:
```suggestion
Not all `STAmount` types are valid in all places. Some fields can only represent XRP, or XRP and fungible tokens but not MPTs. These limitations are defined by the individual transactions and ledger entries.
```
I think writing it this way makes it easier to understand we're talking about defining amount fields. Admittedly very minor, but I could see how someone just learning about this might stumble momentarily.

docs/references/protocol/binary-format.md
```diff
@@ -218,23 +226,40 @@ AccountIDs that appear as stand-alone fields (such as `Account` and `Destination
...
+{% /admonition %}
 
-The following diagram shows the serialization formats for both XRP amounts and token amounts:
+The following diagram shows the serialization formats for both XRP amounts and token amounts: ***TODO: update***
```
oeggert:
```suggestion
The following diagram shows the serialization formats for both XRP amounts and token amounts:***
```
Does this diagram need to be updated still?

docs/references/protocol/binary-format.md
```diff
@@ -271,11 +296,14 @@ The [`rippled` APIs](../http-websocket-apis/index.md) support a **standard forma
 
 The **nonstandard format** is any 160 bits of data as long as the first 8 bits are not `0x00`.
 
+#### MPT Amount Format
+
+The following diagram shows the amount serialization for MPTs. ***TODO: consolidate or not?***
```
oeggert:
```suggestion
The following diagram shows the amount serialization for MPTs.
```
Missed note?

docs/references/protocol/binary-format.md
```diff
@@ -289,25 +317,23 @@ The following example shows the serialization format for an array (the `SignerEn
...
 
-### Hash Fields
-[Hash128]: #hash-fields
-[Hash160]: #hash-fields
-[Hash256]: #hash-fields
+### Currency Fields
+[Currency]: #currency-fields
 
-The XRP Ledger has several "hash" types: Hash128, Hash160, and Hash256. These fields contain arbitrary binary data of the given number of bits, which may or may not represent the result of a hash operation.
+Some fields specify a currency code, which could be an a fungible token, the ticker symbol for an off-ledger asset, or some other identifier for a currency. This field type is currently used only in [Price Oracles](../../concepts/decentralized-storage/price-oracles.md).
```
oeggert:
```suggestion
Some fields specify a currency code, which could be a fungible token, the ticker symbol for an off-ledger asset, or some other identifier for a currency. This field type is currently used only in [Price Oracles](../../concepts/decentralized-storage/price-oracles.md).
```

docs/references/protocol/binary-format.md
```diff
@@ -289,25 +317,23 @@ The following example shows the serialization format for an array (the `SignerEn
...
 
-All such fields are serialized as the specific number of bits, with no length indicator, in big-endian byte order.
+These fields consists of 160 bits of binary data. If the data matches the ["standard" currency code format](#currency-codes), it may be represented as a three-letter currency code string in JSON. Otherwise, it is represented as hexadecimal. Client libraries _may_ attempt to interpret this as a string of ASCII or UTF-8, but it is not guaranteed to be valid.
```
oeggert:
```suggestion
These fields consist of 160 bits of binary data. If the data matches the ["standard" currency code format](#currency-codes), it may be represented as a three-letter currency code string in JSON. Otherwise, it is represented as hexadecimal. Client libraries _may_ attempt to interpret this as a string of ASCII or UTF-8, but it is not guaranteed to be valid.
```
Do we not have a definitive answer here for which client libraries interpret as ASCII or UTF-8? Since odd to leave it so ambiguous.

_code-samples/tx-serialization/py/serialize.py
```diff
@@ -349,6 +392,34 @@ def uint16_to_bytes(i):
...
+    b = hex_to_bytes(i)
+    if len(b) != 24: # 24 bytes = 192 bits
+        raise ValueError("UInt192 is not 192 bits long")
+    return b
+
+def uint384_to_bytes(i):
+    b = hex_to_bytes(i)
+    if len(b) != 8: # 8 bytes = 64 bits
+        raise ValueError("UInt64 is not 64 bits long")
+    return b
```
oeggert:
```suggestion
def uint384_to_bytes(i):
    b = hex_to_bytes(i)
    if len(b) != 48: # 48 bytes = 384 bits
        raise ValueError("UInt384 is not 384 bits long")
    return b
```

oeggert (changes requested):
Mostly good. 1 tiny issue with serialization.py and some suggestions to the binary-format page.

docs/references/protocol/binary-format.md
mDuo13: I am trying to get away from using the `ST` prefix in docs, and I don't see the clarification you're trying to make here.

You can't use MPTs in Escrow, for example. And you can't use fungible tokens until the TokenEscrow amendment gets enabled, either. Similarly, you can't deposit MPTs into AMMs. Those are the types of limitations I'm trying to mention here.

docs/references/protocol/binary-format.md
mDuo13: It's up to different client libraries. Some of them do, some don't, and that may change. I don't have a survey here. But I can link the code sample we have for this purpose.

docs/references/protocol/binary-format.md
mDuo13: It did. Done now.

docs/references/protocol/binary-format.md
mDuo13: Removed this redundant section entirely.

oeggert (approved):
LGTM

PR #3164, opened 2025-06-25
Update internal data types in reference docs

### LINKED ISSUE #3121: Inconsistency in "Internal Type" names
In transaction references, we inconsistently mix and match different names for the same types, such as "STArray" vs "Array" or "STAccount" vs "AccountID". This happens in the transaction references and even within the Binary Format page itself.

## Background
Every field table has an "Internal Type" column that describes the data type used in the canonical binary representation of the transaction or ledger entry. The definition of these types is the [Type List in the Binary Format page](https://xrpl.org/docs/references/protocol/binary-format#type-list).

The "ST" stands for "Standard Type" and is used in the `rippled` codebase for many data types that are not primitives like unsigned integer (UInt) types.

## Proposed Solution

We should modify the Binary Format page to consistently refer to types without the "ST" prefix in the type table and headings. (Maybe we can still mention the ST-prefixed name in the text body for clarity & SEO purposes.) So, in field tables, we should change:

- STIssue → Issue
- STAccount → AccountID
- STArray → Array
- STObject → Object
- STBlob → Blob

This applies to transaction reference and ledger entry reference pages.

Related: #2859

mDuo13:
This PR builds on top of #3163 but I split it off because the diff is very large and mostly consists of uninteresting find-and-replace and whitespace changes. (This has been rebased on `master` after merging that PR.)

- Consistently use the same names for data types (Fix #3121)
- Various other formatting / style updates and corrections for various tables that were affected, including:
    - Align of many (not all) tables in MD source
    - Add "Required?" column instead of having _(Optional)_ in the Description column (in many but not all cases—there are a lot)
    - Further clarify the JSON format for some fields, i.e. Hash, Address, hexadecimal fields
    - Remove common fields from transaction-type specific tables.
    - Fix one case where the internal type was specified incorrectly (`NFTokenMinter` field of AccountSet [should be AccountID](https://github.com/XRPLF/rippled/blob/e9d46f0bfc172f7fcda02d0585f043844c892729/include/xrpl/protocol/detail/sfields.macro#L288) not "Blob")
    - Fix type definition of `Delegate` transaction common field (This fixes one of the problems identified in #3186.)
    - Replace dark heavy check mark (✔️) emoji with white check mark (✅) emoji for visibility in dark mode on the NFTokenAcceptOffer page.
    - Clarify how tokens can be used in Payment Channels (depends on `TokenEscrow` amendment) in some references. This fixes part of the problem identified in #3187.
    - Fix syntax error in JSON sample of MPTokenIssuance and mention XLS-89d on both that page and MPTokenIssuanceCreate.
- Remove unused frontmatter in a couple files.
- Improve wording in a few places, like NFTokenAcceptOffer and MPTokenIssuance references.


@l10n/ja/docs/references/protocol/data-types/nftoken.md
```diff
@@ -27,7 +27,7 @@ _([NonFungibleTokensV1_1 amendment][]により追加されました。)_
 ## NFTokenID
 
 
-NFTokenID, 任意, 文字列, Hash256
```
tequdev, line 30: If based on rippled source code, it is Uint256, but if based on server_definition, it is Hash256.

Do you mean to change it to be based on rippled source code?

https://github.com/XRPLF/xrpl.js/blob/0eb77a9458fa515643f18aae2976e5a125b9ed31/packages/ripple-binary-codec/src/enums/definitions.json#L970

docs/references/protocol/ledger-data/ledger-entry-types/nftokenoffer.md
```diff
@@ -31,18 +31,18 @@ _(Added by the [NonFungibleTokensV1_1 amendment][].)_
...
-| `PreviousTxnID`     | string           | Hash256           | Yes         | Identifying hash of the transaction that most recently modified this object. |
-| `PreviousTxnLgrSeq` | number           | UInt32            | Yes         | Index of the ledger that contains the transaction that most recently modified this object. |
+| Name                |JSON Type             | [Internal Type][] | Required?   | Description |
+|:--------------------|:---------------------|:------------------|:------------|:-----------|
+| `Amount`            | [Currency Amount][]  | Amount         | Yes         | Amount expected or offered for the NFToken. If the token has the `lsfOnlyXRP` flag set, the amount must be specified in XRP. Sell offers that specify assets other than XRP must specify a non-zero amount. Sell offers that specify XRP can be 'free' (that is, the Amount field can be equal to `"0"`). |
+| `Destination`       | String - [Address][] |           | AccountID         | No          | The AccountID for which this offer is intended. If present, only that account can accept the offer. |
```
oeggert:
```suggestion
| `Destination`       | String - [Address][]        | AccountID         | No          | The AccountID for which this offer is intended. If present, only that account can accept the offer. |
```

docs/references/protocol/transactions/types/mptokenissuancecreate.md
```diff
@@ -34,30 +32,26 @@ This example assumes that the issuer of the token is the signer of the transacti
...
+| `TransferFee`      | number             | UInt16            | No        | The value specifies the fee to charged by the issuer for secondary sales of the Token, if such sales are allowed. Valid values for this field are between 0 and 50,000 inclusive, allowing transfer rates of between 0.000% and 50.000% in increments of 0.001. The field _must not_ be present if the tfMPTCanTransfer flag is not set. If it is, the transaction should fail and a fee should be claimed. |
+| `MaximumAmount`   | string              | UInt64            | No        | The maximum asset amount of this token that can ever be issued, as a base-10 number encoded as a string. The current default maximum limit is 9,223,372,036,854,775,807 (2^63-1). _This limit may increase in the future. If an upper limit is required, you must specify this field._ |
+| `MPTokenMetadata` | string              | Blob              | No        | Arbitrary metadata about this issuance, in hex format. The limit for this field is 1024 bytes. |
```
oeggert:
```suggestion
| Field             | JSON Type           | [Internal Type][] | Required? | Description |
|:------------------|:--------------------|:------------------|:----------|:------------|
| `AssetScale`      | Number              | UInt8             | No        | An asset scale is the difference, in orders of magnitude, between a standard unit and a corresponding fractional unit. More formally, the asset scale is a non-negative integer (0, 1, 2, …) such that one standard unit equals 10^(-scale) of a corresponding fractional unit. For example, a US Dollar Stablecoin will likely have an asset scale of _2_, representing 2 decimal places; 1 unit of this MPToken would equal 0.01 US Dollars. If the fractional unit equals the standard unit, then the asset scale is 0. Note that this value is optional, and will default to 0 if not supplied. |
| `TransferFee`      | Number             | UInt16            | No        | The value specifies the fee to charged by the issuer for secondary sales of the Token, if such sales are allowed. Valid values for this field are between 0 and 50,000 inclusive, allowing transfer rates of between 0.000% and 50.000% in increments of 0.001. The field _must not_ be present if the tfMPTCanTransfer flag is not set. If it is, the transaction should fail and a fee should be claimed. |
| `MaximumAmount`   | String - Number | UInt64            | No        | The maximum asset amount of this token that can ever be issued, as a base-10 number encoded as a string. The current default maximum limit is 9,223,372,036,854,775,807 (2^63-1). _This limit may increase in the future. If an upper limit is required, you must specify this field._ |
| `MPTokenMetadata` | String - Hexadecimal | Blob              | No        | Arbitrary metadata about this issuance, in hex format. The limit for this field is 1024 bytes. |
```

docs/references/protocol/ledger-data/ledger-entry-types/mptoken.md
```diff
@@ -43,7 +43,7 @@ The `MPTokenID` is the result of SHA512-Half of the following values, concatenat
 | `MPTokenIssuanceID` | string    | UInt192       | The `MPTokenIssuance` identifier. |
 | `MPTAmount`         | string    | UInt64        | This value specifies a positive amount of tokens currently held by the owner. Valid values for this field are between 0x0 and 0x7FFFFFFFFFFFFFFF. |
 | `Flags`             | number    | UInt32        | (Default) See [MPToken Flags](#mptoken-flags) |
```
oeggert:
```suggestion
| `Flags`             | Number    | UInt32        | (Default) See [MPToken Flags](#mptoken-flags) |
```
Need to update the whole table JSON type column as well.

docs/references/protocol/ledger-data/ledger-entry-types/mptokenissuance.md
```diff
@@ -70,7 +70,7 @@ The `MPTokenIssuanceID` is a 192-bit integer, concatenated in order:
...
 | `TransferFee`       | number    | UInt16        | This value specifies the fee, in tenths of a basis point, charged by the issuer for secondary sales of the token, if such sales are allowed at all. Valid values for this field are between 0 and 50,000 inclusive. A value of 1 is equivalent to 1/10 of a basis point or 0.001%, allowing transfer rates between 0% and 50%. A `TransferFee` of 50,000 corresponds to 50%. The default value for this field is 0. Any decimals in the transfer fee are rounded down. The fee can be rounded down to zero if the payment is small. Issuers should make sure that their MPT's `AssetScale` is large enough. |
 | `MPTokenMetadata`   | string    | Blob          | Arbitrary metadata about this issuance, in hex format. The limit for this field is 1024 bytes. |
-| `PreviousTxnID`     | string    | Hash256       | Transaction ID of the transaction that most recently modified this object. |
+| `PreviousTxnID`     | string    | UInt256       | Transaction ID of the transaction that most recently modified this object. |
```
oeggert:
```suggestion
| `PreviousTxnID`     | String    | UInt256       | Transaction ID of the transaction that most recently modified this object. |
```
Need to update the whole table column.

oeggert (changes requested):
Just a few details need fixing.

@l10n/ja/docs/references/protocol/data-types/nftoken.md
mDuo13: Yes, I'm going based on the source code. I do mention the values used in `server_definitions`.

oeggert (approved):
LGTM

docs/references/protocol/transactions/types/nftokenacceptoffer.md
```diff
@@ -44,9 +42,9 @@ The mode in which the transaction operates depends on the presence of the `NFTok
 
 | `NFTokenSellOffer` | `NFTokenBuyOffer` | Mode     |
 |:-------------------|:------------------|:---------|
-| ✔️                  | ✔️                 | Brokered |
-| ✔️                  | ❌                 | Direct   |
-| ❌                  | ✔️                 | Direct   |
+| ✅                 | ✅                | Brokered |
```
oeggert, line 45: Why the icon change? The old one seemed to match better with the "X"?

mDuo13: I actually did explain this in the PR description above:
> Replace dark heavy check mark (✔️) emoji with white check mark (✅) emoji for visibility in dark mode on the NFTokenAcceptOffer page.

PR #3179, opened 2025-07-02
Replace `react-alert` with `react-hot-toast`

Daryna-del:
Migrate alert handling from `react-alert` to `react-hot-toast` due to React 19 incompatibility


maria-robobug (approved):
LGTM

PR #3183, opened 2025-07-07
feat: adds new event

AKCodez:


mDuo13 (approved):
Looks good, but while we're editing the Events page—isn't it past time to take down the "Save the Date" section for last month's Apex conference?

AKCodez:
> Looks good, but while we're editing the Events page—isn't it past time to take down the "Save the Date" section for last month's Apex conference?
> 
> 

Thanks for flagging, I will bring this up with the team

PR #3184, opened 2025-07-07
Transaction Queue: update normalized order

### LINKED ISSUE #3043: Normalized ordering within the transaction queue
Image

https://xrpl.org/docs/concepts/transactions/transaction-queue#%E3%82%AD%E3%83%A5%E3%83%BC%E5%86%85%E3%81%AE%E9%A0%86%E5%BA%8F:~:text=Transactions%20that%20pay%20the%20same%20transaction%20cost%20are%20ranked%20in%20the%20order%20the%20server%20received%20them.

Currently, in the transaction queue, transactions that pay the same cost are ordered in a normalized order rather than in the order in which they are received by the server.
https://github.com/XRPLF/rippled/releases/tag/1.8.2

mDuo13:
Fix #3043.

The order of same-cost transactions in the queue was updated in [rippled 1.8.2](https://github.com/XRPLF/rippled/releases/tag/1.8.2), released in 2021, to address issues with high fees and full transaction queues on the network at the time.


docs/concepts/transactions/transaction-queue.md
```diff
@@ -60,7 +60,7 @@ If none of the above occur, transactions can stay in the queue for a theoretical
...
+Transactions that were previous proposed in the consensus process but did not achieve the threshold to be validated are the first ones to be added to the next open ledger. After that, transactions are pulled from the queue to be proposed in a normalized order so that, during periods of high traffic, different servers propose similar sets of transactions. Within the transaction queue, transactions are ranked so that transactions paying a higher transaction cost come first. This ranking is not by the transactions' _absolute_ XRP cost, but by costs _relative to the [minimum cost for that type of transaction](transaction-cost.md#special-transaction-costs)_.  Transactions that pay the same transaction cost are ranked in ascending order by hash. Other factors may also affect the order of transactions in the queue; for example, transactions from the same sender are sorted by their `Sequence` numbers so that they are submitted in order.
```
maria-robobug:
```suggestion
Transactions that were previously proposed in the consensus process but did not achieve the threshold to be validated are the first ones to be added to the next open ledger. After that, transactions are pulled from the queue to be proposed in a normalized order so that, during periods of high traffic, different servers propose similar sets of transactions. Within the transaction queue, transactions are ranked so that transactions paying a higher transaction cost come first. This ranking is not by the transactions' _absolute_ XRP cost, but by costs _relative to the [minimum cost for that type of transaction](transaction-cost.md#special-transaction-costs)_.  Transactions that pay the same transaction cost are ranked in ascending order by hash. Other factors may also affect the order of transactions in the queue; for example, transactions from the same sender are sorted by their `Sequence` numbers so that they are submitted in order.
```

maria-robobug (dismissed):
LGTM, just minor spelling correction.

PR #3188, opened 2025-07-09
Add code snippets from xrpl.js and xrpl-py

maria-robobug:
Addresses https://github.com/XRPLF/xrpl-dev-portal/issues/3141 

We already have most/all of these examples in our repository, but I noticed we didn't use them in our docs, so I'm adding snippets where possible in the How To guides.

I also noticed we didn't have a pay channel code snippet for Python, so I added one myself for parity with Javascript. ~~However, it looks like there is an issue when I run the original `claimPayChannel.ts` sample code - it fails with 'tefBAD_AUTH' for the `PaymentChannelClaim` transaction. I'm chasing this up with DGE to understand what the original intention was for the code sample and why it's failing.~~ (**UPDATE**: got a response, we've solved the error in the example code, the wrong wallet was being used to sign the transaction).

### DIFF
--- a/_code-samples/claim-payment-channel/py/claim_pay_channel.py
+++ b/_code-samples/claim-payment-channel/py/claim_pay_channel.py
@@ -13,75 +13,71 @@ from xrpl.transaction import submit, submit_and_wait
 from xrpl.account import get_balance
 
 
-def claim_pay_channel():
-    """The snippet walks us through creating and claiming a Payment Channel."""
+"""The snippet walks us through creating and claiming a Payment Channel."""
 
-    client = JsonRpcClient("https://s.altnet.rippletest.net:51234")
-    
-    # Creating wallets as prerequisite
-    wallet1 = generate_faucet_wallet(client)
-    wallet2 = generate_faucet_wallet(client)
+client = JsonRpcClient("https://s.altnet.rippletest.net:51234")
 
-    print("Balances of wallets before Payment Channel is claimed:")
-    print(f"Balance of {wallet1.address} is {get_balance(wallet1.address, client)} XRP")
-    print(f"Balance of {wallet2.address} is {get_balance(wallet2.address, client)} XRP")
+# Creating wallets as prerequisite
+wallet1 = generate_faucet_wallet(client)
+wallet2 = generate_faucet_wallet(client)
 
-    # Create a Payment Channel and submit and wait for tx to be validated
-    payment_channel_create = PaymentChannelCreate(
-        account=wallet1.address,
-        amount="100",  # 10 XRP
-        destination=wallet2.address,
-        settle_delay=86400,  # 1 day in seconds
-        public_key=wallet1.public_key,
-    )
+print("Balances of wallets before Payment Channel is claimed:")
+print(f"Balance of {wallet1.address} is {get_balance(wallet1.address, client)} XRP")
+print(f"Balance of {wallet2.address} is {get_balance(wallet2.address, client)} XRP")
 
-    print("Submitting a PaymentChannelCreate transaction...")
-    payment_channel_response = submit_and_wait(
-        payment_channel_create,
-        client,
-        wallet1
-    )
-    print("PaymentChannelCreate transaction response:")
-    print(payment_channel_response.result)
+# Create a Payment Channel and submit and wait for tx to be validated
+payment_channel_create = PaymentChannelCreate(
+    account=wallet1.address,
+    amount="100",  # 10 XRP
+    destination=wallet2.address,
+    settle_delay=86400,  # 1 day in seconds
+    public_key=wallet1.public_key,
+)
 
-    # Check that the object was actually created
-    account_objects_request = AccountObjects(
-        account=wallet1.address,
-    )
-    account_objects_response = client.request(account_objects_request)
-    account_objects = account_objects_response.result["account_objects"]
+print("Submitting a PaymentChannelCreate transaction...")
+payment_channel_response = submit_and_wait(
+    payment_channel_create,
+    client,
+    wallet1
+)
+print("PaymentChannelCreate transaction response:")
+print(payment_channel_response.result)
 
-    # Find the PayChannel object to get the correct channel ID
-    channel_id = None
-    for obj in account_objects:
-        if obj["LedgerEntryType"] == "PayChannel":
-            channel_id = obj["index"]
-            break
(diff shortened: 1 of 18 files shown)


_code-samples/claim-payment-channel/py/claim_pay_channel.py
```diff
@@ -0,0 +1,83 @@
+"""
+Create, claim and verify a Payment Channel.
+Reference: https://xrpl.org/paychannel.html
```
mDuo13: We should use the new URL here instead of the old (redirected) one.

```suggestion
Reference: https://xrpl.org/docs/references/protocol/ledger-data/ledger-entry-types/paychannel
```

_code-samples/claim-payment-channel/py/claim_pay_channel.py
```diff
@@ -0,0 +1,83 @@
...
+from xrpl.account import get_balance
+
+
+"""The snippet walks us through creating and claiming a Payment Channel."""
+
+client = JsonRpcClient("https://s.altnet.rippletest.net:51234")
+
+# Creating wallets as prerequisite
+wallet1 = generate_faucet_wallet(client)
+wallet2 = generate_faucet_wallet(client)
```
mDuo13: Since the first thing the code sample does is wait a while for wallets to get funded, it could be helpful to add a print-out to let people know it's actually working.

```suggestion
# Creating wallets as prerequisite
print("Setting up wallets...")
wallet1 = generate_faucet_wallet(client)
wallet2 = generate_faucet_wallet(client)
```

_code-samples/claim-payment-channel/py/claim_pay_channel.py
```diff
@@ -0,0 +1,83 @@
...
+channel_id = None
+for obj in account_objects:
+    if obj["LedgerEntryType"] == "PayChannel":
+        channel_id = obj["index"]
+        break
+
+if not channel_id:
+    raise Exception("PayChannel not found in account objects")
+
+print(f"PayChannel ID: {channel_id}")
```
mDuo13: This _will_ work in this specific scenario because it's using freshly-funded accounts, but it's not a good example of how to do this more generally, because it won't return the _right_ channel ID if an account has multiple payment channels.

I believe you can even have multiple PayChannels with the same source+destination accounts, so even just iterating the list of channels and matching the accounts is not strictly guaranteed to be correct.

Valid approaches include:
1. Calculate the channel ID yourself using the PayChannel ledger entry ID formula. This is what the `claimPayChannel.ts` example does above, in these lines:
    ```ts
     Channel: hashes.hashPaymentChannel(
      wallet1.classicAddress,
      wallet2.classicAddress,
      paymentChannelResponse.result.tx_json.Sequence ?? 0, 
    )
    ```
    Unfortunately xrpl-py does not have the equivalent functionality easily accessible. (Issue: https://github.com/XRPLF/xrpl-py/issues/356 )
2. Iterate through the `ModifiedNodes` array of the PaymentChannelCreate metadata, looking for a `CreatedNode` entry of type `PayChannel`, then take the `index` (ledger entry ID) of that item.

docs/_snippets/wait-for-validation.md
```diff
@@ -1,3 +1,5 @@
 On a live network (including Mainnet, Testnet, or Devnet), you can wait 4-7 seconds for the ledger to close automatically.
 
 If you're running `rippled` in stand-alone mode, use the [ledger_accept method][] to manually close the ledger.
+
+{% raw-partial file="/docs/_snippets/common-links.md" /%}
```
mDuo13: Rather than add this here (which could lead to it being double-included on the relevant pages), change the places that include `wait-for-validation.md` to do so as a `raw-partial`.

docs/tutorials/how-tos/use-specialized-payment-types/use-escrows/send-a-time-held-escrow.md
```diff
@@ -47,27 +47,69 @@ print(release_date_ripple)
...
+xrpl.validate(escrowCreateTransaction);
+
+// Sign and submit the transaction ----------------------------------------
+console.log('Signing and submitting the transaction:',
+            JSON.stringify(escrowCreateTransaction, null,  "\t"), "\n"
+);
+const response = await client.submitAndWait(escrowCreateTransaction, { wallet });
+console.log(`Sequence number: ${response.result.tx_json.Sequence}`);
+console.log(`Finished submitting! ${JSON.stringify(response.result, null, "\t")}`);
+```
```
mDuo13: Why not use `{% code-snippet ... /%}` here, too?

mDuo13 (changes requested):
Generally looks good, but I have some nitpicks.

_code-samples/claim-payment-channel/py/claim_pay_channel.py
maria-robobug: Gotcha! I've opted for option 2, from what I could see in the PaymentChannelCreate metadata, the `AffectedNodes` array contains the relevant entries. Updated now.

docs/tutorials/how-tos/use-specialized-payment-types/use-escrows/send-a-time-held-escrow.md
maria-robobug: Ah yeah, so the existing code example includes a `Condition` field that should be omitted for time-held escrows, which could confuse readers comparing the Websocket snippet. I created a variation that omits this field to better match the time-held escrow use case. 

Though, I could just add a comment in the code example to mention this...I'll update.

_code-samples/claim-payment-channel/py/claim_pay_channel.py
```diff
@@ -53,15 +53,16 @@
...
 channel_id = None
-for obj in account_objects:
-    if obj["LedgerEntryType"] == "PayChannel":
-        channel_id = obj["index"]
-        break
+if 'meta' in payment_channel_response.result and 'AffectedNodes' in payment_channel_response.result['meta']:
+    for node in payment_channel_response.result["meta"]["AffectedNodes"]:
+        if 'CreatedNode' in node and node["CreatedNode"]["LedgerEntryType"] == 'PayChannel':
+            channel_id = node['CreatedNode']['LedgerIndex']
+            break
```
mDuo13, line 60: 💯

mDuo13 (approved):

PR #3194, opened 2025-07-14
Add Crossmark wallet to wallets list

maria-robobug:
While looking into this VOD issue, https://ripplelabs.atlassian.net/browse/VODF-127, I noticed we don't have Crossmark in our current list of wallets.

I've also added a link to the wallets list from the "Crypto Wallets" page for more visibility.


oeggert (approved):
LGTM

PR #3195, opened 2025-07-14
Refactor home hero graphic styles and update image attributes for bet…

AKCodez:
…ter responsiveness. Adjusted width to 100% with a max-width of 856px and set height to auto. Removed width attribute from the image tag in index.page.tsx.


AKCodez:
Screenshot of localhost on firefox with changes vs live site on firefox, 
image

mDuo13 (approved):
Looks better. :+1: 

But looking back, the dimensions on the hero `img` tag were added in #2610 to address usability issues with the page "jumping" as the image or CSS file loaded in.

Is there another approach we can take to compromise between the two concerns? Maybe a `style="width: 100%; max-width: 856px; height: auto;"` attribute directly on the tag would do it?

AKCodez:
> Looks better. 👍
> 
> But looking back, the dimensions on the hero `img` tag were added in #2610 to address usability issues with the page "jumping" as the image or CSS file loaded in.
> 
> Is there another approach we can take to compromise between the two concerns? Maybe a `style="width: 100%; max-width: 856px; height: auto;"` attribute directly on the tag would do it?

Very good point, however this change has not affected the CLS of the page at all

PR #3196, opened 2025-07-15
Fix minor issues in Batch ref

mDuo13:
- Add missing common links snippet. Without it, reference links are unparsed.
    Screenshot
- Fix case of `Fee` field and string type of example value.
- Update internal type names to match #3164 
- Applied to both English and Japanese translations.


oeggert (approved):
LGTM

PR #3214, opened 2025-07-23
Add blog with Frii Pay case study

amarantha-k:


blog/2025/frii-pay-xrpl-case-study-crypto-payment-solution.md
```diff
@@ -0,0 +1,79 @@
+---
+category: 2025
+date: "2025-07-23"
+template: '../../@theme/templates/blogpost'
+seo:
+    description: Learn how Frii Pay built a crypto payment solution on the XRP Ledger (XRPL). This case study details their journey, technical integration, and market impact.
+labels:
+    - Case Study
```
mDuo13, line 8: This is technically fine but we don't have a filter for this category on the blog landing page. It does feel very similar to "Developer Reflections" although I understand if we don't necessarily want to treat this as a continuation of that series.

mDuo13 (approved):

blog/2025/frii-pay-xrpl-case-study-crypto-payment-solution.md
amarantha-k: You're right - this will be a new category and we didn't want to treat this as a continuation of the Developer Reflections series.

PR #3215, opened 2025-07-24
fix broken links and add mailing list links

### LINKED ISSUE #416: API change documentation notifications
Is there any way to learn when the API changes? i.e. when methods are added/deprecated, or parameters are changed.

As I go through the API documentation and find a mix of deprecated arguments, methods, or badges indicating the version a particular method became available, I am seeing the need for an organization relying on the API directly, or a developer maintaining wrapper libraries, to know when changes are made.

If there's something in place, I'd recommend adding documentation about it, perhaps in the API Conventions doc.

An email list where devs can be notified when changes occur would be extremely helpful as well. I think it's especially important to address due to the nature of the API in that it manages both value and trust.

oeggert:
Add links to rippled server updates and client library updates.

### DIFF
--- a/docs/introduction/crypto-wallets.md
+++ b/docs/introduction/crypto-wallets.md
@@ -50,4 +50,4 @@ Next: [Transactions and Requests](transactions-and-requests.md)
 
 ## See Also
 
-For more information on which wallets are supported on the XRP Ledger, see [Wallets](../../about/xrp#wallets).
+For more information on which wallets are supported on the XRP Ledger, see [Wallets](../../about/xrp.page.tsx#wallets).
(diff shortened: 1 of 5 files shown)


docs/introduction/crypto-wallets.md
```diff
@@ -50,4 +50,4 @@ Next: [Transactions and Requests](transactions-and-requests.md)
 
 ## See Also
 
-For more information on which wallets are supported on the XRP Ledger, see [Wallets](../../about/xrp#wallets).
+For more information on which wallets are supported on the XRP Ledger, see [Wallets](../../about/xrp.page.tsx#wallets).
```
mDuo13, line 53: The old link already works in production, except that the anchor doesn't seem to scroll it to the right place. The updated link seems to have the same problem, though, so I'm not sure what the benefit is in changing this.

mDuo13 (commented):
Glad to get this on there. This finally fixes #416.

docs/introduction/crypto-wallets.md
oeggert: I changed it since it was getting flagged as an error in local builds, but I see what you mean. It _was_ going to the right anchor when I tested the change locally, but digging a little further it seems to work if you use that link twice. First time doesn't work, going back and clicking the link again works.

mDuo13: I guess we can ask Redocly if that's fixable. Not a blocker for this PR since it works the same for now.

mDuo13 (approved):

PR #3217, opened 2025-07-25
Fix Permission Delegation docs & add sample code

### LINKED ISSUE #3186: Permission Delegation reference documentation is mixed up
The reference docs for the Permission Delegation amendment seem to have gotten jumbled around somehow.

- The `Delegate` ledger entry type is supposed to be a page under Ledger Entry Types, but it's described on the Transaction Common Fields page instead. (The `Delegate` transaction common field is defined incorrectly and should be an AccountID field.)
- The DelegateSet transaction reference page looks mostly right although it doesn't quite follow our style conventions around code samples and stuff. For example, `tx_json = ` should not be in the sample JSON, variable casing isn't consistent with our other docs, the sample transaction has obvious placeholders instead of real values (`"Account": "rDelegatingAccount"`), etc.
- The "Permission Delegation" concept page goes deep into code samples of how to perform different tasks instead of staying at a high level describing what can/can't be done and why.

Original XLS: https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0075d-permission-delegation

mDuo13:
Fixes #3186.

- Rewrite the Permission Delegation concept page with a lot more contextual and conceptual information about the feature, why you'd use it, and how it compares with related features (namely, multi-signing)
- Add a new page on Permission Values, with the details of all the permissions that can/can't be set, in the data types reference section
- Correct the definition of `Delegate` in transaction common fields (which had a mix of info including stuff that should've been in the ledger entry reference).
- Add the missing `Delegate` ledger entry reference page.
- Rewrite the DelegateSet transaction reference page to better match style guidelines, correct errors, link the other relevant (rewritten) pages, and use a real example of a transaction.
- Add JavaScript code samples that grant permissions to a delegate and then send a sample transaction as a delegate.


_code-samples/delegate-permissions/js/README.md
```diff
@@ -0,0 +1,38 @@
+# Delegate Permissions (JavaScript) Sample Code
+
+These code samples demonstrate how to delegate permissions to another account and how to send an a transaction as a delegate, using xrpl.js 4.3 in Node.js.
+
+## Usage
+
+1. First, install dependencies.
```
oeggert:
```suggestion
1. Install dependencies.
```
No need to say first, if the step is already labeled as `1`.

_code-samples/delegate-permissions/js/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+4. Run `delegate-permissions.js`.
+
+    ```sh
+    node delegate-permissions.js
+    ```
+
+    If it runs successfully, it should output several things including "Delegate successfully set." followed by an [account_objects API method](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/account-methods/account_objects) response showing the delegate permissions.
+
+    Take note of the `account` address in this output. That's the address of the delegating account.
```
oeggert:
```suggestion
    Take note of the `account` address in this output. That's the address of the delegator.
```

_code-samples/delegate-permissions/js/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+5. Run `use-delegate-permissions.js` and provide both the delegating account's address (from the previous step's output) and the delegate's secret key (from the devnet faucet earlier).
```
oeggert:
```suggestion
5. Run `use-delegate-permissions.js` and provide both the delegator's address (from the previous step's output) and the delegate's secret key (from the devnet faucet earlier).
```

_code-samples/delegate-permissions/js/delegate-permissions.js
```diff
@@ -0,0 +1,48 @@
...
+    await client.connect()
+
+    console.log("Funding new wallet from faucet...")
+    const { wallet } = await client.fundWallet()
+
+    // Define the transaction
+    const delegate_address = "r9GAKojMTyexqvy8DXFWYq63Mod5k5wnkT"
+    const delegateset = {
+        "TransactionType": "DelegateSet",
+        "Account": wallet.address,
```
oeggert:
```suggestion
        "Account": wallet.address, // Delegator address
```

_code-samples/delegate-permissions/js/delegate-permissions.js
```diff
@@ -0,0 +1,48 @@
...
+            {
+                "Permission": {
+                    "PermissionValue": "AccountDomainSet"
+                }
+            }
+        ]
+    }
+
+    // Prepare, sign, and submit the transaction
+    const prepared = await client.autofill(delegateset)
```
oeggert: How do we want to "standardize" submitting transactions? Break out steps like this, or more streamlined

```
const result = await client.submitAndWait(transaction, {
    wallet,
    autofill: true
});
```

_code-samples/delegate-permissions/js/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Usage
+
+1. First, install dependencies.
+
+    ```sh
+    npm i
+    ```
+
+2. Go to the [XRP Faucet](https://xrpl.org/resources/dev-tools/xrp-faucets) and generate a **Devnet** account to be your delegate account.
```
oeggert: Can we just bake this step into the `delegate-permissions.js` and output the delegate account? That way the user doesn't have to manually edit this file.

docs/concepts/accounts/permission-delegation.md
```diff
@@ -7,102 +7,66 @@ labels:
 status: not_enabled
 ---
 # Permission Delegation
-XRPL accounts can delegate both transaction permissions and granular permissions to other accounts, enhancing flexibility and enabling use cases such as implementing role-based access control. This delegation is managed using the [`DelegateSet`](../../references/protocol/transactions/types/delegateset.md) transaction.
 
-_(Requires the [PermissionDelegation amendment][] {% not-enabled /%})_
+Permission delegation is the function of granting various permissions to another account to send permissions on behalf of your account. This can be used instead of or alongside techniques such as [multi-signing](./multi-signing.md) to enable flexible security paradigms such as role-based access control.
```
oeggert:
```suggestion
Permission delegation is the function of granting various permissions to another account to send permissions on behalf of your account. This can be used alongside (or in place of) techniques such as [multi-signing](./multi-signing.md) to enable flexible security paradigms, such as role-based access control.
```

docs/concepts/accounts/permission-delegation.md
```diff
@@ -7,102 +7,66 @@ labels:
...
-| `Authorize` | The address of the account that is being granted the permission(s). |
-| `Permissions` | An array of permission objects, specifying the permissions to delegate. Each permission is defined within a `Permission` object, using the `PermissionValue` field. See [XLS-74d, Account Permissions] for a complete list of valid `PermissionValues`. |  
+Permission Delegation can help optimize for these sorts of challenges by granting very limited permissions to separate accounts that can have their keys available online, while keeping the master keys for the main account offline and only using them manually in special cases when rare, important tasks like issuing tokens are necessary. This is especially helpful when using compliance features like [Authorized Trust Lines](../tokens/fungible-tokens/authorized-trust-lines.md) that require a stablecoin issuer to individually approve each user after meeting regulatory requirements like Know Your Customer rules. With a proper configuration, the keys that are used to approve users cannot be used to perform other actions like issuing new tokens or rotating the issuer's key pairs.
```
oeggert:
```suggestion
Permission Delegation alleviates these types of challenges by granting limited permissions to separate, delegate accounts that can have their keys available online for day-to-day tasks. Meanwhile, your master keys for a main account are kept offline and only used for special tasks, such as issuing tokens. This set up is especially helpful when using compliance features like [Authorized Trust Lines](../tokens/fungible-tokens/authorized-trust-lines.md) that require a stablecoin issuer to individually approve each user after meeting regulatory requirements like Know Your Customer rules. With a proper configuration, you can isolate keys to specific tasks, minimizing risks if a delegate account is compromised.
```

docs/concepts/accounts/permission-delegation.md
```diff
@@ -7,102 +7,66 @@ labels:
...
 
-## Revoking Permissions
 
-Permissions can be revoked using the `DelegateSet` transaction. There are two ways to revoke permissions:
+## How It Works
 
-### Revoke All Permissions
+The account on whose behalf transactions are being sent is called the _delegating account_. The account sending the transactions is called the _delegate_.
```
oeggert: Is delegating account the official term? Delegator feels more concise.

docs/references/protocol/data-types/permission-values.md
```diff
@@ -0,0 +1,84 @@
...
+
+For a mapping of transaction types known by a server and their corresponding numeric transaction type values, check the `TRANSACTION_TYPES` field in the [server_definitions method][].
+
+### List of Non-Delegatable Permissions
+
+Some transaction types are not delegatable. If you attempt to grant these permissions to a delegate, the transaction fails with a [result code](../transactions/transaction-results/) such as `tecNO_PERMISSION`. This includes all transaction types that can be used to grant other permissions to different key pairs or accounts. Additionally, all [pseudo-transaction types](/docs/references/protocol/transactions/pseudo-transaction-types/pseudo-transaction-types) are not delegatable since they are not meant to be sent by normal accounts anyway.
```
oeggert:
```suggestion
Some transaction types can't be delegated. If you attempt to grant these permissions to a delegate, the transaction fails with a [result code](../transactions/transaction-results/) such as `tecNO_PERMISSION`. This includes all transaction types that can be used to grant other permissions to different key pairs or accounts. Additionally, all [pseudo-transaction types](/docs/references/protocol/transactions/pseudo-transaction-types/pseudo-transaction-types) can't be delegated since they can't be sent by normal accounts anyway.
```

docs/references/protocol/data-types/permission-values.md
```diff
@@ -0,0 +1,84 @@
...
+{% admonition type="warning" name="Caution" %}
+Not all client libraries support numeric PermissionValue types. In most cases, you should use the string names of the permissions you want to grant.
+{% /admonition %}
+
+- For **transaction type permissions**, the string is the name of the transaction type exactly (case-sensitive). For example, a permission value of `"PaymentChannelClaim"` grants permission to send [PaymentChannelClaim transactions][].
+- For **granular permissions**, the string is the name of the granular permission (case-sensitive). For example, a permission value of `"TrustlineAuthorize"` grants permission to send TrustSet transactions that authorize trust lines (but not ones that modify other settings such as the trust line limit or freeze status).
+
+The numeric value `0` is reserved for "full permissions", meaning permission to send transactions of all types, but it is not possible to delegate full permissions.
+
+## Transaction Type Permissions
```
oeggert, line 31:
```suggestion
### Transaction Type Permissions
```

docs/references/protocol/data-types/permission-values.md
```diff
@@ -0,0 +1,84 @@
...
+| [DelegateSet][]     | `65` |
+| [EnableAmendment][] | `101` |
+| [SetFee][]          | `102` |
+| [UNLModify][]       | `103` |
+
+{% admonition type="warning" name="Known Issue" %}
+With only the PermissionDelegation amendment, it's possible to assign permissions for transaction types that are reserved, unassigned, or part of amendments that are not currently enabled; it's also possible to assign PermissionValue `0` for full permissions. However, these values do not actually grant any permissions. This is a bug, and a future amendment will prevent assigning values outside of currently-enabled, delegatable transaction types or known granular permissions.
+{% /admonition %}
+
+## Granular Permissions
```
oeggert, line 59:
```suggestion
### Granular Permissions
```

docs/references/protocol/data-types/permission-values.md
```diff
@@ -0,0 +1,84 @@
...
+
+The numeric value `0` is reserved for "full permissions", meaning permission to send transactions of all types, but it is not possible to delegate full permissions.
+
+## Transaction Type Permissions
+
+Transaction Type Permissions have numeric values from 1 to 65536 (that is, 2<sup>16</sup>), inclusive. They correspond with known transaction types, except you add 1 when specifying a transaction type as a permission value. For example, the string `"Payment"` corresponds to a `TransactionType` value of `0`, but a `PermissionValue` value of `1`. To grant permissions to make Payment transactions, you can specify either `"PermissionValue": "Payment"` or `"PermissionValue": 1`.
+
+For a mapping of transaction types known by a server and their corresponding numeric transaction type values, check the `TRANSACTION_TYPES` field in the [server_definitions method][].
+
+### List of Non-Delegatable Permissions
```
oeggert, line 37: Maybe move this subsection under granular permissions.

docs/references/protocol/data-types/permission-values.md
```diff
@@ -0,0 +1,84 @@
...
+| [SignerListSet][]   | `13` |
+| [AccountDelete][]   | `22` |
+| [LedgerStateFix][]  | `54` |
+| [DelegateSet][]     | `65` |
+| [EnableAmendment][] | `101` |
+| [SetFee][]          | `102` |
+| [UNLModify][]       | `103` |
+
+{% admonition type="warning" name="Known Issue" %}
+With only the PermissionDelegation amendment, it's possible to assign permissions for transaction types that are reserved, unassigned, or part of amendments that are not currently enabled; it's also possible to assign PermissionValue `0` for full permissions. However, these values do not actually grant any permissions. This is a bug, and a future amendment will prevent assigning values outside of currently-enabled, delegatable transaction types or known granular permissions.
```
oeggert, line 56: Does this mean `0` (full permissions) will never be available?

docs/references/protocol/data-types/permission-values.md
```diff
@@ -0,0 +1,84 @@
...
+| `65542`       | `AccountMessageKeySet`   | [AccountSet][]         | Can set the `MessageKey` field of the account. |
+| `65543`       | `AccountTransferRateSet` | [AccountSet][]         | Can set the [transfer fee of fungible tokens issued by the account](/docs/concepts/tokens/transfer-fees). |
+| `65544`       | `AccountTickSizeSet`     | [AccountSet][]         | Can set the [tick size of fungible tokens issued by the account](/docs/concepts/tokens/decentralized-exchange/ticksize). |
+| `65545`       | `PaymentMint`            | [Payment][]            | Can send payments that mint new fungible tokens or MPTs. |
+| `65546`       | `PaymentBurn`            | [Payment][]            | Can send payments that burn fungible tokens or MPTs. |
+| `65547`       | `MPTokenIssuanceLock`    | [MPTokenIssuanceSet][] | Can lock the balances of a particular MPT issued by the account. _(Requires the [MPTokensV1 amendment][] {% not-enabled /%}.)_ |
+| `65548`       | `MPTokenIssuanceUnlock`  | [MPTokenIssuanceSet][] | Can unlock the balances of a particular MPT issued by the account. _(Requires the [MPTokensV1 amendment][] {% not-enabled /%}.)_ |
+
+### Limitations to Granular Permissions
```
oeggert, line 79: I think we can remove this section. We already say in the previous section that granular permissions are hard-coded in the table above.

oeggert (changes requested):
Looks mostly good, but a few questions to tighten up the language.

_code-samples/delegate-permissions/js/README.md
mDuo13: Yeah. It'll need to output the secret key, but I guess that's no worse than doing so on a website.

_code-samples/delegate-permissions/js/delegate-permissions.js
mDuo13: There are some advantages to splitting it out occasionally (you can display what the autofilled transaction looks like and inspect things before submitting it) but I agree that in most cases including this one it makes sense to condense it like that.

docs/concepts/accounts/permission-delegation.md
mDuo13: We didn't have an official term. I am fine changing it to delegator.

docs/references/protocol/data-types/permission-values.md
mDuo13: This doesn't belong under granular permissions, because the non-delegatable permissions are _transaction type_ permissions.

docs/references/protocol/data-types/permission-values.md
mDuo13: I believe the (current) intent is for `0` (full permissions) to never be delegatable. Under the implementation as it exists now, you can "delegate" that permission but it does nothing. An upcoming fix amendment will make it so that attempting to delegate that permission will become an error.

The same standard for numeric permissions may also, eventually, be applied to multi-signing lists—That's what the standard was originally created for. In that case, the list ID corresponds to what permissions it has, and the default ID (`0`) grants full permissions. But right now, it's unclear if multi-signing list permissions will advance and be implemented, and the _only_ multi-signing list you can create has ID 0 and full permissions.

docs/references/protocol/data-types/permission-values.md
mDuo13: This section does add a little bit more color and answers what I expect to be a common question, which is, "Can I grant permission to just send FOO and not XRP?" (No, you can't), so I'd like to keep it.

docs/references/protocol/data-types/permission-values.md
mDuo13: I don't think it makes sense to move these under the previous heading.

oeggert (approved):

PR #3226, opened 2025-08-05
Update feesettings.md

mvadari:
The `FeeSettings` object has now changed post `XRPFees` amendment, so the data has changed


mDuo13 (changes requested):
Unfortunately we can't just remove the description of the old format as long as the amendment is not unconditionalized, because it'll still appear in its old format on any chain that doesn't have the amendment enabled.

Even if no such chains exist, it's important to document the legacy format for people looking at historical validated ledgers.

I think we should make the current format more prominent now, but maybe leave the legacy format in a tab or something.

mvadari:
@mDuo13 I copied the format on the `SetFee` page (I think - I don't know how to deploy the code so I can't make sure that my formatting is correct)

mvadari:
I checked the project preview, looks to have rendered correctly (I can't view the broken link issue though)

mDuo13 (approved):
The broken link was coming from the link to pseudo-transaction common fields (which is the wrong link here anyway since this is a ledger entry, not a pseudo-transaction).

Fixed that and fixed up the wording that mentioned SetFee transactions due to copy-paste error.

Also removed the common fields from the table of fields specific to this type. (Opened #3238 as a reminder to do this more consistently elsewhere too.)

:+1: with the changes.

mvadari:
Thanks!

mrbeastmcnasty (approved):

PR #3219, opened 2025-07-26
Add Amendment Tracker tag

oeggert:
Amendment tracker tag queries live amendment data off mainnet and displays it in a table.

Table is added using the amendments-table tag.

PR #3227, opened 2025-08-05
Amendment disclaimer

mDuo13:
Extends amendment disclaimer tag (#3223) using the live data from the amendment tracker (#3219).

The result is amendment disclaimers that automatically update with the status of the amendment on Mainnet and are also automatically translated. Before the amendment is enabled, the disclaimer says, "Requires the ... amendment." and the badge shows the voting status. After the amendment is enabled, the disclaimer says, "Added by the ... amendment." and the badge shows the date when the amendment became enabled.

Example usage:

```
{% amendment-disclaimer name="MPTokensV1" /%}
```

Examples of the disclaimer in action:

Note, this PR does not actually replace any the instances of the old hand-written disclaimers with the new markdoc tag. After it gets merged, we can make a follow-up PR to mass-replace instances of it.

### DIFF
--- a/@theme/markdoc/schema.ts
+++ b/@theme/markdoc/schema.ts
@@ -229,11 +229,6 @@ export const amendmentDisclaimer: Schema &  { tagName: string } = {
       type: 'String',
       required: true
     },
-    isVoting: {
-      type: 'Boolean',
-      required: false,
-      default: false
-    }
   },
   render: 'AmendmentDisclaimer',
   selfClosing: true
--- a/@theme/markdoc/schema.ts
+++ b/@theme/markdoc/schema.ts
@@ -221,3 +221,20 @@ export const amendmentsTable: Schema & { tagName: string } = {
   render: 'AmendmentsTable',
   selfClosing: true
 }
+
+export const amendmentDisclaimer: Schema &  { tagName: string } = {
+  tagName: 'amendment-disclaimer',
+  attributes: {
+    name: {
+      type: 'String',
+      required: true
+    },
+    isVoting: {
+      type: 'Boolean',
+      required: false,
+      default: false
+    }
+  },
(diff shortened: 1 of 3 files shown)


mDuo13:
Known issue: If you view a page in the local dev environment and then edit the amendment name in the source page, it live-refreshes with the updated amendment name, but does not update the badge status unless you manually refresh the page.

Also, I'm not sure if the difference in spaces after the opening parenthesis in Japanese is fixable. (I think it's because the Japanese sentence structure puts the link element first, and the link tag has extra whitespace around it due to linebreaks or indentation or something.)

There may be some ways to optimize the code to share more of the functionality for fetching amendment statuses with the other component. I'm also not sure if this makes the same network request multiple times if you have multiple status badges on a page. It might, which would be inefficient but probably not a dealbreaker.

@theme/markdoc/schema.ts
```diff
@@ -221,3 +221,20 @@ export const amendmentsTable: Schema & { tagName: string } = {
...
+  attributes: {
+    name: {
+      type: 'String',
+      required: true
+    },
+    isVoting: {
+      type: 'Boolean',
+      required: false,
+      default: false
+    }
```
tequdev: not necessary

```suggestion
```

maria-robobug (approved):
LGTM! 

I can see the issue with spacing for the Japanese version, but not a dealbreaker.

> There may be some ways to optimize the code to share more of the functionality for fetching amendment statuses with the other component. I'm also not sure if this makes the same network request multiple times if you have multiple status badges on a page. It might, which would be inefficient but probably not a dealbreaker.

I did test the multiple badges scenario, and it does end up calling the API more than once. I like the idea of creating a shared function that "caches" the data (until the web page is refreshed I guess), then all amendment components could use that data. Maybe a separate improvement ticket/PR?

PR #2841, opened 2024-11-05
API v2 Updates

oeggert:
Code samples updates will be included in a separate PR.

PR #3228, opened 2025-08-06
Restore API versioning docs

mDuo13:
While reviewing some other docs, I noticed that the "API versioning" and "API v2" links in several places were broken. Upon further inspection, I saw that #2841 seems to have (inadvertently?) deleted the section that provides details on API versioning.

I have restored it and updated the docs slightly:
- Removed notes on gRPC updates since gRPC is no longer a supported API (except as needed for Clio)
- Added a table showing the default API version by client library. I don't have the data on xrpl4j but I've asked about that.
- Rephrased the sentence about pre-release API versions being subject to change, for clarity.
- Added a note about fixing bugs (to match existing docs/prior behavior) not being a breaking change, since we've used that justification at least a couple times in the past.

The original Japanese translation is [available in the Git history](https://github.com/XRPLF/xrpl-dev-portal/blob/d093c69ae5dbcb2773122e1bbf4fcb1d7a16550a/%40i18n/ja/docs/references/http-websocket-apis/api-conventions/request-formatting.md#api%E3%81%AE%E3%83%90%E3%83%BC%E3%82%B8%E3%83%A7%E3%83%B3%E7%AE%A1%E7%90%86), but it needs updating for the changes I made.

### DIFF
--- a/docs/references/http-websocket-apis/api-conventions/request-formatting.md
+++ b/docs/references/http-websocket-apis/api-conventions/request-formatting.md
@@ -99,4 +99,47 @@ The commandline always uses the latest [API version](#api-versioning).
 
 {% admonition type="warning" name="Caution" %}The commandline interface is intended for administrative purposes only and is _not a supported API_. New versions of `rippled` may introduce breaking changes to the commandline API without warning!{% /admonition %}
 
+## API Versioning
+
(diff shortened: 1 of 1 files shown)


oeggert:
In that PR the information was moved into the [parent index page](https://xrpl.org/docs/references/http-websocket-apis). I believe the info was moved after discussing the API v2 changes.

mDuo13:
Ah. Then, this PR needs to take a very different form (updating all the links to point to the new section). Will open a new one to do that.

PR #3234, opened 2025-08-13
Javascript Get Started with code walkthrough

maria-robobug:
Preview link [here](https://xrpl-dev-portal--stepped-tutorial-demo.preview.redocly.app/docs/tutorials/javascript/build-apps/get-started)


_code-samples/get-started/js/base.html
```diff
@@ -0,0 +1,17 @@
...
+    <script src="https://unpkg.com/xrpl/build/xrpl-latest-min.js"></script>
+    <!-- @chunk-end -->
+  </head>
+  <body>
+    <h1>XRPL.js Example</h1>
+    <!-- @chunk {"steps": ["import-web-tag"]} -->
+    <script>
+      const xrpl = require("xrpl")
+    </script>
+    <!-- @chunk-end -->
```
mDuo13: `require(...)` is Node.js syntax, and you shouldn't need this at all for browsers anyway. The `<script>` tag should provide it on its own.

_code-samples/get-started/js/base.js
```diff
@@ -1,5 +1,5 @@
 // In browsers, use a <script> tag. In Node.js, uncomment the following line:
-// const xrpl = require('xrpl')
+const xrpl = require('xrpl')
```
mDuo13: As discussed, let's migrate to ES Module syntax. That means the full text of `base.js` becomes:

```js
// In browsers, use a <script> tag instead of the following line:
import xrpl from 'xrpl'

// Define the network client
const client = new xrpl.Client("wss://s.altnet.rippletest.net:51233")
await client.connect()

// ... custom code goes here

// Disconnect when done (If you omit this, Node.js won't end the process)
client.disconnect()
```

The `package.json` also needs to be updated to support this:

```json
{
  "dependencies": {
    "xrpl": "^4.4.0"
  },
  "type": "module"
}
```

mDuo13: FWIW I put these changes in a commit here: a46633a3b4a888f737cb4e9b1fde3e8bb69a28ad
Feel free to cherry pick if desired. I didn't change any tutorials that use this code, but I did confirm that the code runs if you add stuff to the "custom code goes here" section.

_code-samples/get-started/js/base.js
```diff
@@ -1,17 +1,11 @@
 // In browsers, use a <script> tag. In Node.js, uncomment the following line:
```
mDuo13: Since the following line is already uncommented, I think we need to adjust this.

```suggestion
// You can also use a <script> tag in browsers or require('xrpl') in Node.js
```

_code-samples/get-started/js/get-acct-info.js
```diff
@@ -1,39 +1,63 @@
...
-  const client = new xrpl.Client(SERVER_URL)
-  await client.connect()
+// @chunk {"steps": ["get-account-create-wallet-tag"]}
+// Create a wallet and fund it with the Testnet faucet:
+console.log("\nCreating a new wallet and funding it with Testnet XRP...")
+const fund_result = await client.fundWallet()
+const test_wallet = fund_result.wallet
+console.log(`Wallet: ${test_wallet.address}`)
+console.log(`Balance: ${fund_result.balance}`)
+console.log(`\nAccount Testnet Explorer URL:\nhttps://testnet.xrpl.org/accounts/${test_wallet.address}`)
```
mDuo13: The fact that this line is long (to col 104) matters more when displaying the code samples side by side, so it's better to break it up:

```suggestion
console.log('Account Testnet Explorer URL:')
console.log(`  https://testnet.xrpl.org/accounts/${test_wallet.address}`)
```

_code-samples/get-started/js/get-acct-info.js
```diff
@@ -1,39 +1,63 @@
...
-}
+// @chunk {"steps": ["listen-for-events-tag"]}
+// Listen to ledger close events
+console.log("\nListening for ledger close events...")
+client.request({
+  "command": "subscribe",
+  "streams": ["ledger"]
+})
+client.on("ledgerClosed", async (ledger) => {
+  console.log(`Ledger #${ledger.ledger_index} validated with ${ledger.txn_count} transactions!`)
```
mDuo13: I suggest splitting up this line for length too.

```suggestion
  console.log(`Ledger #${ledger.ledger_index} validated ` +
              `with ${ledger.txn_count} transactions!`)
```

docs/tutorials/javascript/build-apps/get-started.md
```diff
@@ -9,152 +7,294 @@ labels:
...
+### 1. Install Dependencies
 
-1. [Import the library.](#1-import-the-library)
-1. [Connect to the XRP Ledger.](#2-connect-to-the-xrp-ledger)
-1. [Get an account.](#3-get-account)
-1. [Query the XRP Ledger.](#4-query-the-xrp-ledger)
-1. [Listen for Events.](#5-listen-for-events)
+<!-- Web steps -->
+{% step id="import-web-tag" when={ "environment": "Web" } %}
+To load `xrpl.js` into your project, add a `<script>` tag such as the following to your HTML.
```
mDuo13: The text "such as the following" doesn't make sense anymore. With the way the highlighting works, I think we can just cut it.

```suggestion
To load `xrpl.js` into your project, add a `<script>` tag to your HTML.
```

mDuo13 (dismissed):
This looks pretty good, but I do see some room to polish it up.

1. The CSS could be better. We should save this for a follow-up task, but just making note of what I see:
    * The font color of the code sample's non-focused areas feels a bit _too_ low-contrast to me, especially since the entire code sample is "greyed out" much of the time. Some of the UI elements like the Environment selector are also low visibility in light mode.
    * The columns feel cramped to me and a little uneven too. Especially on my 1440p desktop monitor, there's wasted space due to the max width (1504px) of `.code-walkthrough` elements... I don't know why that's there, but removing it makes the fit a lot better on big monitors.
    * The margins between elements and the space for line numbers in the code sample feel a bit too small.
2. We should add a README to the `js/` folder that sends people to the page where they can download the clean example. We might also want to provide a separate, clean example file in the repository, without the `@chunk` comments. That would involve maintenance to keep them from diverging, though, so maybe not.
3. It seems wrong to have this buried under "Build Applications..." That's something we can and should address later when we move the tutorials around as planned. After we do that, we should also update the `_code-samples/get-started/README.md` file to have the new links to the tutorials.

Otherwise, I noted a few places where the old format or language needs to be updated just a little. This should be good to go with those changes.

mDuo13 (dismissed):

maria-robobug:
Holding off on merging this, as per @amarantha-k's suggestion to wait for feedback from the design team.

mDuo13:
Commit bb80ac5668d1cc52715b333d86972b8273d4b2bb should fix #3256, at least the high priority parts. (It also fixes a weird issue with tab sizing that was caused by a clash with our markdown list item styling.)

maria-robobug:
Thanks @mDuo13, nice work!

mDuo13 (dismissed):

maria-robobug:
A few issues I've noticed when testing:

1. In dark mode, the file tab in the walkthrough preview panel is showing a weird artefact next to it. This doesn't happen in light mode.
   
    dark-mode-artifact

2. Search pop up has the wrong colours:
   - Light Mode
     search-broken-light-mode
   - Dark Mode
     search-broken-dark-mode

3. Language picker seems affected too:
	- Light Mode
          language-picker-light-mode

	- Dark Mode
          language-picker-dark-mode

	
4. The first breadcrumb in the interactive tutorials has been pushed down (dark and light mode)
   Screenshot 2025-09-03 at 13 05 09

maria-robobug:
@amarantha-k / @mDuo13 I've tried to add some CSS to fix the massive gap between the documentation and code panel for widescreen monitors, please have a look and let me know what you think. 

https://xrpl-dev-portal--stepped-tutorial-demo.preview.redocly.app/docs/tutorials/javascript/build-apps/get-started

amarantha-k:
Thanks for fixing the wide screen, @maria-robobug! Looks good now.

amarantha-k (approved):

mDuo13 (approved):
Looks good to me. I don't have a widescreen monitor, but I tried stretching the screen out across two monitors and it seemed to use the space relatively well. :+1:

PR #3236, opened 2025-08-14
Fix blog landing bugs & add case study category

mDuo13:
- Add case study category to the blog. This fixes the issue where the recent Frii Pay article isn't showing up on the homepage now that it's not the hero post anymore.
    - The image is a copy of the one used for Developer Reflections. There's precedent for this in that the Release Notes and Amendments categories use the same image too.
- Fix bugs in the blog landing page where multiple elements were reusing the same unique ID.
- I didn't see a quick way to fix the general issues around blogs with unknown categories, so I added a reminder to the blogs sidebar file to document this limitation.


amarantha-k (approved):
lgtm

PR #3237, opened 2025-08-14
Update account_objects docs

mDuo13:
- Remove some extraneous fields that seem to have been accidentally copied into the account_objects request format
- Clarify that account_objects can return an empty list even if there are more objects, due to filtering.
- Move the list of "short" ledger entry names of types to a separate file, referenced by both account_objects and ledger_data.
    - Add a column for amendment status rather than having a bunch of different amendment disclaimers on the page. Left the column empty for functionality that's part of the current base protocol (including unconditionalized old amendments)
    - Update the amendment-disclaimer to have a "compact" option for cases like this where you want the status badge but not the full disclaimer
- Move the sample API responses out to snippet files.
- Fix incorrect alphabetical order (PayChannel after PermissionedDomain) in ledger entry type list.


amarantha-k (approved):
LGTM

PR #3230, opened 2025-08-12
Revise tutorial and sample code guidelines

mDuo13:
- Update "Tutorial Guidelines" with more specific, actionable guidelines for making tutorials. [⇒ Preview](https://xrpl-dev-portal--979c70.preview.redocly.app/resources/contribute-documentation/tutorial-guidelines)
- Remove the Tutorial Structure page, with relevant content added to the Tutorial Guidelines instead. Redirect from the old page to the new one.
- Add a **Tutorial Template**, representing the best practices for how tutorials should work, which you can copy as a starting point for new tutorials. Because it includes placeholders that Redocly considers broken links, it is an ignored file, similar to the blog template.
    - A previous version of this PR had more of the guidelines embedded into the tutorial as kind of a meta-tutorial. My main takeaway from @oeggert's review feedback was that this structure did not work as intended, so I moved the guidance into the "Tutorial Guidelines" page and made the template file more strictly _just_ a template.
- Refactor the `verify_credential.js` code sample into a linear script that is much less powerful but much simpler, as a demonstration of the guidelines for code samples. Update the tutorial accordingly, consistent with the guidelines.
    - Known issue: The Python code sample and tutorial have not been updated to match. This is left as-is because the two separate tutorials will need to be merged into a new, single tutorial (with tabs) as part of #3245. We will need to update the Python sample when we merge the tutorials.

PR #3241, opened 2025-08-15
Show two possible formats of Escrow sample code

mDuo13:
This is an experiment to collect feedback on preferred formats for sample code.

It creates two alternative code samples for sending a time-based escrow:

1. [`send-timed-escrow-linear.js`](https://github.com/XRPLF/xrpl-dev-portal/blob/91d63dfe090f884ef813c504d789a62a8ae32af2/_code-samples/escrow/js/send-timed-escrow-linear.js) puts as much action as possible at the top level of the file, so that it resolves in a very linear order. This lends itself well to tutorials that have numbered steps, and the control flow is simply "top to bottom".
2. [`send-timed-escrow-functions.js`](https://github.com/XRPLF/xrpl-dev-portal/blob/91d63dfe090f884ef813c504d789a62a8ae32af2/_code-samples/escrow/js/send-timed-escrow-functions.js) splits the action up into several functions and uses a `main()` function to call them. This shows the overall control flow first, and compartmentalizes the steps so that the parameters required at each step are more obvious. It's also a little closer to something you could import as a library/module. This lends itself more to a tutorial with a "code walkthrough" rather than numbered steps.

The third and fourth options, not presented here, are:

3. A commandline utility similar to [`verify_credential.js`](https://github.com/XRPLF/xrpl-dev-portal/tree/6f53d6301678a0b60815bafd06222d8edd225c7d/_code-samples/verify-credential/js) that parses user input for values instead of using hard-coded values.
4. A graphical interface such as the one in [Create Time Based Escrows Using JavaScript](https://xrpl.org/docs/tutorials/javascript/send-payments/create-time-based-escrows).

Currently we are inclined towards options 1 or 2 because they are easier to write and have fewer distractions from the XRPL-specific parts of the code.


mDuo13:
For the record, this is related to #3230 and the objectives and priorities laid out there. Specifically, the goals of this project are:

1. Modernize our sample code and tutorials, and reduce the maintenance burden involved in keeping them updated and consistent going forward.
2. Refine our standard format and improve consistency of the sample code so that it's easier to go from one tutorial to another without getting confused because of how different they are.
3. Produce a larger quantity of tutorials and sample code, with a mind towards enabling large language models (LLMs) to generate more high-quality code that uses the XRP Ledger.

    **Personal Aside:** I _really_ don't recommend "vibe coding" with blockchains; security issues and other mistakes can be subtle and costly, and I've personally seen AIs talk about following proper security practices while simultaneously producing code that violates those practices and puts your secret keys at risk. But since people are going to be using code-generation tools regardless, we may as well do what we can to help them work better.

shortthefomo:
+1 for option one with the STEPS. As previous examples have been like this and switching that up now not the best IMO.
Also makes it clear do this then that.... vs functions one misses that.

krkmu:
+1 for option 2 — that’s the approach I usually take for [my scripts](https://github.com/RippleDevRel/xrpl-js-python-simple-scripts/blob/main/devnet/tokenEscrow.js) and tutorials aimed at developers during hackathons.  
From the feedback I’ve received, developers tend to prefer the multi-step approach with clearly defined functions for each piece of logic. It makes it easier for them to reuse specific functions in their apps and better understand the overall logic flow.  
For simpler flows (e.g. payments or trustlines), option 1 is more straightforward.

pkcs8:
I'll vote for the second approach (`send-timed-escrow-functions.js`) where functions are compartmentalized and have clearly defined scope of what they're doing on-ledger.

mDuo13:
See #3351 and #3358 for the continuation of this work.

PR #3244, opened 2025-08-18
[JA] Fix two issues with tables

mDuo13:
- Fix several instances where tables weren't parsed correctly because of an extra space
- Fix unnecessary horizontal scrolling of tables in Japanese because the word-wrap rules didn't allow breaking of Japanese text.

| Before | After |
|---|---|
|  <br> (Note the horizontal scrollbar and that the text continues offscreen) |  <br> (Text is all visible without scrolling.) |
|  <br> (Table is unparsed) |  (Table appears as intended) |


oeggert (approved):
LGTM

PR #3251, opened 2025-08-20
update mpt docs for permissioneddomains

oeggert:
Updated `MPTTokenIssuanceCreate` and `MPTokenIssuanceSet` transaction docs to include `DomainID` field. Also added additional error codes.

Documents changes to MPT that are gated by Single Asset Vaults: https://github.com/XRPLF/rippled/pull/5509

### DIFF
--- a/docs/references/protocol/transactions/types/mptokenissuancecreate.md
+++ b/docs/references/protocol/transactions/types/mptokenissuancecreate.md
@@ -13,6 +13,7 @@ If the transaction is successful, it creates an [MPTokenIssuance entry][] where 
 
 {% amendment-disclaimer name="MPTokensV1" /%}
 
+
 ## Example MPTokenIssuanceCreate JSON
 
 This example assumes that the issuer of the token is the signer of the transaction.
@@ -24,6 +25,7 @@ This example assumes that the issuer of the token is the signer of the transacti
   "AssetScale": 4,
   "TransferFee": 0,
   "MaximumAmount": "50000000",
+  "Flags": 83659,
   "MPTokenMetadata": "7B2274223A225442494C4C222C226E223A22542D42696C6C205969656C6420546F6B656E222C2264223A2241207969656C642D62656172696E6720737461626C65636F696E206261636B65642062792073686F72742D7465726D20552E532E205472656173757269657320616E64206D6F6E6579206D61726B657420696E737472756D656E74732E222C2269223A226578616D706C652E6F72672F7462696C6C2D69636F6E2E706E67222C226163223A22727761222C226173223A227472656173757279222C22696E223A224578616D706C65205969656C6420436F2E222C227573223A5B7B2275223A226578616D706C657969656C642E636F2F7462696C6C222C2263223A2277656273697465222C2274223A2250726F647563742050616765227D2C7B2275223A226578616D706C657969656C642E636F2F646F6373222C2263223A22646F6373222C2274223A225969656C6420546F6B656E20446F6373227D5D2C226169223A7B22696E7465726573745F72617465223A22352E303025222C22696E7465726573745F74797065223A227661726961626C65222C227969656C645F736F75726365223A22552E532E2054726561737572792042696C6C73222C226D617475726974795F64617465223A22323034352D30362D3330222C226375736970223A22393132373936525830227D7D",
   "Fee": "12",
   "Flags": 122,
@@ -38,6 +40,7 @@ This example assumes that the issuer of the token is the signer of the transacti
 | Field             | JSON Type            | [Internal Type][] | Required? | Description |
 |:------------------|:---------------------|:------------------|:----------|:------------|
 | `AssetScale`      | Number               | UInt8             | No        | Where to put the decimal place when displaying amounts of this MPT. More formally, the asset scale is a non-negative integer (0, 1, 2, …) such that one standard unit equals 10^(-scale) of a corresponding fractional unit. For example, if a US Dollar Stablecoin has an asset scale of _2_, then 1 unit of that MPT would equal 0.01 US Dollars. This indicates to how many decimal places the MPT can be subdivided. If omitted, the default is 0, meaning that the MPT cannot be divided into smaller than 1 unit. |
+| `DomainID`        | String - [Hash][]    | UInt256           | No        | The ledger entry ID of a permissioned domain that grants access to the MPT. You must enable the `tfMPTRequireAuth` flag to use permissioned domains. {% amendment-disclaimer name="PermissionedDomains" /%} {% amendment-disclaimer name="SingleAssetVault" /%} |
 | `TransferFee`     | Number               | UInt16            | No        | The value specifies the fee to charged by the issuer for secondary sales of the Token, if such sales are allowed. Valid values for this field are between 0 and 50,000 inclusive, allowing transfer rates of between 0.000% and 50.000% in increments of 0.001. The field _must not_ be present if the tfMPTCanTransfer flag is not set. If it is, the transaction should fail and a fee should be claimed. |
 | `MaximumAmount`   | String - Number      | UInt64            | No        | The maximum asset amount of this token that can ever be issued, as a base-10 number encoded as a string. The current default maximum limit is 9,223,372,036,854,775,807 (2^63-1). _This limit may increase in the future. If an upper limit is required, you must specify this field._ |
 | `MPTokenMetadata` | String - Hexadecimal | Blob              | No        | Arbitrary metadata about this issuance. The limit for this field is 1024 bytes. By convention, the metadata should decode to JSON data describing what the MPT represents. The [XLS-89 specification](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0089-multi-purpose-token-metadata-schema) defines a recommended format for metadata. |
@@ -45,6 +48,7 @@ This example assumes that the issuer of the token is the signer of the transacti
 {% admonition type="success" name="Tip" %}
 For an example of how to encode metadata for the `MPTokenMetadata` field, see {% repo-link path="_code-samples/issue-mpt-with-metadata/" %}Code Sample: Issue MPT with Metadata{% /repo-link %}.
 {% /admonition %}
+
 
 ## MPTokenIssuanceCreate Flags
 
@@ -58,6 +62,20 @@ Transactions of the MPTokenIssuanceCreate type support additional values in the 
 | `tfMPTCanTrade`    | `0x00000010` | `16`          | If set, indicates that individual holders can trade their balances using the XRP Ledger DEX. |
 | `tfMPTCanTransfer` | `0x00000020` | `32`          | If set, indicates that tokens can be transferred to other accounts that are not the issuer. |
 | `tfMPTCanClawback` | `0x00000040` | `64`          | If set, indicates that the issuer can use the `Clawback` transaction to claw back value from individual holders. |
+
+
+## Error Cases
+
+Besides errors that can occur for all transactions, {% $frontmatter.seo.title %} transactions can result in the following [transaction result codes](../transaction-results/index.md):
+
+| Error Code                | Description |
+|:--------------------------|:------------|
+| `tecDIR_FULL`             | The owner directory of the account creating the `MPTokenIssuance` ledger entry is full. |
+| `temBAD_TRANSFER_FEE`     | The transfer fee specified is greater than the maximum allowed value of 50,000. |
+| `temDISABLED`             | The `MPTokensV1` amendment is disabled. You will also receive this error if you include a `DomainID` field in the transaction, but the `PermissionedDomains` and `SingleAssetVault` amendments are both disabled. |
+| `tecINSUFFICIENT_RESERVE` | The account creating the `MPTokenIssuance` ledger entry doesn't have enough XRP to meet the owner reserve. |
+| `temMALFORMED`            | Besides generally malformed transactions, you can receive this error if:<br>- A non-zero transfer fee is set, but the `tfMPTCanTransfer` flag is _not_ set.<br>- The `DomainID` points to an invalid permissioned domain; you can also receive this error if you include a `DomainID` without setting the `tfMPTRequireAuth` flag.<br>- The `MPTokenMetadata` field is an invalid length (0 or exceeds 1024 bytes).<br>- The `MaximumAmount` field is 0 or exceeds 9,223,372,036,854,775,807 (2^63-1). |
+
 
 ## See Also
 
--- a/docs/references/protocol/transactions/types/mptokenissuanceset.md
+++ b/docs/references/protocol/transactions/types/mptokenissuanceset.md
@@ -11,6 +11,7 @@ Update a mutable property of a [Multi-purpose Token (MPT)](../../../../concepts/
 
 {% amendment-disclaimer name="MPTokensV1" /%}
 
+
 ## Example MPTokenIssuanceSet JSON
 
 This example locks the balances of all holders of the specified MPT issuance.
@@ -32,8 +33,10 @@ This example locks the balances of all holders of the specified MPT issuance.
 
 | Field               | JSON Type            | [Internal Type][] | Required? | Description |
 |:--------------------|:---------------------|:------------------|:----------|-------------|
+| `DomainID`          | String - [Hash][]    | UInt256           | No        | The ledger entry ID of a permissioned domain that grants access to the MPT. An empty value or `0` removes the permissioned domain from the MPT issuance so that only users who are explicitly approved by the issuer can send and receive the MPT. You can only set a `DomainID` if the MPT issuance has [**Require Auth**](/docs/concepts/tokens/fungible-tokens/multi-purpose-tokens#transferability-controls) enabled. {% amendment-disclaimer name="PermissionedDomains" /%} {% amendment-disclaimer name="SingleAssetVault" /%} |
 | `MPTokenIssuanceID` | String - Hexadecimal | UInt192           | Yes       | The identifier of the `MPTokenIssuance` to update. |
 | `Holder`            | String - [Address][] | AccountID         | No        | An individual token holder. If provided, apply changes to the given holder's balance of the given MPT issuance. If omitted, apply to all accounts holding the given MPT issuance. |
+
 
 ### MPTokenIssuanceSet Flags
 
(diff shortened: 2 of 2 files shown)


docs/references/protocol/transactions/types/mptokenissuancecreate.md
```diff
@@ -54,4 +55,13 @@ Transactions of the MPTokenIssuanceCreate type support additional values in the
...
 | `tfMPTCanClawback` | `0x00000040` | `64`          | If set, indicates that the issuer can use the `Clawback` transaction to claw back value from individual holders. |
 
+## Error Cases
+
+Besides errors that can occur for all transactions, {% $frontmatter.seo.title %} transactions can result in the following [transaction result codes](../transaction-results/index.md):
+
+| Error Code                | Description |
+|:--------------------------|:------------|
+| `temDISABLED`             | Either the `PermissionedDomains` or `SingleAssetVault` amendment is disabled. |
+| `temMALFORMED`            | The `DomainID` is `0` or empty; you can also receive this error if you include a `DomainID` without setting the `tfMPTRequireAuth` flag. |
```
mDuo13: I see some other error codes that can occur [in the source](https://github.com/XRPLF/rippled/blob/develop/src/xrpld/app/tx/detail/MPTokenIssuanceCreate.cpp#L29-L111), such as `tecINSUFFICIENT_RESERVE` and `temBAD_TRANSFER_FEE`.

docs/references/protocol/transactions/types/mptokenissuanceset.md
```diff
@@ -39,4 +40,14 @@ Transactions of the `MPTokenIssuanceSet` type support additional values in the `
...
 
+## Error Cases
+
+Besides errors that can occur for all transactions, {% $frontmatter.seo.title %} transactions can result in the following [transaction result codes](../transaction-results/index.md):
+
+| Error Code                | Description |
+|:--------------------------|:------------|
+| `tecNO_PERMISSION`        | You are trying to lock or unlock an MPT, but the `tfMPTCanLock` flag wasn't set in the `MPTokenIssuanceCreate` transaction. | 
+| `temDISABLED`             | Either the `PermissionedDomains` or `SingleAssetVault` amendment is disabled. |
+| `temMALFORMED`            | You specified a `DomainID` and `Holder` value. You can only set one or the other in a single transaction. You can also receive this error if the transaction isn't changing anything; the transaction must either update a flag or modify the `DomainID`. |
```
mDuo13: `tecOBJECT_NOT_FOUND` and `tecNO_DST` are examples of a couple other type-specific errors that can occur with this transaction, based on a quick skim of [the source](https://github.com/XRPLF/rippled/blob/afc05659edbc22d04b13402f157832778556f7b7/src/xrpld/app/tx/detail/MPTokenIssuanceSet.cpp#L138).

mDuo13 (commented):
To be clear, this is set to go into `rippled` version 3.0.0, not version 2.6.0 which is the next planned release?

Bronek:
> To be clear, this is set to go into `rippled` version 3.0.0, not version 2.6.0 which is the next planned release?

Yes. Although the changes are in release 2.6.0, they are amendment gated by `SingleAssetVault` which, in release 2.6.0 is not supported hence cannot be voted (and enabled). This means the changes will not be enabled until **after** 3.0.0 release (which will flip the `Supported` flag) and obviously, after the amendment has been voted.

docs/references/protocol/transactions/types/mptokenissuancecreate.md
Bronek: I am not sure how the reader will interpret `... or empty`. The `DomainID` is an optional field, many users might not even want to use it. **IF** they do use it, it must point to a valid permissioned domain.

docs/references/protocol/transactions/types/mptokenissuancecreate.md
```diff
@@ -54,4 +55,13 @@ Transactions of the MPTokenIssuanceCreate type support additional values in the
 | `tfMPTCanTransfer` | `0x00000020` | `32`          | If set, indicates that tokens can be transferred to other accounts that are not the issuer. |
 | `tfMPTCanClawback` | `0x00000040` | `64`          | If set, indicates that the issuer can use the `Clawback` transaction to claw back value from individual holders. |
 
+## Error Cases
+
+Besides errors that can occur for all transactions, {% $frontmatter.seo.title %} transactions can result in the following [transaction result codes](../transaction-results/index.md):
+
+| Error Code                | Description |
+|:--------------------------|:------------|
+| `temDISABLED`             | Either the `PermissionedDomains` or `SingleAssetVault` amendment is disabled. |
```
Bronek: `PermissionedDomain` does not have to be enabled for `MPTokenIssuanceCreate` to succeed. It only needs to be enabled if the user tried to use `DomainID` field.

docs/references/protocol/transactions/types/mptokenissuanceset.md
```diff
@@ -27,6 +27,7 @@ _(Requires the [MPTokensV1 amendment][] {% not-enabled /%}.)_
 
 | Field               | JSON Type            | [Internal Type][] | Required? | Description |
 |:--------------------|:---------------------|:------------------|:----------|-------------|
+| `DomainID`          | String - [Hash][]    | UInt256           | No        | The ledger entry ID of a permissioned domain that will manage access to the MPT. An empy value or `0` will remove permissioned domain access management. Any accounts that lose access to the MPT because of `DomainID` updates lose the ability to send or receive MPTs they already hold. The `tfMPTRequireAuth` flag must have been set in the `MPTokenIssuanceCreate` transaction to use permissioned domains. _(Requires the [PermissionedDomains amendment][] and [SingleAssetVault amendment][])_ |
```
Bronek: `receive MPTs they already hold` I suggest adding something along the lines "unless they have been explicitly authorized by the MPT issuer"

docs/references/protocol/transactions/types/mptokenissuancecreate.md
Bronek: Also, there are other possible reasons to get `temMALFORMED` error

docs/references/protocol/transactions/types/mptokenissuancecreate.md
oeggert: Updated to differentiate the amendments required when including the `DomainID` field.

docs/references/protocol/transactions/types/mptokenissuancecreate.md
oeggert: @mDuo13 I don't think I need to include `tecINSUFFICIENT_RESERVE` since it's a common error and the reason for getting it in this case isn't unique to MPTs.

I did add `temBAD_TRANSFER_FEE` though.

oeggert: On second thought, I will add in the insufficient reserve since it's nice to see what object is adding to the reserve.

docs/references/protocol/transactions/types/mptokenissuancecreate.md
```diff
@@ -37,6 +37,7 @@ This example assumes that the issuer of the token is the signer of the transacti
 | Field             | JSON Type            | [Internal Type][] | Required? | Description |
 |:------------------|:---------------------|:------------------|:----------|:------------|
 | `AssetScale`      | Number               | UInt8             | No        | Where to put the decimal place when displaying amounts of this MPT. More formally, the asset scale is a non-negative integer (0, 1, 2, …) such that one standard unit equals 10^(-scale) of a corresponding fractional unit. For example, if a US Dollar Stablecoin has an asset scale of _2_, then 1 unit of that MPT would equal 0.01 US Dollars. This indicates to how many decimal places the MPT can be subdivided. If omitted, the default is 0, meaning that the MPT cannot be divided into smaller than 1 unit. |
+| `DomainID`        | String - [Hash][]    | UInt256           | No        | The ledger entry ID of a permissioned domain that will manage access to the MPT. You must set the `tfMPTRequireAuth` flag to use permissioned domains.<br>{% amendment-disclaimer name="PermissionedDomains" /%}<br>{% amendment-disclaimer name="SingleAssetVault" /%} |
```
mDuo13: Instead of using future tense, try phrasing it like this:

```suggestion
| `DomainID`        | String - [Hash][]    | UInt256           | No        | The ledger entry ID of a permissioned domain that grants access to the MPT. You must set the `tfMPTRequireAuth` flag to use permissioned domains.<br>{% amendment-disclaimer name="PermissionedDomains" /%}<br>{% amendment-disclaimer name="SingleAssetVault" /%} |
```

docs/references/protocol/transactions/types/mptokenissuanceset.md
```diff
@@ -27,6 +27,7 @@ _(Requires the [MPTokensV1 amendment][] {% not-enabled /%}.)_
 
 | Field               | JSON Type            | [Internal Type][] | Required? | Description |
 |:--------------------|:---------------------|:------------------|:----------|-------------|
+| `DomainID`          | String - [Hash][]    | UInt256           | No        | The ledger entry ID of a permissioned domain that will manage access to the MPT. An empy value or `0` will remove permissioned domain access management. Any accounts that lose access to the MPT because of `DomainID` updates--unless explicitly authorized by the MPT issuer--lose the ability to send or receive MPTs they already hold. The `tfMPTRequireAuth` flag must have been set in the `MPTokenIssuanceCreate` transaction to use permissioned domains.<br>{% amendment-disclaimer name="PermissionedDomains" /%}<br>{% amendment-disclaimer name="SingleAssetVault" /%} |
```
mDuo13: Same note as on create.

docs/references/protocol/transactions/types/mptokenissuanceset.md
```diff
@@ -39,4 +40,17 @@ Transactions of the `MPTokenIssuanceSet` type support additional values in the `
...
 | `tfMPTUnlock`      | `0x00000002` | 2             | Enable to unlock balances of this MPT issuance. |
 
+## Error Cases
+
+Besides errors that can occur for all transactions, {% $frontmatter.seo.title %} transactions can result in the following [transaction result codes](../transaction-results/index.md):
+
+| Error Code                | Description |
+|:--------------------------|:------------|
+| `temDISABLED`             | The `MPTokensV1` amendment is disabled. You will also receive this error if you include a `DomainID` field in the transaction, but the `PermissionedDomains` and `SingleAssetVault` amendments are both disabled. |
+| `tecNO_DELEGATE_PERMISSION` | You are attempting to use a delegate account with insufficient permissions to submit this transaction. |
```
mDuo13: I think this code is not specific to MPTokenIssuanceSet, and it's been replaced with a `ter` code in the next release also (relating to the bug that is the reason for `PermissionDelegationV1_1`)

docs/references/protocol/transactions/types/mptokenissuanceset.md
```diff
@@ -39,4 +40,17 @@ Transactions of the `MPTokenIssuanceSet` type support additional values in the `
...
+| `tecNO_DST`               | The account specified in the `Holder` field doesn't exist. |
+| `tecNO_PERMISSION`        | - The `lsfMPTCanLock` flag isn't enabled, but you are attempting to lock or unlock an MPT.<br>- The `SingleAssetVault` amendment is disabled and you're trying to modify a `DomainID` field. |
```
mDuo13, line 59: Stray bullet point?

```suggestion
| `tecNO_PERMISSION`        | The `lsfMPTCanLock` flag isn't enabled, but you are attempting to lock or unlock an MPT.<br>- The `SingleAssetVault` amendment is disabled and you're trying to modify a `DomainID` field. | 
```

docs/references/protocol/transactions/types/mptokenissuanceset.md
```diff
@@ -39,4 +40,17 @@ Transactions of the `MPTokenIssuanceSet` type support additional values in the `
...
+
+| Error Code                | Description |
+|:--------------------------|:------------|
+| `temDISABLED`             | The `MPTokensV1` amendment is disabled. You will also receive this error if you include a `DomainID` field in the transaction, but the `PermissionedDomains` and `SingleAssetVault` amendments are both disabled. |
+| `tecNO_DELEGATE_PERMISSION` | You are attempting to use a delegate account with insufficient permissions to submit this transaction. |
+| `tecNO_DST`               | The account specified in the `Holder` field doesn't exist. |
+| `tecNO_PERMISSION`        | - The `lsfMPTCanLock` flag isn't enabled, but you are attempting to lock or unlock an MPT.<br>- The `SingleAssetVault` amendment is disabled and you're trying to modify a `DomainID` field. | 
+| `temMALFORMED`            | - You specified a `DomainID` and `Holder` value--only one can be set in a single transaction.<br>- You specified the same account for both `Acount` and `Holder`.<br>- The transaction isn't changing anything; it must either update a flag or modify the `DomainID`. |
```
mDuo13: `temMALFORMED` is a code that can occur for any transaction type.

```suggestion
```

docs/references/protocol/transactions/types/mptokenissuancecreate.md
```diff
@@ -60,6 +61,20 @@ Transactions of the MPTokenIssuanceCreate type support additional values in the
...
+
+| Error Code                | Description |
+|:--------------------------|:------------|
+| `tecDIR_FULL`             | The owner directory of the account creating the `MPTokenIssuance` ledger entry is full. |
+| `temBAD_TRANSFER_FEE`     | The transfer fee specified is greater than the maximum allowed value of 50,000. |
+| `temDISABLED`             | The `MPTokensV1` amendment is disabled. You will also receive this error if you include a `DomainID` field in the transaction, but the `PermissionedDomains` and `SingleAssetVault` amendments are both disabled. |
+| `tecINSUFFICIENT_RESERVE` | The account creating the `MPTokenIssuance` ledger entry doesn't have enough XRP to meet the owner reserve. |
+| `temMALFORMED`            | - A non-zero transfer fee is set, but the `tfMPTCanTransfer` flag is _not_ set.<br>- The `DomainID` points to an invalid permissioned domain; you can also receive this error if you include a `DomainID` without setting the `tfMPTRequireAuth` flag.<br>- The `MPTokenMetadata` field is an invalid length (0 or exceeds 1024 bytes).<br>- The `MaximumAmount` field is 0 or exceeds 9,223,372,036,854,775,807 (2^63-1). |
```
mDuo13: `temMALFORMED` is one of the errors that can occur for any transaction type.

```suggestion
```

mDuo13 (changes requested):

docs/references/protocol/transactions/types/mptokenissuancecreate.md
oeggert: True, but I think it's helpful to explain what specifically you could be malforming to make it easier for someone to troubleshoot.

Maybe I could preface it with something like: "Besides generally malformed transactions, you can receive this error if ..."

docs/references/protocol/transactions/types/mptokenissuanceset.md
oeggert: Good catch, removing this line.

docs/references/protocol/transactions/types/mptokenissuanceset.md
oeggert: I've got 2 bullet points in this description.

docs/references/protocol/transactions/types/mptokenissuanceset.md
oeggert: Same comment in issuancecreate. I think it's helpful to explain how you'd malform this specific transaction.

docs/references/protocol/transactions/types/mptokenissuancecreate.md
mDuo13: There are infinite ways to creatively malform a transaction, but if there are common uses or patterns that hit the specific cases described here, then I agree it makes sense to note it here with a preface like you suggested.

docs/references/protocol/transactions/types/mptokenissuancecreate.md
```diff
@@ -39,6 +39,7 @@ This example assumes that the issuer of the token is the signer of the transacti
 | Field             | JSON Type            | [Internal Type][] | Required? | Description |
 |:------------------|:---------------------|:------------------|:----------|:------------|
 | `AssetScale`      | Number               | UInt8             | No        | Where to put the decimal place when displaying amounts of this MPT. More formally, the asset scale is a non-negative integer (0, 1, 2, …) such that one standard unit equals 10^(-scale) of a corresponding fractional unit. For example, if a US Dollar Stablecoin has an asset scale of _2_, then 1 unit of that MPT would equal 0.01 US Dollars. This indicates to how many decimal places the MPT can be subdivided. If omitted, the default is 0, meaning that the MPT cannot be divided into smaller than 1 unit. |
+| `DomainID`        | String - [Hash][]    | UInt256           | No        | The ledger entry ID of a permissioned domain that grants access to the MPT. You must set the `tfMPTRequireAuth` flag to use permissioned domains.<br>{% amendment-disclaimer name="PermissionedDomains" /%}<br>{% amendment-disclaimer name="SingleAssetVault" /%} |
```
mDuo13: 1. You don't need `<br>` tags here, since the disclaimer already uses its own `<p>` tag.
2. Use "enable" instead of "set" for clarity when talking about flags. ("Set? To what?")
3. This is currently (as expected) throwing errors because the SingleAssetVault amendment isn't in the VHS data yet. We should make a note to come back and double-check this before merging to the master branch, closer to release—but it's not a blocker for merging to the `rippled-3.0.0` branch.



```suggestion
| `DomainID`        | String - [Hash][]    | UInt256           | No        | The ledger entry ID of a permissioned domain that grants access to the MPT. You must enable the `tfMPTRequireAuth` flag to use permissioned domains. {% amendment-disclaimer name="PermissionedDomains" /%} {% amendment-disclaimer name="SingleAssetVault" /%} |
```

docs/references/protocol/transactions/types/mptokenissuanceset.md
```diff
@@ -32,6 +32,7 @@ This example locks the balances of all holders of the specified MPT issuance.
 
 | Field               | JSON Type            | [Internal Type][] | Required? | Description |
 |:--------------------|:---------------------|:------------------|:----------|-------------|
+| `DomainID`          | String - [Hash][]    | UInt256           | No        | The ledger entry ID of a permissioned domain that grants access to the MPT. An empy value or `0` will remove permissioned domain access management. Any accounts that lose access to the MPT because of `DomainID` updates--unless explicitly authorized by the MPT issuer--lose the ability to send or receive MPTs they already hold. The `tfMPTRequireAuth` flag must have been set in the `MPTokenIssuanceCreate` transaction to use permissioned domains.<br>{% amendment-disclaimer name="PermissionedDomains" /%}<br>{% amendment-disclaimer name="SingleAssetVault" /%} |
```
mDuo13: 1. Typo ("empy")
2. Avoid future tense.
3. Use an actual emdash instead of two hyphens. (Here's one you can copy-paste: — ) Or better yet, avoid phrasing sentences with this type of interjection in the first place.
4. The description feels a little confusing about when you lose the ability to send or receive MPTs.
5. The phrase "permissioned domain access management" feels a little ambiguous or confusing, like it's controlling who has access to manage the domain itself, and I don't think it's defined anywhere in those exact terms. Either rephrase into simpler language or shorten this description and link to a better description on a concept page.
6. Rather than talk about the exact transaction flag from the create transaction, talk about the current state of the MPT issuance (among other reasons, maybe that flag is mutable).

```suggestion
| `DomainID`          | String - [Hash][]    | UInt256           | No        | The ledger entry ID of a permissioned domain that grants access to the MPT. An empty value or `0` removes the permissioned domain from the MPT issuance so that only users who are explicitly approved by the issuer can send and receive the MPT. You can only set a `DomainID` if the MPT issuance has [**Require Auth**](/docs/concepts/tokens/fungible-tokens/multi-purpose-tokens#transferability-controls) enabled.) {% amendment-disclaimer name="PermissionedDomains" /%} {% amendment-disclaimer name="SingleAssetVault" /%} |
```

mDuo13: Note, I opened #3392 since it feels like we need better documentation on how to create and use allow-listing with MPTs (concept/tutorial/code sample/etc.).

mDuo13 (changes requested):

rucettimema32-del (approved):

mDuo13 (approved):

PR #3138, opened 2025-06-03
Updates to support tokens in payment channels and escrows

DennisDawson:
Branch with updates, will remain a draft until the amendment is up for voting. This provides documentation in context that can be linked from opensource.ripple.com.

PR #3255, opened 2025-08-24
Update account_lines.md - fixed the request parameter table

pkcs8:
Removed erroneously appearing `UInt32` from the request parameter table. Also fixed table formatting.


mDuo13 (approved):
Thanks. It looks like a few other errors may also have slipped through as part of #3138 so I'm going to take a closer look and make a follow-up to fix those too.

PR #3257, opened 2025-08-26
Payments page update

AKCodez:


docs/use-cases/payments/index.page.tsx
```diff
@@ -0,0 +1,318 @@
...
+      label: "Ripple",
+      url: "https://ripple.com/solutions/cross-border-payments/",
+      description: "Ripple Payments enables crypto companies, payment service providers and fintech to facilitate real-time cross-border payments using stablecoins, digital assets and local currencies — with XRPL as a foundational transaction layer.",
+      buttonText: "Case Study"
+    },
+    {
+      id: "fiipay",
+      label: "FiiPay",
+      url: "https://xrpl.org/blog/2025/frii-pay-xrpl-case-study-crypto-payment-solution",
+      description: "FiiPay connects XRPL-based crypto wallets to point-of-sale terminals, allowing customers to pay with RLUSD or XRP while helping merchants save costs on card processing fees.",
```
mDuo13: It's supposed to be "FriiPay" (as in, sounds like "free", not like "fee", I think)

```suggestion
      id: "friipay",
      label: "FriiPay",
      url: "https://xrpl.org/blog/2025/frii-pay-xrpl-case-study-crypto-payment-solution",
      description: "FriiPay connects XRPL-based crypto wallets to point-of-sale terminals, allowing customers to pay with RLUSD or XRP while helping merchants save costs on card processing fees.",
```

styles/light/_light-theme.scss
```diff
@@ -1367,3 +1367,112 @@ main article .card-grid {
...
+    .project-logo img {
+      &.coinpayments {
+        content: url('../img/uses/lightmode/payments/coinpayments.png');
+      }
+      
+      &.ripple {
+        content: url('../img/uses/lightmode/payments/ripple-black.png');
+      }
+      
+      &.fiipay {
```
mDuo13: Typo fix matching the above one

```suggestion
      &.friipay {
```

styles/light/_light-theme.scss
```diff
@@ -1367,3 +1367,112 @@ main article .card-grid {
...
+        content: url('../img/uses/lightmode/payments/friipay.png');
+      }
+    }
+  }
+
+
+}
```
mDuo13: File should end in newline

docs/index.page.tsx
```diff
@@ -191,7 +191,7 @@ function FlatCard(props: { href: string; title: string; description: string; lin
   );
 }
 
-function VideoCard(props: { url: string; title: string; src: string }) {
+export function VideoCard(props: { url: string; title: string; src: string }) {
```
mDuo13: I approve of making components that can be reused in multiple places, but it's probably bad practice to export them from random pages for use in other arbitrary pages. This should probably move to the `shared/components/` folder or something like that if you're going to reuse it.

Though, as best I can tell, this component isn't actually being reused anywhere else, so it doesn't really make sense to add `export` at this time.

docs/use-cases/payments/index.page.tsx
```diff
@@ -0,0 +1,318 @@
...
+          description: "",
+        },
+      ],
+    },
+    {
+      id: "reliable-infrastructure",
+      title: "Access Reliable Payments Infrastructure",
+      contents: [
+        {
+          subtitle: "100% uptime since 2012",
```
mDuo13: I don't know who wrote this copy, but I'd say it's misleading. Most people would consider the hour-long halt in February 2025 to be _not_ uptime, because you couldn't confirm new transactions even if the servers were up and you could still check for balances.

docs/use-cases/payments/index.page.tsx
```diff
@@ -0,0 +1,318 @@
...
+      id: "europ",
+      label: "EUROP",
+      url: "https://schuman.io/europ/", 
+      description: "EUROP, issued by Schuman Financial, is the first MiCA-compliant euro stablecoin on XRPL.",
+    },
+    {
+      id: "xsgd",
+      label: "XSGD",
+      url: "https://www.straitsx.com/xsgd",
+      description: "XSGD, from StraitsX, is a Singapore Dollar-backed stablecoin regulated by MAS.",
```
mDuo13: It's debatable, but I'd think it's helpful to clarify _some_ of the jargon and abbreviations in this section.

```suggestion
      description: "XSGD, from StraitsX, is a Singapore Dollar-backed stablecoin regulated by MAS (Monetary Authority of Singapore).",
```

docs/use-cases/payments/index.page.tsx
```diff
@@ -0,0 +1,318 @@
...
+      id: "usdb",
+      label: "USDB", 
+      url: "https://www.brazabank.com.br/en/usdben/",
+      description: "USDB, by Braza Group, is a USD-pegged stablecoin backed by U.S. and Brazilian bonds.",
+    },
+    {
+      id: "europ",
+      label: "EUROP",
+      url: "https://schuman.io/europ/", 
+      description: "EUROP, issued by Schuman Financial, is the first MiCA-compliant euro stablecoin on XRPL.",
```
mDuo13: I don't really know what MiCA-compliance means and the link doesn't really clarify, so probably at least some of our target audience doesn't know either. But that aside, this coin seems to be pretty consistently stylized as EURØP, so we should try to match that:

```suggestion
      label: "EURØP",
      url: "https://schuman.io/europ/", 
      description: "EURØP, issued by Schuman Financial, is the first MiCA-compliant euro stablecoin on XRPL.",
```

styles/_use-cases.scss
```diff
@@ -991,4 +943,557 @@ html.light {
...
+      &.coinpayments {
+        content: url('../img/uses/payments/coinpayments.png');
+        width: 99px;
+        height: 60px;
+      }
+      
+      &.ripple {
+        content: url('../img/uses/payments/ripple-white.png');
+        width: 100px;
+        height: 30px;
```
mDuo13: This image (both the light mode and dark mode versions) is only 26px tall so it gets unpleasantly stretched by this style.

```suggestion
        height: 26px;
```

docs/use-cases/payments/index.page.tsx
```diff
@@ -0,0 +1,318 @@
...
+                <div className="integration-column">
+                  <h3 className="integration-column__title">
+                    {translate("Work with a Partner")}
+                  </h3>
+                  <p className="integration-column__subtitle">
+                    {translate("Ideal for regulated institutions")}
+                  </p>
+                  <ul className="developer-tools__list">
+                    <FeatureItem
+                      link="https://discord.com/invite/KTNmhJDXqa"
```
mDuo13, line 302: I don't like that this link takes people to Discord without warning them. I'm not sure what to expect from the link but it should probably at least be treated as an external link rather than looking the same as the "Build It Yourself" links that stay on xrpl.org.

Aside from that, Discord doesn't strike me as an "institutional DeFi" type of gathering place. It presents itself as very "gamer" centric, and while some community developers are fine with that overlap, I'm sure at least some of them are uncomfortable with it or would see it as mixing professional and non-professional concerns.

So, regardless of whether we change the link, we should at least warn people accordingly.

shared/components/project-cards.tsx
```diff
@@ -0,0 +1,154 @@
...
+                <strong className="first-word">{firstWord}</strong>
+                {restOfText && <span className="rest-text"> {restOfText}</span>}
+              </>
+            );
+          })()}
+        </div>
+      )}
+      {!showCarousel && project.buttonText && (
+        <div className="project-button">
+          <a href={project.url} className="btn-arrow battle-tested-arrow">{translate(project.buttonText)}</a>
```
mDuo13: This is invalid markup and is causing a hydration error because this `<a>` element is a descendant of the card which is also an `<a>` element (started on line 28). Changing it to a span fixes the problem and doesn't affect it visually.

```suggestion
          <span className="btn-arrow battle-tested-arrow">{translate(project.buttonText)}</span>
```

mDuo13 (dismissed):
Mostly looks good, but there are a few typos and errors that should be fixed, and I also had some feedback that isn't with the implementation but should probably be passed up the chain regarding the text itself in some places.

AKCodez:
@mDuo13 resolved all your comments, relayed feedback regarding copy and discord links down the chain, will let you know the response!

docs/use-cases/payments/index.page.tsx
amarantha-k: Change to 99.99% uptime since 2012.

amarantha-k:
@AKCodez There is a merge conflict - could you resolve it please?

amarantha-k:
1. In light mode, the card and background color are the same after clicking the card . .Screenshot 2025-09-10 at 10 35 26 AM

2.  The "Ideal for..." sentence doesn't stand out as a sub-heading. 
Screenshot 2025-09-10 at 10 38 00 AM

amarantha-k (approved):
Thanks for fixing these issues. LGTM

PR #3263, opened 2025-08-29
Remove old/inadvertent TODO notes

mDuo13:
Neither of these TODOs is relevant at the moment and they should not be
visible to readers (even in the HTML comments).

(I think the Japanese page might not be in sync with the English version, also, but that's a separate concern.)


amarantha-k (approved):
lgtm

PR #3268, opened 2025-09-02
Rwa tokenization changes

AKCodez:
Branched off payments update page branch in order to make use of new shared components, this will be merged after the payments page updates


docs/use-cases/tokenization/real-world-assets.page.tsx
```diff
@@ -284,17 +275,16 @@ const marketIntegrationData = [
     description: (
       <>
         Utilize XRP Ledger's native{" "}
-        <Link to="/docs/concepts/tokens/decentralized-exchange">
+        <Link target="_blank" to="/docs/concepts/tokens/decentralized-exchange">
```
mDuo13: I don't think these in-site links should use new tabs.

amarantha-k (dismissed):
lgtm

styles/_pages.scss
```diff
@@ -2105,10 +2197,56 @@ html.light {
...
+      .card-description {
+        margin-bottom: 24px !important;
+        margin-top: 0 !important;
+        flex-grow: 0 !important;
+        padding: 0 !important;
+      }
+      
+      .card-links {
+        margin-top: 0 !important;
+        margin-bottom: 0 !important;
```
mDuo13: Having `!important` in so many instances here smells like bad code. It should work without all those if you just use a more specific wrapping class instead, right?

shared/components/developer-resources-section.tsx
```diff
@@ -24,14 +24,15 @@ export const DeveloperResourcesSection: React.FC<DeveloperResourcesSectionProps>
...
   return (
-    <div className={`developer-resources-section page-community ${className}`}>
+    <div className={` container developer-resources-section page-community ${className} ${isSingleCard ? 'single-card' : ''}`}>
       <section className="bottom-cards-section bug-bounty section-padding">
         {cards.map((card, index) => (
-          <div key={index} className="com-card">
+          <div key={index} className={`com-card ${isSingleCard ? 'single-card' : ''}`}>
             <img
-              className={`${index === 0 ? 'top-right-img' : 'bottom-right-img'} bug-bounty-card-bg${index === 0 ? '' : '-2'} ${card.backgroundClass || ''}`}
+              className={`${isSingleCard ? 'bottom-right-img' : (index === 0 ? 'top-right-img' : 'bottom-right-img')} ${isSingleCard ? 'bug-bounty-card-bg-2' : `bug-bounty-card-bg${index === 0 ? '' : '-2'}`} ${card.backgroundClass || ''}`}
```
mDuo13: Very minor nit: `bug-bounty` seems like an inappropriate name for this class, since on this page the section has nothing to do with a bug bounty.

mDuo13 (changes requested):
The page looks good, but I had just a few nitpick comments.

The only one I'll really insist on changing is the `target="_blank"` for non-external links.

shared/components/developer-resources-section.tsx
AKCodez: Re using existing class names from another page, we can let this one slide

styles/_pages.scss
AKCodez: Did not need them in the first place, good catch

docs/use-cases/tokenization/real-world-assets.page.tsx
AKCodez: removed

mDuo13 (dismissed):
I noticed an issue where the swirly images in the cards at the end of the page were blocking text selection because of the size of the image. You can see the area where text selection is blocked by the image:


I fixed the issue by converting them to be actual background images instead of using `` tags and cropping out the transparency (which also saves a few kilobytes of data). This involved refactoring the `DeveloperResourcesSection` component a little, but it turns out it already had an (unused) `backgroundClass` property so it was more of a matter of applying that property correctly.

With the changes, :+1: from me.

mDuo13 (approved):

PR #3274, opened 2025-09-05
Add downgrade rippled guide

xVet:
Adding downgrade guide path for rippled, currently we have only install and update. In cases where operators may need to downgrade this can be used as an reference.


intelliot:
I think we can change the upgrade guide to call out that you can specify the version number. That should be sufficient to have the existing guide serve as a reference for downgrade

xVet:
sounds good 👍

PR #3299, opened 2025-09-12
Add Get Started Walkthrough for Python

maria-robobug:
Preview here: https://xrpl-dev-portal--python-get-started-code-walkthrough.preview.redocly.app/docs/tutorials/python/build-apps/get-started


docs/tutorials/python/build-apps/get-started.md
```diff
@@ -1,210 +1,227 @@
...
+```
+{% /step %}
 
+### 2. Connect to the XRP Ledger
 
-### 1. Connect to the XRP Ledger
+{% step id="connect-tag" %}
+#### Connect to the XRP Ledger Testnet
 
 To make queries and submit transactions, you need to connect to the XRP Ledger. To do this with `xrpl-py`, use the [`xrp.clients` module](https://xrpl-py.readthedocs.io/en/latest/source/xrpl.clients.html):
```
mDuo13: This line needs to change. It doesn't make sense in context to end the paragraph with a colon anymore.

_code-samples/get-started/py/get-acct-info.py
```diff
@@ -1,34 +1,54 @@
...
 # Define the network client
 from xrpl.clients import JsonRpcClient
+from xrpl.wallet import generate_faucet_wallet
+from xrpl.core import addresscodec
+from xrpl.models.requests.account_info import AccountInfo
+import json
+
 JSON_RPC_URL = "https://s.altnet.rippletest.net:51234/"
 client = JsonRpcClient(JSON_RPC_URL)
+print("Connected to Testnet")
```
mDuo13: This print statement is inaccurate with how JSON-RPC works and potentially misleading in case of failure.

I assume you're copying from the JS version, but having a print statement in the equivalent place only works because the JS example uses WebSocket, which has a specific flow:

1. Create a WS Client instance
2. Establish a persistent connection to the network; if it fails, it'll throw an error
3. Print that you're connected

Since the Python example uses JSON-RPC, you don't have a persistent connection; the client connects separately each time you make an API call. Because of that, nothing happens when you instantiate the client with a particular URL: it won't throw an error here if you pass a URL that isn't a real, online JSON-RPC server.

So, by printing "Connected to Testnet" here, you're potentially misleading people into thinking that the connection actually works, when nothing has even tried connecting to anything yet. Which means if there's a problem with the Testnet servers—or the user changed the URL to one that won't work—it'll confusingly and inaccurately say that it connected even though it didn't, which will complicate the debugging process.

docs/tutorials/python/build-apps/get-started.md
```diff
@@ -1,210 +1,227 @@
...
-
-##### Derive an X-address
-
-You can use `xrpl-py`'s [`xrpl.core.addresscodec`](https://xrpl-py.readthedocs.io/en/latest/source/xrpl.core.addresscodec.html) module to derive an [X-address](https://xrpaddress.info/) from the `Wallet.address` field:
-
-{% code-snippet file="/_code-samples/get-started/py/get-acct-info.py" from="# Derive an x-address from the classic address:" before="# Look up info about your account" language="py" /%}
+You can use `xrpl-py`'s [`xrpl.core.addresscodec`](https://xrpl-py.readthedocs.io/en/latest/source/xrpl.core.addresscodec.html) module to derive an [X-address](https://xrpaddress.info/) from the `Wallet.address` field.
 
 The X-address format [packs the address and destination tag](https://github.com/XRPLF/XRPL-Standards/issues/6) into a more user-friendly value.
+{% /step %}
```
mDuo13: Hot take: we should consider removing this from the "Get Started" tutorial; I rarely see X-addresses being used in practice.

docs/tutorials/python/build-apps/get-started.md
```diff
@@ -1,210 +1,227 @@
...
 
+### 4. Query the XRP Ledger
 
-### 3. Query the XRP Ledger
-
+{% step id="query-xrpl-tag" %}
```
mDuo13: Wouldn't it be more appropriate to mark each step such that it _includes_ the header? (in this case, "4. Query the XRP Ledger") As it is, when someone clicks the text area around the step, the page scrolls so that the relevant header is just out of sight, which seems less than ideal.

docs/tutorials/python/build-apps/get-started.md
```diff
@@ -1,210 +1,227 @@
...
 
-Here, we use `xrpl-py`'s [`xrpl.account`](https://xrpl-py.readthedocs.io/en/latest/source/xrpl.account.html) module to look up information about the [account we got](#2-get-account) in the previous step.
-
-
-{% code-snippet file="/_code-samples/get-started/py/get-acct-info.py" from="# Look up info about your account" language="py" /%}
-
-
+Here, we use `xrpl-py`'s [`xrpl.account`](https://xrpl-py.readthedocs.io/en/latest/source/xrpl.account.html) module to look up information about the account we got in the previous step.
```
mDuo13: The sample code doesn't actually use the `xrpl.account` module anymore (it's not even imported!) so this sentence is out of date. It actually makes an account_info request using the request model to validate the format.

Also, per Microsoft style, we should use second person, not first person, for instructions. So this would more appropriately be written something like:

```suggestion
Use the [account_info method][] to look up information about the account you got in the previous step. Use a request model like `AccountInfo` to validate the request format and catch errors sooner.
```

docs/tutorials/python/build-apps/get-started.md
```diff
@@ -1,210 +1,227 @@
...
-git clone git@github.com:XRPLF/xrpl-dev-portal.git
-cd xrpl-dev-portal/_code-samples/get-started/py/get-acct-info.py
 python3 get-acct-info.py
 ```
 
 You should see output similar to this example:
 
 ```sh
-Classic address:
+Connected to Testnet
```
mDuo13: Remember to update this (per my note earlier about how this print statement is misleading)

docs/tutorials/python/build-apps/get-started.md
```diff
@@ -1,210 +1,227 @@
...
+    "validated": true
 }
 ```
 
-#### Interpreting the response
-
 The response fields that you want to inspect in most cases are:
 
-* `account_data.Sequence` — This is the sequence number of the next valid transaction for the account. You need to specify the sequence number when you prepare transactions. With `xrpl-py`, you can use the [`get_next_valid_seq_number`](https://xrpl-py.readthedocs.io/en/latest/source/xrpl.account.html#xrpl.account.get_next_valid_seq_number) to get this automatically from the XRP Ledger. See an example of this usage in the project [README](https://github.com/XRPLF/xrpl-py#serialize-and-sign-transactions).
+- `account_data.Sequence` — This is the sequence number of the next valid transaction for the account. You need to specify the sequence number when you prepare transactions. With `xrpl-py`, you can use the [`get_next_valid_seq_number`](https://xrpl-py.readthedocs.io/en/latest/source/xrpl.account.html#xrpl.account.get_next_valid_seq_number) to get this automatically from the XRP Ledger. See an example of this usage in the project [README](https://github.com/XRPLF/xrpl-py#serialize-and-sign-transactions).
```
mDuo13: I'm not sure this bullet point even needs to be here, because manually specifying the sequence number is usually not necessary. You do occasionally need to do it—for example, if you're preparing a transaction on an air-gapped, offline machine, or if you're bulk submitting more than one transaction per few seconds—but those use cases are the exception, not the rule, so maybe we can actually just delete this entire bullet point to keep things simple.

mDuo13 (changes requested):
The big picture looks good and I confirmed that you can download and run the code sample, with the expected results, but I noticed a few inaccuracies that should be corrected.

docs/tutorials/python/build-apps/get-started.md
mDuo13: It might actually make sense to note here that the default with xrpl-py is to use JSON-RPC, not WebSocket (because writing `async` code in Python is annoying and kind of tends to make you need to do everything async) but there _is_ an option for a WebSocket client if you need that functionality.

mDuo13 (commented):
Looks good. I think the build failure is probably unrelated to this PR. It does build successfully in local dev at least.

mDuo13 (approved):

PR #3305, opened 2025-09-16
Update community page content and links

AKCodez:


amarantha-k (changes requested):
The heading "Hot Topics Happening Now" seems misplaced after the specific hot topics are replaced with generic links. How about removing that and making "Join the Conversation" the heading instead?

AKCodez:
> The heading "Hot Topics Happening Now" seems misplaced after the specific hot topics are replaced with generic links. How about removing that and making "Join the Conversation" the heading instead?

updated

amarantha-k (approved):
lgtm

Issue #3260, opened 2025-08-27
Fix documentation of TokenEscrow amendment


mDuo13:
PR #3138 introduced a number of errors, only some of which have been corrected yet.

- Functionality added by the TokenEscrow amendment needs to be tagged as such since it's not available in production (#3187)
- Double check all the fields added and modified. Amount fields that can accept XRP, Tokens, and MPTs should be tagged consistently. In some cases the description of two fields was updated to reflect multiple new datatypes, but the data type column was updated for only one field.
- Confirm and correct capabilities of TokenEscrow (i.e. not payment channels)
- Look for other copy-paste errors and fields introduced to places they don't belong. Confirm everything with the source code in the latest stable `rippled` release.

mDuo13:
For reference, here is the PR adding the actual code for TokenEscrow to `rippled`: https://github.com/XRPLF/rippled/pull/5185

PR #3307, opened 2025-09-17
Add 'asfAllowTrustLineLocking' flag to accountset.md

RomThpt:


docs/references/protocol/transactions/types/accountset.md
```diff
@@ -74,6 +74,7 @@ The available AccountSet flags are:
 |:----------------------------------|:--------------|:--------------|
 | `asfAccountTxnID`                 | 5             | Track the ID of this account's most recent transaction. Required for [`AccountTxnID`](../common-fields.md#accounttxnid) |
 | `asfAllowTrustLineClawback`       | 16            | Allow account to claw back tokens it has issued. _(Requires the Clawback amendment.)_ Can only be set if the account has an empty owner directory (no trust lines, offers, escrows, payment channels, checks, or signer lists). After you set this flag, it cannot be reverted. The account permanently gains the ability to claw back issued assets on trust lines. |
+| `asfAllowTrustLineLocking`        | 17            | Allow account to lock trust lines for **IOU** tokens it has issued. **IOU** Tokens: If the issuer's account does not have the `lsfAllowTrustLineLocking` flag set, the transaction fails with `tecNO_PERMISSION`. |
```
Bronek, line 77: "the `EscrowCreate` transaction fails with"  perhaps ? I guess this can be polished.

mDuo13: Yeah, I think this one is missing some important context.

- "IOU" is specifically not a term we encourage since (a) tokens don't have to refer to amounts owed and (b) "IOU" has an unprofessional connotation
- "Locking" on trust lines (allowing amounts to be escrowed)  is completely unrelated to "locking" on MPTs (equivalent to deep freezing on trust lines). It's annoying that the terminology is like that, but we're stuck with it at this point.
- This field is not currently available on mainnet because it [requires the TokenEscrow amendment](https://github.com/XRPLF/rippled/blob/510314d34465324575719ce528bdda185581464b/src/xrpld/app/tx/detail/SetAccount.cpp#L654-L660).

mDuo13:
Noting this one for #3260

mDuo13 (changes requested):
I don't think we should merge this as currently worded, but also, we need to put in a bunch more fixes for the TokenEscrow docs.

docs/references/protocol/transactions/types/accountset.md
RomThpt: > "the `EscrowCreate` transaction fails with" perhaps ? I guess this can be polished.

something like that 

```md
| `asfAllowTrustLineLocking` | 17 | Allow account to lock trust lines for tokens it has issued. If the issuer's account does not have the `lsfAllowTrustLineLocking` flag set, the `EscrowCreate` transaction fails with `tecNO_PERMISSION`. (Requires the TokenEscrow amendment.) |

```

RomThpt:
Which specific wording issues should I address for this entry? And what other TokenEscrow docs need fixes - should we handle those in this PR or separately?

mDuo13:
#3260 describes the issues with TokenEscrow docs in more detail. @maria-robobug is working on that, so I'll let her follow up on this as needed.

maria-robobug:
This should be fixed now, made the change in https://github.com/XRPLF/xrpl-dev-portal/pull/3336, so we can close this PR.

https://xrpl.org/docs/references/protocol/transactions/types/accountset#accountset-flags

Issue #1353, opened 2022-03-02
More detailed docs on community credit use case


mDuo13:
The community credit use case is a place the current docs are a little lacking in describing what the XRPL is really capable of. In particular, some diagrams could really help to show how it works.

For some ideas, check out the "Beyond banking" section of [this old Ripple wiki article](https://web.archive.org/web/20150422094959/https://wiki.ripple.com/Ripple_for_Users) (Note: "Ripple" mostly refers to the XRP Ledger, not the company, in that article.)

Or, here's the start of another, similar diagram [warning, there may be typos here].
![community-credit-1](https://user-images.githubusercontent.com/7515597/156469522-e73105a8-2271-4e54-b1ac-c21b53ac3433.png)

Note, some conventions from these types of diagrams that are useful to understand and maybe preserve:

- Circles generally represent "accounts" while lines represent trust lines
- Arrows on trust lines represent the direction of trust. In the case of a stablecoin, this means that "the arrows point to where the money is" (that is, the real assets that back the stablecoin).

PR #3312, opened 2025-09-24
Rewrite MPTs concept & related docs

### LINKED ISSUE #3311: Miscellaneous MPT doc issues
The MPT concept documentation has a number of flaws that should be corrected:

- It describes the on-chain metadata and URI as separate points, without mentioning XLS-89. Actually, there is no specific URI field but you can put a URI in the metadata if you want to link to off-chain metadata.
- It only describes the properties of MPTs in comparison to the older token standard. It should describe their properties overall, independent of prior features, first.
- It uses the non-preferred term "IOUs" to refer to the original fungible token standard.
- It compares MPTs (v2 fungible tokens) to trust lines and then to IOUs (v1 fungible tokens), which is extra strange because trust lines are just a part of the v1 token implementation.
- It doesn't specifically call out some of the notable differences:
    - MPTs don't have a currency code (i.e. "BTC"), just an issuance ID. Their metadata _may_ contain a currency code, but there is no guarantee that there is one; and if there is, there are no hard rules on its structure, and other MPTs from the same issuer could use the same currency code.
    - MPTs have many properties defined on a per-issuance basis rather than per-issuer basis (i.e. different tokens from the same issuer can have different properties).
    - Some properties like transfer fees have different range and resolution than their trust line token equivalents
- It incorrectly states that issuers are limited to issuing 32 MPTs. That's probably based on a misunderstanding of the 32-bit sequence number that is part of the MPT issuance ID. That puts a limit of 2^32 MPTs issued by one account, but it's more likely the account will hit other soft limits first.

Related issues: #3306, #3262, #2796, #3185

mDuo13:
- Rename "tokens" to "trust line tokens" where necessary to distinguish them from MPTs
- Rewrite the MPT concept doc to fix #3311
- Rewrite the "Tokens" and "Fungible Tokens" landing pages to better explain the differences between all four types of XRPL asset (trust line tokens, MPTs, NFTs, and XRP)
- Move "Transfer Fees" doc into the Fungible Tokens category since that's what it discusses.

(Note: I have fixed the links in the translated pages and moved Japanese version of the transfer fees page, but not implemented any of the textual changes in translation.)


docs/concepts/tokens/index.md
```diff
@@ -32,20 +29,20 @@ For more on this type of usage, see [paths](fungible-tokens/paths.md). <!--{# TO
 
 ## Other Tokens
 
-There are other use cases for tokens issued in the XRP Ledger. For example, you can create an "Initial Coin Offering" (ICO) by issuing a fixed amount of currency to a secondary address, then "throwing away the key" to the issuer.
+There are other use cases for tokens issued in the XRP Ledger, including digital-only tokens that are intended for use in various online systems. It is also common for users to create and trade "meme coins" for novelty and speculative purposes, without any more serious use case in mind. Digital first tokens are sometimes distributed with an "Initial Coin Offering" (ICO) where a fixed amount of tokens are initially created and distributed, with the issuer then "throwing away the key" to their account so that no more tokens can be issued later.
```
oeggert, line 32: Do we want to mention meme coins here? It's technically a use case, but the legitimacy and all the rugs associated with it make it iffy to me.

docs/concepts/tokens/index.md
```diff
@@ -32,20 +29,20 @@ For more on this type of usage, see [paths](fungible-tokens/paths.md). <!--{# TO
...
 
 {% admonition type="danger" name="Warning" %}ICOs might be [regulated as securities](https://www.sec.gov/oiea/investor-alerts-and-bulletins/ib_coinofferings) in the USA. <!-- SPELLING_IGNORE: ico, icos -->{% /admonition %}
 
 Be sure to research the relevant regulations before engaging in any financial service business.
```
oeggert, line 36: Do we want to single out only the US since other countries may have regulations (or will) around ICOs. Maybe merge both these statements into a general warning to always check relevant regulations.

docs/concepts/tokens/fungible-tokens/trust-line-tokens.md
oeggert, line 1: Don't forget to add this into sidebars.yaml

docs/concepts/tokens/fungible-tokens/trust-line-tokens.md
```diff
@@ -0,0 +1,101 @@
...
+Trust lines can be implicitly created by some transactions, such as when you buy a token in the [decentralized exchange](../decentralized-exchange/index.md). In this case, the trust line uses entirely default settings.
+
+
+## Going Above the Limit
+
+There are three cases where you can hold a balance that is _greater_ than your limit on that trust line:
+
+1. When you acquire more of that token through [trading](../decentralized-exchange/index.md).
+2. When you decrease the limit on a trust line that has a positive balance.
+3. When you acquire more of that token by [cashing a Check](../../payment-types/checks.md). (_Requires the [CheckCashMakesTrustLine amendment][]_)
```
oeggert, line 40: Shouldn't this list also include escrows with the tokenescrow amendment?

docs/concepts/tokens/fungible-tokens/trust-line-tokens.md
```diff
@@ -0,0 +1,101 @@
...
+
+
+## Trust Line Settings
+
+In addition to the shared balance, each account has its own settings on the trust line, which consist of the following:
+
+- The **Limit**, a number from 0 to the [maximum token amount](../../../references/protocol/data-types/currency-formats.md). Payments and other accounts' actions cannot cause the trust line's balance (from this account's perspective) to go over the limit. The default is `0`.
```
oeggert, line 47: I'm guessing that `0` means no limit. It doesn't mention the default value and what `0` means in the `TrustSet` transaction docs--might be worth updating that real quick too.

docs/concepts/tokens/fungible-tokens/trust-line-tokens.md
```diff
@@ -0,0 +1,101 @@
...
+- **Quality In** and **Quality Out** settings, which allow the account to value tokens issued by the other account on this trust line at less (or more) than face value. For example, if a stablecoin issuer charges a 3% fee for withdrawing tokens for the equivalent off-ledger assets, you could use these settings to value those tokens at 97% of face value. The default, `0`, represents face value.
+
+
+## Properties
+
+Unlike XRP or MPTs, trust line tokens always exist in _trust lines_, and all transfers move along trust lines. You cannot cause someone else's account to hold more of a token than the _limit_ configured on their trust line. (You _can_ cause your own trust line to go over the limit, for example by buying more of it in the decentralized exchange or by decreasing the limit after you already have a positive balance.)
+
+Anyone can issue trust line tokens by sending a [Payment transaction][] if the necessary trust lines are in place. You can "burn" tokens by sending them back to the issuer. In some cases, cross-currency payments or trades can also create more tokens according to an issuer's settings.
```
oeggert: It might be worth defining what "burning" tokens means, since this is a conceptual topic.

docs/concepts/tokens/fungible-tokens/trust-line-tokens.md
```diff
@@ -0,0 +1,101 @@
...
+
+Trust line tokens use decimal (base-10) math with 15 digits of precision and an exponent that allows them to express very large values (up to 9999999999999999 × 10<sup>80</sup>) and very small values (down to 1.0 × 10<sup>-81</sup>).
+
+## Reserves and Deletion
+
+Since a trust line occupies space in the ledger, [a trust line increases the XRP your account must hold in reserve](../../accounts/reserves.md). Either or both accounts in the trust line may be charged the reserve for the trust line, depending on the status of the trust line: if any of your settings are not the default, or if you hold a positive balance, it counts as one item toward your owner reserve.
+
+Generally, this means that the account that created the trust line is responsible for the reserve and the issuer is not. <!-- STYLE_OVERRIDE: is responsible for -->
+
+Trust lines are automatically deleted if both sides' settings are in the default state and the balance is 0. This means that, to delete a trust line, you need to:
```
oeggert, line 73: This statement initially reads like both the issuer and holder have to have a 0 balance. Specify the holder has to have a 0 balance.

docs/concepts/tokens/fungible-tokens/stablecoins/index.md
```diff
@@ -1,29 +1,27 @@
...
 ---
 # Stablecoins
 
-A Stablecoin holds assets of value outside of the XRP Ledger and issues tokens representing the equivalent value on the ledger. This type of issuer is sometimes called a _gateway_, because currency can move into and out of the XRP Ledger through their service. If the assets that back a token use the same amounts and denomination as the tokens in the ledger, that token can be considered a "stablecoin" because – in theory – the exchange rate between that token and its off-ledger representation should be stable at 1:1.
+A stablecoin is a type of fungible token that is backed by assets of value outside of the blockchain in which is issued. Typically, the stablecoin's issuer holds backing assets representing the equivalent value of the tokens issued on the XRP ledger. A stablecoin issuer is sometimes called a _gateway_, because currency can move into and out of the XRP Ledger through their service. If the assets that back a token use the same amounts and denomination as the tokens in the ledger, that token can be considered a "stablecoin" because—in theory—the exchange rate between that token and its off-ledger representation should be stable at 1:1.
```
oeggert, line 10:
```suggestion
A stablecoin is a type of fungible token that is backed by assets of equal denominations and amounts offchain, such as U.S. dollars. The token is considered stable because it has a 1:1 exchange rate with its offchain asset. Typically, the stablecoin's issuer holds backing assets representing the equivalent value of the tokens issued on the XRP ledger. A stablecoin issuer is sometimes called a _gateway_, because currency can move into and out of the XRP Ledger through their service.
```
Tweaked the wording since the original makes it sound like a stablecoin can be any token that's backed by assets. The last sentence is what makes a stablecoin a stablecoin, and not just another fungible token.

docs/concepts/tokens/fungible-tokens/multi-purpose-tokens.md
```diff
@@ -7,82 +7,98 @@ labels:
...
+- **Issuers are separate from holders.** An issuer cannot hold their own tokens. If you send tokens to the issuer, those tokens are automatically burned. Issuers should use separate [operational accounts](../../accounts/account-types.md)—also called hot wallets—if they want to hold their own tokens.
+- **You can't force someone to hold tokens.** Holders must send a transaction that indicates their willingness to hold a token before they can receive it.
+
+Unlike trust line tokens, each MPT issuance is uniquely identified by an **MPT Issuance ID**, which is a 192-bit number not intended for display to humans, for example `00070C4495F14B0E44F78A264E41713C64B5F89242540EE255534400000000000000`. MPTs can be configured with metadata that affects how they are displayed, but there are no hard guarantees about the presence or uniqueness of currency codes.
 
 ## Key Features of MPTs on XRPL
 
-The following are notable features of MPTs on XRPL.
+MPTs are designed decentralized finance, so they are ready to go with settings and features for institutional use cases:
```
oeggert:
```suggestion
MPTs are designed for decentralized finance, so they are ready to go with settings and features for institutional use cases:
```

docs/concepts/tokens/fungible-tokens/multi-purpose-tokens.md
```diff
@@ -7,82 +7,98 @@ labels:
...
+## Limits on Issuing
 
-It is also simpler for ordinary users to understand what tokens are available, what tokens they have issued, and what they hold in their wallet.  
+There is not intended to be a limit on how many MPT issuances you can create, but the technology does impose some hard and soft limits indirectly:
 
-For both issuers and holders of MPTs, there will typically be a smaller XRP reserve compared to the equivalent representations with IOU trustlines.
+- The ledger entry that defines an MPT issuance counts as one object towards the issuer's [owner reserve](../../accounts/reserves.md#owner-reserves), so the issuer needs to set aside {% $env.PUBLIC_OWNER_RESERVE %} per MPT issuance.
+- Each holder's balance of MPTs is tracked in both the holder's and issuer's [owner directories](/docs/references/protocol/ledger-data/ledger-entry-types/directorynode). The maximum number of items that can be in an owner directory is very large, but finite, which places a hard cap on how many MPT issuances you can issue.
```
oeggert, line 98: Can we just say what the max item limit is?

docs/concepts/tokens/fungible-tokens/index.md
```diff
@@ -1,93 +1,58 @@
...
-- **Freeze**: A true/false value indicating whether an [individual freeze](freezes.md#individual-freeze) is in effect on this trust line. The default is `false`.
-- **DeepFreeze**: A true/false value indicating whether a [deep freeze](deep-freeze.md) is in effect on this trust line. The default is `false`.
-- **Quality In** and **Quality Out** settings, which allow the account to value tokens issued by the other account on this trust line at less (or more) than face value. For example, if a stablecoin issuer charges a 3% fee for withdrawing tokens for the equivalent off-ledger assets, you could use these settings to value those tokens at 97% of face value. The default, `0`, represents face value.
+If you are creating a new token on the XRP Ledger, it may be confusing that there are two standards to choose from with different properties and strengths. If you aren't sure which one to use, the following summary should help you decide:
 
+- If you need **compatibility with the DEX** or your use case is **community credit**, choose trust line tokens.
```
oeggert: Can we define what community credit is supposed to mean? Also maybe de-emphasize the dex compatibility since the reason for this is that MPTs just don't have it yet--don't want to create the idea that MPTs just aren't good for dex in general.

docs/concepts/tokens/fungible-tokens/index.md
```diff
@@ -1,93 +1,58 @@
...
-    - [Decentralized Exchange](../decentralized-exchange/index.md)
-    - [Rippling](rippling.md)
-- **References:**
-    - [account_lines method][] - Look up trust lines attached to a given account
-    - [gateway_balances method][] - Look up an issuer's total balance issued
-    - [RippleState object](../../../references/protocol/ledger-data/ledger-entry-types/ripplestate.md) - The data format for trust lines in the ledger's state data.
-    - [TrustSet transaction][] - The transaction to create or modify trust lines.
-
-{% raw-partial file="/docs/_snippets/common-links.md" /%}
+There are other strengths and weaknesses of each format, such as trust line tokens' ability to represent very large and very small quantities of the same token (greater than 63 orders of magnitude apart), or compatibility with legacy software.
```
oeggert: I would include this in the bullet list above. Also I'd take out the part about other strengths and weaknesses unless we plan on defining what those are here. As is, saying it is kind of an empty statement since we aren't pointing them to other docs are providing them the info here to help them choose.

oeggert (changes requested):

docs/concepts/tokens/fungible-tokens/trust-line-tokens.md
mDuo13: 0 does not mean no limit here, it actually means 0. If your limit is zero then you can't receive tokens issued by the other side of the trust line except through your own actions, i.e. you are probably the issuer and not the holder.

docs/concepts/tokens/index.md
mDuo13: Given how common meme coins are, I decided it was better to acknowledge they exist—without getting _too_ judgemental—than to pretend they don't.

docs/concepts/tokens/fungible-tokens/trust-line-tokens.md
mDuo13: I'm not sure. Will double check the TokenEscrow spec and update accordingly.

docs/concepts/tokens/fungible-tokens/trust-line-tokens.md
mDuo13: There's only one shared balance and it has to be zero for the line to be deleted.

docs/concepts/tokens/fungible-tokens/stablecoins/index.md
mDuo13: Fractional reserve stablecoins are a real (if somewhat frowned-upon) possibility. For a long time, people have suspected that USDT (Tether) is not fully backed—but as long as they have _enough_ USD to process withdrawals, it remains stable.

docs/concepts/tokens/fungible-tokens/multi-purpose-tokens.md
mDuo13: I'll try to find out what it is.

docs/concepts/tokens/fungible-tokens/index.md
mDuo13: Given that MPT DEX compatibility doesn't even have a spec discussion yet, it'll be a while until they're compatible, and that's a major feature gap, so I think it's important to emphasize.

When MPT-DEX is closer to release, we should reword this sentence.

mDuo13: As for your other question, "community credit" is not easy to look up, so I agree that it's an important gap. Unfortunately, it's also a pretty involved topic, somewhat fundamental with the design of the original "Ripple Network" (the circa 2009, pre-XRP one, even), so it's not really something that we can just slip into these docs.

#1353 is the existing ticket for better documenting community credit, so I'd like to continue to leave that one as a separate task.

docs/concepts/tokens/fungible-tokens/trust-line-tokens.md
mDuo13: Should be there as of 1e4bb442b209c3a0c00a432b0a6baa6899a1fe22

docs/concepts/tokens/index.md
mDuo13: That's a good point, but I specifically mentioned the US since the reference link we have is for the US SEC. I think it's OK as an illustrative example.

docs/concepts/tokens/fungible-tokens/stablecoins/index.md
oeggert: In that case I think we should explain the difference between the more generally understood definition of stablecoin, and fractional stablecoins and the risk associated with it.

oeggert (dismissed):
Approving with a few open comments. I wouldn't consider them blockers, though, so you can merge as is.

oeggert (approved):

PR #3313, opened 2025-09-24
Fix XLS Spec links for the XLS-0075 (PermissionDex) and XLS-0056 (Batch)

d-moos-ripple:
removed the `d` character so that links point to the right direction in the spec repo again

### DIFF
--- a/resources/known-amendments.md
+++ b/resources/known-amendments.md
@@ -1584,7 +1584,7 @@ For more information, see the [Payment Channels Tutorial](../docs/tutorials/how-
 
 Allows accounts to delegate some permissions to other accounts.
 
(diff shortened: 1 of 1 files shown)


mDuo13 (approved):
Thanks.

This keeps being a problem, as specs drop the draft-designation "d" at an arbitrary point breaking prior links. I think we probably should consider changing the process so this stops happening.

intelliot:
I think we've decided to stop using "d": https://github.com/XRPLF/XRPL-Standards/issues/294

Additionally, for historical cases, we can allow prior links to work by adding symlinks: https://github.com/XRPLF/XRPL-Standards/pull/381

PR #3324, opened 2025-09-30
Add Amendment Data Caching

### LINKED ISSUE #3231: Improve amendments table & disclaimer components to cache data
The `{% amendments-table /%}` and `{% amendment-disclaimer /%}` components call https://vhs.prod.ripplex.io/v1/network/amendments/vote/main/ to get amendment status data—but if you include multiples of them in a page, it calls that API multiple times per page load.

We should cache that data since it doesn't need to be redundantly fetched and refreshed. (At maximum, the ledger's amendment status can only change about every 15 minutes.) It would also be nice to refactor the two components to share some code for fetching the data as well.

(Continues from #3227 and #3219.)

oeggert:
Address #3231 

This PR:
1. Moves all amendment-related component logic to a separate `Amendments.tsx` file to make the code more manageable.
2. Adds caching for the VHS endpoint. API response is stored locally in the browser and kept for 15 minutes. Amendment Disclaimer and Amendment Table components reference this data. If data is older than 15 minutes, it refreshes it with a new API fetch.


@theme/components/Amendments.tsx
```diff
@@ -0,0 +1,391 @@
...
+
+type AmendmentsCachePayload = {
+  timestamp: number;
+  amendments: Amendment[];
+}
+
+// API data caching
+const amendmentsEndpoint = 'https://vhs.prod.ripplex.io/v1/network/amendments/vote/main/';
+const amendmentsCacheKey = 'xrpl.amendments.mainnet.cache';
+const amendmentsTTL = 12 * 60 * 60 * 1000; // 12 hours in milliseconds
```
mDuo13: Where did 12 hours come from? Seems like overkill to me. Amendment statuses don't update _that_ often, but when they do, it's on a flag ledger which are approximately 15 minutes apart.

Seems to me like it would not cause excessive load to check amendments' status every 15 minutes instead of 12 hours, which _would_ probably give people a more gratifyingly "close to live" view of Mainnet's status.

@theme/components/Amendments.tsx
```diff
@@ -0,0 +1,391 @@
...
+  if (typeof window === 'undefined') return;
+  try {
+    const payload: AmendmentsCachePayload = { timestamp: Date.now(), amendments };
+    window.localStorage.setItem(amendmentsCacheKey, JSON.stringify(payload));
+  } catch {
+    // Ignore quota or serialization errors
+  }
+}
+
+// Sort amendments table
```
mDuo13: This would be a good place for a code comment explaining the intended order, e.g. :

```suggestion
// Sort amendments table by status, then chronologically, then alphabetically
```

@theme/components/Amendments.tsx
```diff
@@ -0,0 +1,391 @@
...
+function sortAmendments(list: Amendment[]): Amendment[] {
+  const getStatusPriority = (amendment: Amendment): number => {
+    if (amendment.eta) return 0;          // Expected
+    if (amendment.consensus) return 1;    // Open for Voting
```
mDuo13: I would think "Expected" should appear _between_ "Open for Voting" and "Enabled" because that's how the states transition in the happy path: Voting → Expected → Enabled.

```suggestion
    if (amendment.consensus) return 0;    // Open for Voting
    if (amendment.eta) return 1;          // Expected
```

@theme/components/Amendments.tsx
```diff
@@ -0,0 +1,391 @@
+import * as React from 'react';
```
mDuo13: This file inconsistently uses semicolons; if we want to be consistent, Javascript Standard Style says no semicolons. Technically this is TS, not JS, but [ts-standard](https://github.com/standard/ts-standard) follows the same convention.

Not strictly important (I recognize you're just carrying over an inconsistency from the prior files) but just something I noticed.

@theme/components/Amendments.tsx
```diff
@@ -0,0 +1,391 @@
...
+      setHref(undefined); // No link for voting amendments
+    }
+  }, [props.amendment, enabledLabel, etaLabel, votingLabel]);
+
+  // Split the status at the colon to create two-color badge
+  const parts = status.split(':');
+  const label = shieldsIoEscape(parts[0]);
+  const message = shieldsIoEscape(parts.slice(1).join(':'));
+  
+  const badgeUrl = `https://img.shields.io/badge/${label}-${message}-${color}`;
```
mDuo13: Not something to do right now, but if some neat future work we could do would be to generate our own custom badges as inline SVG. That would give us a little more flexibility on styling and also save on network calls since the user's browser wouldn't need to fetch the badges from shields.io.

The full SVG data for a badge isn't _too_ big or complicated, after all. For example:

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="150" height="20" role="img" aria-label="Open for Voting: 28.57%"><title>Open for Voting: 28.57%</title><linearGradient id="s" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><clipPath id="r"><rect width="150" height="20" rx="3" fill="#fff"/></clipPath><g clip-path="url(#r)"><rect width="97" height="20" fill="#555"/><rect x="97" width="53" height="20" fill="#80d0e0"/><rect width="150" height="20" fill="url(#s)"/></g><g fill="#fff" text-anchor="middle" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" text-rendering="geometricPrecision" font-size="110"><text aria-hidden="true" x="495" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="870">Open for Voting</text><text x="495" y="140" transform="scale(.1)" fill="#fff" textLength="870">Open for Voting</text><text aria-hidden="true" x="1225" y="150" fill="#ccc" fill-opacity=".3" transform="scale(.1)" textLength="430">28.57%</text><text x="1225" y="140" transform="scale(.1)" fill="#333" textLength="430">28.57%</text></g></svg>
```

@theme/markdoc/components.tsx
```diff
@@ -3,13 +3,13 @@ import { useLocation } from 'react-router-dom';
...
 import { Link } from '@redocly/theme/components/Link/Link';
-import { useThemeHooks } from '@redocly/theme/core/hooks'
+import { useThemeHooks } from '@redocly/theme/core/hooks';
 import { idify } from '../helpers';
 import { Button } from '@redocly/theme/components/Button/Button';
 
 export {default as XRPLoader} from '../components/XRPLoader';
 export { XRPLCard, CardGrid } from '../components/XRPLCard';
-
+export * from '../components/Amendments';
```
mDuo13: This is OK but I think the better practice might be to explicitly list the things that we're exporting, like this:

```suggestion
export { AmendmentsTable, AmendmentDisclaimer, Badge } from '../components/Amendments';
```

@theme/components/Amendments.tsx
```diff
@@ -0,0 +1,403 @@
...
+}
+
+function shieldsIoEscape(s: string) {
+  return s.trim()
+    .replace(/-/g, '--')
+    .replace(/_/g, '__')
+    .replace(/%/g, '%25')
+}
+
+export function Badge(props: {
```
mDuo13, line 344: Strictly speaking, badges are not always related to amendments. Many of them are for API changes instead of protocol changes. But it's probably fine to have this component in this file regardless.

mDuo13 (changes requested):
Overall looks good and I confirmed that it does not fetch statuses from VHS repeatedly. I do think the status sorting should be changed slightly, and maybe the cache TTL reduced, but none of my comments should be considered blockers.

@theme/components/Amendments.tsx
oeggert: At the time I just picked an arbitrary number, but your suggestion makes sense.

@theme/components/Amendments.tsx
oeggert: Good point, I'll make the change.

@theme/components/Amendments.tsx
oeggert: I removed the semicolons from this and `Components.tsx`.

mDuo13 (approved):

PR #3327, opened 2025-10-02
Use amendment disclaimer component for MPTs

maria-robobug:
Updated a few places that were not using the component for MPTs.

Also, I added a minor unrelated change (since it's rather small) to provide a link to the MPTs on DEX discussion in the MPT section of the Fungible Tokens landing page.


oeggert (dismissed):

amarantha-k (changes requested):
The use case page needs to be updated to use the disclaimer component too. https://xrpl-dev-portal--mpt-use-new-amendment-component.preview.redocly.app/docs/use-cases/tokenization/creating-an-asset-backed-multi-purpose-token

oeggert (approved):

amarantha-k (approved):

PR #3328, opened 2025-10-02
Add example JSON for MPTokenAuthorize transaction

### LINKED ISSUE #3185: Update MPTokenAuthorize page to include transaction sample
Transaction sample is missing, add one.

maria-robobug:
Fix [#3185](https://github.com/XRPLF/xrpl-dev-portal/issues/3185).
- Adds MPTokenAuthorize JSON example and Mainnet transaction.
- Updates other MPT transactions to use the Mainnet example.


mDuo13 (commented):
It seems strange to show _some_ of the required common fields (like `Fee`) but not others (like `Sequence`) even though both are auto-fillable. The other transaction references are somewhat inconsistent about how much they provide of common and auto-fillable fields, so that's also something we should try to improve. IMO we should provide everything except the fields added in signing (so include `Fee`, `Sequence`, and `Flags` but omit `SigningPubKey` and `TxnSignature`).

Also, ideally, we would add a `{% tx-example txid="(hash)" /%}` tag for the new transaction, with an example from Mainnet (since it'll never be reset, so any full-history server will be able to find it forever).

It might be easier to submit such a transaction yourself rather than trawl the ledger looking for one. I did write a quick script to find and display transactions of types we don't have examples of, but it hasn't found any MPTokenAuthorize examples... yet.

maria-robobug:
> It might be easier to submit such a transaction yourself rather than trawl the ledger looking for one. I did write a quick script to find and display transactions of types we don't have examples of, but it hasn't found any MPTokenAuthorize examples... yet.

Yeah couldn't find any so I've now created my own MPT issuance and an MPT holder on Mainnet. I've also updated the other MPT transactions and added the `{% tx-example txid="(hash)" /%}` tag for each of them.

mDuo13 (approved):
Fantastic!

PR #3340, opened 2025-10-07
Remove beta AI interface

### LINKED ISSUE #2818: AI answers have broken links for source pages
When using the "Ask AI (Beta)" tool, the "Answer generated from the following sources" links at the bottom are all broken.

For example, "Use case: Payment Authorization" links to:

```
https://xrpl.org/concepts/payment-types/checks.html
```

The title is a little weird (maybe it was a header in the old version of the Checks page?) but the link is broken. It's missing the `/docs/` part of the path and has an extraneous `.html` appended. The correct link should be:

```
https://xrpl.org/docs/concepts/payment-types/checks
```

Also, this is of course typical for AI-generated text, but just for reference the answer in the screenshot is exactly wrong and dangerously so—someone who takes it at face value could end up unintentionally violating compliance rules. Deposit Authorization does not control whose offers can cross yours, which is why the article on Deposit Authorization says not to place offers if you're using it.

mDuo13:
The current AI interface is unmaintained and has started showing as "unavailable". A next-generation AI for XRPL is in progress and will replace this when it's ready. But in the meantime, it's time to delete the old one.

Fixes #2818.


amarantha-k (approved):
LGTM

PR #3355, opened 2025-10-15
remove mentions of Sidechain-Devnet

mvadari:
The Sidechain-Devnet no longer exists.

There are also a couple of tutorial pages under `use-xrpl-sidechains` that use the Sidechain-Devnet, not sure if those should also be removed because they may be good snippets.


oeggert (changes requested):
I think we should also remove the sidechain URLs from the tutorial code samples, too, and add something to show they need to add their own sidechain info.

```suggestion
const WS_URL_issuingchain = 'wss://example-sidechain.net:12345/' // Issuing chain
```

mvadari:
@oeggert is this good to go now?

oeggert (approved):
LGTM

PR #3357, opened 2025-10-16
Update mpt_holders, Get MPTokenIssuance, and Get MPToken  with Mainnet "Try It" example

### LINKED ISSUE #3107: Change mpt_holders Try It example to use Mainnet
Relates to: https://github.com/XRPLF/xrpl-dev-portal/pull/3104

> Having the examples on Devnet is best for now—since MPTs are not even enabled on Mainnet yet—but eventually it would be preferable to switch the examples to use Mainnet so that we don't have to re-create the same data every time there's a Devnet reset. Devnet resets are somewhat irregular but can be as often as multiple times per year, whereas on Mainnet we can generally create sample data once and leave it there forever. So that's a task to note for later, but it can't be done until MPT is enabled and there's no need to do anything until the next Devnet reset.

maria-robobug:
Fix [#3107](https://github.com/XRPLF/xrpl-dev-portal/issues/3107)

- Updates `mpt_holders`, `Get MPTokenIssuance`, and `Get MPToken` with a Mainnet "Try It" example.
- Also changed "Get MPT Issuance Object" to "Get MPT Issuance Entry" for consistency. Same with "Get MPToken Object".

### DIFF
--- a/docs/references/http-websocket-apis/public-api-methods/clio-methods/mpt_holders.md
+++ b/docs/references/http-websocket-apis/public-api-methods/clio-methods/mpt_holders.md
@@ -46,7 +46,7 @@ For a given `MPTokenIssuanceID` and ledger sequence, `mpt_holders` returns all h
 
 {% /tabs %}
 
-{% try-it method="mpt_holders" server="s1" /%}
+{% try-it method="mpt_holders" /%}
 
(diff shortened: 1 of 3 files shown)


oeggert (approved):
LGTM

PR #3360, opened 2025-10-17
Add mainnet AMMCreate and OracleSet transaction examples

mDuo13:
Spotted examples of these transactions in the wild, so I figured I should add them.


maria-robobug (approved):
LGTM

Issue #3363, opened 2025-10-21
Update list of invariant checks


mDuo13:
[Invariant Checking](https://xrpl.org/docs/concepts/consensus-protocol/invariant-checking) describes all the invariant checks that existed when it was written, but more have been added since then.

Invariant Checks are defined in [InvariantCheck.h](https://github.com/XRPLF/rippled/blob/master/src/xrpld/app/tx/detail/InvariantCheck.h) and implemented in the corresponding `.cpp` file.

Update the docs with the new checks.

Several of the new invariant checks may be controlled by amendments and some of them may not be active yet (part of the in-development InvariantsV1_1 amendment), so confirm that and label them accordingly.

mDuo13:
Per @ximinez:

> You can search the file for `rules().enabled` to see where and how many invariants are fully or partially gated on amendments.

PR #3180, opened 2025-07-02
Fix broken documentation references

emmanuel-ferdman:
## PR Summary
This PR is a bit large, as it resolves broken links to the `XRPLF/rippled` repository introduced by [this commit](https://github.com/XRPLF/rippled/commit/e416ee72ca26fa0c09d2aee1b68bdfb2b7046eed). Key changes include:

- Updated all relevant file references to use commit `1e01cd34f7a216092ed779f291b43324c167167a`, which is the latest commit on[ `XRPLF/rippled` master (v2.5.0)](https://github.com/XRPLF/rippled/tree/master).
- Removed links to shard files that were deleted in commit [`7d27b1119092e3195238a1897d6012c3372e6c53`](https://github.com/XRPLF/rippled/commit/7d27b1119092e3195238a1897d6012c3372e6c53).
- Adjusted line references where applicable - some did not require changes.
- Due to the inclusion of translated pages, the scope of this PR is wider than usual.
- used 'xrplf' -> 'XRPLF' as GitHub org.
- There are changes in `blog` directory that I didn't upload because the PR is large enough. Let me know if you want them as separate PR.

Let me know if I missed anything or if further changes are needed.

PR #3364, opened 2025-10-22
Fix source links in concepts and infrastructure sections

mDuo13:
This is the first of several PRs attempting to do the same work as #3180 but in a way that is divided by section so that it can be reviewed and merged piecemeal instead of being an unmanageable mega-commit.

- Fix broken source links in the **Concepts** and **Infrastructure** sections of the site.
    - For links to whole files, link to the master branch of the repo. This way, the link will always point to the latest stable release and will only break if the whole file is moved or renamed.
    - For links to specific lines, use a permalink to the latest commit. This way, the correct lines will be highlighted even if there are small changes elsewhere in the same file, but the link may eventually become outdated if the relevant section changes in the source.
    - In some cases, remove links that did not seem appropriate for the section, and reword the sections accordingly.
- Slightly refactor the Invariant Checks page. Rather than linking each invariant check to the individual lines for that check's definition, I switched it to link once at the top of the section to the source file that contains all of the invariants. The names closely match so it should not be confusing. (Also, opened #3363 to update the list with the new ones.)
- Update and note one case where `rippled` 2.3.1 changed a hard-coded cutoff that was mentioned in the docs.
- No changes have been applied to translated files yet, but it should be possible to apply the same changes to the translated files without _too_ much additional effort.

### DIFF
--- a/docs/infrastructure/troubleshooting/understanding-log-messages.md
+++ b/docs/infrastructure/troubleshooting/understanding-log-messages.md
@@ -120,7 +120,7 @@ A large number of these messages around the same time may indicate a problem, su
 The following log message indicates that a client to the server's public API has been dropped as a result of [rate limiting](../../references/http-websocket-apis/api-conventions/rate-limiting.md):
 
 ```text
-Resource:WRN Consumer entry 169.55.164.21 dropped with balance 15970 at or above drop threshold 15000
+Resource:WRN Consumer entry 169.55.164.21 dropped with balance 25970 at or above drop threshold 25000
 ```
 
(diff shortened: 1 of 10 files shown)


maria-robobug (dismissed):

docs/infrastructure/troubleshooting/understanding-log-messages.md
```diff
@@ -125,7 +123,7 @@ The following log message indicates that a client to the server's public API has
 Resource:WRN Consumer entry 169.55.164.21 dropped with balance 15970 at or above drop threshold 15000
```
amarantha-k: Update to reflect the updated drop threshold.

amarantha-k (dismissed):
Looks good; only comment is to update the log message to reflect the updated drop threshold.

amarantha-k (approved):
LGTM

PR #3368, opened 2025-10-27
Refactor RWA tokenization buttons

AKCodez:
… and link behavior for improved responsiveness and user experience.


docs/use-cases/tokenization/real-world-assets.page.tsx
```diff
@@ -412,13 +408,19 @@ function TokenHeroSection() {
         <h1 className="token-title">
           {translate("Real-World Asset (RWA) Tokenization")}
         </h1>
-        <div className="d-lg-block small-100 ">
-          <Link
-            className="btn btn-primary btn-arrow-out"
+        <div className="button-container">
+          <a
```
mDuo13: Why `<a>` instead of `<Link>` here?

docs/use-cases/tokenization/real-world-assets.page.tsx
```diff
@@ -412,13 +408,19 @@ function TokenHeroSection() {
...
+            className="btn btn-primary p-12"
```
mDuo13: Seems inappropriate to drop the `btn-arrow-out` class since this takes you to a PDF instead of a proper page on the site, and the arrow is supposed to be your warning that that's happening.

mDuo13 (changes requested):
I'm not sure how, but this seems to have broken the formatting of the "Powerful Features" cards of the homepage—notice how the cards' text is underlined and the footer graphic doesn't span the whole card:

(screenshot showing misformatted cards)

On the (non-RWA) "Tokenization" page, the spacing between "Quick Start" and "Learn More" has disappeared for some reason:

(screenshot showing tokenization homepage issue)

There are some other scattered small styling issues, too, that aren't explained by the diff. Maybe you should re-build the CSS to make sure the version committed to the branch matches the `scss` source.

Also, an issue with the original page that I didn't catch before but we should absolutely fix is that the logos in the "Trusted by Leaders..." section are not links, but divs with a click event that acts like a link. That's terrible for usability, accessibility, and SEO—we should make them real links.

AKCodez:
@mDuo13 Nice catches, the random css issues were from me bringing over a bundle from another branch, nice catch I fixed the bundle and pushed all the other changes you requested as well, all the issues are now resolved

AKCodez:
> I'm not sure how, but this seems to have broken the formatting of the "Powerful Features" cards of the homepage—notice how the cards' text is underlined and the footer graphic doesn't span the whole card:
> 
> (screenshot showing misformatted cards)
> On the (non-RWA) "Tokenization" page, the spacing between "Quick Start" and "Learn More" has disappeared for some reason:
> 
> (screenshot showing tokenization homepage issue)
> There are some other scattered small styling issues, too, that aren't explained by the diff. Maybe you should re-build the CSS to make sure the version committed to the branch matches the `scss` source.
> 
> Also, an issue with the original page that I didn't catch before but we should absolutely fix is that the logos in the "Trusted by Leaders..." section are not links, but divs with a click event that acts like a link. That's terrible for usability, accessibility, and SEO—we should make them real links.

resolved both issues

docs/use-cases/tokenization/real-world-assets.page.tsx
mDuo13: This comment still applies. "Get Started Now" takes you to a PDF, so it should be marked with an arrow or something comparable to warn people.

mDuo13 (changes requested):
1. Still need to put the arrow back on "Get Started Now"
2. The fake links in "Trusted by Leaders..." section should still be turned into real links

Can confirm the other style issues are indeed resolved.

AKCodez:
@mDuo13 added both changes requested

mDuo13 (approved):

PR #3380, opened 2025-11-13
Add minor improvements for Javascript code walkthrough

### LINKED ISSUE #1795: Include links to js.xrpl.org from Get Started Using JavaScript Library
Are we able to include links to js.xrpl.org in the Docs? That way, devs see the links directly and do not need to go and search for them.

maria-robobug:
Adds some minor improvements to the Getting Started with Javascript walkthrough tutorial:

- Make sure header is included in each step, where possible.
- Add links to xrpl.js reference documentation where relevant. Fixes #1795

### DIFF
--- a/_code-samples/get-started/js/get-acct-info.js
+++ b/_code-samples/get-started/js/get-acct-info.js
@@ -1,7 +1,9 @@
+// @chunk {"steps": ["import-node-tag"]}
 // Import the library
-// @chunk {"steps": ["connect-tag"]}
 import xrpl from "xrpl"
+// @chunk-end
 
+// @chunk {"steps": ["connect-tag"]}
 // Define the network client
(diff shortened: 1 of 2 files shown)


mDuo13 (approved):
Good changes, although it made me realize some things I don't like about the code walkthrough widget:

1. When you click on a section, it scrolls just past the top of that section. (Probably something to do with our fixed header, or something.)
2. If you try to click a link in a section, it does the scroll thing instead of opening the link unless you're already scrolled to where it wants you to be.

These may be issues for us to bring up with Redocly, but they're not problems with this PR.

maria-robobug:
Yeah agreed, these are Redocly issues that need to be addressed as it does hinder the code walkthrough experience.

PR #3381, opened 2025-11-15
Add code sample showing how to walk an owner directory

mDuo13:
This is not the _most_ useful code sample but it does provide a useful sample of using the `ledger_entry` command to iterate over an account's owner directory and can be used to show owner directory fragmentation where it occurs.


_code-samples/walk-owner-directory/js/iterate-owner-directory.js
```diff
@@ -0,0 +1,55 @@
...
+// Iterate over an account's owner directory and display how many ledger entries
+// are in each page. In cases of highly active accounts, it can demonstrate
+// the extent of "fragmentation" with skipped page numbers and non-full pages.
+
+import xrpl from 'xrpl'
+
+const client = new xrpl.Client('wss://s.altnet.rippletest.net:51233')
+await client.connect()
+
+const owner = "rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe" // Testnet faucet
```
pdp2121: constant should be upper case, same with 2 constants below

_code-samples/walk-owner-directory/js/iterate-owner-directory.js
```diff
@@ -0,0 +1,55 @@
...
+
+// Set initial values for iterating
+let sub_index = 0 // Directory root
+let ledger_index = "validated"
+
+// Query pages from the owner directory until they run out
+console.log("Page #\t\t\tEntry count")
+console.log("-----------------------------------")
+while (true) {
+    // console.log(`Getting directory page ${sub_index}`)
```
pdp2121: I would prefer we keeping the logs (instead of commenting out), so that users are aware of what's going on. Also, for log before an async call, probably consider adding ... so users know to wait for the call to finish

_code-samples/walk-owner-directory/js/iterate-owner-directory.js
```diff
@@ -0,0 +1,55 @@
...
+        console.error("ledger_entry failed with error",resp.error)
+        break
+    }
+
+    // Consistently iterate the same ledger: query by index after the first
+    if (ledger_index === "validated") {
+        ledger_index = resp.result.ledger_index
+    }
+
+    console.log(`${sub_index}\t\t\t${resp.result.node.Indexes.length}`)
```
achowdhry-ripple, line 41: nit: I'd prefer this have a sentence format for clarity or a descriptor of some kind, similar to the commented out log below

_code-samples/walk-owner-directory/py/iterate-owner-directory.py
```diff
@@ -0,0 +1,55 @@
...
+        break
+
+    # The 'ledger_index' is consistently set after the first successful query.
+    # This ensures subsequent pages are read from the same ledger version.
+    if ledger_index == "validated":
+        ledger_index = response.result["ledger_index"]
+
+    # The entries are stored in the 'Indexes' field of the 'DirectoryNode'
+    entry_count = len(response.result["node"]["Indexes"])
+    print(f"{sub_index}\t\t\t{entry_count}")
```
achowdhry-ripple, line 45: same here as other comment

achowdhry-ripple (dismissed):
aside from a few nits looks good!

_code-samples/walk-owner-directory/js/iterate-owner-directory.js
mDuo13: I kept them commented out because that was how the original version of the script worked, but the updated version produces output that is, I think, much easier to understand. It looks like this:

```text
Page #			Entry count
-----------------------------------
0				0
4532			2
4697			1
5121			2
5519			2
6203			1
6660			3
6672			2
6673			1
6681			2
7220			1
7505			2
8468			2
12989			2
14202			2
20696			4
22556			4
23156			1
23779			1
23780			9
23784			9
```

pdp2121 (approved):
LGTM

PR #3399, opened 2025-12-01
Component/button

AKCodez:


static/img/icons/button/button-arrow-hovered.svg
Calvinjwala, line 1: :question: **Question**: are we using this asset since it's an SVG (HTML) in the Button component?

shared/components/Button/Button.tsx
```diff
@@ -0,0 +1,111 @@
...
+    .filter(Boolean)
+    .join(' ');
+
+  return (
+    <button
+      type={type}
+      className={classNames}
+      onClick={onClick}
+      disabled={disabled}
+      aria-disabled={disabled}
```
Calvinjwala, line 128: :see_no_evil: **Nitpick**: We should probably add an `aria-label` for our buttons to improve accessibility

Calvinjwala (approved):
Couple clean up things, but otherwise approved

PR #3403, opened 2025-12-03
Add `hash` request parameter to `server_definitions`

### LINKED ISSUE #3167: Update `server_definitions` docs to include the `hash` field
https://github.com/XRPLF/rippled/blob/develop/src/xrpld/rpc/handlers/ServerInfo.cpp#L312-L318

It is provided in the response. If provided in the request, it will only return the hash and not the definitions (so you know nothing has changed).

mvadari:
Updated server_definitions.md to include request parameters.

Fixes #3167


oeggert (approved):
LGTM

PR #3405, opened 2025-12-04
Add Divider component and documentation

AKCodez:
- Introduced the Divider component following the XRPL Brand Design System, supporting horizontal and vertical orientations, three stroke weights (thin, regular, strong), and three color variants (gray, black, green).
- Created a comprehensive showcase page for the Divider component, demonstrating its usage and variations.
- Added detailed documentation for the Divider component, including props API, usage examples, and best practices.
- Included styles for the Divider component in SCSS format, ensuring compatibility with light and dark themes.


Calvinjwala (changes requested):
Looks good, but I am going to push back on calling one of the color options "black" when in light mode is turns "white" and we call it a different color or even "bnw" or "flat"

PR #3408, opened 2025-12-08
Update public-servers.md with dRPC NodeCloud

dRPCMarketing:
Recently started supporting XRPL so added it to the docs.


dRPCMarketing:
Hi! Was there a problem with the PR? Been trying to get us listed for a while. Thanks

PR #3416, opened 2025-12-09
COIN DEVELOPMENT

SOULDIER1:


mDuo13:
There seems to have been some kind of mixup here. While you appear to be doing some kind of AI-driven development on a coin, this repository is for the XRPL documentation website. So, you're welcome to continue using your fork of this repository for development if it's useful to you, but this is not something that would make sense for us to accept as a contribution to the site. 

You may also want to re-evaluate some things because the commits seem to have a lot of AI hallucinations and I'm not sure they're accomplishing what you're intending, either.

PR #3424, opened 2025-12-19
Remove unnecessary common fields from tables

### LINKED ISSUE #3238: Remove common fields from ledger entry reference tables
Right now, the tables of fields in [ledger entry references](https://xrpl.org/docs/references/protocol/ledger-data/ledger-entry-types) say, "In addition to the common fields..." but then they usually list common fields too.

We should update all the ledger entry reference pages to remove `LedgerEntryType` and `Flags` from the parameter tables, since those are common fields.

oeggert:


maria-robobug (approved):
LGTM

amarantha-k (approved):

PR #3427, opened 2025-12-20
Add lsfMutable flag documentation to NFToken

nabe3m:
In this document only, documentation for the lsfMutable flag (0x00000010) was missing, so it has been added.

### DIFF
--- a/docs/references/protocol/data-types/nftoken.md
+++ b/docs/references/protocol/data-types/nftoken.md
@@ -57,6 +57,7 @@ Flags are properties or other options associated with the `NFToken` object.
 | `lsfTrustLine`    | `0x0004`   | **DEPRECATED** If enabled, automatically create [trust lines](../../../concepts/tokens/fungible-tokens/index.md) to hold transfer fees. Otherwise, buying or selling this `NFToken` for a fungible token amount fails if the issuer does not have a trust line for that token. The [fixRemoveNFTokenAutoTrustLine amendment][] makes it invalid to enable this flag. |
 | `lsfTransferable` | `0x0008`   | If enabled, this `NFToken` can be transferred from one holder to another. Otherwise, it can only be transferred to or from the issuer. |
 | `lsfReservedFlag` | `0x8000`   | This flag is reserved for future use. Attempts to set this flag fail. |
+| `lsfMutable`      | `0x00000010` | If enabled, the `URI` field of the `NFToken` can be updated using the `NFTokenModify` transaction. |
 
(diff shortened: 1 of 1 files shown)


docs/references/protocol/data-types/nftoken.md
```diff
@@ -57,6 +57,7 @@ Flags are properties or other options associated with the `NFToken` object.
 | `lsfTrustLine`    | `0x0004`   | **DEPRECATED** If enabled, automatically create [trust lines](../../../concepts/tokens/fungible-tokens/index.md) to hold transfer fees. Otherwise, buying or selling this `NFToken` for a fungible token amount fails if the issuer does not have a trust line for that token. The [fixRemoveNFTokenAutoTrustLine amendment][] makes it invalid to enable this flag. |
 | `lsfTransferable` | `0x0008`   | If enabled, this `NFToken` can be transferred from one holder to another. Otherwise, it can only be transferred to or from the issuer. |
 | `lsfReservedFlag` | `0x8000`   | This flag is reserved for future use. Attempts to set this flag fail. |
+| `lsfMutable`      | `0x00000010` | If enabled, the `URI` field of the `NFToken` can be updated using the `NFTokenModify` transaction. |
```
mDuo13: Looking at the [source code](https://github.com/XRPLF/rippled/blob/70d5c624e8cf732a362335642b2f5125ce4b43c1/include/xrpl/protocol/nft.h#L53-L57) the flags for NFTokens don't use the `lsf` prefix (which makes sense, since they're not in the `Flags` field of a ledger entry—the flags of an NFT are incorporated into the `NFTokenID`), but are just prefixed `flag` instead. So, that's an error in the existing documentation.

But also, unlike `lsf` flags which are in a 32-bit space, NFT flags are in a 16-bit space. Fixing the `lsf` prefix throughout the docs would be a bigger task, but let's at least fix the size issue:

```suggestion
| `lsfMutable`      | `0x0010`   | If enabled, the `URI` field of the `NFToken` can be updated using the `NFTokenModify` transaction. |
```

mDuo13 (changes requested):
Thanks for catching this. One minor suggestion.

Also, I opened #3433 for following up on the flag naming scheme.

nabe3m:
@mDuo13 
Sorry, I have corrected the incorrect flag value.

Although there is an open issue, if prioritizing the addition of `lsfMutable` is preferred, please proceed with the merge.

mDuo13 (approved):

PR #3440, opened 2026-01-09
Update reserve requirements in genesis ledger documentation

### LINKED ISSUE #3439: Stand-alone mode docs have out-of-date numbers for default reserve values
https://xrpl.org/docs/infrastructure/testing-and-auditing/start-a-new-genesis-ledger-in-stand-alone-mode#settings-in-new-genesis-ledgers

The doc has the old numbers of 200 XRP per account and 50 XRP per item owned, but the latest source has them hard-coded to 10/2: https://github.com/XRPLF/rippled/blob/develop/src/xrpld/core/Config.h#L46-L59

lmaisons:
Taken from latest release at:
https://github.com/XRPLF/rippled/blob/3.0.0/src/xrpld/core/Config.h#L60-L78

Resolves #3439


mDuo13 (approved):
🎉

PR #3441, opened 2026-01-12
add prettier, lint all files

### LINKED ISSUE #3202: Add Markdown formatter
Use a formatter to ensure consistency in Markdown format.

- Local development environment formatting
- GitHub Action formatting check

mvadari:
Fixes #3202 

The only non-automated changes made here are in `.prettierrc.yaml`.


mDuo13 (commented):
I appreciate the idea, but I think this is too much and not quite the right time.

I don't think we can reasonably spot-check all the changes that this makes to 1600+ files, and knowing what I do about the quirks of Markdown parsers and Markdoc formatting I would not be surprised if this introduced some new bugs somewhere in the docs because it deleted load-bearing whitespace or replaced double quotes in code that were being referenced in docs, or something along those lines. (In fact, I'd be more surprised if it _didn't_ break something somewhere in the site, even knowing that linters are supposed to be functionally harmless.)

Also, there are two big reorganizations currently underway that are bound to cause merge conflicts: one with the organization of the tutorials section (moving and merging dozens of files) and one that refreshes the visual style and marketing portion of the site.

So maybe it's best to put this on hold for a little bit and then do it right after those two are merged.

One rule I think we will definitely want to change is regarding the padding of markdown tables. We frequently have a long final "Description" column that is expected to get word-wrapped, so padding out the header rows to match it results in multiple pointless lines' worth of hyphens, as in this example:

mvadari:
@mDuo13 thanks for the feedback. I just did `prettier` support for a couple other repos in the past couple of months, so I figured it'd be easy enough to just copy that here and open a PR for discussion (easier than someone who doesn't have that experience doing it, at least).

> I appreciate the idea, but I think this is too much and not quite the right time.
> 
> I don't think we can reasonably spot-check all the changes that this makes to 1600+ files, and knowing what I do about the quirks of Markdown parsers and Markdoc formatting I would not be surprised if this introduced some new bugs somewhere in the docs because it deleted load-bearing whitespace or replaced double quotes in code that were being referenced in docs, or something along those lines. (In fact, I'd be more surprised if it _didn't_ break something somewhere in the site, even knowing that linters are supposed to be functionally harmless.)



> Also, there are two big reorganizations currently underway that are bound to cause merge conflicts: one with the organization of the tutorials section (moving and merging dozens of files) and one that refreshes the visual style and marketing portion of the site.

The beauty of auto-generated changes is that it's quite easy to handle changes afterwards - just take all your changes in the merge conflicts and re-rerun the script/command/etc to re-lint.

> So maybe it's best to put this on hold for a little bit and then do it right after those two are merged.
> 
> One rule I think we will definitely want to change is regarding the padding of markdown tables. We frequently have a long final "Description" column that is expected to get word-wrapped, so padding out the header rows to match it results in multiple pointless lines' worth of hyphens, as in this example:

Yeah, that's fair. As I was working on this I discovered that prettier is rather opinionated and doesn't have a lot of options, so it may not be the right option for this repo (at least, for the Markdown files). [markdownlint](https://github.com/DavidAnson/markdownlint) might be a better option here, it's got a lot more options - so we might be able to toggle them a bunch to get a really small diff and slowly play with it to get something that both compiles and is more consistent.

Closing this PR based on this feedback. I think it's potentially worth revisiting using prettier just on the JS files, though.

PR #3443, opened 2026-01-13
Brand update header

AKCodez:


@theme/components/Navbar/Navbar.tsx
```diff
@@ -1,462 +1,1395 @@
...
+        { label: "Python", href: "/docs/references/xrpl-py" },
+        { label: "PHP", href: "/docs/references/xrpl-php" },
+        { label: "Go", href: "/docs/references/xrpl-go" },
+      ],
+    },
+  ],
+};
+
+// Use Cases submenu data structure
+const useCasesSubmenuData: {
```
Calvinjwala: :bulb: **Suggestion**: We should parse these out into smaller subnavs and then import into this file

@theme/components/Navbar/Navbar.tsx
```diff
@@ -1,462 +1,1395 @@
...
+// Network submenu icons
+import insightsIcon from "../../../static/img/navbar/insights.svg";
+import resourcesIcon from "../../../static/img/navbar/resources.svg";
+
+// Network submenu pattern images
+import resourcesPurplePattern from "../../../static/img/navbar/resources-purple.svg";
+import insightsGreenPattern from "../../../static/img/navbar/insights-green.svg";
+import darkInsightsGreenPattern from "../../../static/img/navbar/dark-insights-green.svg";
+import darkLilacPattern from "../../../static/img/navbar/dark-lilac.svg";
+
```
Calvinjwala: :bulb: **Suggestion**: It might be best to move these images + the subnav to a separate file

Calvinjwala (changes requested):
@AKCodez I think it's best if the Navbar is broken out into smaller files. This file is almost ~1400 lines, can we separate into more manageable file system

Calvinjwala:
@AKCodez `search button` + `language picker` don't work for me in local.

Also, the spacing/margin under "Payments" is off compared to the other tiles.
Screenshot 2026-01-12 at 7 33 36 PM

@theme/components/Navbar/constants/icons.ts
```diff
@@ -0,0 +1,85 @@
...
+export { default as resourcesIcon } from "../../../../static/img/navbar/resources.svg";
+
+// Network submenu pattern images
+export { default as resourcesPurplePattern } from "../../../../static/img/navbar/resources-purple.svg";
+export { default as insightsGreenPattern } from "../../../../static/img/navbar/insights-green.svg";
+export { default as darkInsightsGreenPattern } from "../../../../static/img/navbar/dark-insights-green.svg";
+export { default as darkLilacPattern } from "../../../../static/img/navbar/dark-lilac.svg";
+
+// Wallet icon mapping for dynamic icon lookup
+import greenWallet from "../../../../static/img/navbar/green-wallet.svg";
```
Calvinjwala, line 48: :question: **Question**: It seems like these icons are imported twice? can't we use `greenWallet` from the one above to the `walletIcons`?

@theme/components/Navbar/Navbar.tsx
```diff
@@ -1,462 +1,139 @@
...
+      // Start closing animation
+      const currentSubmenu = activeSubmenu;
+      if (currentSubmenu) {
+        setClosingSubmenu(currentSubmenu);
+        setActiveSubmenu(null);
+        
+        // After animation completes (300ms), clear closing state
+        closingTimeoutRef.current = setTimeout(() => {
+          setClosingSubmenu(null);
+        }, 350); // Slightly longer than animation to ensure completion
```
Calvinjwala, line 75: :see_no_evil: **Nitpick**: might be better to parse out the animation times into a constants/animation.ts file and the export timing const from there

@theme/components/Navbar/icons/CloseIcon.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from "react";
+
+/**
+ * Close Icon Component for Mobile Menu
+ */
+export function CloseIcon() {
+  return (
+    <svg width="28" height="28" viewBox="0 0 28 28" fill="none" xmlns="http://www.w3.org/2000/svg">
+      <line x1="7" y1="7" x2="21" y2="21" stroke="#141414" strokeWidth="2" strokeLinecap="round" />
```
Calvinjwala, line 9: :bulb: **Suggestion**: don't hard code the color and instead pass it with "currentColor"

Calvinjwala (changes requested):
Looks good overall, but a couple UI bugs I spotted:
1. `Tab` is not working for the nav - ideally we can make sure this is ARIA-compliant
2. Should lang picker be available if nav is open?
Image
3. Should we lock the background when nav is open. You can currently scroll and there is a white div that follows
Image

Calvinjwala:
More of a design question, but I'd think this should span 100% on mobile.
Screenshot 2026-01-13 at 12 04 53 PM

@theme/components/Navbar/icons/CloseIcon.tsx
AKCodez: makes no difference as the color will be overriden

AKCodez:
> More of a design question, but I'd think this should span 100% on mobile. Screenshot 2026-01-13 at 12 04 53 PM

added

Calvinjwala (approved):
LGTM!

PR #3446, opened 2026-01-14
Pattern/feature two column

AKCodez:


shared/patterns/FeatureTwoColumn/FeatureTwoColumn.scss
```diff
@@ -0,0 +1,482 @@
...
+
+  // Mobile media column
+  &__media-col--mobile {
+    display: flex;
+    flex-direction: column;
+  }
+
+  // Content
+  &__content {
+    display: flex;
```
Calvinjwala, line 210: :bulb: **Suggestion**: Let's use `@extend` where applicable

shared/patterns/FeatureTwoColumn/FeatureTwoColumn.scss
```diff
@@ -0,0 +1,482 @@
...
+// =============================================================================
+// Design Tokens
+// =============================================================================
+
+// Background colors for each color variant (light mode)
+// From styles/_colors.scss - Design tokens
+$bds-feature-neutral-bg: $gray-100; // #F0F3F7 (Neutral-100)
+$bds-feature-lilac-bg: $lilac-200; // #D9CAFF (Primary-Lilac-200)
+$bds-feature-yellow-bg: $yellow-100; // #F3F1EB (Secondary-Yellow-100)
+$bds-feature-green-bg: $green-300; // #21E46B (Primary-Green-300default)
```
Calvinjwala: :bulb: **Suggestion**: Let's add a mapping function for these variations

shared/patterns/FeatureTwoColumn/FeatureTwoColumn.scss
```diff
@@ -0,0 +1,482 @@
...
+// Grid gutter - consistent with PageGrid
+$bds-grid-gutter: 8px;
+
+// =============================================================================
+// Base Styles
+// =============================================================================
+.bds-feature-two-column__button-group {
+  .bds-btn--tertiary {
+    padding-top: 0px !important;
+    padding-bottom: 0px !important;
```
Calvinjwala, line 72: :no_entry_sign: **Must fix**: This is causing issues - I don't think we should be removing default padding for the Button component. This causing other issues:

Image

shared/patterns/FeatureTwoColumn/FeatureTwoColumn.scss
```diff
@@ -0,0 +1,482 @@
...
+      // Match height of media column (which has aspect-ratio 1/1)
+      aspect-ratio: 1 / 1;
+    }
+  }
+
+  // Content grid - positions content within column using grid system
+  // Desktop: 6-column grid within the content half, content spans cols 2-5 (offset-1, span-4)
+  // Tablet: 8-column grid, content spans cols 2-7 (offset-1, span-6)
+  // Mobile: 4-column grid, content spans cols 1-4 (no offset, span-4)
+  &__content-grid {
```
Calvinjwala, line 148: :question: **Question**: seems like there was no way to make this layout work with the current grid?

shared/patterns/FeatureTwoColumn/FeatureTwoColumn.tsx
```diff
@@ -0,0 +1,214 @@
...
+  const buttonColor = 'black';
+  const forceColor = true;
+
+  // Render content section with appropriate CTA layout based on link count
+  // For 3-5 links, items are direct children for space-between distribution
+  const renderContent = () => {
+    const linkCount = links.length;
+
+    // 1 link: Secondary button
+    if (linkCount === 1) {
```
Calvinjwala, line 81: :bulb: **Suggestion**: Let's combine these two to say if `linkCount < 3` since it should be the same layout with 1 or 2 buttons.

shared/patterns/FeatureTwoColumn/FeatureTwoColumn.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+    // 2 links: Primary + Tertiary in a row
+    if (linkCount === 2) {
+      return (
+        <div className="bds-feature-two-column__content">
+          <div className="bds-feature-two-column__text-group">
+            <h2 className="bds-feature-two-column__title">{title}</h2>
+            <p className="bds-feature-two-column__description">{description}</p>
+          </div>
+          <div className="bds-feature-two-column__cta bds-feature-two-column__cta--double">
```
Calvinjwala, line 105: :no_entry_sign: **Must fix**: The layout shouldn't change between 1 or 2 buttons

Calvinjwala (changes requested):
Looks good, but there are some issues with the button layout. Ideally, we could also consolidate some of the CSS but due to time I'm fine with the CSS grid being as is.

Image

AKCodez:
> Looks good, but there are some issues with the button layout. Ideally, we could also consolidate some of the CSS but due to time I'm fine with the CSS grid being as is.
> 
> Image

Did demo and coding session with Ricky buttons are exactly how they are suppose to be. Can consult with Ricky

shared/patterns/FeatureTwoColumn/FeatureTwoColumn.scss
AKCodez: I'm removing the padding from within the section implementation not the button component itself

shared/patterns/FeatureTwoColumn/FeatureTwoColumn.scss
AKCodez: Couldn't do it, Opus gave up

shared/patterns/FeatureTwoColumn/FeatureTwoColumn.tsx
AKCodez: The layout logic came directly from Z they are unique to the amount of links passed

shared/patterns/FeatureTwoColumn/FeatureTwoColumn.tsx
```diff
@@ -0,0 +1,214 @@
...
+          </div>
+          {/* Secondary button */}
+          {links[2] && (
+            <Button variant="secondary" color={buttonColor} forceColor={forceColor} href={links[2].href}>
+              {links[2].label}
+            </Button>
+          )}
+          {/* Remaining tertiary links */}
+          {links.slice(3).map((link, index) => (
+            <Button key={index} variant="tertiary" color={buttonColor} forceColor={forceColor} href={link.href}>
```
Calvinjwala: :no_entry_sign: **Must fix**: let's avoid `index` as key. use **key={`${link.href}-${link.label}`}** instead

PR #3459, opened 2026-01-22
Add support for inactive amendments in tracker

oeggert:
This PR fixes an issue where missing amendments from the VHS Vote endpoint caused an error. This also includes minor cleanup to the PermissionDelegation info in the known amendments page.

[Example of error](https://xrpl.org/docs/concepts/accounts/permission-delegation)

**Old Behavior**
image
**New Behavior**
image

The fix now queries the VHS Info endpoint as a secondary check if the first one fails to fetch the amendment. If the amendment is found on this second list, its treated as "Inactive" with a link to get further details on what that means (Obsolete vs In Development). The VHS API doesn't have the info required to programmatically determine if an amendment is inactive due to being in development or obsolete.

@mDuo13 Japanese translations were pulled from google, please check they are correct.

### DIFF
--- a/@l10n/ja/translations.yaml
+++ b/@l10n/ja/translations.yaml
@@ -155,7 +155,7 @@ amendment.table.status: ステータス
 amendment.status.enabled: 有効
 amendment.status.eta: 予定
 amendment.status.openForVoting: 投票中
-amendment.status.inactive: 非アクティブ
+amendment.status.inactive: 無効
 amendment.status.inactiveButton: 詳細を取得する
 
 # index.page.tsx
--- a/@l10n/ja/translations.yaml
+++ b/@l10n/ja/translations.yaml
@@ -155,6 +155,8 @@ amendment.table.status: ステータス
 amendment.status.enabled: 有効
 amendment.status.eta: 予定
 amendment.status.openForVoting: 投票中
+amendment.status.inactive: 非アクティブ
+amendment.status.inactiveButton: 詳細を取得する
 
 # index.page.tsx
 home.hero.h1part1: ビジネスのための
--- a/@theme/components/Amendments.tsx
+++ b/@theme/components/Amendments.tsx
@@ -23,6 +23,7 @@ type AmendmentsCachePayload = {
 
 // API data caching
 const amendmentsEndpoint = 'https://vhs.prod.ripplex.io/v1/network/amendments/vote/main/'
+const amendmentsInfoEndpoint = 'https://vhs.prod.ripplex.io/v1/network/amendments/info/main/'
 const amendmentsCacheKey = 'xrpl.amendments.mainnet.cache'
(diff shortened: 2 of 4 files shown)


@l10n/ja/translations.yaml
```diff
@@ -155,6 +155,8 @@ amendment.table.status: ステータス
 amendment.status.enabled: 有効
 amendment.status.eta: 予定
 amendment.status.openForVoting: 投票中
+amendment.status.inactive: 非アクティブ
+amendment.status.inactiveButton: 詳細を取得する
```
mDuo13: I think 無効 might be a better translation than 非アクティブ for "Inactive", but maybe @tequdev would like to weigh in?

resources/known-amendments.md
```diff
@@ -1683,11 +1683,11 @@ For more information, see the [Payment Channels Tutorial](../docs/tutorials/how-
 | Amendment    | PermissionDelegation |
 |:-------------|:---------------------|
 | Amendment ID | AE6AB9028EEB7299EBB03C7CBCC3F2A4F5FBE00EA28B8223AA3118A0B436C1C5 |
-| Status       | Open for Voting |
+| Status       | Disabled |
```
mDuo13: We don't need to invent a new "Disabled" status for PermissionDelegation; the existing "Obsolete" status covers it. If you look at the list of [Obsolete Amendments](https://xrpl.org/resources/known-amendments#obsolete-amendments) several of them were buggy and replaced by a renamed amendment with the same overall purposes. (SusPay→Escrow, FlowV2→Flow, Tickets→TicketBatch)


```suggestion
| Status       | Obsolete |
```

We should add PermissionDelegation to the Obsolete Amendments table though.

mDuo13 (changes requested):

mDuo13 (approved):
I'm OK with the changes now but I have reached out to @tequdev in case he wants to weigh in on the translation. If we don't hear back in a day or two I suggest we merge and follow up on the translation with a new PR as needed.

@l10n/ja/translations.yaml
mDuo13: Got a reply from tequ, who says "無効" looks more natural.


```suggestion
amendment.status.inactive: 無効
amendment.status.inactiveButton: 詳細を取得する
```

@theme/components/Amendments.tsx
```diff
@@ -389,6 +415,8 @@ export function Badge(props: {
       "更新": "blue", // ja: updated in
       "in development": "lightgrey",
       "開発中": "lightgrey", // ja: in development
+      "inactive": "lightgrey",
+      "非アクティブ": "lightgrey" // ja: inactive
```
mDuo13:
```suggestion
      "無効": "lightgrey" // ja: inactive
```

mDuo13 (changes requested):

mDuo13 (approved):

PR #3462, opened 2026-01-23
Update padding for CTA buttons in HeaderHeroPrimaryMedia component

gabriel-ortiz:
- Added !important to padding for hover and focus states of tertiary buttons to ensure consistent styling across interactions.


Calvinjwala (approved):
Approved!

PR #3489, opened 2026-02-11
Add Bitget wallet to XRP and Uses pages

amarantha-k:


mDuo13 (approved):
I noticed on the Use Cases page the filter option isn't expanding anymore, but that's also true in production.

The changes from this PR look fine though.

Issue #3486, opened 2026-02-09
Add missing code sample readmes


mDuo13:
Our [contributor guidelines](https://xrpl.org/resources/contribute-documentation/tutorial-guidelines#folder-layout) suggest that every code sample should have one top-level `README.md` file explaining the sample itself (which is used by the Code Samples page for the description of the code sample card), and another `README.md` file in each language subdirectory with instructions on how to install dependencies and run that code sample in that language. However, many code samples are missing these, something like 97 instances at the moment.

Run the following Python code from the top of the repo to get an updated list of directories that are missing a README:

```py
import os
for dirpath, dirnames, filenames in os.walk("./_code-samples/"):
     pathlist = dirpath.split(os.path.sep)
     if len(pathlist) < 3 or len(pathlist) > 4:
         continue
     if set(pathlist) & {"node_modules", ".venv", "venv", "__pycache__"}:
         continue
     if "README.md" not in filenames:
         print(dirpath[16:])
```

mDuo13:
As of #3493 this is done for JS and Python so it's only the Go and other miscellaneous code samples that don't have it at this point.

amarantha-k:
This ticket can be closed as JS and Python code samples have been updated. 

I've created https://github.com/XRPLF/xrpl-dev-portal/issues/3588 to track the remaining tasks, which can be addressed as part of our effort to increase code coverage for other languages.

PR #1684, opened 2023-01-20
Python/JS code sample for account deletion

wojake:
Delete (or blackhole) an account

This is work for a XRPL Bounty @ [XRPLBounties/Proposals](https://github.com/XRPLBounties/Proposals/blob/main/open_bounties/0079%20Python%20Code%20Samples%202.md) for Milestone 5 and [XRPLBounties/Proposals](https://github.com/XRPLBounties/Proposals/blob/main/open_bounties/0077%20JavaScript%20Code%20Samples%202.md) for Milestone 8

PR #3493, opened 2026-02-12
Add missing READMEs for JS and Python code samples

maria-robobug:
Fixes issue #3486

- Adds missing READMEs for JS and Python code samples **only**. Figured they should be the ones we focus on for now, and it's already a lot of files.
- Adds package.json (JS) and requirements.txt (Python) where relevant.

**Note**: We have some Typescript examples that didn't actually seem runnable, so I've made sure users can run these now.


_code-samples/delete-account/js/README.md
mDuo13, line 1: There's something weird going on here. The folder and this new README say that it shows how to _delete_ an account, but actually it shows how to blackhole an account, which is different.

I guess this confusion dates back to #1684 which was merged before anyone pointed out the discrepancy. 🤦‍♂

mDuo13: I suggest we remove these faux "delete account" code samples entirely:

- They don't embody best practices: for example, the JS one uses the `if (typeof module !== "undefined")` thing which was a hack meant for scripts to run in browsers, but it's not used in a browser.
- They handle flags completely incorrectly. The Python code defines `asfDisableMaster = 1114112` (which is wrong on multiple levels) and the JavaScript code checks to see if account's flags are equal to the value 1114112 (the combination of `lsfDisableMaster` and `lsfPasswordSpent`)—this gives the _incorrect_ result in several circumstances such as if the account has received a payment of XRP (resetting its "password spent" flag) or if it had any other flags enabled (like Default Ripple—a flag you would often want enabled in the same use case where you want to blackhole an account). 
- The (recently updated) "Disable Master Key Pair" tutorial covers this same process but with more updated conventions and more accuracy. Even though blackholing an account is not the main goal of that tutorial & code sample, it does have explicit notes on what to do if you want to black hole an account.

We should (separately) add code samples and a tutorial of how to _actually_ delete an account.

mDuo13 (changes requested):
Mostly looks good, but this brought my attention to a (previously-existing) problem with the "delete account" code sample.

mDuo13 (approved):

PR #3494, opened 2026-02-12
Move translations to match English tutorials reorg & fix broken links

mDuo13:
- Move translations to match their places in the new IAv4 for tutorials, and fix broken links resulting from the move. The changes are _almost_ entirely in the Japanese localization, but one or two Spanish translations are also affected.
- Fix a couple typos in the English versions of some files.
- Remove the outdated "Require Destination Tags" translation. The page has been almost entirely rewritten since the translated version, which does not represent best practices.


amarantha-k (approved):
LGTM

PR #3495, opened 2026-02-13
Lending Protocol Tutorials and Python Code Samples

oeggert:
Changes:

- Add Python code samples for Lending
- Update JavaScript LoanSet code samples to use local signing
- Add Python to Lending tutorials


_code-samples/lending-protocol/py/cover_clawback.py
```diff
@@ -0,0 +1,130 @@
...
+
+from xrpl.asyncio.clients import AsyncWebsocketClient
+from xrpl.asyncio.transaction import submit_and_wait
+from xrpl.models import LedgerEntry, LoanBrokerCoverClawback, LoanBrokerCoverDeposit, MPTAmount
+from xrpl.wallet import Wallet
+
+
+async def main():
+    # Connect to the network ----------------------
+    async with AsyncWebsocketClient("wss://s.devnet.rippletest.net:51233") as client:
```
maria-robobug: I thought for Python we recommended using the JSON-RPC client, is there a reason we're using AsyncWebsocketClient for these examples?

oeggert: So my thinking was that in a production environment you'd use async clients since you need to scale large numbers of transactions. That's also why I did the websocket connection since it enables you to handle websocket only features like subscribe.

I figured it's better to pick the most flexible client, even if the scripts don't take full advantage of it. I'm not sure if that's considered bad practice though and I should use clients that match the needs of the code.

maria-robobug: Not a bad practice, but I guess I would only consider using the `AsyncWebsocketClient` if the example code actually needs to make use of it. Otherwise, I would avoid using it just because it adds complexity with Python.

https://xrpl-py.readthedocs.io/en/latest/source/xrpl.asyncio.clients.html#xrpl.asyncio.clients.AsyncWebsocketClient

oeggert: Alrighty I'll change it to the json-rpc client.

docs/tutorials/how-tos/set-up-lending/use-the-lending-protocol/create-a-loan-broker.md
```diff
@@ -40,45 +41,69 @@ You can find the complete source code for this tutorial's examples in the {% rep
...
 {% tabs %}
 {% tab label="JavaScript" %}
 {% code-snippet file="/_code-samples/lending-protocol/js/createLoanBroker.js" language="js" from="// Prepare LoanBrokerSet" before="// Submit, sign" /%}
 {% /tab %}
+{% tab label="Python" %}
+{% code-snippet file="/_code-samples/lending-protocol/py/create_loan_broker.py" language="py" from="# Prepare LoanBrokerSet" before="# Submit, sign" /%}
+{% /tab %}
 {% /tabs %}
 
 The `ManagementFeeRate` is set in 1/10th basis points. A value of `1000` equals 1% (100 basis points).
```
maria-robobug: We may need to repeat this for Python with `management_fee_rate`.

Or, just keep this generic:


```suggestion
The management fee rate is set in 1/10th basis points. A value of `1000` equals 1% (100 basis points).
```

docs/tutorials/how-tos/set-up-lending/use-the-lending-protocol/claw-back-cover.md
```diff
@@ -92,6 +117,9 @@ Create the [LoanBrokerCoverDeposit transaction][] object.
 {% tab label="JavaScript" %}
 {% code-snippet file="/_code-samples/lending-protocol/js/coverClawback.js" language="js" from="// Prepare LoanBrokerCoverDeposit" before="// Sign, submit, and wait for deposit validation" /%}
 {% /tab %}
+{% tab label="Python" %}
+{% code-snippet file="/_code-samples/lending-protocol/py/cover_clawback.py" language="py" from="# Prepare LoanBrokerCoverDeposit" before="# Sign, submit, and wait for deposit validation" /%}
+{% /tab %}
 {% /tabs %}
 
 The `Amount` field specifies the MPT and amount to deposit as first-loss capital. If the transaction succeeds, the amount is deposited and held in the pseudo-account associated with the `LoanBroker` entry.
```
maria-robobug: Move this in both JS and Python tabs:

JS: The `Amount` field specifies the MPT and amount to deposit as first-loss capital.

Python: The `amount` field specifies the MPT and amount to deposit as first-loss capital.

We can leave "If the transaction..." as it's generic enough for both.

docs/tutorials/how-tos/set-up-lending/use-the-lending-protocol/deposit-and-withdraw-cover.md
```diff
@@ -42,45 +43,69 @@ You can find the complete source code for this tutorial's examples in the {% rep
...
 {% tabs %}
 {% tab label="JavaScript" %}
 {% code-snippet file="/_code-samples/lending-protocol/js/coverDepositAndWithdraw.js" language="js" from="// Prepare LoanBrokerCoverDeposit" before="// Sign, submit, and wait for deposit" /%}
 {% /tab %}
+{% tab label="Python" %}
+{% code-snippet file="/_code-samples/lending-protocol/py/cover_deposit_and_withdraw.py" language="py" from="# Prepare LoanBrokerCoverDeposit" before="# Sign, submit, and wait for deposit" /%}
+{% /tab %}
 {% /tabs %}
 
 The `Amount` field specifies the MPT and amount to deposit as first-loss capital. If the transaction succeeds, the amount is deposited and held in the pseudo-account associated with the `LoanBroker` entry.
```
maria-robobug: Move this in both JS and Python tabs:

JS: The `Amount` field specifies the MPT and amount to deposit as first-loss capital.

Python: The `amount` field specifies the MPT and amount to deposit as first-loss capital.

We can leave "If the transaction..." as it's generic enough for both.

docs/tutorials/how-tos/set-up-lending/use-the-lending-protocol/create-a-loan.md
```diff
@@ -43,45 +44,69 @@ You can find the complete source code for this tutorial's examples in the {% rep
...
 {% tabs %}
 {% tab label="JavaScript" %}
 {% code-snippet file="/_code-samples/lending-protocol/js/createLoan.js" language="js" from="// Prepare LoanSet" before="// Loan broker signs first" /%}
 {% /tab %}
+{% tab label="Python" %}
+{% code-snippet file="/_code-samples/lending-protocol/py/create_loan.py" language="py" from="# Prepare LoanSet" before="# Loan broker signs first" /%}
+{% /tab %}
 {% /tabs %}
 
 The `Account` field is the loan broker, and the `Counterparty` field is the borrower. These fields can be swapped, but determine the signing order: the `Account` signs first, and the `Counterparty` signs second.
```
maria-robobug, line 107: We may have to duplicate this for the Python tab:

The `account` field is the loan broker, and the `counterparty` field is the borrower. These fields can be swapped, but determine the signing order: the `account` signs first, and the `counterparty` signs second.

maria-robobug (commented):
LGTM, minor comments/suggestions.

_code-samples/lending-protocol/py/lending_setup.py
```diff
@@ -0,0 +1,366 @@
...
+    VaultCreate,
+    VaultDeposit,
+)
+from xrpl.models.transactions.vault_create import VaultCreateFlag
+from xrpl.models.transactions.deposit_preauth import Credential
+from xrpl.utils import encode_mptoken_metadata, str_to_hex
+
+
+async def main():
+    async with AsyncWebsocketClient("wss://s.devnet.rippletest.net:51233") as client:
```
maria-robobug, line 35: I guess we can leave this Async if it helps speed up the set up?

oeggert: Yeah doing it async cuts the set up time by like half.

maria-robobug (approved):

PR #3498, opened 2026-02-13
Upgrade Redocly & remove unused htmltojsx

mDuo13:
- Update to the latest Redocly version (0.130.4). (**Updated**)
- Remove the `htmltojsx` devDependency that was the source of all the alerts about critical security vulnerabilities. As far as I know we don't actually use it now—Roman had it in the initial version of our Redocly config file, so maybe it was used for some migration scripts or something and it's been carried over unchanged since then, but as far as I can tell nothing references it and nothing broke when I removed it.
- Remove now-redundant redirects for Japanese pages. In the new Redocly version, these work implicitly. (e.g. `/ja/rippling.html` → `/ja/docs/concepts/tokens/fungible-tokens/rippling` works based on the `/rippling.html`→ `/docs/concepts/tokens/fungible-tokens/rippling` redirect without needing a separate entry for the Japanese version.)
- Remove many instances of `type: 301` from the redirects file. This field is optional and the default is 301. I am hoping that by removing this we will have fewer merge conflicts later because the diff won't find "matching" lines in the middle of a change.

Tested and working in local dev:

- Homepage
- Code Samples page 
- All the built-in dev tools
- Translated pages, switching between them
- Search - results seem a little less good than on production but it's not outright broken
- :new: Implicit redirects for localized files

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "@codemirror/state": "6.5.2",
         "@codemirror/view": "^6.22.2",
         "@lezer/highlight": "^1.2.0",
-        "@redocly/realm": "0.126.0",
+        "@redocly/realm": "0.130.2",
         "@uiw/codemirror-themes": "4.21.21",
         "@uiw/react-codemirror": "^4.21.21",
         "@xrplf/isomorphic": "^1.0.0-beta.1",
@@ -31,7 +31,6 @@
       },
       "devDependencies": {
         "bootstrap": "^4.6.2",
-        "htmltojsx": "^0.3.0",
         "sass": "1.26.10"
       }
     },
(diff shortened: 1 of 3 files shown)


amarantha-k:
npm flags security issues (details below); waiting for Redocly's response. 
```
...
ajv <8.18.0
Severity: moderate
ajv has ReDoS when using `$data` option - https://github.com/advisories/GHSA-2g4f-4pwh-qvx6
fix available via `npm audit fix --force`
Will install @redocly/realm@0.128.1, which is a breaking change
...
fast-xml-parser 4.1.3 - 5.3.5
Severity: high
fast-xml-parser affected by DoS through entity expansion in DOCTYPE (no expansion limit) - https://github.com/advisories/GHSA-jmr7-xgp7-cmfj
fix available via `npm audit fix --force`
Will install @redocly/realm@0.128.1, which is a breaking change
...
```

mDuo13:
Switched it to version 0.130.4 now that that version is out and updates `ajv` and `fast-xml-parser` to resolve the security alerts on those.

amarantha-k (approved):
Looks good.

PR #3499, opened 2026-02-14
Update & combine 'Verify Credentials' tutorials

mDuo13:
- Update the Python version of the sample code to match the simplified JS version introduced in #3230
- Remove the separate Python version of this tutorial and add tabs to the combined version
- Update the sample code in both versions to look up an example Credential on Mainnet that should remain in place until it expires in 2038. The sample data that the code samples previously used was on Devnet and is no longer there, probably due to a Devnet reset. Using Mainnet for long-lasting sample read-only data is consistent with our new persistent data guidelines.

Not covered in this PR: adding redirects. Since this is going to the non-deployed IAv4 branch, I want to wait and not add redirects to/from transient URLs that won't be deployed to production. My script to generate redirects should be able to handle this all at once when we're ready.


_code-samples/verify-credential/py/verify_credential.py
```diff
@@ -1,167 +1,65 @@
...
+# Confirm that the credential is not expired -----------------------------------
+if credential.get("Expiration"):
+    expiration_time = ripple_time_to_datetime(credential["Expiration"])
+    print("Credential has expiration:", expiration_time.isoformat())
+    print("Looking up validated ledger to check for expiration.")
+
+    ledger_response = client.request(Ledger(ledger_index="validated"))
+    if ledger_response.status != "success":
+        print("Error looking up most recent validated ledger:", ledger_response.result)
+        exit(3)
```
maria-robobug, line 56: "Error looking up most recent validated ledger:"

This error is missing in the Javascript example, we should add it in.

maria-robobug (commented):
Looks good, just minor comments

docs/tutorials/compliance-features/verify-credentials.md
```diff
@@ -26,7 +26,7 @@ To complete this tutorial, you should:
 - Have a basic understanding of the XRP Ledger.
 - Have an XRP Ledger client library, such as [xrpl.js](../get-started/get-started-javascript.md), installed.
```
maria-robobug:
```suggestion
- Have an [XRP Ledger client library](../../references/client-libraries.md), such as **xrpl.js**, installed.
```

maria-robobug:
The existing Python README needs updating as we can no longer run this as a commandline utility.

maria-robobug (approved):
LGTM

PR #3502, opened 2026-02-17
update gpg key expected output

oeggert:


mDuo13 (approved):
It's probably also wise to publish a blog post about updating the key.

PR #3509, opened 2026-02-23
Update Tutorials Landing Page

maria-robobug:
Updated the Tutorials landing page:

- Updated Tutorials `index.md` with an `index.page.tsx` page that shows a curated list of tutorials under different categories.
- Created a plugin that scans tutorial pages for code tabs and displays the detected language icons on cards. Maybe we can expand what this does in future?
- Used Japanese translations that already existed for the old `index.md` page.
- Also fixed the order of the Lending tutorials and removed the landing pages since we are not using these anymore.
- Matched tutorial headings to language suffix in filename (there were a couple I'd missed in a previous change).

image

image

I decided to show the following categories, but let me know if too many or if need to change:

1. **Get Started** - All the SDKs (JavaScript, Python, Go, Java, PHP, HTTP/WebSocket APIs)
2. **Tokens**
    - Issue a Multi-Purpose Token
    - Issue a Fungible Token
    - Mint and Burn NFTs Using JavaScript
3. **Payments**
    - Send XRP
    - Sending MPTs in JavaScript
    - Create Trust Line and Send Currency (JS & Python)
    - Send a Conditional/Timed Escrow
4. **DeFi**
    - Create an Automated Market Maker
    - Trade in the Decentralized Exchange
    - Lending tutorials (Loan Broker, Create a Loan, Vaults)
5. **Best Practices**
    - API Usage
    - Use Tickets
    - Batch Transactions
    - Key Management (Regular Key, Multi-Signing)
6. **Sample Apps**
    - Browser/Desktop Wallets (JS & Python)
    - Credential Issuing Service (JS & Python)


docs/tutorials/index.page.tsx
```diff
@@ -0,0 +1,353 @@
...
+      {
+        title: "Issue a Multi-Purpose Token",
+        body: "Issue new tokens using the v2 fungible token standard.",
+        path: "/docs/tutorials/tokens/mpts/issue-a-multi-purpose-token/",
+      },
+      {
+        title: "Issue a Fungible Token",
+        body: "Issue new tokens using the v1 fungible token standard.",
+        path: "/docs/tutorials/tokens/fungible-tokens/issue-a-fungible-token/",
+      },
```
mDuo13, line 97: I'm concerned about the maintenance burden of these big JSON objects. It feels inevitable that they will fall out of sync with the sidebar and contents of the files.

Since we already autodetect programming languages the tutorial files, I think we should plan to make this source as much more data as possible in the same way, like the titles and body blurbs.

Perhaps that can be done in a follow-up task (with a ticket) rather than being a blocker stopping us from moving forward with this PR though.

docs/tutorials/index.page.tsx
```diff
@@ -0,0 +1,353 @@
...
+  detectedLanguages,
+  showFooter = false,
+  translate,
+}: {
+  tutorial: Tutorial
+  detectedLanguages?: string[]
+  showFooter?: boolean
+  translate: (text: string) => string
+}) {
+  // Get icons: auto-detected languages take priority, then manual icon, then XRPL fallback
```
mDuo13: Nit: As a best practice, I tend to think the priority should be manual > auto-detected > fallback. That way, you can override incorrect auto-detections with a manual icon.

docs/tutorials/index.page.tsx
```diff
@@ -0,0 +1,353 @@
...
+  const icons = detectedLanguages && detectedLanguages.length > 0
+      ? detectedLanguages.map((lang) => langIcons[lang]).filter(Boolean)
+      : tutorial.icon && langIcons[tutorial.icon]
+        ? [langIcons[tutorial.icon]]
+        : [langIcons.xrpl]
+
+  return (
+    <a href={tutorial.path} className="card">
```
mDuo13: Using `<a href=...>` is not ideal. Notably, it switches you back to English if you're viewing a translated page. Instead, use the `Link` component (provided by Redocly) for a more context-aware link.

First, you need to import the component:

```js
import { Link } from '@redocly/theme/components/Link/Link';
```

Then, use the Link class here:

```suggestion
    <Link href={tutorial.path} className="card">
```

Third, you need to change the closing tag (currently on line 290) from `</a>` to match:

```
    </Link>
```

mDuo13 (changes requested):
The landing page looks good in English, but there's one major issue for Japanese that I think is a blocker.

I would also like if we followed up to automate fetching more info for the various sections so we don't have to worry about them falling out of date, especially since as of right now I don't think Redocly link-checks links in `.page.tsx` files.

One other change I recommend is to add an in-page table of contents below the first h1, so that people can see the list of sections ("Get Started with SDKs", "Tokens", etc.) up-front and jump ahead if they're looking for one in particular.

docs/tutorials/index.page.tsx
maria-robobug: Yeah definitely something we should improve upon in the near future, I will add a ticket to track this.

mDuo13 (approved):
Updated page looks great and works well on mobile too. Nicely done!

PR #3516, opened 2026-02-24
Add tab syncing within and across pages

oeggert:
This PR replaces the default Tabs component shipped with Redocly.

New functionality:

- All tabs on a page will now update when any tab is clicked.
- The clicked tab is saved in local storage as a preference; new pages load tabs from stored preference.
- Tab loading priority is: stored preference -> initial tab parameter -> first available tab
- Tab label matching is case insensitive for cases such as "JavaScript" vs "Javascript".
- Tabs only sync coding languages (JS, Py, Go, etc) and connection protocols (Websocket, JSON-RPC, Commandline). All other tab labels are ignored (including typos of tracked labels).


mDuo13 (approved):
It works really well!

PR #3519, opened 2026-02-25
Don't use Batch for SAV tutorials setup

maria-robobug:
Since Batch is being disabled in Devnet, we shouldn't use it for our setup scripts. This uses Tickets instead to retain set up speed.

Also noticed we were missing the amendment status badge in the Batch concept doc so I've added it.

### DIFF
--- a/_code-samples/vaults/js/vaultSetup.js
+++ b/_code-samples/vaults/js/vaultSetup.js
@@ -25,6 +25,28 @@ const [
 process.stdout.write('Setting up tutorial: 1/5\r')
 
 const credType = 'VaultAccess'
+
+// Create tickets for domain owner to submit transactions concurrently
+const ticketCreateResult = await client.submitAndWait(
+  {
+    TransactionType: 'TicketCreate',
+    Account: domainOwner.address,
+    TicketCount: 2
+  },
+  { wallet: domainOwner, autofill: true }
+)
+
+// Get the ticket sequences
+const ticketsResponse = await client.request({
+  command: 'account_objects',
+  account: domainOwner.address,
+  type: 'ticket',
+  ledger_index: 'validated'
+})
+const ticketSequences = ticketsResponse.result.account_objects.map(
+  (obj) => obj.TicketSequence
+)
+
 const [mptCreateResult] = await Promise.all([
   client.submitAndWait(
     {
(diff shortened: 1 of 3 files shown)


_code-samples/vaults/js/vaultSetup.js
```diff
@@ -120,31 +136,46 @@ const domainId = domainOwnerObjects.result.account_objects.find(
...
-const [, vaultCreateResult] = await Promise.all([
+// Create tickets for depositor to submit transactions concurrently
+const depositorTicketCreateResult = await client.submitAndWait(
+  {
+    TransactionType: 'TicketCreate',
+    Account: depositor.address,
+    TicketCount: 2
+  },
+  { wallet: depositor, autofill: true }
+)
```
oeggert: This can be sent concurrently with the first `TicketCreate` transaction. It'd be good to update the variable names of the first `TicketCreate` transaction to also be less generic.

_code-samples/vaults/js/vaultSetup.js
```diff
@@ -25,6 +25,28 @@ const [
...
+// Get the ticket sequences
+const ticketsResponse = await client.request({
+  command: 'account_objects',
+  account: domainOwner.address,
+  type: 'ticket',
+  ledger_index: 'validated'
+})
+const ticketSequences = ticketsResponse.result.account_objects.map(
+  (obj) => obj.TicketSequence
+)
```
oeggert:
```suggestion
const ticketCreateResult = await client.submitAndWait(
  {
    TransactionType: 'TicketCreate',
    Account: domainOwner.address,
    TicketCount: 2
  },
  { wallet: domainOwner, autofill: true }
)

const ticketSequences = ticketCreateResult.result.meta.AffectedNodes
  .filter(node => node.CreatedNode?.LedgerEntryType === 'Ticket')
  .map(node => node.CreatedNode.NewFields.TicketSequence)
```
`ticketCreateResult` is declared here but never used. You could extract the tickets from it though, which would eliminate the need for the subsequent `account_objects` command.

oeggert:
```suggestion
const ticketSequences = ticketCreateResult.result.meta.AffectedNodes
  .filter(node => node.CreatedNode?.LedgerEntryType === 'Ticket')
  .map(node => node.CreatedNode.NewFields.TicketSequence)
```
`ticketCreateResult` is declared above but never used. You could extract the tickets from it though, which would eliminate the need for the subsequent `account_objects` command.

_code-samples/vaults/js/vaultSetup.js
```diff
@@ -120,31 +136,46 @@ const domainId = domainOwnerObjects.result.account_objects.find(
...
+// Get the ticket sequences for depositor
+const depositorTicketsResponse = await client.request({
+  command: 'account_objects',
+  account: depositor.address,
+  type: 'ticket',
+  ledger_index: 'validated'
+})
+const depositorTicketSequences = depositorTicketsResponse.result.account_objects.map(
+  (obj) => obj.TicketSequence
+)
```
oeggert: We can eliminate this API call if you extract the tickets from the unused `depositorTicketCreateResult`. See my suggestion with the first `TicketCreate` transaction.

_code-samples/vaults/js/vaultSetup.js
```diff
@@ -25,6 +25,28 @@ const [
...
+  command: 'account_objects',
+  account: domainOwner.address,
+  type: 'ticket',
+  ledger_index: 'validated'
+})
+const ticketSequences = ticketsResponse.result.account_objects.map(
+  (obj) => obj.TicketSequence
+)
+
 const [mptCreateResult] = await Promise.all([
```
oeggert:
```suggestion
const [mptCreateResult, domainSetResult] = await Promise.all([
```
If we capture the result of the `PermissionedDomainSet` transaction here, we can eliminate the API call at line 127 and just extract the domain ID from the result.

Eliminating the call saves a negligible amount of time, but it's probably best practice to reduce unnecessary calls. If a person (or LLM) looks at this script we want to teach them the optimal way of doing things.

_code-samples/vaults/py/vault_setup.py
oeggert, line 1: Since this is a 1:1 mirror of the JS setup script, the same suggestions apply.

1. Unused variables declared.
2. Extracting info from transaction results (tickets and domain ID) instead of making separate API calls.
3. Submitting both `TicketCreate` transactions concurrently.

oeggert (changes requested):
Setup scripts work, just some minor changes requested to clean up the code and reduce API calls.

oeggert (approved):
LGTM

PR #3536, opened 2026-03-04
Update llmstxt parameters

oeggert:
This PR cleans up the llmstxt object parameters to generate a cleaner llms.txt file.

- It mostly follows the `sidebars.yaml` hierarchy with a few modifications where it makes sense.
- `index.md` files are only kept if they have substantial content in them. If not and they only contain the `{ child-pages }` component (which currently doesn't render anything in generated .md files, making these pages pointless), they are excluded from the txt file.
- Redocly doesn't generate an additional `llms-full.txt` file, or `llms.txt` files for foreign languages. If we add a script to generate these later, this file can also be updated to point to them.

### DIFF
--- a/redocly.yaml
+++ b/redocly.yaml
@@ -43,6 +43,81 @@ metadataGlobs:
     redocly_category: Community
 seo:
   siteUrl: https://xrpl.org/
+  llmstxt:
+    hide: false
+    title: XRPL Developer Portal & Documentation
+    description: >-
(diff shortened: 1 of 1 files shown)


mDuo13 (approved):
Huge improvement to the file. :+1: 

It's still 116k characters, which is a little bigger than the 100k recommendation to fit in context windows comfortably, but that's still much better than the 384k or whatever it was before.

Issue #2333, opened 2024-01-10
Community page: conversation topic text should be link


mDuo13:
The "Hot topics happening now" list describes various topics in "active" discussion on various platforms, with links out to those platforms.

![2024-01-10_102258_739359614](https://github.com/XRPLF/xrpl-dev-portal/assets/7515597/621cb4c1-53e3-406b-89a7-270b1c3cda26)

However, only the small purple arrow on the right edge is clickable. It seems like it would be much better UX if the topic text itself, and maybe the platform icon on the left too, was a link to the same thing.

rachelflynn:
Addressed this issue in PR [3547](https://github.com/XRPLF/xrpl-dev-portal/pull/3547).

PR #3547, opened 2026-03-23
Fix: Make community icons and text clickable links

rachelflynn:
Makes the icons and text in the "Join the Conversation" section clickable, not just the purple arrow. Addresses the UX feedback in good first issue [#2333 ](https://github.com/XRPLF/xrpl-dev-portal/issues/2333). 

In testing, all 8 new links successfully opened in new tabs. 

Screenshot 2026-03-23 at 2 42 35 PM

### DIFF
--- a/community/index.page.tsx
+++ b/community/index.page.tsx
@@ -730,12 +730,22 @@ const CommunityPage: React.FC = () => {
           <tbody>
             <tr>
               <td className="td-img">
+              <a 
+                href="https://discord.gg/xrpl" 
+                target="_blank"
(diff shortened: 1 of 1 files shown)


mDuo13 (approved):
:+1: Tested locally and works as expected.

If it weren't for our impending redesign, I would suggest refactoring the template some more to use a component that's shared with other similar elements (we have several kind of similar lists of links on, for example, the Documentation landing page) but since we have a bigger redesign already in progress this is a perfectly acceptable fix.

PR #3552, opened 2026-03-24
Revert change so Redocly doesn't ignore CODE-OF-CONDUCT.md

amarantha-k:


mDuo13 (approved):
With the broken link removed this is :+1: from me.

The link only worked in the GitHub view of the file, so I figured it was best to remove it; readers on the live site can use the language toggle as before, but readers on GitHub will have to figure out on their own (or from reading the contributor guide) that the localized versions of the code of conduct are in the `@l10n` folder.

Issue #1709, opened 2023-01-25
Short tutorial for looking up / calculating reserves


mDuo13:
Looking up / calculating an account's reserve is a basic task many apps need to do, to show the user how much XRP they need to set aside and why. You may also want to look up the incremental reserve so that you know how much to pay for an AccountDelete or AMMCreate transaction.

A quick tutorial with sample code in Python, JavaScript, and Java would help devs get started on this. The Python Desktop Wallet tutorial [has an example already](https://xrpl.org/build-a-desktop-wallet-in-python.html#codeblock-17), but it's deep in the context of a larger task, so it's not as easy to follow as a separate tutorial would be.

The basics are already mentioned in the [Reserves](https://xrpl.org/reserves.html) page, but we'd like to split off the task information from the concept article so that readers who don't need to know the exact code details can skip past it.

rachelflynn:
Added tutorial in PR [#3573 ](https://github.com/XRPLF/xrpl-dev-portal/pull/3573). Closing this issue.

PR #3573, opened 2026-04-01
Add calculate reserves tutorial

rachelflynn:
Adding a brief tutorial about looking up/calculating reserves as part of good first issue [#1709](https://github.com/XRPLF/xrpl-dev-portal/issues/1709). Used the new tutorial format outlined [here](https://xrpl.org/docs/tutorials/defi/lending/use-the-lending-protocol/create-a-loan-broker).


package.json
```diff
@@ -30,7 +30,7 @@
     "react18-json-view": "^0.2.6",
     "smol-toml": "^1.3.1",
     "use-query-params": "^2.2.1",
-    "xrpl": "^4.2.5"
+    "xrpl": "^4.6.0"
```
mDuo13: This change doesn't relate to the code samples (it's for the top level, the site project itself), so let's undo both this and the corresponding `package-lock.json` change.

docs/tutorials/best-practices/account-management/calculate-reserves.md
```diff
@@ -0,0 +1,133 @@
...
+{% code-snippet file="/_code-samples/calculate-reserves/py/calculate_reserves.py" language="py" from="# Look up reserve values" to="# Look up owner count"/%}
+{% /tab %}
+{% tab label="Go" %}
+{% code-snippet file="/_code-samples/calculate-reserves/go/calculate_reserves.go" language="go" from="// Look up reserve values" to="// Look up owner count"/%}
+{% /tab %}
+{% /tabs %}
+
+### 4. Determine owner reserve
+
+Once you have the incremental reserve value, multiply that by the number of objects the account owns to determine the owner reserve. You can call [`account_info`](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/account-methods/account_info) and take `account_data.OwnerCount` to retrieve an account's total number of objects.
```
mDuo13: The heading and the description describe how to calculate the owner reserve but the code actually stops short of doing the calculation here. Perhaps you should have the code do the multiplication and output the subtotal in this step; alternatively, you could rephrase this step to "Determine owner count" and remove mention of multiplying from this step.

docs/tutorials/best-practices/account-management/calculate-reserves.md
```diff
@@ -0,0 +1,133 @@
...
+{% tab label="Go" %}
+{% code-snippet file="/_code-samples/calculate-reserves/go/calculate_reserves.go" language="go" from="// Look up owner count" to="// Calculate total reserve" /%}
+{% /tab %}
+{% /tabs %}
+
+### 5. Calculate total reserve
+
+Use the following formula to calculate an account's total reserve requirement: 
+
+`total reserve = reserve_base_xrp + (OwnerCount × reserve_inc_xrp)`
```
mDuo13: This pseudo-code mixes spaces and underscores, CamelCase and snake_case, and uses the × symbol (not typically valid in code) despite being in `code font`. Since it's immediately followed by true code samples that have the exact same formula in more consistent syntax, perhaps it's best to describe the formula in prose instead. For example:

```suggestion
To calculate an account's total reserve requirement, calculate the owner reserve by multiplying the incremental reserve value by the owner count, then add the base reserve.
```

_code-samples/calculate-reserves/go/calculate_reserves.go
```diff
@@ -0,0 +1,56 @@
...
+
+	ownerCount := accountInfo.AccountData.OwnerCount
+
+// Calculate total reserve ----------------------
+
+	totalReserve := float64(baseReserve) + (float64(ownerCount) * float64(reserveInc))
+
+	fmt.Printf("Owner count: %v\n", ownerCount)
+	fmt.Printf("Total reserve: %v XRP\n", totalReserve)
+}
```
mDuo13: Nit: Missing newline at end of file

_code-samples/calculate-reserves/js/calculate_reserves.js
```diff
@@ -0,0 +1,33 @@
...
+const ownerCount = accountInfo.result.account_data.OwnerCount
+
+// Calculate total reserve ----------------------
+
+const totalReserve = baseReserve + (ownerCount * reserveInc)
+
+console.log(`Owner count: ${ownerCount}`)
+console.log(`Total reserve: ${totalReserve} XRP`)
+
+await client.disconnect()
```
mDuo13: Nit: missing newline at end of file

_code-samples/calculate-reserves/py/calculate_reserves.py
```diff
@@ -0,0 +1,34 @@
...
+address = "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh"  # replace with any address
+response = client.request(AccountInfo(account=address))
+owner_count = response.result["account_data"]["OwnerCount"]
+
+# Calculate total reserve ----------------------
+
+total_reserve = base_reserve + (owner_count * reserve_inc)
+
+print(f"Owner count: {owner_count}")
+print(f"Total reserve: {total_reserve} XRP")
```
mDuo13: Nit: missing newline at end of file

_code-samples/calculate-reserves/py/calculate_reserves.py
```diff
@@ -0,0 +1,34 @@
+
+# Set up client ----------------------
+
+from xrpl.clients import JsonRpcClient
+
+client = JsonRpcClient("https://s.devnet.rippletest.net:51234")
```
mDuo13: Rather than using Devnet, which is somewhat regularly reset so data is unreliable, a read-only tutorial like this is better off using Mainnet and a sample address (such as rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn) that has more interesting values.

Here's the line but connecting to Mainnet:

```suggestion
client = JsonRpcClient("https://xrplcluster.com")
```

_code-samples/calculate-reserves/js/calculate_reserves.js
```diff
@@ -0,0 +1,33 @@
...
+
+// Look up owner count ----------------------
+
+const address = "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh" // replace with any address
+const accountInfo = await client.request({ command: "account_info", account: address })
+const ownerCount = accountInfo.result.account_data.OwnerCount
+
+// Calculate total reserve ----------------------
+
+const totalReserve = baseReserve + (ownerCount * reserveInc)
```
mDuo13: Unfun fact: doing XRP math like this using native JavaScript numbers tends to produce rounding issues due to floating point inaccuracy. For example, just using the Node.js interpreter:

```node
> 1+(19*0.2)
4.800000000000001
```

The same thing happens when I modify the code sample to use my sample address on Mainnet.

The truly best practice is _probably_ to use a big-number library (like Decimal.js) rather than native JS here.

Another option would be to do the math using integer drops of XRP, which would avoid the problem in this case, but risk having the same problem if the XRP amounts get too large. (JavaScript's native number type is a double-precision float, which can only represent 32 bits' worth of integers at full precision; it _can_ represent larger numbers but precision losses start to occur, and XRP caps at 10^17 drops which requires at least 57 bits to represent fully.)

It's _maybe_ OK to do the math using native JS and just leave a warning (both in the tutorial text _and_ as a code comment in case people only see the code) that the results may be imprecise.

_code-samples/calculate-reserves/py/calculate_reserves.py
```diff
@@ -0,0 +1,34 @@
...
+
+from xrpl.models.requests import AccountInfo
+
+address = "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh"  # replace with any address
+response = client.request(AccountInfo(account=address))
+owner_count = response.result["account_data"]["OwnerCount"]
+
+# Calculate total reserve ----------------------
+
+total_reserve = base_reserve + (owner_count * reserve_inc)
```
mDuo13, line 28: Python (well, the typical C Python implementation) has the same issue as JS here when you use the decimal drops, since it also defaults to double-precision floats to represent `0.2`.

Conveniently if you use the integer drops from `server_state`, Python _won't_ have the same issue since its integers are arbitrary precision by default. (Aside: I still remember back in Python 2.x when that wasn't the case.)

If you want help converting the samples to use `server_state`, don't hesitate to ask, since I tested converting them over already. You'll need to change line 13 to use `["state"]` instead of `["info"]` and lines 15-16 to drop the `_xrp` suffix, and then you have the option of either importing `drops_to_xrp` from `xrpl.utils` and converting back to XRP for display, or simply changing the outputs to say `{total_reserve} drops of XRP` for example. (If you convert using the helper function, you need to do `drops_to_xrp(str(total_reserve))` because it assumes drops of XRP should be formatted as a string. Annoyingly, this is one of the only cases in the API where drops of XRP _aren't_ formatted as a string...)

_code-samples/calculate-reserves/py/calculate_reserves.py
```diff
@@ -0,0 +1,34 @@
+
+# Set up client ----------------------
+
+from xrpl.clients import JsonRpcClient
+
+client = JsonRpcClient("https://s.devnet.rippletest.net:51234")
+
+# Look up reserve values ----------------------
+
+from xrpl.models.requests import ServerInfo
```
mDuo13: Nit: it's considered [good style](https://peps.python.org/pep-0008/#imports) to put all the imports at the start of the file.

_code-samples/calculate-reserves/py/calculate_reserves.py
```diff
@@ -0,0 +1,34 @@
+
```
mDuo13, line 4: Nit: It's not good style to start with a blank line. (This can be fixed automatically with a formatter.)

_code-samples/calculate-reserves/js/calculate_reserves.js
```diff
@@ -0,0 +1,33 @@
...
+
+const baseReserve = validatedLedger.reserve_base_xrp
+const reserveInc = validatedLedger.reserve_inc_xrp
+
+console.log(`Base reserve: ${baseReserve} XRP`)
+console.log(`Incremental reserve: ${reserveInc} XRP`)
+
+// Look up owner count ----------------------
+
+const address = "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh" // replace with any address
```
mDuo13: Nit: most of the code sample uses single quotes (for regular strings) or backticks (for format strings), but this and the following line use double quotes for a regular string. The JS Standard recommendation is single quotes unless the string contains a single quote. (This can be automatically fixed with a formatter.)

docs/tutorials/best-practices/account-management/calculate-reserves.md
```diff
@@ -0,0 +1,133 @@
...
+{% /tab %}
+{% tab label="Go" %}
+{% code-snippet file="/_code-samples/calculate-reserves/go/calculate_reserves.go" language="go" from="// Calculate total reserve" /%}
+{% /tab %}
+{% /tabs %}
+
+## See Also
+
+Concepts: 
+- [Reserves](https://xrpl.org/docs/concepts/accounts/reserves)
```
mDuo13: Don't use fully qualified URLs (starting with `https://xrpl.org/`) when linking between different pages of XRPL.org; this can cause translations to send people back to English, and it may also cause the links to be misclassified as "external" links. 

Instead, either use an absolute path (starting with `/docs` in this case) or a relative path to the `.md` source file, such as the following (I did not test that I used the correct number of `../`'s, so double-check that if you use this):

```suggestion
- [Reserves](../../../concepts/accounts/reserves.md)
```

For certain things that get linked a lot, like API methods, we have a "Common Links" file you can include so you can link them using Markdown's reference-style links. For more info, see [Contribute Documentation: Links](https://xrpl.org/resources/contribute-documentation#links).

mDuo13 (changes requested):
This is off to a good start, but I noticed a few issues.

I confirmed that both the Python and JS code work (minor caveats apply, see below). However, the code samples are missing the recommended `README.md` files (one at the `calculate-reserves/` level and another one in each language folder) plus the `package.json` (JS) and `requirements.txt` (Py) files specifying the dependencies. Due to the missing files, the Code Samples page doesn't properly show this sample. See [Tutorial Guidelines: Folder Layout](https://xrpl.org/resources/contribute-documentation/tutorial-guidelines#folder-layout).

I also recommend running the code samples through the formatter / style enforcers noted in the style guide, which should fix the inconsistencies in white space and quotation marks in the samples as written. ("Black" for Python and "Standard JS" for JavaScript.)

I haven't tested the Go sample code; I don't have a Go dev environment set up (yet).

_code-samples/calculate-reserves/py/calculate_reserves.py
rachelflynn: One thing I noticed: when using drops_to_xrp() for display, Python outputs trailing zeros (e.g. 1.000000 XRP instead of 1 XRP). Not seeing this with JS or Go. I think it's fine as it is just a display difference, but let me know if you think otherwise!

mDuo13: Yes, this is fine. The Python behavior quirk is actually [my fault](https://github.com/XRPLF/xrpl-py/pull/188)—since XRP has fixed precision to 6 decimal places I thought it was a good default to display the full precision, since it's easier to eyeball the difference between, for example, 10 drops and 100 drops when they're represented as `0.000010` and `0.000100` than if you omit the trailing zeroes.

_code-samples/calculate-reserves/README.md
```diff
@@ -0,0 +1,7 @@
+# Calculate Reserves
+
+This code sample demonstrates how to look up and calculate an XRP Ledger account's reserve requirements.
```
mDuo13: For brevity and consistency with other code samples' READMEs:

```suggestion
Look up and calculate an account's reserve requirements.
```

(This first paragraph is the one that gets used as the card text on the [Code Samples page](https://xrpl.org/resources/code-samples).)

_code-samples/calculate-reserves/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+See the [Calculate Account Reserves tutorial](https://xrpl.org/docs/tutorials/best-practices/account-management/calculate-reserves) for a detailed walkthrough.
+
+Samples are provided in JavaScript, Python, and Go.
```
mDuo13: I'd rather not include a line like this, which will need to be updated if we add more code samples, when both places where you are likely to read this will already show you what languages are available in a different way. (On the Code Samples page, there are icons per language, and in the GitHub interface you can see the language folders just above the Readme.)

_code-samples/calculate-reserves/js/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "calculate-reserves",
+  "version": "1.0.0",
+  "license": "MIT",
+  "type": "module",
+  "dependencies": {
+    "xrpl": "^4.4.0"
+  }
+}
```
mDuo13: Nit: missing newline at end of file

_code-samples/calculate-reserves/py/requirements.txt
```diff
@@ -0,0 +1 @@
+xrpl-py>=3.0.0
```
mDuo13: Nit: missing newline at end of file

mDuo13 (changes requested):
This looks almost perfect, but I had a few more nitpicky comments. And this time I found time to test the Go code, which I was able to confirm works as expected too. :tada:

_code-samples/calculate-reserves/py/requirements.txt
rachelflynn: Addressed the newline. 

Also noticed that I had bumped the version to 4.4.0 (per other tutorials), but it didn't get committed. Let me know if you recommend otherwise!

mDuo13 (approved):

PR #3574, opened 2026-04-01
Claude Code Release Notes Skill

oeggert:
This PR adds:

- A new Claude Code Skill: `generate-release-notes`. It leverages a `generate-release-notes.py` script to do the heavy lifting of fetching release diffs and populating relevant context, as well as basic amendment sorting and instructions.
- Project-level rules for Claude Code: Blocks all `git push` attempts by Claude (user must manually do them), and provides lightweight project-level context that is auto-loaded.


tools/generate-release-notes.py
```diff
@@ -0,0 +1,697 @@
...
+import base64
+import json
+import os
+import re
+import subprocess
+import sys
+from datetime import date, datetime
+
+
+# Emails to exclude from credits (Ripple employees not using @ripple.com).
```
amarantha-k, line 26: Checking if the user is a member of the ripple org in GitHub may help future-proof.

amarantha-k (approved):
LGTM.
Added a note to future-proof by checking for org membership instead using a manual listing of addresses to exclude.

tools/generate-release-notes.py
oeggert: Good idea. I created #3615 to follow up on this task.

PR #3575, opened 2026-04-02
PR to fix websocket tool error

amarantha-k:


package-lock.json
```diff
@@ -36,7 +36,7 @@
     },
     "node_modules/@ai-sdk/gateway": {
       "version": "3.0.63",
-      "resolved": "https://registry.npmjs.org/@ai-sdk/gateway/-/gateway-3.0.63.tgz",
+      "resolved": "https://artifactory.ops.ripple.com/artifactory/api/npm/ripple-npm/@ai-sdk/gateway/-/gateway-3.0.63.tgz",
```
mDuo13: Resolving these from Ripple's Artifactory is potentially a problem since last I checked it's not available outside of Ripple's company networks

mDuo13 (changes requested):

package-lock.json
amarantha-k: Agree - we don't want these to be resolving from Ripple's Artifactory! My bad that I missed this change.

amarantha-k:
@mDuo13 This has been fixed and verified the preview URL. Ready for re-review.

mDuo13 (commented):
Can confirm looks good now :+1:

PR #3583, opened 2026-04-07
Fix redundant text in CredentialCreate ref

### LINKED ISSUE #3570: Duplicate text on CredentialCreate page
[This page](https://github.com/XRPLF/xrpl-dev-portal/blob/master/docs/references/protocol/transactions/types/credentialcreate.md) has the following text twice: `In addition to the [common fields](https://xrpl.org/docs/references/protocol/transactions/common-fields), CredentialCreate transactions use the following fields:`

mDuo13:
Fix #3570.

The deleted line is part of the partial already.


oeggert (approved):
LGTM

PR #3591, opened 2026-04-09
Update Transaction Set Handling blog post with additional credits

bthomee:
The reporter asked me to give credits to additional members of their team, and to add a hyperlink to their company website, which is a fair request.


oeggert (approved):
LGTM

PR #3594, opened 2026-04-09
Fix to TileLogo and fix to svg for external link

Calvinjwala:
This PR corrects dark-mode styling for the green TileLogo variant so the default tile uses green-300, the hover/focus overlay uses green-200, and the pressed state stays green-400, including the instant overlay path on small viewports. It also updates TileLogo.md and the TileLogo component JSDoc with light/dark neutral and green token tables, mobile behavior, and a pointer to TileLogo.scss so future changes stay aligned with design.


gabriel-ortiz (approved):
LGTM

PR #3597, opened 2026-04-10
Fix Gem Wallet url

gememerald8:
Gem Wallet is an open-source, self-custody mobile wallet for iOS and Android that supports XRP and XRPL tokens. https://gemwallet.com/xrp-wallet/

The official website of the wallet is Gemwallet.com


maria-robobug (approved):
LGTM, thanks!

PR #3604, opened 2026-04-13
Update FeatureTwoColumn component and related styles

gabriel-ortiz:
- Enhanced FeatureTwoColumn component with improved layout and responsive behavior.
- Updated SCSS for consistent spacing and layout adjustments across breakpoints.
- Revised documentation in FeatureTwoColumn.md for clarity on layout behavior and props.

image


Calvinjwala (changes requested):
Code looks good, I am noticing on medium, there is no spacing between the text and links (when 3+) for medium size. I'd confirm with Design what the right spacing should be. Image

Calvinjwala (changes requested):
Change looks good. Did one final pass and found a weird break point (998px) where if 5 links, it looks like this:

Image

Calvinjwala (approved):
lgtm!

PR #3609, opened 2026-04-15
Fix: Link positioning after Redocly upgrade

rachelflynn:
From Redocly: "I also checked out some older commits and it seems that the behavior changed during upgrade from 0.122.3 to 0.126.0 [here](https://github.com/XRPLF/xrpl-dev-portal/commit/9e40756dd1c447891b95f189fbf37316606fc0b2). There was an update of the Markdoc dependency between those versions, so this might be it."

**Before**
image

**After**
image


mDuo13:
The "before" behavior here was intentional. It's supposed to be floated to the right (with a margin preventing text from overlapping it too much), not on its own line.

rachelflynn:
> Member

I tested this on a local dev server from an older commit (Internet Archive would not load) and saw [Source] rendering on its own separate line, so this PR aimed to restore that. However, after looking across multiple historic versions, I see that the float-right-into-description behavior has been present for a long time and is not a clean regression. The fix proposed in this PR is no longer relevant. 

That said, I believe a better UX is separating the [Source] link from the introductory sentences to make the prose easier to read and less visually disruptive. Floating [Source] to the right interrupts the flow of the text, particularly for longer introductions. 

This example shows the [Source] link cutting the introduction mid-sentence. This occurs in a majority of the transaction type docs: 

Screenshot 2026-04-16 at 10 37 37 AM

Docs like [SciPy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.norm.html) and [AWS boto3](https://docs.aws.amazon.com/boto3/latest/reference/core/session.html) float right as well. This works for them because the source link is anchored to a fixed-width code signature, not variable length prose, so it can't flow into the middle of a sentence. 

[Solana](https://solana.com/docs/rpc/http/sendtransaction) links source code in a callout below the title, and [Python](https://docs.python.org/3/library/multiprocessing.html) links it directly below the title. 

There also appear to be wrapping issues which may be impacting the current [Source] link wrapping as well, exemplified by the longest title from the set below: 

Screenshot 2026-04-16 at 10 40 50 AM

styles/_content.scss
```diff
@@ -87,6 +87,15 @@
     padding-left: 20px;
   }
 
+  // [[Source]] renders as its own <p> after the heading (Redocly 0.131.0+).
+  // overflow: hidden creates a block formatting context so the <p> contains
+  // its floated child instead of collapsing to zero height and bleeding into
+  // the next paragraph.
+  p:has(> a[title="Source"]),
+  p:has(> a[title="ソース"]) {
+    overflow: hidden;
```
mDuo13, line 96: This is more roundabout than necessary for removing the `float: right` effect of the source links. The more elegant way would be to modify the existing rule (on lines 83-87 of this same file) to remove the `float` and `padding`, for example, replacing them with this:

```css
a[title="Source"],
a[title="ソース"] {
  display: block;
  text-align: right;
}
```

(We might also want to add a negative bottom margin, to compensate for the p element's default margin.)

But overall, putting the source links on their own line is not my preferred solution.

mDuo13 (commented):
I agree that the current UX is not ideal. The way I see it, the problem isn't exactly that it's floated, but rather that the "[Source]" text is too similar to the paragraph text that wraps around it, so it becomes visually confusing, making it unclear if the source link is supposed to be part of the paragraph text.

Rather than having the source links display on a separate line, pushing down the content, I would rather add new styles to better distinguish them visually from the content they're floated in. For example, adding a visible border and making the font smaller still allows the text to flow around the floated source link without the confusion:

Screenshot of source link with a border floating to the right of the text
This is just a quick mockup, but in case you're curious here are the styles, replacing the "Source links float right" styles on lines 83-88 of `_content.scss` :

```scss
a[title="Source"],
a[title="ソース"] {
	float: right;
	padding: 5px 10px;
	border: 1px solid #666;
	border-radius: var(--border-radius-md);
	margin: 0 0 12px 16px;
	font-size: 12px;
}
```

Contrast with putting the source link on its own line:

Screenshot of the source link on a separate line

Ultimately this is just my opinion but I prefer having the source link floated.

mDuo13:
P.S. that's a good catch with the wrapping issue of long single-word page names but I don't think it impacts this issue (other than for us to be aware that the space above where the source link is currently won't necessarily be available)

mDuo13:
P.P.S. Thanks for checking what other projects do; in the case of SciPy, I discovered that their source links _can_ actually end up inline with their paragraph texts, but it's not a problem because the font and color of the source link is quite distinct from the paragraph text:

mDuo13:
I think #3715 resolved this same issue, so maybe we can close this?

Issue #3339, opened 2025-10-07
New label-based navigation / groupings


mDuo13:
Create & use a component for landing pages to display pages grouped by particular labels, so for example the Transactions reference can show related transactions grouped together in a box or something.

For v1, it can be used just for children of the current page, but we would eventually like to be able to make label landing pages that show every page on the site with a specified label.

It would also be good to be able to iterate over all labels that exist in the child pages so that the entire landing page can be populated automatically.

Issue #3035, opened 2025-03-18
Revise reference landing page for Transactions Types to group transactions by feature.


amarantha-k:
Issue to track feedback that it's not easy to find all transactions related to a feature and the UX on the reference page could be improved.

mDuo13:
Related: #3339.

mDuo13:
Fixed in #3622.

PR #3623, opened 2026-04-23
Revised transaction types landing page

rachelflynn:
For good issues [#3035](https://github.com/XRPLF/xrpl-dev-portal/issues/3035) and [#3339](https://github.com/XRPLF/xrpl-dev-portal/issues/3339): 

- Grouped transaction types by feature area
- Tightened descriptions and intro text
- Added tables for improved readability

Transaction types are broadly grouped to match the top-level nav categories. The _Tokens_ section could be broken down further by sub-feature (e.g., AMM, Lending, NFTs) at the cost of roughly doubling the number of sections.

### DIFF
--- a/docs/references/protocol/transactions/types/index.md
+++ b/docs/references/protocol/transactions/types/index.md
@@ -8,13 +8,108 @@ labels:
 ---
 # Transaction Types
 
-The type of a transaction (`TransactionType` field) is the most fundamental information about a transaction. This indicates what type of operation the transaction is supposed to do.
+The `TransactionType` field determines what operation a transaction performs. All transactions share [common fields](../common-fields.md). 
 
-All transactions have certain fields in common:
+Each transaction type includes additional fields specific to its operation. The following tables list all transaction types, grouped by feature area.
 
-* [Common Fields](../common-fields.md)
+## Accounts
 
-Each transaction type has additional fields relevant to the type of action it causes.
+| Transaction | Description |
+|:------------|:------------|
+| [`AccountDelete`](accountdelete.md) | Delete an account and its objects. |
+| [`AccountSet`](accountset.md) | Modify an account's settings. |
+| [`DelegateSet`](delegateset.md) | Grant or revoke another account's permission to transact on your behalf. |
+| [`DepositPreauth`](depositpreauth.md) | Preauthorize an account to send you payments. |
+| [`SetRegularKey`](setregularkey.md) | Assign or remove a regular key pair for signing. |
+| [`SignerListSet`](signerlistset.md) | Create, replace, or remove a multi-signing list. |
+| [`TicketCreate`](ticketcreate.md) | Reserve sequence numbers as tickets. |
 
+## Payments
 
-{% child-pages /%}
+| Transaction | Description |
+|:------------|:------------|
+| [`Payment`](payment.md) | Send funds, convert currencies, or create an account. |
+| [`CheckCancel`](checkcancel.md) | Cancel an unredeemed check. |
+| [`CheckCash`](checkcash.md) | Redeem a check. |
+| [`CheckCreate`](checkcreate.md) | Create a check. |
+| [`EscrowCancel`](escrowcancel.md) | Return expired escrow funds to the sender. |
+| [`EscrowCreate`](escrowcreate.md) | Lock funds in escrow until specific conditions are met. |
+| [`EscrowFinish`](escrowfinish.md) | Deliver escrowed funds to the recipient. |
+| [`PaymentChannelClaim`](paymentchannelclaim.md) | Claim funds from a payment channel. |
+| [`PaymentChannelCreate`](paymentchannelcreate.md) | Open a new payment channel. |
+| [`PaymentChannelFund`](paymentchannelfund.md) | Add more funds to a payment channel. |
+
+## Tokens
+
+| Transaction | Description |
+|:------------|:------------|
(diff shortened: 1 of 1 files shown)


docs/references/protocol/transactions/types/index.md
```diff
@@ -8,13 +8,108 @@ labels:
...
+| [`EscrowFinish`](escrowfinish.md) | Deliver escrowed funds to the recipient. |
+| [`PaymentChannelClaim`](paymentchannelclaim.md) | Claim funds from a payment channel. |
+| [`PaymentChannelCreate`](paymentchannelcreate.md) | Open a new payment channel. |
+| [`PaymentChannelFund`](paymentchannelfund.md) | Add more funds to a payment channel. |
+
+## Tokens
+
+| Transaction | Description |
+|:------------|:------------|
+| [`AMMBid`](ammbid.md) | Bid on an AMM's auction slot for a discounted fee. |
```
mDuo13, line 46: I suggest splitting "Decentralized Exchange (DEX)" transactions off from "Tokens" as a way to split this category into more "right-sized" groups. The DEX category would include AMM, Loan, Offer, and probably Vault transactions; the "Token" category would include TrustSet, Clawback, MPT and NFT transactions.

docs/references/protocol/transactions/types/index.md
```diff
@@ -8,13 +8,108 @@ labels:
...
 
-* [Common Fields](../common-fields.md)
+## Accounts
 
-Each transaction type has additional fields relevant to the type of action it causes.
+| Transaction | Description |
+|:------------|:------------|
+| [`AccountDelete`](accountdelete.md) | Delete an account and its objects. |
+| [`AccountSet`](accountset.md) | Modify an account's settings. |
+| [`DelegateSet`](delegateset.md) | Grant or revoke another account's permission to transact on your behalf. |
```
mDuo13, line 21: One thing the old version does better is to provide forewarning about which features are actually available on the XRP Ledger Mainnet, and which ones are gated by a not-yet-enabled amendment, via the little flask icon:



But nowadays, we can do even better with the `{% amendment-disclaimer ... /%}` component. For cases like this where the transaction types come from a mix of amendments, you could have a "Related amendment" column in the table which shows if/when the amendment became enabled. See [Short Names of Ledger Entries](https://xrpl.org/docs/references/http-websocket-apis/api-conventions/ledger-entry-short-names) as an example. For cases like sidechains where the entire group is gated by the same amendment, you could also try using the non-compact version of the amendment disclaimer in a line above the table.

docs/references/protocol/transactions/types/index.md
```diff
@@ -8,13 +8,108 @@ labels:
...
+| [`NFTokenAcceptOffer`](nftokenacceptoffer.md) | Accept an offer to buy or sell an NFT. |
+| [`NFTokenBurn`](nftokenburn.md) | Permanently destroy an NFT. |
+| [`NFTokenCancelOffer`](nftokencanceloffer.md) | Cancel one or more offers to buy or sell an NFT. |
+| [`NFTokenCreateOffer`](nftokencreateoffer.md) | Create an offer to buy or sell an NFT. |
+| [`NFTokenMint`](nftokenmint.md) | Mint an NFT. |
+| [`NFTokenModify`](nftokenmodify.md) | Modify a dynamic NFT. |
+| [`OfferCancel`](offercancel.md) | Cancel an offer in the decentralized exchange. |
+| [`OfferCreate`](offercreate.md) | Create a limit order in the decentralized exchange. |
+| [`PermissionedDomainDelete`](permissioneddomaindelete.md) | Delete a permissioned domain. |
+| [`PermissionedDomainSet`](permissioneddomainset.md) | Create or update a permissioned domain. |
```
mDuo13, line 76: I'd move these to the Decentralized Storage category. Permissioned domains aren't exclusively for use with tokens, they—like other things in that category—don't have built-in semantic meaning, and they are closely associated with Credentials in usage, so I think it makes sense to group them there.

mDuo13 (approved):
Overall, this is a nice improvement over the existing list, although I have a couple suggestions on how it could be further improved and one concern.

Long term, my main concern is that this replaces an automatically-populated list with a manually-maintained page. We've already seen with the manually managed API Methods landing pages that these tend to get out of sync with the list of pages under them. It would be preferable to populate this more automatically, perhaps using frontmatter data from the individual transaction pages; that's what #3339 is about. But, as noted in that ticket, doing so probably involves writing a Redocly plugin and custom React components; [the new tutorials landing](https://github.com/XRPLF/xrpl-dev-portal/blob/master/docs/tutorials/index.page.tsx) has some components that might be a good starting point for the React components at least. _That_ page also has the problem of the actual data being hardcoded and manually maintained, but if we built a way to detect the data automatically then we could plug the data in without having to redo the React components themselves, which is an approach we could also take on a page like this.

As far as creating a more customized landing page for transactions, we might also want to experiment with grouping Create/Delete/Cancel functions of related transaction types into more of a "logical order" rather than alphabetical, since it's a bit weird to start with how to cancel something before you talk about how to create or use it in the first place. [The Explorer](https://livenet.xrpl.org/) has a legend of shapes and colors it uses to communicate different types of operations on different groupings of features, which we could potentially adapt if we go the route of building fully custom components for this page. But for just rewriting the Markdown version of the page, that's not necessary.

I have some other, more immediate suggestions regarding this page, including an idea of how to break up the overly large Tokens section, as noted on in my other comments.

amarantha-k:
Closing this PR without merge as #3626 builds on this PR and incorporates feedback noted in this PR.

PR #3631, opened 2026-05-05
Update browser wallet tutorial dependencies

mDuo13:
Update the "Build a Browser Wallet in JavaScript" sample app / code to use Vite 8.x (it was previously on Vite 4.x) and make some related optimizations/cleanup.

- With xrpl.js v3.x and up, a custom vite bundler is no longer needed
- Environment variables can be accessed by default since vite includes dotenv now.
    - Added a note about passing sensitive values this way. The old way was perhaps unwise (passing _all_ env vars to the client can become a security hole).
- Switch from Yarn to NPM for consistency with other tutorials.

Reword some parts of the tutorial to better follow style guidelines (like avoiding first person).

I still have some doubts about the overall architecture and usefulness of this sample app but with these changes it is at least less outdated.


oeggert (approved):
LGTM

PR #3637, opened 2026-05-10
Update {repo-link} component and refactor old/broken links

oeggert:
This PR updates the {repo-link} component to now accept 2 new parameters: `repo` and `format`.

- `repo` reads env vars to create links to either devportal or xrpld (Note: I'm using xrpld in preparation for the rippled rename in 3.2; we'll only need to update the env var URLs when the change happens). The new `repo` parameter now fetches the branch from the env file, which enables us to update the env to keep all our links pointed to the latest release branch. New repos can be added as env vars.
- `format` replaces the old CSS hook which looked at the "Source" title in links.  Using `format:"source-link"` now attaches the `source-link` class to the link, which will align-right the source link. This also fixes the current broken right float source links. Additional formats can be added to the CSS style sheet and called via `format`.

Additional changes made:
- Documented component updates in `markdoc-tags.md`.
- Refactored old {repo-link} components to use new params.
- Updated all hard-coded source links to use {repo-link}.
- Fixed numerous broken links in Japanese docs.

@mDuo13 Can you take a look at the `RepoLink.tsx` file? The code comment says the component was meant to point to PR URLs, but was never implemented--I want to make sure that's still the case and my additions don't mess with that.


oeggert:
Closing this PR. New link component merged in #3696 and links fixed by #3658

PR #3652, opened 2026-05-19
New Blackholed Account concepts doc

rachelflynn:
Related to good first issues #https://github.com/XRPLF/xrpl-dev-portal/issues/2088 and #https://github.com/XRPLF/xrpl-dev-portal/issues/1240.

- **Adds a new concept doc to explain blackholed accounts.** Covers blackholes as a concept, rather than a tutorial. Intentionally excludes direct steps for how to blackhole an account.
- **Adds a cross-reference to blackholed accounts in existing, relevant docs.** Other docs that mention blackholing can now point to the concept page.
- **Standardizes usage across the docs to blackhole instead of black hole.** Verb forms ("blackholed," "blackholing") are already compound; keeping the noun one-word matches that pattern and aligns with systems precedent ("blackhole route" in IETF).

### DIFF
--- a/docs/concepts/accounts/blackholed-accounts.md
+++ b/docs/concepts/accounts/blackholed-accounts.md
@@ -1,0 +1,58 @@
+---
+seo:
+    description: A blackholed account is an XRP Ledger account that can never send transactions again because its signing authority has been permanently removed.
+labels:
+  - Accounts
+---
+# Blackholed Accounts
+
+A blackholed account is an account whose signing authority has been permanently and verifiably removed. Once an account is blackholed, no one can ever send transactions from it again.
+
+## Why Blackhole an Account
+
+Blackholing an account is an extreme measure that most account owners will never need to consider. Use cases include:
+- **Proving a fixed token supply.** By blackholing an issuing account, the owner proves to users and token holders that the account can no longer mint additional tokens, change transfer fees, or freeze balances.
+- **Decommissioning an account that can't be deleted.** If an account has [deletion blockers](deleting-accounts.md#deletion-blockers), it cannot be deleted. Blackholing permanently removes the account's signing authority, preventing further activity. Unlike deletion, the account remains on the ledger and its reserves stay locked.
+
+## Requirements
+
+To be considered blackholed, an account must meet all of the following conditions:
+- The master key is disabled.
+- The regular key is set to a known [blackhole address](addresses.md#special-addresses).
+- The account has no [signer list](multi-signing.md).
+
+## Checking Whether an Account Is Blackholed
+
+There is no cryptographic certificate of blackholing. You can only verify that an account meets the conditions above.
+
+To check, call the [account_info method][] and confirm all of the following:
+1. The `lsfDisableMaster` flag is set.
+2. The `RegularKey` field is set to a known [blackhole address](addresses.md#special-addresses) (either `rrrrrrrrrrrrrrrrrrrrrhoLvTp` for ACCOUNT_ZERO or `rrrrrrrrrrrrrrrrrrrrBZbvji` for ACCOUNT_ONE).
+3. No `SignerList` object exists for the account, confirmed by calling the [account_objects method][].
+
+ACCOUNT_ZERO and ACCOUNT_ONE are considered known blackhole addresses because they are derived from the integers 0 and 1 instead of a private key. This makes it effectively impossible for anyone to produce a matching private key.
+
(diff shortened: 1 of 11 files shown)


docs/concepts/accounts/blackholed-accounts.md
```diff
@@ -0,0 +1,58 @@
...
+- **Proving a fixed token supply.** By blackholing an issuing account, the owner proves to users and token holders that the account can no longer mint additional tokens, change transfer fees, or freeze balances.
+- **Decommissioning an account that can't be deleted.** If an account has [deletion blockers](deleting-accounts.md#deletion-blockers), it cannot be deleted. Blackholing permanently removes the account's signing authority, preventing further activity. Unlike deletion, the account remains on the ledger and its reserves stay locked.
+
+## Requirements
+
+To be considered blackholed, an account must meet all of the following conditions:
+- The master key is disabled.
+- The regular key is set to a known [blackhole address](addresses.md#special-addresses).
+- The account has no [signer list](multi-signing.md).
+
```
mDuo13, line 24: To future-proof this section, add mention of Permission Delegation:

```suggestion
To be considered blackholed, an account must meet all of the following conditions:
- The master key is disabled.
- The regular key is set to a known [blackhole address](addresses.md#special-addresses).
- The account has no [signer list](multi-signing.md).
- The account has no [delegates](permission-delegation.md). {% amendment-disclaimer name="PermissionDelegation" /%}

```

docs/concepts/accounts/blackholed-accounts.md
```diff
@@ -0,0 +1,58 @@
...
+A blackholed account's transaction history will continue to show activity from other accounts using its tokens or completing transactions involving it. The blackholed account itself is not sending these transactions.
+{% /admonition %}
+
+## Risks and Limitations
+
+Blackholing is irreversible. Before blackholing an account, be aware of the following:
+- Any XRP held by the account is permanently lost and cannot be retrieved.
+- The account cannot send any transactions after blackholing. Any misconfigured settings (such as transfer fees, freeze authority, or rippling) are permanently locked in.
+- Existing ledger entries created before blackholing (such as open offers, escrows, payment channels, or checks) can still complete, meaning additional tokens could be issued or XRP could leave the account after blackholing.
+- Other blackholing conditions not yet met before the master key is disabled can no longer be met afterward, leaving the account only partially blackholed.
```
mDuo13: I'm not sure I understand the intended meaning of this sentence. Can you clarify, maybe with an example?

docs/concepts/accounts/blackholed-accounts.md
```diff
@@ -0,0 +1,58 @@
...
+- The account has no [signer list](multi-signing.md).
+
+## Checking Whether an Account Is Blackholed
+
+There is no cryptographic certificate of blackholing. You can only verify that an account meets the conditions above.
+
+To check, call the [account_info method][] and confirm all of the following:
+1. The `lsfDisableMaster` flag is set.
+2. The `RegularKey` field is set to a known [blackhole address](addresses.md#special-addresses) (either `rrrrrrrrrrrrrrrrrrrrrhoLvTp` for ACCOUNT_ZERO or `rrrrrrrrrrrrrrrrrrrrBZbvji` for ACCOUNT_ONE).
+3. No `SignerList` object exists for the account, confirmed by calling the [account_objects method][].
```
mDuo13: 1. Future-proofing again, check for the existence of a Delegate entry
2. Our style guide prefers using "entry" over "object" for this case. You can make it a common link too, actually—`[SignerList entry][]`.
    - Similarly, `[Delegate entry][]` _should_ work but actually there's an inconsistency in the common links file and three recent types—Delegate, Loan, and LoanBroker—have their common links formatted like `[Delegate ledger entry][]` instead.

mDuo13 (changes requested):
Largely looks good; left couple minor suggestions.

Can you also update the preferred terms in the Contributing Docs page to add guidance on "blackhole" per the changes made here?

docs/concepts/accounts/blackholed-accounts.md
rachelflynn: Good point. 

I can't actually construct an example that holds up. I was reaching for something like "if the regular key is set to a non-blackhole address before the master key is disabled, the account isn't really blackholed because someone still holds a usable key"...but that's just the inverse of the Requirements section.

Going to remove the bullet.

mDuo13 (approved):
:+1:

PR #3655, opened 2026-05-20
Implement external link checker

mDuo13:
Adds a more powerful link checker Python tool that catches many broken links not regularly checked by Redocly, and reports which ones, if any, are currently broken. Features:

- Two modes, with different speeds & tradeoffs:
    - The default mode runs by parsing Markdown files from the file system and checks external links only. This is fast, but cannot check links in custom Markdoc components, fully custom pages, or components of the template.
    - "Live" mode uses Selenium's Web Driver tool to open a headless Chrome browser and check pages as built by the Redocly Realm server running in local dev mode. This mode is much more powerful since it checks "the actual pages, as seen by users" but it is much slower. (A full run can take well over an hour.)
- Automatically retry with automatic backoff so transient failures have a chance to resolve themselves.
- Sleep for 500ms between calls to the same hostname so that it doesn't get flagged for spamming (for example if there are a lot of GitHub PR links in a row).
- Cache results using a local file (`tools/link-cache.json`) so it doesn't excessively re-check the same URLs. Successful results are cached for a week; unsuccessful results are retried once per run.
    - If you interrupt a run with Ctrl-C, save the cache so that it can make progress when you run it again, rather than starting over.
    - Add this file to the `.gitignore` so it doesn't cause excessive churn in the repo.
- Skip a list of "known broken" links in a text file, so you can exclude false positives. This file supports comments and wildcards at the end of a URL, so for example you can exclude all of 𝕏 with `https://x.com/*`
- (Live mode only) Check for unparsed reference links like `[Hash][]` in the user-visible output. Typically this occurs when you forgot to include the common links file, or you used a common link inside a partial.
- (Live mode only) For links _within_ the site that go to specific anchors, check that the linked anchor exists in the linked page.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,6 @@ _code-samples/*/js/package-lock.json
 _code-samples/*/go/go.sum
 _code-samples/*/java/target/
 _code-samples/*/*/*[Ss]etup.json
-
+tools/link-cache.json
 # PHP
 composer.lock
--- a/redocly.yaml
+++ b/redocly.yaml
@@ -6,6 +6,7 @@ ignore:
   - _code-samples/create-amm/ts/tsconfig.json
   - resources/contribute-blog/_blog-template.md
   - resources/contribute-documentation/_tutorial-template.md
+  - tools/*
 l10n:
   defaultLocale: en-US
   locales:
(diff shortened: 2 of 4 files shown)


mDuo13:
Combined into #3658

PR #3658, opened 2026-05-22
Selenium link checker workflow

oeggert:
Combines #3655 with #3656 to run the more thorough Selenium-powered link checker with GitHub Actions

PR #3662, opened 2026-05-27
Fix broken links identified by Selenium-based checker

mDuo13:
Fix links identified as broken in #3658

The following broken link reports are false positives and should be added to the "known broken links" file in #3658 (not in this PR, which is based directly on master):
- `https://sologenic.org/trade`
- `https://papers.ssrn.com/sol3/papers.cfm?abstract_id=4133897`
- `https://www.usenix.org/conference/usenixsecurity21/presentation/torres`
- `https://s1.ripple.com/` (this is a weird one—you don't actually want to use this link normally, but navigating your browser to this page can sometimes fix an issue with TLS cert caching)
- `https://linux.die.net/man/8/logrotate`
- `https://jsfiddle.net/*`


mDuo13:
Closing this in favor of fixing the links as part of #3658

PR #3670, opened 2026-05-29
Update Redocly to Realm v0.133.1

mDuo13:
- Update Redocly and dependencies. This addresses several dependabot reports and fixes a bug where clicking certain links would navigate you as if you had also clicked the first link on the target page.
- Fix some minor style problems in the new version (color clashes, header alignment).

Note that the new version includes minor changes to the styles for link underlines. Since these changes seemed benign, I mostly left them as-is in the new version. The only change I made was to fix a typo (missing semicolon) that had stopped an existing rule from setting the underline color to match our color scheme.

| Before (Redocly 0.132.1) | After (Redocly 0.133.1) |
|---|---|
|  |  |
| <ul><li>Underline color matches text (whether hovered or not)</li><li>Descenders cut through underline</li></ul> | <ul><li>Underline is purple even before hovering the link</li><li>Underline is placed below descenders</li></ul> |

### DIFF
--- a/@theme/styles.css
+++ b/@theme/styles.css
@@ -18,6 +18,16 @@ ul.nav.navbar-nav {
 	border-radius: var(--border-radius);
 }
 
+#content .heading-anchor h1,
+#content .heading-anchor h2,
+#content .heading-anchor h3,
+#content .heading-anchor h4,
+#content .heading-anchor h5,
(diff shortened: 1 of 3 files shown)


oeggert (approved):
LGTM

PR #3675, opened 2026-06-03
Update reserve values (rippled #6382)

rachelflynn:
Updates docs for the new default base/owner reserve of 1 XRP / 0.2 XRP in rippled 3.2.0 (https://github.com/XRPLF/rippled/pull/6382).
- _Start a New Genesis Ledger in Standalone Mode_: default reserve now **1 XRP** base / **0.2 XRP** owner increment; dropped the stale "higher than production" note (the default now matches mainnet) and pointed to Fee Voting.
- _Transaction Cost_: AccountDelete/AMMCreate special cost updated to **200,000 drops** (0.2 XRP). It equals the owner reserve increment, and this table was missed when the other reserve pages moved to 0.2 XRP.


amarantha-k (approved):
LGTM

PR #3679, opened 2026-06-05
Add tls support to grpc (rippled #6374)

rachelflynn:
Documents the optional TLS/mTLS settings added to the [port_grpc] stanza in [rippled #6374](https://github.com/XRPLF/rippled/pull/6374).


amarantha-k (approved):
LGTM

PR #3677, opened 2026-06-04
Rename rippled to xrpld (rippled #6098 and #6347)

rachelflynn:
Replace any references of `rippled.cfg` with `xrpld.cfg` ([XRPLF/rippled#6098](https://github.com/XRPLF/rippled/pull/6098)) and rename the default config directory from `ripple` to `xrpld` ([XRPLF/rippled#6347](https://github.com/XRPLF/rippled/pull/6347)), both part of 3.2.0.

The same upstream change (#6098) also renamed the example config file, so this PR additionally renames `rippled-example.cfg` → `xrpld-example.cfg` to match. #6347 renames the system name from `ripple` to `xrpld`, which changes the default config directory (e.g. `$HOME/.config/ripple/` → `$HOME/.config/xrpld/`). The `rippled` name will be changed in a separate PR.

## Changes

**English docs (`docs/`)**
- Changed `rippled.cfg` → `xrpld.cfg` and `rippled-example.cfg` → `xrpld-example.cfg`
- Renamed the config directory `ripple` → `xrpld` in XDG/home paths (e.g. `$HOME/.config/xrpld/xrpld.cfg`); left package paths (`/etc/opt/ripple/`, `/var/lib/rippled`) unchanged
- Added a note to the Conf File Location snippet about the renaming and backwards compatibility
- Fixed a pre-existing typo in Commandline Usage

**Translations (`@l10n/ja`, `@l10n/es-ES`)**
- Renamed the same config filename strings in the corresponding Japanese and Spanish docs
- Renamed the config directory `ripple` → `xrpld` in the Japanese docs (the Spanish docs had no such paths)
- Did not update translated prose

PR #3680, opened 2026-06-05
Add configurable NuDB block size option (rippled #5468)

rachelflynn:
Adds the new `nudb_block_size` field as part of 3.2.0 (https://github.com/XRPLF/rippled/pull/5468).

Conflicts with #3677 in _Capacity Planning_ (both edit the NuDB intro). Whichever merges second: resolve by dropping the "NuDB does not have performance-related configuration options…" sentence. This PR removes it because the new nudb_block_size option makes it false. Don't keep the rename branch's version of that line.


oeggert (approved):
LGTM

PR #3682, opened 2026-06-05
[feat][site] adding 3 new events to events page

Calvinjwala:
Adding three new events as requested.


community/events.page.tsx
```diff
@@ -1427,6 +1427,41 @@ const events = [
...
+
+  {
+    name: "XRPL Developer Training",
+    description:
+      "​XRPL Commons is hosting an online training open to everyone, whether you are a developer or not. Over two days, dive into the XRP Ledger, one of the world’s leading blockchains, designed for real-world applications and known for its low carbon footprint and efficiency.",
+    type: "conference",
+    link: "https://luma.com/building-on-the-xrpl",
+    location: "Online",
+    date: "June 22-23, 2026",
+    image: require("../static/img/events/commons-online.jpg"),
```
amarantha-k: The build fails as the images in the PR are in a different format - expecting .jpg but PR has .png files.

community/events.page.tsx
```diff
@@ -1427,6 +1427,41 @@ const events = [
...
+      "​Two days. Three communities. One shared question: what does the next generation of ledger technology look like — and how do we build it responsibly?",
+    type: "conference",
+    link: "https://luma.com/XRPL-Blockchain-Research-Summit",
+    location: "Paris, France",
+    date: "June 16-17, 2026",
+    image: require("../static/img/events/commons-online.jpg"),
+    end_date: "June 17, 2026",
+  },
+  {
+    name: "XRPL Meetup Lisbon",
```
amarantha-k, line 1454: Since this event is already in the past, I'm not sure what value adding it would provide. Could you clarify whether we still want to include it?

amarantha-k (changes requested):
Need to update images for build failure, and a question about Lisbon meetup.

community/events.page.tsx
Calvinjwala: Since we have a list of past events, let's add in case someone wants to revisit the details

amarantha-k (approved):
Looks good now. Thanks!

PR #3685, opened 2026-06-05
Empty permissions list is delete-only; add tecNO_ENTRY (rippled #6542)

rachelflynn:
Documents the validation change from rippled PR #6542 (XRPLF/rippled#6542), which disallows an empty `Permissions` list when no Delegate ledger entry exists.

Updates _Delegate Set_:
- Clarify the `Permissions` field: empty array to revoke all permissions is only valid when a Delegate entry already exists.
- Note in the explanatory paragraph that an empty list with no existing entry fails with `tecNO_ENTRY`.
- Add a `tecNO_ENTRY` row to the Error Cases table.

### DIFF
--- a/docs/references/protocol/transactions/types/delegateset.md
+++ b/docs/references/protocol/transactions/types/delegateset.md
@@ -44,9 +44,9 @@ txIcon: modify
 | Field         | Required? | JSON Type            | Internal Type | Description |
 |:--------------|-----------|----------------------|---------------|-------------|
 | `Authorize`   | Yes       | String - [Address][] | AccountID     | The account being granted permissions, also called the _delegate_. |
-| `Permissions` | Yes       | Array                | Array         | A list of up to 10 [Permission objects](#permission-objects) each specifying a different permission granted to the delegate. The delegate's permissions are updated to match this set of permissions exactly. To revoke all permissions, use an empty array. |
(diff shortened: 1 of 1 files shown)


docs/references/protocol/transactions/types/delegateset.md
```diff
@@ -44,9 +44,9 @@ txIcon: modify
...
 
-If a [Delegate ledger entry][] does not exist to record the granted permissions, this transaction creates one. If it already exists, the transaction updates the set of permissions to match the list in the transaction: any permissions not listed are revoked. If all permissions are revoked, the transaction deletes the Delegate ledger entry.
+If a [Delegate ledger entry][] does not exist to record the granted permissions, this transaction creates one. If it already exists, the transaction updates the set of permissions to match the list in the transaction: any permissions not listed are revoked. If all permissions are revoked, the transaction deletes the Delegate ledger entry. An empty `Permissions` list is only valid for deleting an existing Delegate ledger entry; if no such entry exists, the transaction fails with `tecNO_ENTRY`.
```
amarantha-k, line 49: Not a blocker for this PR, but as a follow-up, adding code samples would definitely be helpful.

amarantha-k (approved):
LGTM

PR #3689, opened 2026-06-08
New Send an MPT Tutorial

maria-robobug:
Adds a new Send MPT tutorial (JS, Python, and Go) and deletes the old modular tutorial.
 
 - [x] Add new Send MPT tutorial.
 - [x] Remove modular tutorial version and update references.
 - [x] Update redirects
 - [x] Update **pinned** Payments tutorials in Tutorials landing page.


_code-samples/send-an-mpt/README.md
```diff
@@ -0,0 +1,13 @@
+# Send an MPT Examples
+
+Sample code for the [Send a Multi-Purpose Token (MPT)](https://xrpl.org/docs/tutorials/payments/send-an-mpt) tutorial. Each tutorial script funds a fresh receiver, authorizes it to hold the MPT, snapshots both accounts' MPT balances, has a preconfigured sender forward 100 MPT units to the receiver, and reads both balances again from the validated ledger to confirm the transfer.
```
mDuo13: This first paragraph is what gets picked up and displayed on the "Code Samples" page. I suggest revising it as follows:

- Move the link and mention of the tutorial to a later paragraph. The descriptions in the Code Samples page can't contain links, since it's on a card that is itself a link to the code sample.
- Rephrase to imperative mood.

```suggestion
Send a Multi-Purpose Token (MPT) payment, including funding a fresh receiver, authorizing it to hold the MPT, and transferring 100 MPT units from a preconfigured sender to the receiver. Reads both balances before and after the transaction to confirm the transfer.
```

_code-samples/send-an-mpt/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+| Language | Folder |
```
mDuo13: I suggest omitting this table. I don't think it adds much, since the GitHub interface shows the folders anyway (and I think the language abbreviations are probably clear/specific enough) and it's extra work to update the table if we add or remove languages for this code sample.

_code-samples/send-an-mpt/go/README.md
```diff
@@ -0,0 +1,61 @@
+# Send an MPT Examples (Go)
```
mDuo13: The word "examples" here is both unnecessary and confusing.

```suggestion
# Send an MPT (Go)
```

_code-samples/send-an-mpt/go/send-mpt-setup/main.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/Peersyst/xrpl-go/xrpl/queries/account"
+	"github.com/Peersyst/xrpl-go/xrpl/queries/common"
+	"github.com/Peersyst/xrpl-go/xrpl/rpc"
+	rpctypes "github.com/Peersyst/xrpl-go/xrpl/rpc/types"
+	"github.com/Peersyst/xrpl-go/xrpl/transaction"
+	"github.com/Peersyst/xrpl-go/xrpl/transaction/types"
+	"github.com/Peersyst/xrpl-go/xrpl/wallet"
+)
+
+const total = 4
```
mDuo13: Not a very descriptive variable name. Perhaps `totalSteps` would be better?

_code-samples/send-an-mpt/py/send_mpt_setup.py
```diff
@@ -0,0 +1,102 @@
+import asyncio
```
mDuo13, line 1: Why is this Python code async? Our [tutorial guidelines](https://xrpl.org/resources/contribute-documentation/tutorial-guidelines#language-specific-guidelines) say Python code samples should use the `JsonRpcClient` unless asynchronous functionality is needed.

_code-samples/send-an-mpt/py/send_mpt.py
```diff
@@ -0,0 +1,102 @@
...
+
+# Set up client ----------------------
+client = JsonRpcClient("https://s.altnet.rippletest.net:51234")
+
+# Load setup data ----------------------
+# This step checks for the necessary setup data to run the tutorial.
+# If missing, send_mpt_setup.py will generate it.
+if not os.path.exists("send_mpt_setup.json"):
+    print("\n=== Setup data doesn't exist. Running setup script... ===\n")
+    subprocess.run([sys.executable, "send_mpt_setup.py"], check=True)
```
mDuo13: Using `subprocess` to run another Python script for setup steps is not very Pythonic. If anything goes wrong, it makes it harder to debug, and it's more fragile if you adapt the code in general. Instead, import the setup function and call it natively with Python.

docs/tutorials/payments/send-an-mpt.md
```diff
@@ -0,0 +1,231 @@
...
+
+The MPT issuance in this example uses an [asset scale](../../references/protocol/data-types/currency-formats.md#mpt-precision) of `2`, so applications shift the displayed amount two decimal places. A _value_ of `"100"` therefore shows as `1.00` units of the token.
+
+The example MPT has a `TransferFee` of `0`, which means the sender's debit matches the payment value exactly. With a non-zero `TransferFee`, the sender would have to pay extra so the receiver gets the full value.
+
+{% admonition type="warning" name="Caution" %}
+If the payment fails, it could be for one of the following reasons:
+
+- `tecNO_AUTH`: the issuance does not have **Can Transfer** enabled, or (with allow-listing) the issuer hasn't approved the holder.
+- `tecINSUFFICIENT_FUNDS`: the sender's MPT balance is less than the value being sent.
```
mDuo13: Given the discussion of transfer fees above, it might be worth acknowledging how the sender might also need to send a higher amount if there were a transfer fee.

_code-samples/send-an-mpt/js/sendMPT.js
```diff
@@ -0,0 +1,118 @@
...
+// Connect to the network ----------------------
+const client = new xrpl.Client('wss://s.altnet.rippletest.net:51233')
+await client.connect()
+
+// Load setup data ----------------------
+// This step checks for the necessary setup data to run the tutorial.
+// If missing, sendMPTSetup.js will generate it.
+if (!fs.existsSync('sendMPTSetup.json')) {
+  console.log(`\n=== Setup data doesn't exist. Running setup script... ===\n`)
+  execSync('node sendMPTSetup.js', { stdio: 'inherit' })
```
mDuo13: Similar to the Python script, I think a more idiomatic way to do this would be for the setup script to export a setup function and this script to import and call it using native JS rather than using `execSync`.

Among other reasons to do it that way, on some platforms the Node.js executable is not named `node` so this line is more fragile than it needs to be.

_code-samples/send-an-mpt/js/sendMPT.js
```diff
@@ -0,0 +1,118 @@
...
+    command: 'account_objects',
+    account: address,
+    ledger_index: 'validated',
+    type: 'mptoken'
+  })
+  const node = response.result.account_objects.find(
+    o => o.MPTokenIssuanceID === mptIssuanceID
+  )
+  return node?.MPTAmount ?? '0'
+}
```
mDuo13, line 82: **Important:** This function does not properly handle pagination. If the account owns more than one page's worth of ledger entries (even if most of them aren't MPTokens—or if most of them _are_ different MPTokens) this function may incorrectly return a balance of 0 instead of finding the matching MPToken entry on a later page of results.

You won't see this bug in the test scripts because they use fresh accounts that don't own a lot of ledger entries, so the MPToken you're looking for will always be on the first page of results, but that assumption doesn't hold in real-world usage.

A more direct and reliable way is to [look up an MPToken entry using the ledger_entry method](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/ledger-methods/ledger_entry#get-mptoken-entry) with the issuance ID and address. This should always find the MPToken entry you want with no pagination required. I believe it is also more efficient on the server since it doesn't need to fetch the owner directory at all.

```suggestion
async function getMPTBalance(address, mptIssuanceID) {
  try {
    const response = await client.request({
      command: 'ledger_entry',
      mptoken: {
        account: address,
        mpt_issuance_id: mptIssuanceID
      },
      ledger_index: 'validated',
    })
    return response.result.node?.MPTAmount ?? '0'
  } catch (e) {
    if (e.data?.error === 'entryNotFound') {
      return '0'
    } else {
      throw e
    }
  }
}
```

_code-samples/send-an-mpt/py/send_mpt.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    response = client.request(AccountObjects(
+        account=address,
+        ledger_index="validated",
+        type="mptoken",
+    ))
+    for node in response.result["account_objects"]:
+        if node["MPTokenIssuanceID"] == issuance_id:
+            return node.get("MPTAmount", "0")
+    return "0"
```
mDuo13: The same pagination problem with `sendMPT.js` applies here. Refactoring to use `LedgerEntry(...)` in the same way should fix it.

_code-samples/send-an-mpt/go/send-mpt/main.go
```diff
@@ -0,0 +1,193 @@
...
+				continue
+			}
+			if id, ok := entry["MPTokenIssuanceID"].(string); ok && id == mptIssuanceID {
+				if amt, ok := entry["MPTAmount"].(string); ok {
+					return amt, nil
+				}
+			}
+		}
+		return "0", nil
+	}
```
mDuo13, line 149: I believe this also suffers the same pagination problem as the JS and Python versions, with the same fix possible.

docs/tutorials/tokens/mpts/issue-a-multi-purpose-token.md
```diff
@@ -220,13 +220,12 @@ The decoding utility function converts the metadata back to a JSON object and ex
...
 - **Concepts**:
 	- [Multi-Purpose Tokens (MPT)](../../../concepts/tokens/fungible-tokens/multi-purpose-tokens.md)
+- **Tutorials**:
+	- [Send a Multi-Purpose Token (MPT)](../../payments/send-an-mpt.md)
 - **References**:
 	- [MPTokenIssuance entry][]
 	- [MPTokenIssuanceCreate transaction][]
 	- [MPTokenIssuanceDestroy transaction][]
 	- [MPTokenIssuanceSet transaction][]
-<!-- TODO: Add when the tutorial on sending MPTs is ready. -->
```
mDuo13, line 228: :tada: Great work catching and fixing this TODO.

_code-samples/send-an-mpt/js/README.md
```diff
@@ -0,0 +1,59 @@
+# Send an MPT Example (JavaScript)
```
mDuo13:
```suggestion
# Send an MPT (JavaScript)
```

_code-samples/send-an-mpt/py/README.md
```diff
@@ -0,0 +1,63 @@
+# Send an MPT Example (Python)
```
mDuo13:
```suggestion
# Send an MPT (Python)
```

mDuo13 (changes requested):
Seems like a nice improvement over the existing tutorial. I was able to run the samples as-is, but I identified a couple important changes that should be made so that the code runs robustly in more cases; it seems like the problems apply to all three languages.

In the future: this tutorial's setup steps could be significantly simplified if we used a well-known MPT (with issuance setup in advance, similar to the TST token) which users could quickly acquire. The problem is it would require either MPT-DEX (so holders could buy the token from the DEX) or an MPT faucet, so that improvement will have to be placed on the backburner.

_code-samples/send-an-mpt/README.md
maria-robobug: Ah good to know, I didn't realise this was used on that page. I'll shorten and simplify further. 👍🏾

_code-samples/send-an-mpt/py/send_mpt_setup.py
maria-robobug: Basically the idea is that setup scripts are technically not part of the tutorial, so we thought using the async client would help to make these steps faster for Python.

I think this is something we need to update in the Tutorial guidelines. We've only added it to our Claude rules recently.
https://github.com/XRPLF/xrpl-dev-portal/blob/master/.claude/rules/python-code-samples.md

_code-samples/send-an-mpt/js/sendMPT.js
maria-robobug: Good shout! I'll update 👍🏾

docs/tutorials/payments/send-an-mpt.md
maria-robobug: Will mention this. Also, I've realised tecINSUFFICIENT_FUNDS is actually incorrect. 
Should be tecPATH_PARTIAL. Confusing...

maria-robobug:
> In the future: this tutorial's setup steps could be significantly simplified if we used a well-known MPT (with issuance setup in advance, similar to the TST token) which users could quickly acquire. The problem is it would require either MPT-DEX (so holders could buy the token from the DEX) or an MPT faucet, so that improvement will have to be placed on the backburner.

Agreed, we really do need to look into test data for these tutorials, would be nice to avoid the setup steps/process entirely where possible.

_code-samples/send-an-mpt/js/sendMPTSetup.js
```diff
@@ -1,90 +1,99 @@
...
-  TransactionType: 'MPTokenIssuanceCreate',
-  Account: issuer.address,
-  AssetScale: 2,
-  MaximumAmount: '1000000',
-  TransferFee: 0,
-  Flags: xrpl.MPTokenIssuanceCreateFlags.tfMPTCanTransfer,
-  MPTokenMetadata: xrpl.encodeMPTokenMetadata(metadata)
-}, { wallet: issuer, autofill: true })
+  // Fund issuer and sender in parallel
+  process.stdout.write(`Setting up tutorial: ${step++}/${TOTAL_STEPS}\r`)
```
mDuo13, line 14: Is there a good reason to use `process.stdout.write(...)` instead of `console.log(...)` here and in the late setup steps?

docs/tutorials/payments/send-an-mpt.md
```diff
@@ -147,7 +150,7 @@ If the issuance uses allow-listing (the **Require Auth** flag), this step isn't
 
 ### 4. Check initial balances
 
-Before sending the payment, check each account's MPT holdings using the [account_objects method][]. For each account, filter by the object type `MPToken` and locate the entry whose `MPTokenIssuanceID` matches this issuance.
+Before sending the payment, check each account's MPT holdings using the [ledger_entry method][]. Pass the `mptoken` filter with the holder's `account` and the `mpt_issuance_id` to look up the `MPToken` entry directly. This avoids paginating through the account's owner directory and works reliably for accounts that own many ledger entries.
```
mDuo13: Nit: `mptoken` isn't a filter here.

mDuo13 (dismissed):
Looks good; had some minor nitpicks that you could optionally address, but aren't blockers. If you do, ping me for a quick re-review.

_code-samples/send-an-mpt/js/sendMPTSetup.js
maria-robobug: Purely visual, "Setting up tutorial: X/4" line overwrites the previous one on the same line, so it has a single in-place progress counter.

`console.log` appends \n, so it would print each step on a separate line and break that effect.

mDuo13 (approved):

Issue #3197, opened 2025-07-16
Update channel_authorize to reflect that it's restricted like other signing methods


mDuo13:
This change was included in `rippled` 2.5.0: https://github.com/XRPLF/rippled/pull/5385

The `channel_authorize` API method takes a secret key and uses it to produce a signature that can be used to claim funds from a payment channel. As a security measure against people sending secret keys to servers they don't control, this has been added to the list of methods that are admin-only unless the server admin enables public signing. 

The following pages need to be updated:

- [channel_authorize API reference](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/payment-channel-methods/channel_authorize) - the method is now admin-only by default, or public if public signing is enabled. It should be moved to the admin methods category.
    - Possibly, the "Payment Channel Methods" category, which would then contain only one other method (channel_verify) should be removed and redirected, while the channel_verify method is moved into another category (TBD)
-  [Enables public signing](https://xrpl.org/docs/infrastructure/configuration/enable-public-signing) - should list `channel_authorize` in the affected methods.
- Possibly add a note in the [Use Payment Channels how-to](https://xrpl.org/docs/tutorials/how-tos/use-specialized-payment-types/use-payment-channels) to clarify that the method is admin-only. (But also, that whole tutorial should probably be updated: see issue #1055.)

These changes should probably be tagged with an "Updated in: rippled 2.5.0" badge too.

SirJosh1987:
So what do I need to do exactly in simplest form?

PR #3708, opened 2026-06-15
Move channel_authorize to admin-only API methods

akshitj11:
## Summary

Implements the changes described in issue #3197 (good first issue). The `channel_authorize` method was made admin-only by default in rippled 2.5.0 (XRPLF/rippled#5385), but the documentation still listed it primarily under public API methods.

### Changes

1. **Moved `channel_authorize.md`** from `public-api-methods/payment-channel-methods/` to `admin-api-methods/signing-methods/` (alongside `sign` and `sign_for`)
2. **Updated `common-links.md`** - Changed link references to point to the new admin API path
3. **Updated `admin-api-methods/index.md`** - Changed cross-reference to local path
4. **Updated `public-api-methods/index.md`** - Simplified entry to cross-reference the admin API location with admin-only note
5. **Updated `enable-public-signing.md`** - Added `channel_authorize` to the list of methods affected by public signing
6. **Updated `sidebars.yaml`** - Moved sidebar entry from public payment-channel-methods to admin signing-methods
7. **Updated `redirects.yaml`** - Updated redirect target to new admin API path


docs/references/http-websocket-apis/admin-api-methods/signing-methods/channel_authorize.md
mDuo13, line 1: It looks like you've copied the file rather than moving it. That's not preferable, since now there are two instances that need to be kept in sync.

I think if you had deleted the old file (`docs/references/http-websocket-apis/public-api-methods/payment-channel-methods/channel_authorize.md`) in the same commit then git would recognize it as renamed and carry over the commit/blame history correctly. But that might not work if you do so in a new commit. So maybe you should amend or rebase and squash to get the change into the history correctly.

mDuo13 (changes requested):

akshitj11:
Thanks for the review. I've amended the commit to properly use git mv, git now recognizes it as a rename. The PR should be updated.
cc: @mDuo13

docs/references/http-websocket-apis/public-api-methods/index.md
```diff
@@ -74,9 +74,7 @@ Payment channels are a tool for facilitating repeated, unidirectional payments,
 
 * **[`channel_verify`](payment-channel-methods/channel_verify.md)** - Check a payment channel claim's signature.
 
-By default, the following method is [admin-only](../admin-api-methods/index.md). It can be used as a public method if the server admin has [enabled public signing](../../../infrastructure/configuration/enable-public-signing.md).
-
-* **[`channel_authorize`](payment-channel-methods/channel_authorize.md)** - Sign a claim for money from a payment channel.
+* **[`channel_authorize`](../admin-api-methods/signing-methods/channel_authorize.md)** - Sign a claim for money from a payment channel. (This method is [admin-only](../admin-api-methods/index.md) by default. It can be used as a public method if the server admin has [enabled public signing](../../../infrastructure/configuration/enable-public-signing.md).)
```
mDuo13: This change makes this section inconsistent with how the Transaction Methods section handles the same type of thing. The doc already separates out and notes the admin method, so I don't think you need to modify this file at all.

mDuo13 (changes requested):
I have a nitpick regarding the change to the public methods index page, and there's the merge conflict that needs to be fixed too, but that's all that's preventing this from being merged now.

akshitj11:
Thanks for the review! I've:

1. Reverted changes to `public-api-methods/index.md` so the Payment Channel section keeps the same admin-only structure as Transaction Methods.
2. Rebased onto latest `master` to resolve the merge conflict.

The doc move, sidebar, common-links, admin index, enable-public-signing, and redirect updates are unchanged. Happy to adjust further if needed.

mDuo13 (approved):

PR #3711, opened 2026-06-16
Migration guide v2

xVet:
Updates the rippled → xrpld migration guide to make the backup and data steps safer and easier to follow. Same page format as before (admonitions, badge, links).

## What changed

**Step 1 - Back up first**
- Adds a clear checklist of which files to back up and where they live, grouped by how important they are: config and keys (can't be recovered), ledger data (can be re-synced), and logs (optional).
- Adds a warning to back up before removing the package, with commands to copy the important files and move them off the server.
- Makes the full ledger backup (`tar`) optional, since Step 5 keeps your data in place anyway - only worth doing if you're unsure.

**Step 2 - Install xrpld**
- Stops the service after install, so nothing is running while you migrate.

**Steps 3-4 - Restore config**
- Restores your config and validators file from the backup you made in Step 1.

**Step 5 - Move your data**
- Splits the two options by who they're for: re-sync from scratch (validators and small nodes), or keep your data (full and large-history nodes).
- Removes the risky delete step and uses a safe move instead - the default folders are renamed, not deleted, so it can be undone.

**Step 6 - Restart xrpld**
- Uses `restart` instead of `start`, so the service actually picks up your new config.

Nothing was removed from the Post-Migration or See Also sections.

copy created via my friend claude

### DIFF
--- a/docs/infrastructure/installation/migrate-to-xrpld.md
+++ b/docs/infrastructure/installation/migrate-to-xrpld.md
@@ -33,16 +33,74 @@ These steps assume a typical Debian/Ubuntu `rippled` install with:
 
 Run these steps on each host you are migrating and adjust the paths to match your environment.
 
-### 1. Stop and remove rippled
+### 1. Back up, then stop and remove rippled
 
-Back up your config and data files (for example, `wallet.db`). Then stop the service and remove the package:
+{% admonition type="warning" name="Back up first - or risk losing your config and data" %}
+The next step removes the `rippled` package. While `apt-get remove` is *designed* to leave your config and data in place, a single slip - `purge` instead of `remove`, a wrong path, a failed disk, or an interrupted migration - can permanently destroy your configuration, your node identity, and your validator keys, with no way to recover them. Copy the files below to a location **off this host** before you run anything in this step.
+{% /admonition %}
+
+Back up the following files. In a default install they live here:
+
+**Configuration and identity** (cannot be regenerated - losing these loses your settings and, for validators, your on-ledger identity):
+
+- `/opt/ripple/etc/rippled.cfg` - main server config, including your `[validator_token]` on validators.
+- `/opt/ripple/etc/validators.txt` - your trusted validator list (UNL). Skip if your `[validators]` section lives inside `rippled.cfg`.
+- `/var/lib/rippled/db/wallet.db` - your node's persistent identity (node keypair) and peer reservations.
+- `validator-keys.json` (validators only) - your validator master keys, stored wherever you generated them with `validator-keys-tool` (often kept offline). Without this file you can never renew or rotate your validator token.
+
+**Ledger data** (large, but re-syncable from the network - back up only to skip a long re-sync):
+
+- `/var/lib/rippled/db/nudb/` - the NuDB node store, the bulk of your ledger data.
+- `/var/lib/rippled/db/*.db` - the SQLite databases, such as `transaction.db` and `ledger.db`.
+- `/var/lib/rippled/db/` - the whole data directory covers all of the above in one step.
+
+**Logs** (optional):
+
+- `/var/log/rippled/debug.log`
+
+{% admonition type="info" name="Find your own paths" %}
+The locations above are the defaults. Your real paths are whatever `[database_path]`, `[node_db]`, and `[debug_logfile]` point to in your `rippled.cfg`. If you customized them, back up those locations instead.
+{% /admonition %}
+
+Copy the config and identity files into a backup directory:
+
+```sh
+sudo mkdir -p /root/rippled-backup
+sudo cp /opt/ripple/etc/rippled.cfg     /root/rippled-backup/
+sudo cp /opt/ripple/etc/validators.txt  /root/rippled-backup/   # if present
+sudo cp /var/lib/rippled/db/wallet.db   /root/rippled-backup/
+```
+
+Then copy that directory **off the host**, so a disk failure or a botched migration can't take your backup down with it. For example, pull it to your own machine over SSH:
+
+```sh
+# Replace user@your-node with your SSH login and host
+scp -r user@your-node:/root/rippled-backup ./rippled-backup
+```
+
+{% admonition type="info" name="Optional: full-history and large-history nodes" %}
+This is optional. In Step 5 you keep your existing data by moving the directory into the new location, so the ledger store is preserved in place - not re-downloaded - and a separate snapshot usually isn't needed.
+
+That said, moving data directories is the riskiest part of the migration. If you have the disk space and any doubt, take a snapshot first as a safety net - it's cheap insurance. Stop the node so the copy is consistent:
+
+```sh
+sudo systemctl stop rippled
+sudo tar -czf /root/rippled-data-$(hostname).tar.gz -C /var/lib rippled
+```
+
+Validators and small-history nodes (a day or so) can skip this either way - they re-sync from the network in minutes.
+{% /admonition %}
+
+Once your backups are safely off the host, stop the service and remove the package:
 
 ```sh
 sudo systemctl stop rippled
 sudo apt-get remove -y rippled
 ```
 
-{% admonition type="info" name="Note" %}`apt-get remove` leaves the config files and your data directory in place.{% /admonition %}
+{% admonition type="info" name="Note" %}
+`apt-get remove` leaves the config files and your data directory in place. (By contrast, `apt-get purge` would delete the config - which is exactly why you back up first.)
+{% /admonition %}
 
 ### 2. Install xrpld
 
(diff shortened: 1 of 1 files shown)


docs/infrastructure/installation/migrate-to-xrpld.md
```diff
@@ -102,30 +174,42 @@ path=/var/lib/xrpld/db/nudb
...
+{% /admonition %}
+
 If your data or logs are in a non-default location, you only need to update ownership:
 
 ```sh
 sudo chown -R xrpld:xrpld [path to your data directory]
 sudo chown -R xrpld:xrpld [path to your log directory]
 ```
 
-### 6. Start xrpld
```
rachelflynn, line 124: This change is awaiting dev team review. Hold off for now.

docs/infrastructure/installation/migrate-to-xrpld.md
```diff
@@ -33,16 +33,74 @@ These steps assume a typical Debian/Ubuntu `rippled` install with:
 
 Run these steps on each host you are migrating and adjust the paths to match your environment.
 
-### 1. Stop and remove rippled
+### 1. Back up, then stop and remove rippled
 
-Back up your config and data files (for example, `wallet.db`). Then stop the service and remove the package:
+{% admonition type="warning" name="Back up first - or risk losing your config and data" %}
```
rachelflynn, line 38: Change the tone in "a single slip … can permanently destroy your configuration … with no way to recover them." The point is right, but I'd tighten it to something like: apt-get remove leaves your config and data in place, but one wrong flag or path can irreversibly destroy them for good. Copy the files below off this host first.

docs/infrastructure/installation/migrate-to-xrpld.md
```diff
@@ -60,31 +118,45 @@ sudo yum install xrpld
 
 This installs the `xrpld` binary at `/usr/bin/xrpld`, creates the `xrpld` user and group, and installs a default mainnet config at `/etc/xrpld/xrpld.cfg`. The `xrpld` service starts automatically.
 
+Stop it before going further. The auto-started service is running on the default config and syncing into a fresh `/var/lib/xrpld/`. Stop it so nothing is writing to those directories while you restore your config and move your data into place:
```
rachelflynn: Good catch adding this. Only concern is it's tucked at the end of Step 2, where someone skimming for "install" might miss it. Worth a bolded lead-in like "Important: stop the auto-started service before continuing." so it doesn't get skipped. (Promoting it to its own numbered step would be cleaner but requires renumbering - your call.)

rachelflynn:
Some of the new prose drifts more conversational than the surrounding guide ("it's cheap insurance", "just point xrpld at the new default paths and let it rebuild"). A bit more neutral language would match the rest of the docs.

docs/infrastructure/installation/migrate-to-xrpld.md
```diff
@@ -60,31 +118,45 @@ sudo yum install xrpld
...
+{% admonition type="info" name="Note" %}
+Some operators do not have a separate `validators.txt` file and instead keep the `[validators]` and `[validator_xxx]` sections embedded within `xrpld.cfg`. In that case, delete the generic config file instead.
+{% /admonition %}
 
 ### 5. Migrate the data directories
 
-Follow the scenario that matches your setup.
+Two scenarios, depending on whether you need to keep your ledger data. Pick the one that matches your node.
+
+#### Re-sync from the network (validators and small-history nodes)
```
rachelflynn: Suggestion: #### Re-sync small-history nodes and validators from the network

docs/infrastructure/installation/migrate-to-xrpld.md
```diff
@@ -102,30 +174,42 @@ path=/var/lib/xrpld/db/nudb
 /var/log/xrpld/debug.log
 ```
 
-#### Keep your existing data
+#### Keep your existing data (full-history and large-history nodes)
```
rachelflynn: Suggestion: #### Keep existing data on full-history nodes

docs/infrastructure/installation/migrate-to-xrpld.md
```diff
@@ -33,16 +33,74 @@ These steps assume a typical Debian/Ubuntu `rippled` install with:
 
 Run these steps on each host you are migrating and adjust the paths to match your environment.
 
-### 1. Stop and remove rippled
+### 1. Back up, then stop and remove rippled
```
rachelflynn: Suggest creating step 1 with back up only, then keeping stop and remove in step 2 so it doesn't get too lengthy.

docs/infrastructure/installation/migrate-to-xrpld.md
```diff
@@ -33,16 +33,74 @@ These steps assume a typical Debian/Ubuntu `rippled` install with:
...
+Once your backups are safely off the host, stop the service and remove the package:
 
 ```sh
 sudo systemctl stop rippled
 sudo apt-get remove -y rippled
 ```
 
-{% admonition type="info" name="Note" %}`apt-get remove` leaves the config files and your data directory in place.{% /admonition %}
+{% admonition type="info" name="Note" %}
+`apt-get remove` leaves the config files and your data directory in place. (By contrast, `apt-get purge` would delete the config - which is exactly why you back up first.)
```
rachelflynn: Suggestion: apt-get remove leaves your config and data in place. apt-get purge would delete the config, which is why you back up first.

docs/infrastructure/installation/migrate-to-xrpld.md
```diff
@@ -60,31 +118,45 @@ sudo yum install xrpld
...
+Stop it before going further. The auto-started service is running on the default config and syncing into a fresh `/var/lib/xrpld/`. Stop it so nothing is writing to those directories while you restore your config and move your data into place:
+
+```sh
+sudo systemctl stop xrpld
+```
+
 ### 3. Migrate the binary config
 
-Your existing `rippled.cfg` already contains your tuning, validators, and peers. Copy it to the new location:
+Restore the config you backed up in Step 1 - it already holds your tuning, validators, and peers - into the new `xrpld` location:
```
rachelflynn: Suggest reducing the sentence complexity. Something like "Restore the config you backed up in Step 1 into the new xrpld location. It already holds your tuning, validators, and peers:"

docs/infrastructure/installation/migrate-to-xrpld.md
```diff
@@ -60,31 +118,45 @@ sudo yum install xrpld
...
 
-{% admonition type="info" name="Note" %}Some operators do not have a separate `validators.txt` file and instead have the `[validators]` and `[validator_xxx]` sections embedded within the `xrpld.cfg` file. In that case, delete the generic config file instead.{% /admonition %}
+{% admonition type="info" name="Note" %}
+Some operators do not have a separate `validators.txt` file and instead keep the `[validators]` and `[validator_xxx]` sections embedded within `xrpld.cfg`. In that case, delete the generic config file instead.
+{% /admonition %}
 
 ### 5. Migrate the data directories
 
-Follow the scenario that matches your setup.
+Two scenarios, depending on whether you need to keep your ledger data. Pick the one that matches your node.
```
rachelflynn: Fragment - suggest something like "There are two scenarios, depending on whether you need to keep your ledger data. Pick the one that matches your node."

rachelflynn (changes requested):

xVet:
Thanks @rachelflynn - addressed the review in the latest push. Summary:

**Editorial / structural (your comments):**
- Split Step 1 into **Step 1 (Back up)** and **Step 2 (Stop and remove rippled)**; renumbered to 8 steps and updated every cross-reference.
- Tightened the backup-warning tone and simplified the `remove` vs `purge` note to your wording.
- Split the binary-config sentence (now Step 4) and fixed the "two scenarios" fragment.
- Adopted your heading rewrites for both data scenarios.
- Neutralized the conversational phrasing ("cheap insurance", "let it rebuild", "park the dirs").
- Added the bold **"Important: stop the auto-started service before continuing."** lead-in.

**Flagging for dev-team review (2 technical items):**

1. **Stop-after-install (now Step 3)** - the line you marked "awaiting dev review." I kept it: the packaging confirms the service auto-starts on install (`debian/rules` runs `dh_installsystemd` for `xrpld.service` with no `--no-start`/`--no-enable`, and `50-xrpld.preset` enables it). Please confirm this is the intended behaviour.

2. **Data step (Step 6)** - I changed the "keep your data" path from *moving* the directory to `/var/lib/xrpld` to **leaving it in place and just `chown`-ing it to `xrpld`**. The shipped unit uses `ProtectSystem=full` (not `strict`) with no `ReadWritePaths=`, so `/var/lib/rippled` stays writable and the config restored in Step 4 already points there - no path edits needed. This also closes a latent failure where moving the data without repointing the config (`[node_db]`, `[database_path]`, `[debug_logfile]`) breaks the node on restart. Would appreciate a dev sanity-check on the approach.

Happy to adjust either if the dev team prefers something different. Ready for another look when you have a moment.

rachelflynn (approved):

docs/infrastructure/installation/migrate-to-xrpld.md
```diff
@@ -102,33 +180,31 @@ path=/var/lib/xrpld/db/nudb
...
+For full-history and large-history nodes, where re-downloading the ledger store from peers would take hours or days. The simplest and safest option is to leave your data where it is and hand ownership to the new `xrpld` user. The config you restored in Step 4 already points at `/var/lib/rippled` and `/var/log/rippled`, so no paths need to change:
 
 ```sh
-sudo mv /var/lib/rippled /var/lib/xrpld
-sudo chown -R xrpld:xrpld /var/lib/xrpld
-
-sudo mv /var/log/rippled /var/log/xrpld
-sudo chown -R xrpld:xrpld /var/log/xrpld
+sudo chown -R xrpld:xrpld /var/lib/rippled
+sudo chown -R xrpld:xrpld /var/log/rippled
```
legleux, line 189: Was this tested? When strictly run as a service this should be fine.
This does address a concern I had that `sudo mv /var/lib/rippled /var/lib/xrpld` was _not_ going to do what was intended.

There's some very annoying bugs regarding how xrpld locates the config and database directories that I'm going to address ASAP.
I'm pretty sure they'd only present when running `xrpld` locally as an RPC client _and_ not specifying the config but they would also present in several other places. I'm not sure how to make it know without adding additional confusion though.

Otherwise this looks like a decent way to kick the can.

docs/infrastructure/installation/migrate-to-xrpld.md
```diff
@@ -102,33 +180,31 @@ path=/var/lib/xrpld/db/nudb
...
+
+### 7. Restart xrpld
 
-### 6. Start xrpld
+You stopped `xrpld` in Step 3 to migrate safely. Bring it up now so it loads your restored config and reads your data with the right ownership:
 
 ```sh
 sudo systemctl daemon-reload
-sudo systemctl start --now xrpld
+sudo systemctl restart xrpld
```
legleux, line 204: @rachelflynn re:
https://github.com/XRPLF/xrpl-dev-portal/pull/3711/changes#r3421748479

The `--now` was redundant and you were right that start would have not reloaded the config since the service was running already.
This change with the modification I commented on above should be good.

legleux (approved):

docs/infrastructure/installation/migrate-to-xrpld.md
legleux: This got handled down in the change https://github.com/XRPLF/xrpl-dev-portal/pull/3711#discussion_r3423557607 referred to.

PR #3714, opened 2026-06-17
Remove outdated page and add redirect

amarantha-k:


redirects.yaml
```diff
@@ -320,7 +320,9 @@ contribute.html:
 events.html:
     to: /community/events
 ambassadors.html:
-    to: /community/ambassadors
+    to: https://ripple.com/impact/ubri/
+community/ambassadors:
+    to: https://ripple.com/impact/ubri/
```
rachelflynn: trailing whitespace

styles/_pages.scss
```diff
@@ -591,30 +591,6 @@
...
-
-      {
-      ##{$eligibility} {
-        content: url("../img/ambassadors/#{$eligibility}.svg");
-      }
-    }
-  }
-
   // Button has link out arrow.
   .btn {
```
rachelflynn, line 595: Quick question - were .btn and below nested under .page-ambassadors? Just wondering if removing the wrapper braces means they'd be in the global scope now.

styles/_pages.scss
rachelflynn, line 580: Optionally could remove this now since we removed the image from developer-funding.page.tsx.

rachelflynn (dismissed):

styles/_pages.scss
amarantha-k: No, they weren't nested under .page-ambassadors.

redirects.yaml
amarantha-k: good catch, will fix it.

rachelflynn (approved):

PR #3725, opened 2026-06-23
Fix label name typo for standby seed and amount

CarloDotGG:
Minor typo correction


darlenelangdon34-sys (approved):

mDuo13 (approved):
Reasonable fix, although not very important. We do plan to replace this code sample entirely at some point.

mDuo13 (approved):

PR #3728, opened 2026-06-24
Update release notes script

oeggert:
This PR updates the `generate-release-notes.py` script:

1. Streams and calculates SHA 256 checksum from `xrpld` packages. Falls back to `TODO` if download fails.
2. Fixes a case-sensitivity matching issue when creating the Amendments key for an AI to sort.


maria-robobug (approved):
LGTM

PR #3731, opened 2026-06-25
Update graphics in Crypto Wallets

rachelflynn:

- Refresh the graphics in [Crypto Wallets](https://xrpl.org/docs/introduction/crypto-wallets) (and translations) with new SVGs


docs/introduction/crypto-wallets.md
```diff
@@ -15,12 +15,12 @@ Crypto wallets provide a way to manage your account and funds on the XRP Ledger.
 
 A major factor when choosing a wallet is if you want a custodial or non-custodial wallet.
 
+![Custodial vs. Non-custodial Wallets](/docs/img/custodial-non-custodial.svg)
```
mDuo13, line 18: This image has too much text, which becomes illegible in the mobile view. Same for the other one.

Maybe we can make this into two images with _just_ the bank/key images and then put the textual parts of the graphic into a table or some kind of React component?

mDuo13 (changes requested):
These images are really a bit too text-heavy to function as they are.

mDuo13 (approved):
The updated tables are great and look better on mobile. :tada:

mDuo13:
Oh yeah, I don't know why the project preview got stuck but I tested in local build and it's fine.

PR #3735, opened 2026-07-01
Update graphics in Trading NFTs

rachelflynn:

- Refresh the graphics in [Trading NFTs](https://xrpl.org/docs/concepts/tokens/nfts/trading) (and translations) with new SVGs


mDuo13 (changes requested):
The new graphics are illegible on mobile: 
> 

Honestly this type of text-heavy graphic might be best converted into a custom React component, which will display the actual text in the page (which has better searchability, accessibility, etc.), and can be styled to display differently on mobile.

Also, the original used different "swim lanes" to clearly distinguish who does which action, whereas the text-only display for the new style makes it a little less obvious that different people are taking action at different steps. Maybe you could add icons representing the different participants? (Not that I have a great idea of what icons you'd use to distinguish buyer/seller/broker.)

@l10n/ja/docs/concepts/tokens/nfts/trading.md
```diff
@@ -67,13 +67,15 @@ XRP Ledger上のアカウント間で`NFToken`オブジェクトを転送する
 最も単純なワークフローでは、クリエイターが新しい`NFToken`を発行します。クリエイターは売却オファーを作成する際、最低売却価格を入力し、売却先にブローカーを設定します。購入希望者はブローカーを経由して`NFToken`に入札を行います。ブローカーは落札者を選び、取引を完了させ、ブローカー手数料を受け取ります。ベストプラクティスとして、ブローカーは`NFToken`に対して残っている購入オファーをすべてキャンセルします。
 
 
-![Brokered Mode with Reserve](/docs/img/nft-brokered-mode-with-reserve.png)
+<img class="d-none d-md-block" src="/docs/img/nft-brokered-mode-with-reserve.svg" alt="リザーブありのブローカーモード：出品者がNFTを発行して最低価格で出品し、購入者が入札し、ブローカーが売却を成立させて手数料を受け取り、残りの入札をキャンセルします。" />
+<img class="d-block d-md-none" src="/docs/img/nft-brokered-mode-with-reserve.mobile.svg" alt="" />
```
mDuo13: Leaving the alt text blank for the mobile version is not quite ideal. I believe most screenreaders won't read text that's set to `display: none` so the alt text will only be used on desktop, not on mobile. That might be OK, but in theory I think the alt text should be present in both elements. Of course, that means it has to be duplicated, which kind of sucks for us.

Also, inlining HTML is OK but not ideal, since it creates one-offs and exceptions that we have to double-check haven't broken when we do broader site updates. For example, `d-md-block` is a Bootstrap class, so if we ever migrate away from Bootstrap we would have to remember that this doc exists and adjust it, or it would silently break with the update.

I suspect both of these can be addressed with a React component.

mDuo13 (dismissed):
Not a bad solution! I still think a fully custom component would probably be the _best_ UX, but this is a lot more lightweight as a solution and probably makes more sense since I doubt we'd reuse the component much if at all. The use of colors is a nice touch.

The desktop graphic is still a little small / hard to read in some places, though. With a lot of our other graphics, we present them as links to the image file so people can click them to enlarge.

I think we can work around the various limitations of this approach with a quick Markdoc tag for responsive graphics. I'll try that and get back to you.

mDuo13:
Take a look at the markdoc tag I threw together for this — I think that's an improvement over the inlined HTML. Should probably get a review from someone other than me, though.

oeggert (approved):
Approving, with one minor caveat to check. This may just be an issue on my end, but when viewing the image on mobile (iPhone Safari) the image tiles randomly switch between dark/light mode on page load/refresh rather than switching based on site light/dark mode selection.

mDuo13:
> This may just be an issue on my end, but when viewing the image on mobile (iPhone Safari) the image tiles randomly switch between dark/light mode on page load/refresh rather than switching based on site light/dark mode selection.

I don't see this effect when viewing on my own (Android Firefox) device though it does seem to get "stuck" in dark mode display even if I toggle to light mode. It's still legible regardless so I think it's fine, but agree it's worth keeping an eye on.

PR #3736, opened 2026-07-01
Add simulate method limitation for Batch transactions

rachelflynn:
As part of issue #[3667](https://github.com/XRPLF/xrpl-dev-portal/issues/3667), migrates [Batch and Simulate](https://github.com/XRPLF/rippled/wiki/Batch-and-Simulate) wiki content to XRPL.org: 
- Added a limitations section to [Batch Transactions](https://xrpl.org/docs/concepts/transactions/batch-transactions)
- Added new possible error to [simulate](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/transaction-methods/simulate)

### DIFF
--- a/docs/concepts/transactions/batch-transactions.md
+++ b/docs/concepts/transactions/batch-transactions.md
@@ -172,7 +172,7 @@ Inner transactions cannot be broadcast (and won't be accepted if they happen to 
 
 ## Limitations
 
-`Batch` transactions are incompatible with the [simulate method][]. `simulate` operates locally as a dry run and does not interact with the peer protocol. Therefore, the transaction is neither applied nor broadcast over the network.
+`Batch` transactions are incompatible with the [simulate method][]. Since `simulate` operates as a dry run for a single transaction only, it cannot simulate the effect of adding inner transactions to the consensus set.
 
 ## Integration Considerations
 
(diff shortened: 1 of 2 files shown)


docs/concepts/transactions/batch-transactions.md
```diff
@@ -165,10 +165,15 @@ If Alice provides a fully autofilled and signed transaction to Bob, Bob can subm
...
 
 ### Inner Transaction Safety
+
 An inner batch transaction is a special case. It doesn't include a signature or a fee (since those are both included in the outer transaction). Therefore, they must be handled carefully to ensure that someone can't somehow directly submit an inner `Batch` transaction without it being included in an outer transaction.
 
 Inner transactions cannot be broadcast (and won't be accepted if they happen to be broadcast, for example, from a malicious node). They must be generated from the `Batch` outer transaction instead. Inner transactions cannot be directly submitted via the submit RPC.
 
+## Limitations
+
+`Batch` transactions are incompatible with the [simulate method][]. `simulate` operates locally as a dry run and does not interact with the peer protocol. Therefore, the transaction is neither applied nor broadcast over the network.
```
mDuo13: This phrasing feels a bit weird to me. It doesn't fully explain why Batch is incompatible with `simulate`, and it almost seems to imply that simulate can apply or broadcast other types of transactions (even though, on careful reading, the intended meaning is the opposite).

I suggest this, instead:

```suggestion
`Batch` transactions are incompatible with the [simulate method][]. Because `simulate` operates as a dry run for a single transaction only, it cannot simulate the effect of adding inner transactions to the consensus set.
```

mDuo13 (changes requested):
Suggested a rephrase for clarity, otherwise on board

mDuo13 (approved):

PR #3739, opened 2026-07-02
docs: update OS support documentation with glibc requirement

shichengripple001:
## Summary

- Replaces vague OS descriptions in system-requirements.md with links to the install pages
- Updates install pages to use minimum glibc requirement (>= 2.31) instead of a hardcoded version table, so the doc stays accurate as new distro versions are released
- Lists minimum supported versions per distro family

## Changes

**system-requirements.md**: replaced hardcoded OS list with links to install pages

**install-xrpld-on-ubuntu.md**:
- Ubuntu 20.04 (Focal) or later
- Debian 11 (Bullseye) or later
- Added `focal`, `trixie` codenames to apt sources list

**install-xrpld-on-rhel.md**:
- Red Hat Enterprise Linux 9 or later
- Rocky Linux 9 or later
- AlmaLinux 9 or later

## Test plan

- [ ] Verify rendered pages look correct on the docs site
- [ ] Confirm codenames in apt sources list match the repo at repos.ripple.com

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/infrastructure/installation/system-requirements.md
+++ b/docs/infrastructure/installation/system-requirements.md
@@ -12,15 +12,7 @@ The following system requirements apply to both the core XRP Ledger server, `xrp
 
 For reliable performance in production environments, it is recommended to run a server on bare metal with the following characteristics or better:
 
-- Operating System: One of the following supported Linux distributions:
-
-    | Distribution | Versions |
-    |---|---|
-    | Ubuntu | 22.04 (Jammy), 24.04 (Noble), 26.04 (Resolute) |
-    | Debian | 12 (Bookworm), 13 (Trixie) |
-    | Rocky Linux | 9 |
-    | AlmaLinux | 9 |
-    | Fedora | 42 |
+- Operating System: A supported Linux distribution. See [Install on Ubuntu or Debian](install-xrpld-on-ubuntu.md) or [Install on RHEL](install-xrpld-on-rhel.md) for the full list of recommended versions.
 - CPU: 3+ GHz 64-bit x86_64 processor with 8+ cores.
 - Disk: SSD / NVMe (10,000 IOPS sustained - not burst or peak - or better). Minimum 50 GB for the database partition. Do not use Amazon Elastic Block Store (AWS EBS) because its latency is too high to sync reliably.
 - RAM: 64 GB.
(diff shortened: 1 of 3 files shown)


docs/infrastructure/installation/system-requirements.md
```diff
@@ -12,7 +12,15 @@ The following system requirements apply to both the core XRP Ledger server, `xrp
 
 For reliable performance in production environments, it is recommended to run a server on bare metal with the following characteristics or better:
 
-- Operating System: Ubuntu (LTS), Red Hat Enterprise Linux (latest release), or a compatible Linux distribution.
+- Operating System: One of the following supported Linux distributions:
+
+    | Distribution | Versions |
+    |---|---|
+    | Ubuntu | 22.04 (Jammy), 24.04 (Noble), 26.04 (Resolute) |
```
mathbunnyru: We do support Ubuntu 20.04

docs/infrastructure/installation/system-requirements.md
```diff
@@ -12,7 +12,15 @@ The following system requirements apply to both the core XRP Ledger server, `xrp
...
+    | Debian | 12 (Bookworm), 13 (Trixie) |
+    | Rocky Linux | 9 |
+    | AlmaLinux | 9 |
+    | Fedora | 42 |
```
mathbunnyru: Hardcoded list like this gets out-of-date fast (as soon as the new distro is released), but as a rule of thumb only libc versions matters.

It would be better to say minimal libc version we require (2.31), and list minimal versions of distros, like: `Ubuntu >= 20.04`

docs/infrastructure/installation/system-requirements.md
```diff
@@ -26,7 +34,7 @@ For a validator in AWS, consider `z1d.2xlarge` with an extra 1 TB disk for loggi
 
 For testing purposes, you can run an XRP Ledger server on commodity hardware with the following minimum requirements:
 
-- Operating System: macOS, Windows (64-bit), or most Linux distributions (Red Hat, Ubuntu, and Debian supported).
+- Operating System: macOS, Windows (64-bit), or a supported Linux distribution (see recommended specs above).
```
mathbunnyru: Linux should be mentioned first, as we target Linux

mathbunnyru (changes requested):

docs/infrastructure/installation/system-requirements.md
legleux: Absolutely don't put a Fedora number on that list!

docs/infrastructure/installation/system-requirements.md
legleux: Yes, I would argue we _don't_ support macOS and Windows.

legleux:
The specifics are on these pages:
https://xrpl.org/docs/infrastructure/installation/install-xrpld-on-ubuntu
https://xrpl.org/docs/infrastructure/installation/install-xrpld-on-rhel

@mathbunnyru You're right that it doesn't really matter (and never really has) but we need to specify to at least have _some_ line to say "Sorry, we can't support that".

mathbunnyru (approved):

mathbunnyru (approved):

maria-robobug (approved):

PR #3746, opened 2026-07-07
Show tree nav on mobile

### LINKED ISSUE #3323: Tree nav isn't accessible on mobile
On desktop, the tree nav (circled in red) provides a critical way to navigate the docs:
Image

On mobile, the tree nav is hidden, and the hamburger menu only brings up (the mobile version of) the top nav, not left-side tree nav. There's no way to access the tree nav anymore.
screenshot of mobile view screenshot of mobile hamburger menu

On Redocly's official site, the hamburger menu defaults to tree nav with a "< Main menu / Docs" link that takes you to a mobile version of the top nav:

Image

I suspect that what happened is Redocly changed to how/where the default page layout shows the tree nav—they moved it so the way you access it is through the hamburger menu, but that's lost on us because we have a totally custom hamburger menu.

mDuo13:
Fix #3323 so that the tree nav (on pages that have it) displays following the main page content on mobile, instead of being completely gone.

This isn't necessarily the prettiest or _most_ usable solution, but it is a fairly small set of changes that makes the tree navigation usable _at all_ on mobile. With the big redesign of #3362 still pending, I didn't want to put in a lot of effort that would have to be thrown out.

| Before | After |
|---|---|
|  |  |
| Feedback widget and prev/next page buttons are followed by footer. | Feedback widget and prev/next page buttons are followed by tree nav, and footer is below that. |

### DIFF
--- a/@theme/styles.css
+++ b/@theme/styles.css
@@ -380,6 +380,23 @@ button[role="tab"].active {
   line-height: var(--md-table-line-height);
 }
 
+@media screen and (max-width: 959.9px) {
+  [data-component-name="Layout/PageLayout"] {
+    flex-direction: column-reverse;
+  }
+  [data-component-name="Sidebar/Sidebar"] {
+    display: block;
+    position: static;
+    height: auto;
+  }
+  #content {
+    width: auto;
+  }
+  [data-component-name="Menu/MenuContainer"] {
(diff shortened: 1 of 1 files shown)


rachelflynn (approved):
LGTM :)

PR #3747, opened 2026-07-08
Update diagrams in Running an NFT Auction

rachelflynn:
- Refresh the diagrams in [Running an NFT Auction](https://xrpl.org/docs/concepts/tokens/nfts/running-an-nft-auction) (and translations) with new SVGs


amarantha-k (approved):
LGTM

PR #3748, opened 2026-07-08
Fix small issues in landings

mDuo13:
- Fix an issue where arrow buttons become white-text-on-white-background when hovered in light mode. (Examples of affected buttons include "Read the Docs →" on the docs landing page and "Get Started →" on the Agentic Transactions page)
- On the Agentic Transactions page, add appropriate links for cards 1-3 of "What an Agent Needs"
- In the same "What an Agent Needs" section, use an up-right arrow for the Context7 link to indicate that it's external, as opposed to the other links which stay on xrpl.org.


oeggert (approved):
LGTM

PR #3774, opened 2026-07-10
Update {amendments-table} component

oeggert:
This PR updates the `{amendments-table}` component in the following ways:

1. Adds a manual exclusion list for amendments that should be excluded from the table but aren't due to incorrect endpoint data.
2. Adds LLM .md export functionality. `renderForLlms` is a synchronous function, so we can't have it fetch live data at site build time. Current workaround is to introduce a snapshot of amendment data in the repo.
3. Adds a Py script to fetch amendment data and write to the repo with date and time of fetch.
4. Adds Github workflow to run Py script once a day and commit snapshot data to `master`. This _should_ kick off a new build. Note: I expect this method of refreshing live data to apply to other components such as amendment badges.

I'm fairly certain the workflow is configured correctly, but I can't do a live run until this PR is merged to `master`.

### DIFF
--- a/@theme/components/Amendments.tsx
+++ b/@theme/components/Amendments.tsx
@@ -3,6 +3,7 @@
 import * as React from 'react'
 import { Link } from '@redocly/theme/components/Link/Link'
 import { useThemeHooks } from '@redocly/theme/core/hooks'
+import amendmentsSnapshot from '../data/amendments-snapshot.json'
 
 type Amendment = {
   name: string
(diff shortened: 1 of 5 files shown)


tools/fetch-amendments-snapshot.py
```diff
@@ -0,0 +1,55 @@
...
+    OUT.write_text(json.dumps(snapshot, indent=2, ensure_ascii=False) + "\n", encoding="utf-8")
+    print(f"Wrote {len(amendments)} amendments to {OUT} (fetched_at {snapshot['fetched_at']}).")
+
+
+if __name__ == "__main__":
+    try:
+        main()
+    except Exception as err:
+        print(err, file=sys.stderr)
+        sys.exit(1)
```
mDuo13: Why is this try/catch block here? Isn't writing the error to stderr and exiting with an error code what Python normally does when an exception occurs anyway? Unless you're going to add retries or catch specific exceptions here, the only difference I can see from naively running `main()` is you won't see the traceback when an error occurs.

@theme/components/Amendments.tsx
```diff
@@ -93,6 +106,46 @@ function sortAmendments(list: Amendment[]): Amendment[] {
   })
 }
 
+// Get amendment status from snapshot data.
+function amendmentStatusForLlms(a: Amendment): string {
```
mDuo13, line 110: :+1: The table looks good in my testing.

mDuo13 (dismissed):
Not sure why the build failed on CI but it worked locally and looks good.

I have a couple other changes I'd like to add on to this project in a bit, but that can come in a separate PR after this is merged.

tools/fetch-amendments-snapshot.py
oeggert: Oops--looks like some boilerplate snuck through. Updated this to just run `main()` and removed the unnecessary `sys` import.

mDuo13 (approved):

PR #3775, opened 2026-07-14
Amendment component fixes

### LINKED ISSUE #3539: Known Amendments Page scrolling issue
When navigating to the Known Amendments page via an anchor link (e.g. from a concept page like [ https://xrpl.org/docs/concepts/tokens/lending-protocol](https://xrpl.org/docs/concepts/tokens/lending-protocol), the page loads at the correct URL with the correct hash, but does not scroll to the target amendment.

Follow these steps:
1. Go to https://xrpl.org/docs/concepts/tokens/lending-protocol
2. Click the LendingProtocol amendment link, which points to https://xrpl.org/resources/known-amendments#lendingprotocol

The Known Amendments page loads, the URL is correct, but the page doensn't scroll to the #lendingprotocol anchor.

The Known Amendments page renders amendment data dynamically, so it could be that those tables are not fully rendered when the browser attempts to resolve the anchor; the tables might not have a height yet. 

Proposed solutions:
- Option 1: Fix the scroll issue. Investigate whether the scroll is firing before the table content is rendered.

- Option 2: Dedicated page per amendment detail. Move amendment details out of a single long table page and into individual pages under the lfh nav. This should eliminate the anchor scroll problem entirely and potentially make it easier for LLMs to parse amendment details.

mDuo13:
- Pre-calculate height of amendment status table and set the loader element to that size so that the page doesn't jump significantly when the table properly loads in. Fixes #3539.
- Add a status-only mode to `{% amendment-disclaimer ... /%}` so that it can be used in the individual amendments' info tables on the Known Amendments page. Note, this does not work for amendments that are not in the VHS data, namely In Development and Obsolete amendments, so those tables still need to manually specify their status.
    - This fixes a couple cases where the individual amendment table is not up-to-date with the actual status of the amendment  (like it says "Open for Voting" even though the amendment has been enabled already).
    - Note, we are _close_ to being able to replace these detail tables entirely with a Markdoc tag that handles them automatically. The two limitations that are stopping us from going there yet are: (a) We don't have data for "In Development" amendments this way, and (b) VHS doesn't have "default vote" data.
- Add a `renderForLlms` function for `{% amendment-disclaimer ... /%}` so that the component renders properly in the Markdown-for-LLMs view. Note, it currently does not support translation. (That appears to be a Redocly limitation, which I've asked them about.)
- Extend the manually-curated list of obsolete amendments into a more thorough JSON file, and use it with a new Markdoc tag to populate the Obsolete Amendments table much in the same way as the main Amendments table (except it doesn't have to rely on live data from VHS).

### DIFF
--- a/@theme/components/Amendments.tsx
+++ b/@theme/components/Amendments.tsx
@@ -14,46 +14,96 @@ type Amendment = {
   date?: string
   id: string
   eta?: string
-  deprecated?: boolean
+  deprecated: boolean
 }
 
 type AmendmentsResponse = {
   amendments: Amendment[]
 }
 
-type AmendmentsCachePayload = {
+type AmendmentsCache = {
   timestamp: number
-  amendments: Amendment[]
+  vote: Amendment[]
+  info: Amendment[]
 }
 
 // API data caching
-const amendmentsEndpoint = 'https://vhs.prod.ripplex.io/v1/network/amendments/vote/main/'
+const amendmentsVoteEndpoint = 'https://vhs.prod.ripplex.io/v1/network/amendments/vote/main/'
 const amendmentsInfoEndpoint = 'https://vhs.prod.ripplex.io/v1/network/amendments/info/main/'
 const amendmentsCacheKey = 'xrpl.amendments.mainnet.cache'
 const amendmentsTTL = 15 * 60 * 1000 // 15 minutes in milliseconds
 
-function readAmendmentsCache(): Amendment[] | null {
+function readCache(): AmendmentsCache | null {
   if (typeof window === 'undefined') return null
   try {
     const raw = window.localStorage.getItem(amendmentsCacheKey)
     if (!raw) return null
-    const parsed: AmendmentsCachePayload = JSON.parse(raw)
-    if (!parsed || !Array.isArray(parsed.amendments)) return null
+    const parsed: AmendmentsCache = JSON.parse(raw)
+    if (!parsed || !Array.isArray(parsed.vote) || !Array.isArray(parsed.info)) return null
     const fresh = Date.now() - parsed.timestamp < amendmentsTTL
-    return fresh ? parsed.amendments : null
+    return fresh ? parsed : null
   } catch {
     return null
   }
 }
 
-function writeAmendmentsCache(amendments: Amendment[]) {
+function writeCache(cache: AmendmentsCache) {
   if (typeof window === 'undefined') return
   try {
-    const payload: AmendmentsCachePayload = { timestamp: Date.now(), amendments }
-    window.localStorage.setItem(amendmentsCacheKey, JSON.stringify(payload))
+    window.localStorage.setItem(amendmentsCacheKey, JSON.stringify(cache))
   } catch {
     // Ignore quota or serialization errors
   }
+}
+
+// Fetch a single amendments endpoint (vote or info).
+async function fetchAmendmentList(endpoint: string): Promise<Amendment[]> {
+  const response = await fetch(endpoint)
+  if (!response.ok) {
+    throw new Error(`HTTP error ${response.status} from ${endpoint}`)
+  }
+  const data: AmendmentsResponse = await response.json()
+  return data.amendments
+}
+
+// Deduplicated fetching for the amendment data; concurrent callers share the same promise.
+let inFlight: Promise<AmendmentsCache> | null = null
+
+// Cache-first: return the fresh combined cache, or fetch both endpoints in parallel,
+// stamp them with a single timestamp, and store them under one cache key.
+function getAmendmentsCache(): Promise<AmendmentsCache> {
+  const cached = readCache()
+  if (cached) return Promise.resolve(cached)
+  if (inFlight) return inFlight
+  const request = (async () => {
+    const [vote, info] = await Promise.all([
+      fetchAmendmentList(amendmentsVoteEndpoint),
+      fetchAmendmentList(amendmentsInfoEndpoint),
+    ])
+    const cache: AmendmentsCache = { timestamp: Date.now(), vote, info }
+    writeCache(cache)
+    return cache
+  })()
+  inFlight = request
+  // Release the slot once settled so a later cold load can refresh.
+  const release = () => { if (inFlight === request) inFlight = null }
+  request.then(release, release)
+  return request
+}
+
+// Vote list for the amendments table (the caller filters out obsolete amendments).
+function getAmendments(): Promise<Amendment[]> {
+  return getAmendmentsCache().then(cache => cache.vote)
+}
+
+// Look up a single amendment for a disclaimer: the vote list first, then the info
+// list for obsolete/never-enabled amendments.
+async function findAmendment(name: string): Promise<Amendment> {
+  const { vote, info } = await getAmendmentsCache()
+  const found = vote.find(a => a.name === name) ?? info.find(a => a.name === name)
+  if (found) return found
+
+  throw new Error(`Couldn't find ${name} amendment in status tables.`)
 }
 
 // Obsolete amendments (deprecated and never enabled, e.g. Batch) are documented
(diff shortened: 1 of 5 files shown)


@theme/components/Amendments.tsx
```diff
@@ -449,6 +490,61 @@ export function AmendmentDisclaimer(props: {
...
+  }
+
+  let text1, text2
+  if (mode === "updated") {
+    if (amendment.date) {
+      text1 = translate("component.amendment-status.updated.1", "Updated by the ")
+      text2 = translate("component.amendment-status.updated.2", ".")
+    } else {
+      text1 = translate("component.amendment-status.requires.1", "Requires the ")
+      text2 = translate("component.amendment-status.requires.2", ".")
```
oeggert:
```suggestion
      text1 = translate("component.amendment-status.updates.1", "The ")
      text2 = translate("component.amendment-status.updates.2", " updates this.")
```
The text here should match the base version of the component. Also side note: the Japanese `translations.yaml` is missing entries for these 2 lines.

resources/known-amendments.md
```diff
@@ -244,7 +244,7 @@ Although this amendment is enabled, it has no effect unless the [SusPay](#suspay
 | Amendment    | CryptoConditionsSuite |
 |:-------------|:----------------------|
 | Amendment ID | 86E83A7D2ECE3AD5FA87AB2195AE015C950469ABF0B72EAACED318F74886AE90 |
-| Status       | Obsolete |
+| Status       | {% amendment-disclaimer name="CryptoConditionsSuite" statusOnly=true /%} |
```
oeggert, line 233: This is rendering as 0% voting status. The vote endpoint hasn't updated the deprecated value, so maybe use the `isObsolete()` function to filter this to the correct status. Though this does highlight an issue around rendering Obsolete amendments. The endpoint data can either be mislabeled like in this case, or be on the fallback info endpoint where it's still not labelled correctly as deprecated.

I don't have a great solution for this and the `Batch` issue. In #3782 we cache both endpoints. So maybe we can treat amendments that don't show up in the vote endpoint as `Inactive`. Then look at the deprecated field in the info endpoint to determine if it's `Obsolete`; to get around mislabelling we could include an extra manual list like the one I added to the `isObsolete()` check.

resources/known-amendments.md
```diff
@@ -111,7 +111,7 @@ For details, see the [XLS-73: AMMClawback specification](https://github.com/XRPL
 | Amendment    | Batch |
 |:-------------|:------|
 | Amendment ID | 894646DD5284E97DECFE6674A6D6152686791C4A95F8C132CCA9BAF9E5812FB6 |
-| Status       | Obsolete |
+| Status       | {% amendment-disclaimer name="Batch" statusOnly=true /%} |
```
oeggert, line 100: This renders as `Inactive: Get details`, which just links to itself. It should be showing Obsolete. Originally when I added this Inactive status it was because the endpoint doesn't really provide enough info to make the determination why something was inactive, so the link was meant to go to our manually maintained entry for details. Since {amendment-disclaimer} wasn't being used on the Known Amendments page, it was an okay solution, but now that we use it here we should probably figure something out.

I think we can do some logic using the vote and info endpoints to bucket these as Obsolete, Inactive, etc. The problem is those endpoints don't include the dates when things are removed.

oeggert (changes requested):
Mostly looks good to me. There are 2 required fixes:

1. Dupe API calls. #3782 points to this branch and fixes it. If you decide to do more work to resolve the optional issue with Obsolete amendments, maybe review this PR first and merge it in beforehand.
2. Minor fix around wording for `mode=updated` in `{amendment-disclaimer}`.

The only other issue, which maybe should be a separate PR, revolve around Obsolete/Inactive amendments. See review comments for details.

resources/known-amendments.md
mDuo13: Solved by switching to a more extensive manually-curated JSON file for obsolete amendments.

@theme/data/obsolete-amendments.json
```diff
@@ -0,0 +1,82 @@
...
+        "id": "3C43D9A973AA4443EF3FC38E42DD306160FBFFDAB901CD8BAA15D09F2597EB87",
+        "name": "NonFungibleTokensV1",
+        "rippled_version": "1.9.0",
+        "status": "To Be Removed"
+    },
+    {
+        "id": "86E83A7D2ECE3AD5FA87AB2195AE015C950469ABF0B72EAACED318F74886AE90",
+        "name": "CryptoConditionsSuite",
+        "rippled_version": "0.60.0",
+        "status": "To Be Removed"
```
oeggert: This amendment is marked retired in `features.macro`, which seems incorrect? Retired is supposed to be mean functioning code active for at least two years, but the details of this amendment on the Known Amendments page says otherwise.

@theme/data/obsolete-amendments.json
```diff
@@ -0,0 +1,82 @@
...
+        "id": "0285B7E5E08E1A8E4C15636F0591D87F73CB6A7B6452A932AD72BBC8E5D1CBE3",
+        "name": "fixNFTokenDirV1",
+        "rippled_version": "1.9.1",
+        "status": "To Be Removed"
+    },
+    {
+        "id": "3C43D9A973AA4443EF3FC38E42DD306160FBFFDAB901CD8BAA15D09F2597EB87",
+        "name": "NonFungibleTokensV1",
+        "rippled_version": "1.9.0",
+        "status": "To Be Removed"
```
oeggert: Should this be marked as already removed? It was replaced by `NonFungibleTokensV1_1` and isn't in the `features.macro` file or up for voting.

@theme/data/obsolete-amendments.json
```diff
@@ -0,0 +1,82 @@
...
+        "id": "36799EA497B1369B170805C078AEFE6188345F9B3E324C21E9CA3FF574E3C3D6",
+        "name": "fixNFTokenNegOffer",
+        "rippled_version": "1.9.2",
+        "status": "To Be Removed"
+    },
+    {
+        "id": "0285B7E5E08E1A8E4C15636F0591D87F73CB6A7B6452A932AD72BBC8E5D1CBE3",
+        "name": "fixNFTokenDirV1",
+        "rippled_version": "1.9.1",
+        "status": "To Be Removed"
```
oeggert: Same issue as `NonFungibleTokensV1`. I think should be marked as Obosolete/Removed in xyz since it's not up for voting or even showing up in `features.macro`.

@theme/data/obsolete-amendments.json
```diff
@@ -0,0 +1,82 @@
...
+        "name": "PermissionDelegation",
+        "rippled_version": "2.5.0",
+        "status": "Removed in 2.6.1",
+        "href": "https://xrpl.org/blog/2026/rippled-2.6.1"
+    },
+    {
+        "id": "36799EA497B1369B170805C078AEFE6188345F9B3E324C21E9CA3FF574E3C3D6",
+        "name": "fixNFTokenNegOffer",
+        "rippled_version": "1.9.2",
+        "status": "To Be Removed"
```
oeggert: Same issue as `NonFungibleTokensV1`. I think should be marked as Obosolete/Removed in xyz since it's not up for voting or even showing up in `features.macro`.

@theme/data/obsolete-amendments.json
```diff
@@ -0,0 +1,82 @@
...
+        "rippled_version": "2.5.0",
+        "status": "Removed in 3.1.1",
+        "href": "https://xrpl.org/blog/2026/rippled-3.1.1"
+    },
+    {
+        "id": "AE6AB9028EEB7299EBB03C7CBCC3F2A4F5FBE00EA28B8223AA3118A0B436C1C5",
+        "name": "PermissionDelegation",
+        "rippled_version": "2.5.0",
+        "status": "Removed in 2.6.1",
+        "href": "https://xrpl.org/blog/2026/rippled-2.6.1"
```
oeggert:
```suggestion
        "href": "https://xrpl.org/blog/2025/rippled-2.6.1"
```

@theme/components/Amendments.tsx
```diff
@@ -449,6 +536,64 @@ export function AmendmentDisclaimer(props: {
...
+  // For now, shim translate with a placeholder function that does nothing
+  // but matches the syntax of the real translate function, in preparation for
+  // when we can actually use it.
+  // const { useTranslate } = useThemeHooks()
+  // const { translate } = useTranslate()
+  const translate = (key:string, s:string) => s; // Shim version of translate
+
+  let amendment : Amendment | ObsoleteAmendment | any = (
+    isObsolete(name) ||
+    amendmentsSnapshot.amendments.find( (a) => a.name === name )
```
oeggert, line 556:
```suggestion
    amendmentsSnapshot.amendments.find( (a) => a.name === name ) ||
    isObsolete(name)
```
I think we need to switch the order here to match the selection order in the browser version of `amendmentDisclaimer`

oeggert (changes requested):
Mostly looks good, just some minor fixes required.

@theme/data/obsolete-amendments.json
mDuo13: Setting its official status to [Removed in 3.2.0](https://github.com/XRPLF/rippled/pull/6036) along with other NFT amendments.

@theme/components/Amendments.tsx
mDuo13: This ordering is by design, so that our list of obsolete amendments can override incorrect/outdated VHS data. Instead, I've changed the ordering in the web version (in `findAmendment` on lines 110-114)

oeggert (approved):
Great work! LGTM

PR #3778, opened 2026-07-14
Update account references in transaction documentation

lanfeng2:
更新示例地址⁠


mDuo13:
It's not valid to change addresses from an example transaction like this. Not sure if this is an accident or spam, but it's not a useful change.

PR #3781, opened 2026-07-15
Fix NFT flags prefix [en+ja]

### LINKED ISSUE #3433: NFT flags are not lsf
In #3427 I noted that NFToken flags don't use the `lsf` prefix [in the source code](https://github.com/XRPLF/rippled/blob/70d5c624e8cf732a362335642b2f5125ce4b43c1/include/xrpl/protocol/nft.h#L53-L57), since they're not `Flags` in an actual ledger object: NFT flags are incorporated into the `NFTokenID` instead, and they use the prefix `flag` instead (for example, `flagBurnable` not `lsfBurnable`).

The documentation consistently refers to all NFT flags with the `lsf` prefix, but ideally the naming would line up with the naming scheme used in the source code.

mDuo13:
- Fix #3433. Since I don't think changes to the code and libraries are immediately likely, I simply renamed the NFT flags to match their prefix in the code which is `flag` instead of `lsf`.
- The URI field is natively hex, so adjusted example `NFToken` with real data from the sample transaction.
- Removed the "Reserved Flag" from the table, since _any_ flag values that aren't defined should be reserved for future use, and it's weird to call out a specific bit in this way.
- Updated the Japanese translation to add `flagMutable` in addition to reflecting the flag name changes.


docs/references/protocol/data-types/nftoken.md
```diff
@@ -1,69 +1,61 @@
...
-parent: basic-data-types.html
 seo:
     description: Introduction to XRPL NFTs.
 labels:
   - Non-fungible Tokens, NFTs
 ---
 # NFToken
 
-The `NFToken` object represents a single non-fungible token (NFT). It is not stored on its own, but is contained in a [NFTokenPage object][] alongside other `NFToken` objects.
+A `NFToken` object represents a single non-fungible token (NFT) in the XRP Ledger. It is not stored on its own, but is contained in a [NFTokenPage entry][] alongside other `NFToken` objects.
```
rachelflynn: Nit: If read as "en-eff-token," grammatically, this should be "An NFToken...". But the repo mostly writes "a NFToken", so this is consistent with repo conventions.

docs/references/protocol/data-types/nftoken.md
```diff
@@ -1,69 +1,61 @@
...
+| `flagTrustLine`    | `0x0004`   | **DEPRECATED** If enabled, automatically create [trust lines](../../../concepts/tokens/fungible-tokens/index.md) to hold transfer fees. Otherwise, buying or selling this `NFToken` for a fungible token amount fails if the issuer does not have a trust line for that token. The [fixRemoveNFTokenAutoTrustLine amendment][] makes it invalid to enable this flag. |
+| `flagTransferable` | `0x0008`   | If enabled, this `NFToken` can be transferred from one holder to another. Otherwise, it can only be transferred to or from the issuer. |
+| `flagMutable`      | `0x0010`   | If enabled, the `URI` field of the `NFToken` can be updated using a [NFTokenModify transaction][]. |
 
 `NFToken` flags are immutable: they can only be set during the [NFTokenMint transaction][] and cannot be changed later.
 
 ### Example
 
-The example sets three flags: `lsfBurnable` (`0x0001`), `lsfOnlyXRP` (`0x0002`), `lsfTransferable` (`0x0008`). 1+2+8 = 11, or `0x000B` in big endian format.
+The example sets three flags: `flagBurnable` (`0x0001`), `flagOnlyXRP` (`0x0002`), `flagTransferable` (`0x0008`). 1+2+8 = 11, or `0x000B` in big endian format.
```
rachelflynn: Nit: big-endian

rachelflynn (changes requested):
What's the scope of this PR? Some files still contain `lsf` references (e.g., tef codes, NFTokenOffer, Non-Transferable Tokens). 

Otherwise, LGTM aside from merge conflicts.

docs/references/protocol/data-types/nftoken.md
mDuo13: Yeah, that's a remnant of earlier style guidance that tried to convince people "NFT" was pronounced "nifty".

It didn't catch on and we nixed it a while back but not all the existing usage has been updated. I'll update this page but I don't want to expand the scope of this PR to fix all such instances.

mDuo13:
Fixed what I think are the remaining references to the old old flag names (not sure how I missed them the first time) and also addressed the nits.

rachelflynn (changes requested):
Looking good! I found a few remaining instances of `lsfTransferable` in: 
- NFTokenCreateOffer
- tef Codes
- Non-Transferable Tokens
- NFTokenCreateOffer (ja)
- NFTokenOffer (ja)
- Non-Transferable Tokens (ja)

mDuo13:
d'oh! Fixed.

rachelflynn (approved):
LGTM

PR #3806, opened 2026-07-28
Update Batch docs for BatchV1_1

rachelflynn:
Updates the Batch docs for `BatchV1_1` (activating in 3.3.0): points the amendment references at `BatchV1_1`, the bug-fixed replacement for the `Batch` amendment disabled in v3.1.1, and documents the new requirement that the `BatchSigners` array be sorted by `Account` in ascending order and contain no duplicates.


maria-robobug (approved):
LGTM

PR #3810, opened 2026-07-30
Update docs for BatchV1_1

oeggert:
Updates:

- Batch transactions concept
- Batch reference
- Known Amendments
- Transaction Cost concept
- Adds a base fee entry to .env


resources/known-amendments.md
```diff
@@ -99,13 +99,26 @@ For details, see the [XLS-73: AMMClawback specification](https://github.com/XRPL
...
+[BatchV1_1]: #batchv1_1
+
+| Amendment    | BatchV1_1 |
+|:-------------|:----------|
+| Amendment ID | 9F287AED3CDB50A7BD1ACEC24296A30C9B5230CCD136219317AC790E3B884377 |
+| Status       | {% amendment-disclaimer name="BatchV1_1" statusOnly=true /%} |
+| Default Vote (Latest stable release) | No |
+| Pre-amendment functionality retired? | No |
+
+Allows multiple transactions to be bundled into a batch that's processed all together. This amendment replaces [Batch](#batch), fixing a critical bug discovered in the original implementation of the feature. Standard: [XLS-56](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0056-batch)
```
rachelflynn: Nit: Add a period here. Other entries are split, but recent ones include a period.

docs/references/protocol/transactions/types/batch.md
```diff
@@ -168,13 +168,31 @@ Transactions of the `Batch` type support additional values in the [`Flags` field
...
+| `temINVALID_FLAG` | <li>The `Flags` field isn't set to exactly one of the supported [batch modes](#batch-flags).</li><li>One of the inner transactions doesn't have the `tfInnerBatchTxn` flag set.</li> |
+| `temINVALID_INNER_BATCH` | The `RawTransactions` field contains a disabled or invalid `TransactionType`, such as `Batch`. Currently-disabled transactions include: `LoanBrokerCoverClawback`, `LoanBrokerCoverDeposit`, `LoanBrokerCoverWithdraw`, `LoanBrokerDelete`, `LoanBrokerSet`, `LoanDelete`, `LoanManage`, `LoanPay`, `LoanSet`, `VaultCreate`, `VaultSet`, `VaultDelete`, `VaultDeposit`, `VaultWithdraw`, and `VaultClawback`. |
+| `temREDUNDANT` | There is a duplicate transaction in the `RawTransactions` field. |
+| `temSEQ_AND_TICKET` | One of the inner transactions sets both `TicketSequence` and a non-zero `Sequence`, or sets neither. Exactly one is required. |
```
rachelflynn: Nit: "Exactly one" reads weird. Consider "Only one" or something similar.

docs/references/protocol/transactions/types/batch.md
```diff
@@ -168,13 +168,31 @@ Transactions of the `Batch` type support additional values in the [`Flags` field
...
-| `temINVALID_INNER_BATCH`  | - An inner transaction is malformed.<br>- You are attempting to submit a lending or vault-related transaction, which are currently disabled. Invalid transactions include: `LoanBrokerCoverClawback`, `LoanBrokerCoverDeposit`, `LoanBrokerCoverWithdraw`, `LoanBrokerDelete`, `LoanBrokerSet`, `LoanDelete`, `LoanManage`, `LoanPay`, `LoanSet`, `VaultCreate`, `VaultSet`, `VaultDelete`, `VaultDeposit`, `VaultWithdraw`, and `VaultClawback`. |
-| `temSEQ_AND_TICKET`       | The transaction contains both a `TicketSequence` field and a non-zero `Sequence` value. A transaction can't include both fields, but must have at least one. |
+Besides errors that can occur for all transactions, {% $frontmatter.seo.title %} transactions can result in the following [transaction result codes][]:
+
+| Error Code | Description |
+|:-----------|:------------|
+| `temARRAY_EMPTY` | There are fewer than two transactions in the `RawTransactions` field. A batch must contain at least two inner transactions. |
+| `temARRAY_TOO_LARGE` | There are more than 8 entries in the `RawTransactions` or `BatchSigners` field. |
```
rachelflynn: In the [source code](https://github.com/XRPLF/rippled/blob/3.3.0-rc1/include/xrpl/protocol/Protocol.h#L406), it looks like `BatchSigners` has a limit of 24 since `kMaxBatchTxCount` is 8.

```
/**
 * The maximum number of batch signers.
 */
constexpr std::size_t kMaxBatchSigners = kMaxBatchTxCount * 3;
```

docs/references/protocol/transactions/types/batch.md
```diff
@@ -168,13 +168,31 @@ Transactions of the `Batch` type support additional values in the [`Flags` field
...
+| `temINVALID_FLAG` | <li>The `Flags` field isn't set to exactly one of the supported [batch modes](#batch-flags).</li><li>One of the inner transactions doesn't have the `tfInnerBatchTxn` flag set.</li> |
+| `temINVALID_INNER_BATCH` | The `RawTransactions` field contains a disabled or invalid `TransactionType`, such as `Batch`. Currently-disabled transactions include: `LoanBrokerCoverClawback`, `LoanBrokerCoverDeposit`, `LoanBrokerCoverWithdraw`, `LoanBrokerDelete`, `LoanBrokerSet`, `LoanDelete`, `LoanManage`, `LoanPay`, `LoanSet`, `VaultCreate`, `VaultSet`, `VaultDelete`, `VaultDeposit`, `VaultWithdraw`, and `VaultClawback`. |
+| `temREDUNDANT` | There is a duplicate transaction in the `RawTransactions` field. |
+| `temSEQ_AND_TICKET` | One of the inner transactions sets both `TicketSequence` and a non-zero `Sequence`, or sets neither. Exactly one is required. |
+| `tefBAD_AUTH` | <li>The `BatchSigners` field is required but isn't present.</li><li>A `BatchSigners` entry references a pseudo-account, which can't sign batch entries.</li> |
```
rachelflynn: - Other entries are alphabetical, so this one should technically be the first entry.
- The second bullet is accurate but reads as batch-specific. Suggest changing to "...which can't sign any transaction."
- The first bullet looks like it should be `temBAD_SIGNER`. The check at [L488-L495](https://github.com/XRPLF/rippled/blob/3.3.0-rc1/src/libxrpl/tx/transactors/system/Batch.cpp#L488-L495) returns that code and has the comment "Also covers sfBatchSigners being absent while inner txns require signers." `tefBAD_AUTH` doesn't appear anywhere in that file. It also looks like a duplicate of the last bullet in the `temBAD_SIGNER` row, so I think it can just be removed.

.env
```diff
@@ -2,4 +2,5 @@ PUBLIC_GITHUB_FORK=https://github.com/XRPLF/xrpl-dev-portal
 PUBLIC_GITHUB_BRANCH=master
 PUBLIC_OWNER_RESERVE=0.2 XRP
 PUBLIC_BASE_RESERVE=1 XRP
+PUBLIC_BASE_FEE=10 drops
```
rachelflynn, line 5: Optional nit: This is a style preference, but this line doesn't match the others in labeling XRP. Would it be helpful to have 0.00001 XRP instead for consistency?

rachelflynn (changes requested):
Small request from an adjacent PR: Common Links still references the old Batch amendment. Since this PR adds the ### BatchV1_1 heading, could you repoint it to #batchv1_1 in Common Links? That way the anchor and the link land together and we avoid a merge conflict.

docs/references/protocol/transactions/types/batch.md
oeggert: Good catch--updating error message.

.env
oeggert: I don't think we've ever defined at what point we start referring to drops vs fractions of XRP, but saying 10 drops feels more readable than .00001 XRP--also since this is being pulled into pages like transaction-cost, I think it makes sense to use the language already there.

docs/references/protocol/transactions/types/batch.md
```diff
@@ -184,15 +184,15 @@ Besides errors that can occur for all transactions, {% $frontmatter.seo.title %}
 | Error Code | Description |
 |:-----------|:------------|
 | `temARRAY_EMPTY` | There are fewer than two transactions in the `RawTransactions` field. A batch must contain at least two inner transactions. |
-| `temARRAY_TOO_LARGE` | There are more than 8 entries in the `RawTransactions` or `BatchSigners` field. |
+| `temARRAY_TOO_LARGE` | There are more than 8 entries in `RawTransactions`, or more than 24 signatures in `BatchSigners`. |
+| `tefBAD_AUTH` | A `BatchSigners` entry references a pseudo-account, which can't sign any transactions. |
```
rachelflynn: This should be at the top since it's tef, not tem.

rachelflynn (approved):
Minor nit, otherwise LGTM

PR #3785, opened 2026-07-18
Add BatchV1_1, LendingProtocolV1_1, and fixCleanup3_3_0 to Known Amendments

joshuahamsa:
## Summary

Adds the three amendments that are merged on rippled's `develop` branch but not yet in a stable release to the manually-maintained **Amendments in Development** table on the Known Amendments page, each with a detail section:

- **BatchV1_1** ([rippled#6446](https://github.com/XRPLF/rippled/pull/6446)) — the reworked replacement for the original Batch amendment ([XLS-56](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0056-batch)), which was removed in v3.1.1. Also updated the warning in the Batch section to link to the new BatchV1_1 section.
- **LendingProtocolV1_1** ([rippled#6324](https://github.com/XRPLF/rippled/pull/6324)) — revision of the Lending Protocol ([XLS-66](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0066-lending-protocol)).
- **fixCleanup3_3_0** — collection of small fixes targeted at a future release (e.g. [rippled#7685](https://github.com/XRPLF/rippled/pull/7685), [rippled#7666](https://github.com/XRPLF/rippled/pull/7666)).

Amendment IDs computed as SHA-512Half of the amendment name (verified against the published Sponsor ID).

Issue #2666, opened 2024-07-08
Backfill Error Cases for all transaction types


mDuo13:
Some transaction types have their error cases documented, such as [AccountDelete error cases](https://xrpl.org/docs/references/protocol/transactions/types/accountdelete/#error-cases). Other transaction types in the reference don't, and it would be nice to provide that level of detail for all transaction types.

Of course, depending on how complex the transaction type is, it can be pretty difficult to track down all of the possible error codes, which may be buried in different levels or sub-functions of transaction processing. All the more reason that it's worth doing so that users don't have to discover by experimentation what the possible error or make guesses to what they mean.

Also, pay careful attention to the prefix since it's easy to get them mixed up (`terNO_RIPPLE` vs `tecNO_RIPPLE` for example).

PR #3813, opened 2026-08-03
Update Known Amendments for 3.3.0

rachelflynn:
Update known-amendments.md for the 3.3.0 amendments: 
- Adds a fixCleanup3_3_0 entry describing the fixes it includes for Single Asset Vaults, the Lending Protocol, AMMs, the permissioned DEX, Checks, and pseudo-accounts.
- Marks pre-amendment functionality as retired for Clawback, fixDisallowIncomingV1, fixInnerObjTemplate, fixNFTokenReserve, and fixUniversalNumber, which are retired in 3.3.0.
- Restores the Sponsor status so it reads from live amendment data instead of a hardcoded value, matching DynamicMPT and ConfidentialTransfer.


docs/references/protocol/transactions/types/checkcancel.md
```diff
@@ -35,7 +35,7 @@ Cancels an unredeemed [check](../../../../concepts/payment-types/checks.md), rem
 
 ## Error Cases
 
-- If the `CheckID` is an all-zero value, the transaction fails with the result `temMALFORMED`. (Requires the `fixCleanup3_3_0` amendment. Before that amendment is enabled, the transaction instead passes preflight and fails during ledger lookup with `tecNO_ENTRY`.)
+- If the `CheckID` is an all-zero value, the transaction fails with the result `temMALFORMED`. (Requires the {% amendment-disclaimer name="fixCleanup3_3_0" compact=true /%} amendment. Before that amendment is enabled, the transaction instead passes preflight and fails during ledger lookup with `tecNO_ENTRY`.)
```
mDuo13: This type of case is what `mode="updated"` in amendment disclaimers is for.

```suggestion
- If the `CheckID` is an all-zero value, the transaction fails with the result `temMALFORMED`. Previously, the transaction would fail with the result `tecNO_ENTRY`. {% amendment-disclaimer name="fixCleanup3_3_0" mode="updated" /%}
```

Also, the phrase "during ledger lookup" is _so_ precise that it is probably misleading. What that means is the transaction would fail during the part of transaction processing where it attempts to look up a check with the corresponding ID and finds that there is no such ledger entry¹; but people might well think you're talking about a different type of "ledger lookup" such as asking your server to retrieve information about a past ledger.

Also, out of scope for this PR, but I was reminded that "Error Cases" here is not in the preferred table format. The ticket to update the transaction references for consistency is #2666.

¹ Fun cryptography note: as best I can tell, it is not conclusively proven that a check's ID _couldn't_ have legitimately been all zeroes. The SHA-512Half would have to return that value, so this is basically the same challenge as Bitcoin mining except orders of magnitude more difficult. As of early this year, Bitcoin's difficulty was equivalent to producing a SHA-256 hash with about 79 leading zero bits; at the Bitcoin network's massive hashrate, this happens about 150 times a day. Finding a check with a full-zeroes ID means finding the SHA-512 hash has to start with 256 zero bits, and each additional bit doubles the difficulty. So if the entire Bitcoin network's mining power were pointed towards finding a check with an ID of all zeroes, it might find one in about 10^51 years, give or take. (The universe is thought to be on the order of 10^10 years old, so I wouldn't hold my breath.) Unless the SHA-512 hash function is broken, the chances of matching any given ledger entry ID with any specific predetermined value are just as low, but a value of all zeroes specifically is overwhelmingly likely to have been produced by a bug ("oops, forgot to fill that field out when encoding") or someone messing around instead.

docs/references/protocol/transactions/types/checkcash.md
```diff
@@ -43,7 +43,7 @@ The transaction ***must*** include either `Amount` or `DeliverMin`, but not both
 ## Error Cases
 
 - If the sender of the CheckCash transaction is not the `Destination` of the check, the transaction fails with the result code `tecNO_PERMISSION`.
-- If the `CheckID` is an all-zero value, the transaction fails with the result `temMALFORMED`. (Requires the `fixCleanup3_3_0` amendment. Before that amendment is enabled, the transaction instead passes preflight and fails during ledger lookup with `tecNO_ENTRY`.)
+- If the `CheckID` is an all-zero value, the transaction fails with the result `temMALFORMED`. (Requires the {% amendment-disclaimer name="fixCleanup3_3_0" compact=true /%} amendment. Before that amendment is enabled, the transaction instead passes preflight and fails during ledger lookup with `tecNO_ENTRY`.)
```
mDuo13: Similar to CheckCancel.

```suggestion
- If the `CheckID` is an all-zero value, the transaction fails with the result `temMALFORMED`. Previously, the transaction would fail with the result `tecNO_ENTRY`. {% amendment-disclaimer name="fixCleanup3_3_0" mode="updated" /%}
```

resources/known-amendments.md
```diff
@@ -834,6 +834,30 @@ This amendment is a collection of fixes for Single Asset Vaults, the Lending Pro
 - Adds invariant `AccountRootsDeletedClean`, which checks that a deleted account doesn't leave any directly accessible artifacts behind.
 
 
+### fixCleanup3_3_0
```
mDuo13, line 837: Keep an eye out for a conflict with #3785 — even if Git doesn't tag one, we don't want the amendment listed twice on this page.

resources/known-amendments.md
```diff
@@ -834,6 +834,30 @@ This amendment is a collection of fixes for Single Asset Vaults, the Lending Pro
...
+| Default Vote (Latest stable release) | No |
+| Pre-amendment functionality retired? | No |
+
+This amendment is a collection of fixes for Single Asset Vaults, the Lending Protocol, Automated Market Makers, the permissioned DEX, Checks, and pseudo-accounts:
+
+- Unifies freeze and deep freeze checks for transfers to and from pseudo-accounts in the `VaultDeposit`, `VaultWithdraw`, `AMMDeposit`, `AMMWithdraw`, `LoanBrokerCoverDeposit`, and `LoanBrokerCoverWithdraw` transactions.
+- Changes `CheckCash` and `CheckCancel` to reject an all-zero `CheckID` with `temMALFORMED` during preflight instead of `tecNO_ENTRY` during processing.
+- Fixes hybrid offers being removed from the open order book when the account that placed them loses access to the permissioned domain.
+- Fixes Automated Market Maker liquidity being included in quality estimates for permissioned DEX order books.
+- Changes `AMMWithdraw` to return `tecAMM_FAILED` instead of dividing by zero when the withdrawal specifies an `EPrice`.
```
mDuo13: This phrasing makes it seem like specifying _any_ `EPrice` causes a divide-by-zero error, although I am pretty sure that's not the case. Also, what is the result of the divide-by-zero without the amendment—invalid ledger state? Transaction failing with a different code?

mDuo13 (approved):
Looks good, though I think a couple things could be improved.

rachelflynn:
@mDuo13 Addressed all four comments. While verifying against staging/3.3.x-private I also added two missing fixCleanup3_3_0 fixes (lines 859-860), corrected DynamicMPT's now-inverted description (line 386) and the stale fixBatchInnerSigs note (line 779), and merged release-3.3.0 in since this branch was 5 commits behind.

On #3785: keeping this version since master's copy hardcodes In Development and misattributes the AMM/vault-shares fix, which is actually gated on featureSingleAssetVault, so the duplicate is best resolved at the release-3.3.0 → master merge.

mDuo13 (approved):

mainewilliams275-creator (commented):
yea

PR #3818, opened 2026-08-04
Document 3.3.0 sponsorship and pseudo-account behavior

rachelflynn:
Catch-all PR documenting:
- Note that a pseudo-account can't sign, be a delegate, or be a sponsor or sponsee
- Note that some ledger entries can have their owner reserve covered by a sponsoring account
- Add Sponsorship to the deletion blockers table, and note that an account sponsoring reserves can't be deleted
- Add the missing tecPSEUDO_ACCOUNT row
- Correct tecNO_PERMISSION to tecPSEUDO_ACCOUNT (based on staging/3.3.x-private)

Unlinked amendment names in pseudo-accounts.md. LendingProtocol, BatchV1_1, and fixCleanup3_3_0 are plain text because BatchV1_1 needs #3810 and fixCleanup3_3_0 needs #3813 before those known-amendments anchors exist, and LendingProtocol has no common-links definition. Worth linking in a follow-up.


maria-robobug (approved):
LGTM

PR #3825, opened 2026-08-06
Create SECURITY.md for security policy

mainewilliams275-creator:
Added a security policy document outlining supported versions and vulnerability reporting.


oeggert:
Nothing added by this PR. This is a copy of a template with made up versions.

PR #3826, opened 2026-08-06
Add author to account_channels documentation

mainewilliams275-creator:
Added author information to account_channels.md


maria-robobug:
Adding author information on docs pages is not on our radar right now, but we'll certainly revisit if priorities shift.

PR #3847, opened 2026-08-18
Add OpenChainBench latency benchmarks to public-servers page

Flotapponnier:
Adds a **Benchmarks** section at the bottom of the Public Servers page, before the Test Networks section.

**What it does:** [OpenChainBench](https://openchainbench.com/benchmarks/xrp-rpc) probes the public XRPL servers listed on this page (Ripple s1, XRPL Labs/xrplcluster.com, PublicNode) every 60 seconds from US-East, EU-West and Singapore, publishing live p50/p90/p99 latency data.

**Why it fits here:** This page is the reference for XRPL public server endpoints. Developers choosing between servers naturally also want to know which one responds fastest from their region. The Benchmarks section is a direct extension of the server listing — same audience, same decision context.

Open source: https://github.com/ChainBench/OpenChainBench

### DIFF
--- a/docs/tutorials/public-servers.md
+++ b/docs/tutorials/public-servers.md
@@ -24,6 +24,12 @@ If you don't [run your own `xrpld` server](../infrastructure/installation/index.
 | [QuickNode](https://www.quicknode.com/chains/xrpl) | Testnet/Mainnet | N/A | QuickNode provides hosted XRPL RPC mainnet and testnet under their free and paid plans, granting flexible and reliable access to the network.
 
 
+## Benchmarks
+
+| Tool | Notes |
+|:-----|:------|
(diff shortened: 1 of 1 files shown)


Flotapponnier:
Friendly bump. Happy to adjust the wording, section, or the linked page if that helps this land. Thanks for your work on this project!

mDuo13 (commented):
For starters, I'm not sure what the need or use for latency information on public servers is; this data will vary based on where you're querying from, so simply checking the latency of the servers oneself seems like it would be both faster and more individually-relevant than separately looking up OpenChainBench's measurements. If there's a use case here I'm missing, please fill me in.

Also, I took a look at the site and it seems to show contradictory results. The summary says Ripple s1 has the lowest latency but the chart shows XRPL Labs' latency as being half as much, while PublicNode is unresponsive. It seems like you have the InFTF servers attributed to XRPL Labs, and I'm not even sure who PublicNode is (you'll notice they're not listed in this public servers page), while other servers that _are_ on this page don't have data available.

We maintain a high bar for content published on XRPL.org to ensure it is accurate, trustworthy, and relevant for XRPL developers and the broader community. As it stands, this PR doesn't meet that bar.

PR #3859, opened 2026-08-19
Add warning admonition about installing unstable releases

rachelflynn:
The installation docs already cover how to install beta or RC versions addressed in this wiki page: https://github.com/XRPLF/rippled/wiki/Installing-Unstable-Releases. Just added warning admonitions about using these versions in production. 

Part of issue #[3667](https://github.com/XRPLF/xrpl-dev-portal/issues/3667)


amarantha-k (approved):
LGTM

PR #3861, opened 2026-08-19
Update panel stack to have one heading for the section and remove obs…

amarantha-k:
…olete slide heading


zackrdavis:
Non-blocking design issues - I suggest we either diversify the headings or merge this change as-is

Larger gap above than below the heading:
Screenshot 2026-08-21 at 11 03 44 AM

Lost horizontal rules between text in these cards:
Screenshot 2026-08-21 at 11 03 16 AM

PR #3865, opened 2026-08-20
Update first-loss capital info

oeggert:
Adds a couple lines to clarify first-loss capital mechanics. Also fixes an incorrect data type in reference pages and improves the sample values to more realistic ones.


mDuo13 (approved):
I was amused to see that First-Loss Capital first lost its capitals.

PR #3866, opened 2026-08-21
New icon components

zackrdavis:
# Icon set and animation API

- Eighteen icon components behind one barrel at shared/components/Icons. 
-  /icon-demo as the reference page for the set, hidden from sitemap

Screenshot 2026-08-21 at 11 12 11 AM

The two link arrows carry motion. Motion is self-contained to icon-specific .scss and animates within the viewbox, avoiding layout shift.

Containers drive icon animation with the `xrpl-icon-engaged` mixin. Where the trigger is React state rather than a CSS state there is no pseudo-class to hang the mixin off, so the arrows take an optional `engaged` prop.

Fourteen of the icons are Google Material artwork under Apache-2.0. Each file names its exact upstream source and commit, and lists the modifications made to it; the licence entry is in LICENSE at the root.


Calvinjwala (approved):
lgtm

PR #3867, opened 2026-08-21
Document key StatsD metrics on the Configure StatsD page

naavyas:
Adds a 'Key Metrics to Monitor' section to the Configure StatsD page, documenting the key StatsD metrics the core server exports- Ledger, Server State, and Peers. Plus how to list all exported metrics. Previously the page only pointed to the `rippledmon` repository for metric descriptions.

Details:
- **Server State intro:** explains the core server tracks five accounting states (Disconnected, Connected, Syncing, Tracking, Full), while the [xrpld Server States](https://xrpl.org/docs/references/http-websocket-apis/api-conventions/xrpld-server-states) reference lists seven. `validating` and `proposing` aren't tracked separately, so that time is counted under `Full_duration`.
- **Ledger note:** both ledger-age metrics report `1209600` (two weeks) before the server records its first ledger. Excluded that value from alert thresholds.

### DIFF
--- a/docs/infrastructure/configuration/configure-statsd.md
+++ b/docs/infrastructure/configuration/configure-statsd.md
@@ -60,7 +60,55 @@ To enable StatsD on your `xrpld` server, perform the following steps:
 
     You should periodically see messages indicating outbound traffic to the configured address and port of your `rippledmon` instance.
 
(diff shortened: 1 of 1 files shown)


amarantha-k (approved):
lgtm

PR #3868, opened 2026-08-21
Link component design update

alexmahan:
## Summary

Rebuilds the site's anchor link styling from scratch as a new `Link` component (`shared/components/Link/Link.tsx`), matching the BDS Link spec (`link.md` / `link.json` / `link-examples.md`), and rolls it out across the site — both for JSX call sites and for plain markdown/`<a>` content that doesn't use the component directly. The old component (`BdsLink`) is kept intact but unused, under `shared/components/Link/legacy/`.

This branch DOES NOT replace links across rebranded pages, that will be worked on after this branch is merged.

127 0 0 1_4000_link-demo (2)

## The new Link component

`shared/components/Link/Link.tsx` — full docs in [`Link.md`](shared/components/Link/Link.md).

- **Props:** `intention` (`brand` | `neutral`), `context` (`on-theme` | `on-inverse` | `on-saturated`), `variation` (`inline` | `standalone`), `size` (`sm`|`md`|`lg`, standalone only), `iconEnd` (standalone only).
- **No `disabled` state** — the old component had one; the new spec forbids it (a link that can't be followed shouldn't render as a link).
- **Color tokens** are centralized in `styles/_link-tokens.scss` (`xrpl-link-colors()`, `xrpl-link-underline`, `xrpl-link-focus()` mixins), resolved from the spec's color matrix onto existing `_colors.scss` variables — no new hex values introduced. The component's own stylesheet (`_link.scss`) and the plain-`<a>` fallback (below) both consume these same mixins, so a token change only has to happen in one place.
- `neutral`'s hover/active/visited intentionally match its rest color (no distinct treatment), per design direction.

## Styling plain `<a>` tags (not just the component)

Most links on the site are markdown `[text](url)`, not JSX, so two additional mechanisms make sure those get themed without touching every `.md` file:

1. **`@theme/components/MarkdownLink.tsx`** — overrides Redocly's default markdown link renderer to stamp `Link`'s class scheme onto markdown `[text](url)` links (`intention="neutral"`, keeps Redocly's client-side routing).
2. **`styles/_content.scss`**'s `a:not(.xrpl-link)` rule — a CSS-only safety net for raw HTML `<a>` embedded directly in `.md` files and tag components that render their own `<a>` (e.g. `{% child-pages %}`).

Both call the same `_link-tokens.scss` mixins the component uses.

## Why Redocly's own `Link` is still used in places

Redocly ships its own `Link` (`@redocly/theme/components/Link/Link`), which handles client-side routing (SPA navigation without a full page reload) but carries no color/underline/focus styling of its own — call sites are expected to supply a class.

This PR only migrates **prose/content links** — text links embedded in body copy — to the new `Link`. It deliberately leaves Redocly's `Link` in place wherever an anchor is a **structural navigation element** rather than a piece of content: buttons (`btn btn-primary`), cards, quick-reference boxes, in-page TOC/anchor-jump links, and similar. Those already have their own dedicated styling (matching the same precedent as the site's nav/sidebar links, which also stay outside the brand/neutral system), so routing them through the new component would just mean fighting that styling for no visual benefit.

A few files import both: Redocly's `Link` for those structural cases, and this repo's `Link` (aliased `XrplLink` to avoid the name clash) for actual content links on the same page.

The new `Link` component doesn't replace Redocly's routing internals, either — `@theme/components/MarkdownLink.tsx` (the markdown override, see below) still renders through Redocly's `Link` under the hood, just stamped with this component's class scheme, so client-side routing keeps working for markdown links.

## Migration

Swapped real anchor usage over to the new component across ~14 page files — `about/*`, `community/*`, `docs/index`, `docs/use-cases/*`, `resources/*`. Roughly 40 links classified `brand` (marketing/landing-style content) or `neutral` (docs/technical body copy), per explicit page-by-page review. `TileLink.tsx` and `CardIcon.tsx` got import-path fixes after the legacy component moved to `legacy/`.

## `/link-demo` reference page

New page (`link-demo.page.tsx`, `styles/_link-demo.scss`), not linked from nav and excluded from `sitemap.xml` (`excludeFromSearch` in its frontmatter). Shows every combination of size × color group × with/without icon, in both themes, each example labeled with the spec entry it matches — useful for visually diffing against Figma.

## Testing

Manually verified in-browser (light/dark, `/link-demo`, and each migrated page) via computed-style and CSSOM inspection — including forcing real `:hover`/`:focus` state and inspecting matched stylesheet rules for `:visited` (which can't be triggered by tooling directly, since browsers block script access to real visited history). No automated test suite for this area.

Some routes where the new links can be viewed and tested:
```
/link-demo (all link styles)
/about/impact (brand and inline links)
/about/xrp (inline brand links)
/docs/use-cases/payments (inline brand links)
/docs/concepts/networks-and-servers/peer-protocol (inline links)
/docs/tutorials/tokens/fungible-tokens/issue-a-fungible-token (inline links)
Most docs and tutorials pages have the updated inline links
```

Run `npm run contrast-check` to check contrast on the branch. There currently are a few errors unrelated to this link work, here's a screenshot of the test:
Screenshot 2026-08-24 at 11 19 47 AM


alexmahan:
TBD: incorporate new Icons once Zack's icon PR is merged

zackrdavis:
Looks like you need to add `/^xrpl-link--/ ` to the postCSS safelist, prod build is losing some styles (like the gap btw text and icon)
Screenshot 2026-08-21 at 4 20 23 PM

zackrdavis:
Something is off about the demo page's link:visited state. I see all of the `brand` links as black due to the visited state, but the `neutral` links (which are also visited) don't have the off-black color I see in Figma.

The caveat on the demo page says I shouldn't even be able to see :visited states, so that was confusing and we might want a way to "unvisit" if it's easy - maybe a react button that updates all the link destinations. I ended up using an incognito window where the brand links appeared correctly green.

Screenshot 2026-08-21 at 4 42 29 PM

Screenshot 2026-08-21 at 4 36 50 PM

Screenshot 2026-08-21 at 4 38 06 PM

zackrdavis:
We might want this link to be `brand` instead of `neutral`, purple on the brand branch
[/community/developer-funding
](http://localhost:4000/community/developer-funding)Screenshot 2026-08-21 at 5 06 47 PM

Almost definitely want `brand` for these, green on brand branch:
[/docs/usecases/payments](http://localhost:4000/docs/use-cases/payments)
Screenshot 2026-08-21 at 5 19 37 PM

PR #3870, opened 2026-08-24
Button updates

zackrdavis:
## Replace Button with the spec-built implementation

Button is rebuilt from the Figma-derived specification in [`pd-xrpl-developer-docs`](https://github.com/samiamdesigns/pd-xrpl-developer-docs) and swapped in everywhere the old one was used. No page gains or loses a button; all 109 across nine pages render as before, with the geometry and colour differences the spec calls for.

### How it differs

- **You describe the situation, not the look.** Instead of `variant` + `color` + `forceColor`, three axes: whose action it is (`intention`), what it sits on (`context`), how important it is (`emphasis`). The tokens decide the appearance.
- **Invalid combinations don't compile.** The props are a discriminated union, so a combination with no tokens behind it is a type error rather than a plausible-looking mistake.
- **Geometry is fixed** — same padding, 40px minimum target, and a border on every emphasis, so a button doesn't resize when its emphasis changes. `forceNoPadding` is gone; containers compensate.
- **`loading` / `inactive` / `disabled` are behaviours, not looks.** They resolve to identical colours in places but differ in ARIA and tab order.
- **Colours bind to palette variables**, so palette changes propagate on their own.

### Where to test:
- Redesigned pages
  - /
  - /docs/
  - /develop/
  - /resources/
  - /community/
  - /community/developer-funding
  - /docs/use-cases/payments/
  - /docs/use-cases/trading/
  - /docs/use-cases/tokenization/
- non-redesigned pages
  - JSX pages:
    - /blog
    - /resources/dev-tools
    - /about
    - /about/history
    - /about/impact
    - /abut/uses
    - /docs/agentic-transactions
    - /community/events
  - Docs pages with `interactive-block`, ex:
    - /docs/tutorials/payments/send-xrp
    - /docs/tutorials/tokens/fungible-tokens/issue-a-fungible-token
  - Docs pages with "Try it!" button links:
    - /docs/references/http-websocket-apis/public-api-methods/clio-methods/server_info-clio
    - /docs/references/http-websocket-apis/public-api-methods/transaction-methods/submit

### Also here

- **New page `/button-demo`** — every combination and state on all three surfaces, unlinked and out of search. The content pages exercise almost none of the component: one `standard` button between them, and no `loading`, `inactive` or `disabled` at all.

Screenshot 2026-08-21 at 3 35 20 PM

### Not fixed here

Complex ui-integrated buttons are not replaced, as it would break the UI. Generally these are square buttons that match other nearby UI:
Screenshot 2026-08-24 at 1 58 35 PM
Screenshot 2026-08-24 at 2 01 20 PM


styles/_pages.scss
```diff
@@ -1509,48 +1509,22 @@ html.light {
       .button-container {
         display: flex;
         flex-wrap: wrap;
-        gap: 32px;
+        gap: 16px;
```
alexmahan: Should this pixel value be hard-coded or is there a spacing token that can be used here?

shared/components/Button/button-vs-link-candidates.md
```diff
@@ -0,0 +1,65 @@
+# Places that should probably be a Link, not a Button
```
alexmahan: 👏

alexmahan (commented):
One small non-blocking comment. Looking good!

amarantha-k:
@zackrdavis Testing the /blog page on a local build, I noticed that the updated button is applied to the page banner, but the "Read more" buttons on individual cards still use the old design. This results in an inconsistent UI across the page.  
Screenshot 2026-08-24 at 17 23 32

zackrdavis:
> * /docs/references/http-websocket-apis/public-api-methods/transaction-methods/submit

@amarantha-k I've now implemented standalone links in those places per Sam's replacement map

amarantha-k:
@zackrdavis Thanks for the updates — the labels look good now.
One thing I noticed: in light mode, the card borders aren’t visible, while they display correctly in dark mode. Could you take a look?

Screenshot 2026-08-25 at 09 50 02
Screenshot 2026-08-25 at 09 49 48

zackrdavis:
> @zackrdavis Thanks for the updates — the labels look good now. One thing I noticed: in light mode, the card borders aren’t visible, while they display correctly in dark mode. Could you take a look?
> 

@amarantha-k This was the existing style on xrpl-brand-update-2026 when we began working on it. Noting this for phase 2!

PR #3886, opened 2026-08-27
Add xrpl-trading skill and OWS signing path

amarantha-k:


.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,464 @@
...
+
+### Option B: OWS MCP/REST server
+
+```bash
+# Start the OWS MCP server (built into OWS)
+ows mcp
+
+# Or REST server
+ows rest --port 8080
+```
```
oeggert, line 361: I checked the `ows` repo and I don't see an `mcp` or `rest` cli command.

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,464 @@
...
+| :---- | :---- | :---- |
+| Auth | Vault passphrase | API key (token) |
+| Policy binding | All registered policies | Policies scoped to the API key |
+| Key exposure | Passphrase in agent process | Agent never sees passphrase or key |
+| Setup complexity | Low | Requires running OWS MCP/REST server |
+| Recommended for | Development, trusted processes | Production agents |
+
+**Open question (Starter Kit v3):** Whether to ship the OWS MCP server as part
+of the Starter Kit deployment is an outstanding decision. The MCP server
+(`ows mcp`) is built into OWS. This decision gates v3 architecture.
```
oeggert, line 416: Should remove the MCP / REST column. I don't see either implemented in the `ows` repo.

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,707 @@
...
+    account="rYourAddress",
+    taker_pays={ ... },
+    taker_gets=xrp_to_drops(100),
+    memos=[
+        MemoWrapper(memo=Memo(
+            memo_type=str_to_hex("agent/trade"),
+            memo_data=str_to_hex('{"skill":"xrpl-trading","reason":"user-requested"}'),
+        ))
+    ],
+)
```
oeggert, line 575:
```suggestion
from xrpl.utils import str_to_hex
from xrpl.models.transactions import Memo

offer = OfferCreate(
    account="rYourAddress",
    taker_pays={ ... },
    taker_gets=xrp_to_drops(100),
    memos=[
        Memo(
            memo_type=str_to_hex("agent/trade"),
            memo_data=str_to_hex('{"skill":"xrpl-trading","reason":"user-requested"}'),
        )
    ],
)
```
No `MemoWrapper` exists in xrpl-py.

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,464 @@
...
+- **Payments skill:** trust line existence, destination tag requirements, reserve checks
+
+These are application-layer checks — not ledger-enforced. They only apply when
+the skill is loaded. OWS v3 will add equivalent enforcement at the vault layer.
+
+---
+
+## 7. Python Integration
+
+The OWS SDK is Node.js only. Python agents have two options:
```
oeggert, line 323: OWS is also available in Python. This whole step can probably be removed.

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,707 @@
...
+
+## 7. Simulate before handoff
+
+Use `simulate` on the raw transaction object (before `autofill`) to catch errors
+without spending fees or triggering the signing ceremony. This is especially
+important for new currency pairs, first-time trust line interactions, or large
+offers.
+
+```python
+from xrpl.models.requests import SimulateTransaction  # xrpl-py ≥ 4.x
```
oeggert, line 616:
```suggestion
from xrpl.models.requests import Simulate
```
`SimulateTransaction` is an incorrect import

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,707 @@
...
+
+# Build raw tx (no Fee/Sequence)
+raw_offer = OfferCreate(
+    account="rYourAddress",
+    taker_pays={ ... },
+    taker_gets=xrp_to_drops(100),
+)
+
+sim_result = client.request(SimulateTransaction(tx_blob=raw_offer.to_xrpl()))
```
oeggert, line 625:
```suggestion
sim_result = client.request(Simulate(transaction=raw_offer))
```

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,707 @@
...
+        client.request(asks_req),
+        client.request(bids_req),
+    )
+    return asks.result["offers"], bids.result["offers"]
+```
+
+### 3.3 Computing mid price and slippage
+
+```typescript
+import { dropsToXrp } from "xrpl";
```
oeggert, line 357: `dropsToXrp` imported but not used.

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,707 @@
...
+flags=OfferCreateFlag.TF_FILL_OR_KILL,
+```
+
+**Never combine `tfImmediateOrCancel` and `tfFillOrKill`.** Reject at construction.
+
+### 1.5 Post-only (passive)
+
+`tfPassive`: posts to the order book without crossing against existing offers,
+even if a crossing opportunity exists at the time of submission. Use for maker-only
+strategies.
```
oeggert, line 160:
```suggestion
`tfPassive`: does not consume offers whose rate exactly matches this offer's
limit rate. It will consume offers priced better than the limit rate, so the
offer can execute immediately on submission. Use it to add liquidity at a specific
rate without removing existing liquidity at that same rate.
```

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
oeggert, line 1: This is missing `python` samples.

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,464 @@
...
+
+Then call the signing endpoint from Python via `httpx` or `requests`. See the
+OWS documentation at https://docs.openwallet.sh for the API spec.
+
+---
+
+## 8. Migrating from Env-Var to OWS
+
+Migrating from Pattern 1 (env-var) to Pattern 3 (OWS) does **not** require
+generating a new XRPL wallet. Import the existing seed:
```
oeggert, line 371: Add a note _not_ to use Ed25519 seeds; only use secp256k1.

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,464 @@
...
+  const signedBlob = encode(signedTx as Record<string, unknown>);
+
+  // 6. Submit
+  return client.submitAndWait(signedBlob);
+}
+```
+
+---
+
+## 4. Policy Setup
```
oeggert, line 233: This whole section seems incorrect. There are no executable files, and instead it uses a `rules` array with `allowed_chains` and `expires_at`.

I found a better sample on the repo: https://github.com/open-wallet-standard/core/blob/main/examples/agent-with-policy.js

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,464 @@
...
+  0,             // derivation index (default: 0)
+);
+
+// Sign and broadcast in one call (skips Wallet skill submission step).
+// Prefer signTransaction + manual submitAndWait in agentic flows to preserve
+// the Wallet skill's hash-persistence and error-handling discipline.
+const sent = signAndSend("xrpl-agent", "xrpl", txHex, passphrase, 0, "wss://xrplcluster.com");
+```
+
+### Policy management
```
oeggert, line 142: This is formatted incorrectly. No exe is referenced by `ows`. The sample at this link is the correct shape: https://github.com/open-wallet-standard/core/blob/main/examples/agent-with-policy.js

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,707 @@
...
+import { dropsToXrp } from "xrpl";
+
+// Each offer in book_offers has `quality` = taker_pays / taker_gets (as decimal string)
+const askOffers = asks.result.offers;
+const bidOffers = bids.result.offers;
+
+const bestAsk = askOffers.length ? parseFloat(askOffers[0].quality) : Infinity;
+const bestBid = bidOffers.length  ? parseFloat(bidOffers[0].quality)  : 0;
+const midPrice = (bestAsk + bestBid) / 2;
+const spreadPct = midPrice > 0 ? ((bestAsk - bestBid) / midPrice) * 100 : null;
```
oeggert, line 366: I don't think we can do these calculations using only `quality`. When doing comparisons for `midPrice` and `spreadPct`, bid and ask numbers have to be expressed in the same units, and since `quality` changes depending on which side of the trade you're looking from--it would produce a meaningless result.

Ex)
- Buyer wants to pay 10 USD for 5 XRP; quality is 0.5 = 2 XRP/USD
- Seller wants to pay 6 XRP for 10 USD; quality is 1.66 = 1.66 XRP/USD

So the `midPrice` and `spreadPct` can't use the `quality` value as is, and has to normalize the values first.

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,707 @@
...
+## 8. Error codes
+
+| Engine result | Fee charged? | Meaning | Resolution |
+| :---- | :---- | :---- | :---- |
+| `tesSUCCESS` | Yes | Transaction accepted and applied | Parse fill status from `meta.AffectedNodes` |
+| `tecUNFUNDED_OFFER` | Yes | Account XRP balance insufficient to fund the offer | Check balance. Account needs XRP ≥ offer value + fee + reserve buffer. |
+| `tecEXPIRED` | Yes | `Expiration` already passed when ledger closed | Reject at construction. If boundary race, re-offer with future expiry. |
+| `tecKILLED` | Yes | `tfFillOrKill` offer could not fill completely | Do not retry. Ask user to retry with `tfImmediateOrCancel` or adjusted price. |
+| `tecNO_LINE` | Yes | No trust line for an IOU in the offer | User must establish trust line (`TrustSet`) before retrying. |
+| `tecINSUF_RESERVE_OFFER` | Yes | Insufficient XRP reserve to create new offer object | Each resting offer requires 2 XRP owner reserve. Cancel existing offers or fund account. |
```
oeggert, line 652:
```suggestion
| `tecINSUF_RESERVE_OFFER` | Yes | Insufficient XRP reserve to create new offer object | Each resting offer requires 0.2 XRP owner reserve. Cancel existing offers or fund account. |
```

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,707 @@
...
+    # No flags: remainder rests on the book if not immediately crossed
+)
+# → hand to XRPL Agent Wallet skill
+```
+
+---
+
+## 5. AMM interaction
+
+XRPL AMM pools (introduced in the Clawback amendment) are integrated directly
```
oeggert, line 471:
```suggestion
XRPL AMM pools (introduced in the AMM amendment) are integrated directly
```

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,464 @@
...
+  "xrpl-agent",  // wallet name or ID
+  "xrpl",        // chain identifier
+  txHex,         // XRPL canonical binary hex (from xrpl.js encode())
+  passphrase,    // vault passphrase — load from env, never hardcode
+  0,             // derivation index (default: 0)
+);
+
+// Sign and broadcast in one call (skips Wallet skill submission step).
+// Prefer signTransaction + manual submitAndWait in agentic flows to preserve
+// the Wallet skill's hash-persistence and error-handling discipline.
```
oeggert, line 138: signAndSend() produces a submittable tx blob, however signTransaction() doesn't because it omits the `SigningPubKey` field.

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,464 @@
...
+```
+
+### Signing
+
+```typescript
+import { signTransaction, signAndSend } from "@open-wallet-standard/core";
+
+// Sign only — returns raw DER signature.
+// Caller applies TxnSignature and re-encodes (see §3).
+const { signature }: SignResult = signTransaction(
```
oeggert, line 128: Do we need to include `SignResult` type annotation? I think `signTransaction` is already declared as a `SignResult`.

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,464 @@
...
+);
+// key.token is shown once — save to your secrets manager immediately.
+```
+
+---
+
+## 3. XRPL Signing Flow
+
+OWS returns the raw DER-encoded secp256k1 signature, not a complete signed blob.
+The Wallet skill assembles the signed transaction:
```
oeggert, line 177: I don't think this is correct. The above code sample used `signTransaction` which doesn't return the `SigningPubKey`, and I don't think the following steps adds it back in for submission.

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,464 @@
...
+  "xrpl",        // chain identifier
+  txHex,         // XRPL canonical binary hex (from xrpl.js encode())
+  passphrase,    // vault passphrase — load from env, never hardcode
+  0,             // derivation index (default: 0)
+);
+
+// Sign and broadcast in one call (skips Wallet skill submission step).
+// Prefer signTransaction + manual submitAndWait in agentic flows to preserve
+// the Wallet skill's hash-persistence and error-handling discipline.
+const sent = signAndSend("xrpl-agent", "xrpl", txHex, passphrase, 0, "wss://xrplcluster.com");
```
oeggert, line 139: I don't think ows supports wss.

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,464 @@
...
+    ▼
+OWS signTransaction("xrpl", txHex)
+    │
+    ├── Policy Engine  ← registered policy executables
+    │       │
+    │       ├── PASS → decrypt key → sign → wipe key → return { signature }
+    │       └── FAIL → throw (key never decrypted)
+    │
+    ▼
+Wallet skill applies TxnSignature → encodes signed tx → submitAndWait
```
oeggert, line 57: This line should flow after the Policy Engine -> PASS step, right?

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,464 @@
...
+### Signing
+
+```typescript
+import { signTransaction, signAndSend } from "@open-wallet-standard/core";
+
+// Sign only — returns raw DER signature.
+// Caller applies TxnSignature and re-encodes (see §3).
+const { signature }: SignResult = signTransaction(
+  "xrpl-agent",  // wallet name or ID
+  "xrpl",        // chain identifier
```
oeggert, line 130: Shouldn't this and all subsequent network calls be `"xrpl-testnet"`? This resolves to Mainnet

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,707 @@
...
+const bestBid = bidOffers.length  ? parseFloat(bidOffers[0].quality)  : 0;
+const midPrice = (bestAsk + bestBid) / 2;
+const spreadPct = midPrice > 0 ? ((bestAsk - bestBid) / midPrice) * 100 : null;
+```
+
+```python
+# Each offer in book_offers has "quality" = taker_pays / taker_gets (decimal string)
+def compute_mid_spread(ask_offers: list, bid_offers: list):
+    best_ask = float(ask_offers[0]["quality"]) if ask_offers else float("inf")
+    best_bid = float(bid_offers[0]["quality"]) if bid_offers else 0.0
```
oeggert, line 373: `ask_offers` and `bid_offers` aren't declared in 3.2 step.

.claude/skills/xrpl-skills/xrpl-trading/SKILL.md
```diff
@@ -0,0 +1,290 @@
...
+To receive:   [TakerPays amount and currency]
+Limit price:  [TakerPays / TakerGets, expressed as base/quote rate]
+Mid price:    [computed from order book]
+Est. fill:    [% of order expected to cross immediately]
+Slippage:     [execution price vs mid, as ±% and absolute amount]
+Expiry:       [human-readable datetime, or "none"]
+Network fee:  [estimated drops]
+```
+
+**Step 4 — Run built-in guardrails and hand off to Wallet skill**
```
oeggert: Shouldn't these guardrails be applied before step 3? It would be a better flow that a user can only acknowledge trade parameters once it's confirmed by the guardrails it would work, rather than them approving it and then potentially failing anyway.

.claude/skills/xrpl-skills/xrpl-trading/SKILL.md
```diff
@@ -0,0 +1,290 @@
...
+- **Transaction submission:** Handled entirely by the XRPL Agent Wallet skill. This skill builds transaction objects; it does not call `submit_and_wait` or `submitAndWait` directly.
+- **Signing path:** Determined by the XRPL Agent Wallet skill configuration — env-var (development), external signer (HSM/KMS), or OWS (Open Wallet Standard). The Trading skill does not sign directly; all signing is delegated to the Wallet skill. See [xrpl-agent-wallet](/resources/dev-tools/ai-tools) for setup.
+- **Amount handling:** XRP amounts are always strings in drops — use `xrp_to_drops()` / `xrpToDrops()`. Never pass floats or raw XRP values. IOU amounts use `{currency, issuer, value}` objects with `value` as a decimal string.
+- **Source tag:** The XRPL Agent Wallet skill (OWS) automatically applies `SourceTag = 20260530` to every transaction that passes through the signing ceremony. Override by setting `SourceTag` on the transaction object before handoff; the Wallet skill respects any value already present. Do not omit `SourceTag` — it is required for on-chain attribution and usage tracking.
```
oeggert, line 66: `ows.md` specifically says it never enforces a source tag. Mabe drop the `(OWS)`?

.claude/skills/xrpl-skills/xrpl-trading/SKILL.md
```diff
@@ -0,0 +1,290 @@
...
+
+| Engine result | Fee charged? | Meaning | Agent action |
+| :---- | :---- | :---- | :---- |
+| `tesSUCCESS` | Yes | Transaction accepted | Parse fill status per Flow 1 Step 5 |
+| `tecUNFUNDED_OFFER` | Yes | Account balance insufficient to fund the offer | Surface balance vs required. Do not retry automatically. |
+| `tecEXPIRED` | Yes | `Expiration` already passed when processed | Reject at construction. If timing causes slip-through, inform user and request new expiry. |
+| `tecKILLED` | Yes | `tfFillOrKill` offer could not be fully filled | Inform user the order was not executed. Ask whether to retry with different flag or price. Do not retry automatically. |
+| `tecNO_LINE` | Yes | No trust line for an IOU in the offer | Surface issuer and currency. Prompt user to establish trust line before retrying. |
+| `tecINSUF_RESERVE_OFFER` | Yes | Account reserve too low to create a new offer object | Explain reserve requirements. Do not retry. |
+| `temBAD_OFFER` | No | Malformed transaction — invalid amounts, missing fields, zero amounts | Log raw error. Surface for debugging. Do not retry without fixing the transaction. |
```
oeggert, line 220: This is missing a handful of errors from the OfferCreate/OfferCancel page.

.claude/skills/xrpl-skills/xrpl-trading/SKILL.md
```diff
@@ -0,0 +1,290 @@
...
+| `SourceTag` | Applied by Wallet skill | uint32 | OWS Wallet skill applies `20260530` automatically. Override by setting before handoff. |
+| `Expiration` | Optional | uint32 | **XRPL epoch seconds** — not Unix seconds. `XRPL_epoch = Unix_s − 946,684,800`. Offer auto-cancels on first ledger close after this time. |
+| `OfferSequence` | Optional | uint32 | If set, cancel this existing offer atomically when this OfferCreate is processed. |
+| `Flags` | Optional | uint32 | Bitfield. See flags table below. |
+
+**OfferCreate flags:**
+
+| Flag | Hex | Effect |
+| :---- | :---- | :---- |
+| `tfPassive` | `0x00010000` | Do not consume matching offers. Post to book only. Use for post-only orders. Never set by default. |
```
oeggert, line 248: This isn't a book-only flag. It only skips over matching orders, but does consume better-priced orders.

oeggert (changes requested):
There's still some polishing to be done on these skills, mainly in `ows.md` and `trading.md`. This was a pretty large pass and I tried to address the errors I ran into while running the skill, but I'm sure there will be some new ones that arise after addressing this first round of comments.

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
amarantha-k: This line depicts the next step in the flow, which only proceeds if the Policy Engine step passes, and a failure will throw an error.

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,708 @@
...
+```typescript
+Flags: OfferCreateFlags.tfPassive,
+```
+
+```python
+flags=OfferCreateFlag.TF_PASSIVE,
+```
+
+**Never set `tfPassive` by default.** Only set when the user explicitly requests
+"post only" or "maker only" behaviour.
```
oeggert: I wonder if it's worth nothing that the passive behavior slightly deviates from post-only behavior a market maker expects. My understanding is that a true post-only will outright cancel if posting the trade will also execute it (regardless if the reason is because it matches or is better)--the intent is still to add liquidity, but with passive it **will** execute a trade if it's priced better than the limit rate.

.claude/skills/xrpl-skills/xrpl-trading/SKILL.md
```diff
@@ -0,0 +1,290 @@
...
+Stop and surface the specific error if any check fails. Do not proceed to the Wallet skill until all checks pass.
+
+Construct the `OfferCreate` transaction object with all required fields (see Field Reference). Do not set `Fee`, `Sequence`, or `LastLedgerSequence`. Hand the transaction object to the **XRPL Agent Wallet skill**. The Wallet skill will autofill, optionally evaluate OWS policies (time-window, network restrictions), sign, and submit. If the Wallet skill rejects, surface the rejection reason — do not retry without user instruction.
+
+**Step 5 — Parse result**
+
+Check `engine_result`. Any result other than `tesSUCCESS` is a failure. Inspect `meta.AffectedNodes`:
+
+- **Fully filled:** No `CreatedNode` with `LedgerEntryType: "Offer"` matching the signing account and transaction sequence.
+- **Resting / partially filled:** `CreatedNode` exists. Extract `NewFields.TakerPays` and `NewFields.TakerGets` for remaining amounts. Compare against original amounts to distinguish partial from unfilled.
```
oeggert: I believe that `ImmediateOrCancel` never creates a node. It can partially fill an order, be treated as successful, but won't leave a node entry to determine how much was filled.

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,708 @@
...
+| `tecKILLED` | Yes | `tfFillOrKill` offer could not fill completely | Do not retry. Ask user to retry with `tfImmediateOrCancel` or adjusted price. |
+| `tecNO_LINE` | Yes | No trust line for an IOU in the offer | User must establish trust line (`TrustSet`) before retrying. |
+| `tecINSUF_RESERVE_OFFER` | Yes | Insufficient XRP reserve to create new offer object | Each resting offer requires 0.2 XRP owner reserve. Cancel existing offers or fund account. |
+| `tecDIR_FULL` | Yes | Offer directory is full (too many offers from this account) | Cancel some existing offers before creating new ones. |
+| `temBAD_OFFER` | No | Malformed transaction (zero amounts, invalid fields, bad flags) | Fix construction and re-simulate. |
+| `temBAD_EXPIRATION` | No | `Expiration` field value is invalid | Recompute using `XRPL_epoch = Unix_s − 946,684,800`. |
+| `temINVALID_FLAG` | No | Invalid flag combination (`tfImmediateOrCancel` + `tfFillOrKill`) | Reject at construction. |
+| `temREDUNDANT` | No | `TakerPays` == `TakerGets` after quality normalization | Adjust amounts. |
```
oeggert:
```suggestion
| `temREDUNDANT` | No | The transaction would trade a token for the same token (same issuer and currency code). | Adjust token. |
```

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,708 @@
...
+transaction and simulate again before handing to the Wallet skill.
+
+---
+
+## 8. Error codes
+
+| Engine result | Fee charged? | Meaning | Resolution |
+| :---- | :---- | :---- | :---- |
+| `tesSUCCESS` | Yes | Transaction accepted and applied | Parse fill status from `meta.AffectedNodes` |
+| `tecUNFUNDED_OFFER` | Yes | Account XRP balance insufficient to fund the offer | Check balance. Account needs XRP ≥ offer value + fee + reserve buffer. |
```
oeggert:
```suggestion
| `tecUNFUNDED_OFFER` | Yes | Account XRP or IOU balance insufficient to fund the offer | Check balance. Account needs XRP ≥ offer value + fee + reserve buffer, or IOU ≥ offer value. |
```

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,708 @@
...
+fills asynchronously.
+
+```typescript
+await client.request({
+  command: "subscribe",
+  accounts: ["rYourAddress"],
+});
+
+client.on("transaction", (tx) => {
+  if (tx.transaction.TransactionType === "OfferCreate") {
```
oeggert:
```suggestion
  if (tx.tx_json.TransactionType === "OfferCreate") {
```
`xrpl.js` defaults to api v2, which won't return `tx.transaction`, use `tx.json`. Also this should monitor `Payment` I believe, since cross-currency payments can consume offers.

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,708 @@
...
+        limit=20,
+    )
+    asks, bids = await asyncio.gather(
+        client.request(asks_req),
+        client.request(bids_req),
+    )
+    return asks.result["offers"], bids.result["offers"]
+```
+
+### 3.3 Computing mid price and slippage
```
oeggert, line 416: The trading `SKILL.md` mentions considering slippage several times, but it's not actually demonstrated in this section how to calculate.

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,708 @@
...
+offer = OfferCreate(
+    account="rYourAddress",
+    taker_pays={
+        "currency": "USD",
+        "issuer": "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B",
+        "value": "50",
+    },
+    taker_gets=xrp_to_drops(100),   # "100000000"
+    # Fee, Sequence, LastLedgerSequence: set by Wallet skill autofill
+    # SourceTag: applied by Wallet skill (OWS) automatically
```
oeggert: Agent wallet skill applies source tag.

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,708 @@
...
+| `AMMBid` | Bid for the auction slot (reduces trading fees for the winner) |
+| `AMMVote` | Vote on the AMM trading fee (weighted by LP token holdings) |
+
+---
+
+## 6. Agentic patterns
+
+### 6.1 Source tag
+
+The XRPL Agent Wallet skill (OWS) applies `SourceTag = 20260530` to every
```
oeggert: `ows.md` specifically says the agent wallet skill apples it, but this line conflates the two.

.claude/skills/xrpl-skills/xrpl-trading/SKILL.md
```diff
@@ -0,0 +1,290 @@
...
+| Field | Required | Type | Notes |
+| :---- | :---- | :---- | :---- |
+| `TransactionType` | Yes | string | `"OfferCreate"` |
+| `Account` | Yes | string | Signing account's XRPL address |
+| `TakerPays` | Yes | Amount | What the creator wants to receive. XRP as drops string; IOU as `{currency, issuer, value}`. |
+| `TakerGets` | Yes | Amount | What the creator is offering. Same format. |
+| `Fee` | Autofilled | string (drops) | Set by Wallet skill autofill. Do not set manually. |
+| `Sequence` | Autofilled | uint32 | Set by Wallet skill autofill. The offer's identifier if it rests on the book. |
+| `LastLedgerSequence` | Autofilled | uint32 | Set by Wallet skill autofill. Transaction invalid after this ledger index. |
+| `SourceTag` | Applied by Wallet skill | uint32 | OWS Wallet skill applies `20260530` automatically. Override by setting before handoff. |
```
oeggert, line 308: I think there's some confusion on responsibilities here. There are two distinct files, one is `ows.md` and the other is `xrpl-agent-wallet-skill.md`. We should decide which one specifically owns the sourcetag work (agent wallet skill makes the most sense to me), and normalize across the rest of these skill files.

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,636 @@
...
+- **Payments skill:** trust line existence, destination tag requirements, reserve checks
+
+These are application-layer checks — not ledger-enforced. They only apply when
+the skill is loaded. OWS v3 will add equivalent enforcement at the vault layer.
+
+---
+
+## 7. Python Integration
+
+The OWS SDK is Node.js only, so Python agents shell out to the `ows` CLI.
```
oeggert: I'm still unsure why we're using the CLI here. The Python library does exist and it seems to have the same functionality as TS?

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,636 @@
...
+  vaultPath,      // optional; defaults to ~/.ows
+);
+
+// Sign an arbitrary 32-byte hash. Used to recover the account public key (§3).
+const { signature } = signHash("xrpl-agent", "xrpl", hashHex, credential, 0, vaultPath);
+```
+
+`signAndSend` exists but has no XRPL transport — use `signTransaction` plus
+`client.submitAndWait()` so the Wallet skill keeps its hash-persistence and
+error-handling discipline.
```
oeggert: Maybe tweak the language here, since `signAndSend` does sign and submit to the ledger--the issue is it doesn't poll for a validated response, it just returns the preliminary response.

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
oeggert: Right, but the way I'm reading the diagram looks like the agent wallet skill is an entirely separate fork from the Policy Engine that happens regardless of the policy engine result.

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
```diff
@@ -0,0 +1,636 @@
...
+
+```typescript
+import { signTransaction, signHash } from "@open-wallet-standard/core";
+
+// Sign a transaction. The tx hex MUST NOT contain SigningPubKey — OWS supplies
+// it internally when it builds the signing payload. Passing one throws
+// "unsigned transaction must not contain SigningPubKey".
+const { signature } = signTransaction(
+  "xrpl-agent",   // wallet name or ID
+  "xrpl",         // chain identifier
```
oeggert: Might be worth noting this resolves to mainnet, and to use `xrpl-testnet` if testing trades.

oeggert (changes requested):
I found a few more errors this time around--some technical and a few on clarity. Also some of the issues from the last pass still remain.

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,759 @@
...
+
+> **Never use `quality` as a price.** (The previous version of this section
+> imported `dropsToXrp` and then never called it — that unused import was the
+> tell: the drops conversion is mandatory, not optional.)
+> `quality` is `TakerPays / TakerGets` in
+> *protocol* units, so XRP appears in **drops** — an XRP-denominated `quality` is
+> off by 1,000,000 — and the two sides of a book are quoted in inverse
+> orientations. Feeding `quality` straight into a mid-price calculation produces
+> prices ~10^6 too large and negative spreads. Always derive price from the
+> amounts, converting drops to XRP first.
```
oeggert:
```suggestion
Each order book reports the quality of an offer as `TakerPays / TakerGets`. Never compare the two sides of a market using their quality values as-is. `TakerPays` and `TakerGets` swap places depending on which book you read, so the two values are reciprocals of each other.

Before calculating mid price or spread, decide on a base asset, then invert one book so both bid and ask are expressed in the same base unit.
```
Current description emphasizes the wrong thing. Drops was a symptom of the old formula, where the 1,000,000 unit conversion made the issue more apparent. We should keep the `dropsToXRP` function since it makes values more readable, but instead emphasize inverting one of the books.

oeggert (changes requested):
Almost good, just a handful of issues to resolve.

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,759 @@
...
+});
+const target = result.result.offers.find(o => o.seq === targetSequence);
+if (!target) { /* inform user — offer not found */ }
+```
+
+---
+
+## 3. Order book reads
+
+### 3.1 `book_offers` — asks side (offers to sell base)
```
oeggert, line 349: Just discovered a bug in the docs, which impacts this part of the skill. `book_offers` can return completely unfunded offers. Both partially and completely unfunded offers will show up with an additional `taker_gets_funded` and `taker_pays_unfunded` field. These show up alongside the regular `taker_gets` and `taker_pays` fields, so this skill will get incorrect calculations looking at only the latter.

oeggert (commented):
Found another bug--critical for `book_offers`.

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
amarantha-k: Python to be added in a separate PR.

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
amarantha-k: Added

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
amarantha-k: Added note

.claude/skills/xrpl-skills/xrpl-agent-wallet/references/ows.md
amarantha-k: Removing this section for now; Python SDK to be added in a separate PR.

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,924 @@
...
+});
+const target = result.result.offers.find(o => o.seq === targetSequence);
+if (!target) { /* inform user — offer not found */ }
+```
+
+---
+
+## 3. Order book reads
+
+### 3.0 Funding: `book_offers` advertises more than it can deliver
```
oeggert:
```suggestion
### 3.0 Funding: `book_offers` can advertise more than it can deliver
```

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,924 @@
...
+
+**`book_offers` returns offers the owner cannot currently fund, and they look
+identical to real ones unless you check.** An offer's `TakerGets`/`TakerPays`
+are the amounts it was *created* with. What it can actually deliver right now
+depends on the owner's present balance.
```
oeggert:
```suggestion
**`book_offers` can return offers the owner cannot currently fund, and they look
identical to real ones unless you check.** An offer's `TakerGets`/`TakerPays`
are the amounts it was *created* with. What it can actually deliver depends
on the owner's present balance.
```

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,924 @@
...
+
+When an offer is not fully funded, rippled adds two fields alongside the
```
oeggert:
```suggestion
When an offer is not fully funded, `xrpld` adds two fields alongside the
```

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,924 @@
...
+| :---- | :---- |
+| `taker_gets_funded` | What the taker can actually get from this offer now |
+| `taker_pays_funded` | What the taker would actually pay |
+| `owner_funds` | The owner's spendable balance of the `TakerGets` currency. Only on the owner's **highest-ranked** offer in this book. |
+
+Two properties matter, and both are easy to get wrong:
+
+1. **`taker_gets_funded` can be `0`.** A completely unfunded offer is still
+   returned, with a real-looking price and size. It is a phantom price level.
+2. **One owner's offers share one balance.** rippled allocates the owner's funds
```
oeggert:
```suggestion
2. **An owner's offers share one balance.** `xrpld` allocates the owner's funds
```

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,924 @@
...
+   to their best-ranked offer and returns the rest as `0` — so summing
+   `taker_gets_funded` across a book is safe and never double-counts.
+
+Observed on Mainnet: of 300 offers in the USD.Bitstamp→XRP book, **174 were
+completely unfunded**. In the XAH→XRP book the *best* offer advertised 130,000
+XAH but only 1,574 was funded (1.2%), and the third-best was 100% phantom —
+a 5,000 XAH buy costs 3.7% more than the raw book implies.
+
```
oeggert:
```suggestion
```

oeggert (changes requested):
Initial changes requested. Small fixes, may be more when I finish the rest of the review.

amarantha-k:
Thanks for the detailed review and suggestions, @oeggert! Requested a review after applying the suggestions.

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,919 @@
...
+```
+
+### 3.4 Estimated fill and slippage — walk the book by funded size
+
+The pre-trade summary's `Est. fill` and `Slippage` must be computed from
+**funded** amounts. Walking raw `TakerGets` overstates available liquidity —
+measured at 47× on a real Mainnet book.
+
+Walk offers in book order, taking the funded size of each until the order is
+filled or the book runs out. Anything left over does not cross.
```
oeggert:
```suggestion
### 3.4 Estimated fill and slippage — walk the book by funded size

The pre-trade summary's `Est. fill` and `Slippage` must be computed from **funded**
amounts. Walking raw `TakerGets` overstates available liquidity.

Two things determine correctness, and both are easy to get wrong:

1. **Base-side capacity lives on a different field per book.** The base sits on
   `TakerGets` for an ask book but on `TakerPays` for a bid book. Reading
   `taker_gets_funded` on a bid book compares *quote* units against a *base*
   order size.
2. **The walk must stop at the limit price.** Levels beyond the limit cannot
   cross, so counting them reports fill the order will never get.

Walk offers in book order, taking the funded base-side size of each until the
order is filled, the limit price is passed, or the book runs out. Anything left
over does not cross.
```

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,919 @@
...
+  }
+  if (got === 0) return { fillable: 0, avgPrice: null, slippagePct: null };
+  const avgPrice = paid / got;
+  const best     = offerPrice(offers.find(o => amt(fundedGets(o)) > 0), bookGetsBase);
+  return {
+    fillable:    got / wantGets,                        // 0..1
+    avgPrice,
+    slippagePct: ((avgPrice - best) / best) * 100,      // vs best FUNDED price
+  };
+}
```
oeggert, line 530:
```suggestion
/** Estimate execution against a live book.
 *  wantBase   — how much BASE to trade, in human units.
 *  limitPrice — QUOTE per BASE. Omit for a market-style estimate.
 *  midPrice   — from §3.3, for the mid-relative slippage the summary requires. */
function estimateFill(
  offers: any[], wantBase: number, bookGetsBase: boolean,
  limitPrice?: number, midPrice?: number,
) {
  let remaining = wantBase, quote = 0, base = 0, bestCrossing: number | null = null;

  for (const o of offers) {
    // Base-side capacity: TakerGets on an ask book, TakerPays on a bid book.
    const avail = bookGetsBase ? amt(fundedGets(o)) : amt(fundedPays(o));
    if (!(avail > 0)) continue;                      // phantom level — skip

    const price = offerPrice(o, bookGetsBase);        // price from ORIGINAL amounts
    if (!Number.isFinite(price) || price <= 0) continue;

    // Buying base: pay no more than the limit. Selling base: receive no less.
    if (limitPrice !== undefined &&
        (bookGetsBase ? price > limitPrice : price < limitPrice)) break;

    if (bestCrossing === null) bestCrossing = price;
    const take = Math.min(remaining, avail);
    base += take;
    quote += take * price;
    remaining -= take;
    if (remaining <= 1e-12) break;
  }

  if (base === 0) {
    return { fillable: 0, avgPrice: null, slippageVsBestPct: null, slippageVsMidPct: null };
  }
  const avgPrice = quote / base;
  return {
    fillable: base / wantBase,                                        // 0..1
    avgPrice,
    slippageVsBestPct: ((avgPrice - bestCrossing!) / bestCrossing!) * 100,
    slippageVsMidPct: midPrice ? ((avgPrice - midPrice) / midPrice) * 100 : null,
  };
}
```

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,919 @@
...
+        remaining -= take
+        if remaining <= 1e-12:
+            break
+    if got == 0:
+        return {"fillable": 0.0, "avg_price": None, "slippage_pct": None}
+    avg   = paid / got
+    live  = live_offers(offers)
+    best  = _offer_price(live[0], book_gets_base)
+    return {"fillable": got / want_gets, "avg_price": avg,
+            "slippage_pct": (avg - best) / best * 100}
```
oeggert:
```suggestion
def estimate_fill(offers, want_base: float, book_gets_base: bool,
                  limit_price: float | None = None, mid_price: float | None = None):
    remaining, quote, base, best_crossing = want_base, 0.0, 0.0, None

    for o in offers:
        # Base-side capacity: TakerGets on an ask book, TakerPays on a bid book.
        avail = _amt(funded_gets(o)) if book_gets_base else _amt(funded_pays(o))
        if not avail > 0:                    # phantom level - skip
            continue

        price = _offer_price(o, book_gets_base)   # price from ORIGINAL amounts
        if price <= 0:
            continue

        # Buying base: pay no more than the limit. Selling base: receive no less.
        if limit_price is not None and (
                price > limit_price if book_gets_base else price < limit_price):
            break

        if best_crossing is None:
            best_crossing = price
        take = min(remaining, avail)
        base += take
        quote += take * price
        remaining -= take
        if remaining <= 1e-12:
            break

    if base == 0:
        return {"fillable": 0.0, "avg_price": None,
                "slippage_vs_best_pct": None, "slippage_vs_mid_pct": None}

    avg = quote / base
    return {
        "fillable": base / want_base,
        "avg_price": avg,
        "slippage_vs_best_pct": (avg - best_crossing) / best_crossing * 100,
        "slippage_vs_mid_pct": ((avg - mid_price) / mid_price * 100) if mid_price else None,
    }
```

.claude/skills/xrpl-skills/xrpl-trading/references/trading.md
```diff
@@ -0,0 +1,919 @@
...
+```
+
+**Report a partial walk honestly.** If `fillable < 1`, the book cannot fill the
+order at any price — say so rather than extrapolating past the last offer.
```
oeggert:
```suggestion
Report `slippage_vs_mid_pct` in the pre-trade summary. `slippage_vs_best_pct`
isolates depth impact from the spread and is useful for diagnostics. A negative
value means later levels priced better than the first one that crossed.

If `fillable` is `0`, nothing crosses at the limit price — say so, rather than
reporting a slippage figure. If `fillable` is less than `1`, the remainder will
rest on the book (or be cancelled, with `tfImmediateOrCancel`).
```

oeggert (changes requested):
One last fix for an issue with sell side calculations.

oeggert (approved):
LGTM

PR #3914, opened 2026-09-07
Rename hero asset to xrp-icon-timed

xVet:
The homepage hero graphic is the XRP X mark, but the file is called `ripple-icon-timed.png`. Renamed it to `xrp-icon-timed.png` and updated the two places that reference it.

Matches how the other XRP marks in the repo are already named: `xrp-x-logo.png`, `logos/xrp-mark.svg`, `navbar/xrp-symbol-black.svg`.


mDuo13 (approved):
Good catch. Confirmed the changes work as expected. :+1:

PR #3931, opened 2026-09-14
AMM doc updates from 3.4.0 release

rachelflynn:
- Zero-trading-fee pool can no longer acquire the slot at no cost [PR #7430](https://github.com/XRPLF/rippled/pull/7430)
- Equal LP-token amount now takes the withdraw-all path; a clawback rounding to zero on either pool side now fails instead of silently no-op'ing; the reserve check is bypassed for clawback; individual/deep freeze on AMM trust lines no longer blocks it PRs: [#7373](https://github.com/XRPLF/rippled/pull/7373), [#7704](https://github.com/XRPLF/rippled/pull/7704), [#7796](https://github.com/XRPLF/rippled/pull/7796), [#6959](https://github.com/XRPLF/rippled/pull/6959)
- Out-of-range deposit/withdraw amounts fail cleanly [PR #7537](https://github.com/XRPLF/rippled/pull/7537)


docs/concepts/tokens/decentralized-exchange/automated-market-makers.md
mDuo13, line 1: I think line 35 should be updated to clarify the distinction between MPTs and trust line tokens in AMMs, especially that depositing MPTs requires the `MPTokensV2` amendment.

We should also clarify that LP Tokens are always trust line tokens regardless of what type of assets are in the AMM. (Maybe future amendments may change this, but last I heard there were no concrete plans to change the nature of LP Tokens.)

docs/concepts/tokens/decentralized-exchange/automated-market-makers.md
```diff
@@ -135,6 +135,20 @@ Frozen LP tokens affect the following transactions:
...
+
+**Payment**
+
+- An account can't send these LP tokens to another non-issuer holder. The transaction fails with `tecNO_AUTH`.
+
+**OfferCreate**
+
+- Offers to sell these LP tokens are treated as unfunded, the same as an offer to sell a frozen LP token.
+
+{% amendment-disclaimer name="MPTokensV2" /%}
```
mDuo13: As it stands, this disclaimer reads almost like users could deposit MPTs in AMMs without these restrictions before this amendment.

If we add clarity that depositing MPTs into AMMs in the first place requires the amendment (see my comment about editing ~line 35), we can move this disclaimer there, which I hope removes the ambiguity. Even though this restriction is "technically" also added as part of the same amendment, by _not_ calling out the amendment here we make it more obvious that the restriction comes part-and-parcel with the ability to deposit MPTs in the first place.

Does that make sense?

docs/concepts/tokens/fungible-tokens/deep-freeze.md
```diff
@@ -74,13 +74,17 @@ Receipt of funds in a deep-frozen trust line as a result of a rippling step will
...
+## OfferCreate Transaction
 
 Deep Freeze requires a change to the `OfferCreate` transaction:
 
 `OfferCreate` returns `tecFROZEN` if the `TakerPays` (buy amount) token has been deep-frozen by the issuer.
 Moreover, any existing offers where the owner has been deep-frozen on the `TakerPays` token can no longer be consumed. It is considered an unfunded offer that is implicitly cancelled by new offers that cross it.
 
+## AMMClawback Transaction
+
+`AMMClawback` transactions can override ordinary freeze settings. Prior to `fixCleanup3_4_0`, this didn't extend to trust lines connected to an AMM pool. With `fixCleanup3_4_0` enabled, `AMMClawback` can override individual and deep freeze on AMM trust lines too, preventing a freeze-clawback deadlock. {% amendment-disclaimer name="fixCleanup3_4_0" mode="updated" /%}
```
mDuo13: What is "a freeze-clawback deadlock"?

docs/references/protocol/transactions/types/ammbid.md
```diff
@@ -84,8 +84,12 @@ If successful, the transaction automatically outbids the previous slot owner and
     ```
 
     - `M` is the minimum bid.
-    - `L` is the total number of LP Tokens currently issued by the AMM
-    - `F` is the trading fee, as a decimal
+    - `L` is the total number of LP Tokens currently issued by the AMM.
+    - `F` is the trading fee, as a decimal.
+
+{% admonition type="info" name="Note" %}
+If the trading fee (F) is 0, both the minimum bid (M) and the price to outbid the current holder (P) enforce a calculated floor price in LP Tokens instead of evaluating toward zero, preventing free or near-free slot acquisition. {% amendment-disclaimer name="fixCleanup3_4_0" /%}
```
mDuo13: What _is_ the minimum calculated floor price? This is the portion of the docs where I would expect it to say so people know how much to bid.

mDuo13 (changes requested):
The changes mostly look good, but I think some important information is missing regarding the floor price for AMM bidding.

mDuo13 (approved):
Supplied the fixes to my own requests.

PR #3936, opened 2026-09-14
Update Redocly to Realm v0.136.2 w/ fixes

mDuo13:
Update Redocly to Realm v0.136.2 (currently the latest version) while applying some fixes mostly written by Oleksii from Redocly, which I then adapted for the post-redesign version of the files:

- Add explicit dependencies on `@redocly/theme` and some codemirror stuff that's no longer implicitly included in the `@redocly/realm` dependency.
- Switch from `react-router-dom` to `react-router` package for compatibility with new React
- Drop `use-query-params` (incompatible with the updated React Router) and refactor WebSocket tool accordingly.

This PR supersedes #3838, #3839, #3840, #3908, and #3915 which separately attempt to bump one dependency and Redocly without the compatibility fixes.


rachelflynn (approved):
LGTM

PR #3940, opened 2026-09-15
Signing updates 3.4

rachelflynn:
- `signature_target` now accepts `SponsorSignature` in addition to `CounterpartySignature`, and any other value returns an invalidParams error [PR #8162](https://github.com/XRPLF/rippled/pull/8162)
- Adds four new hash-prefix entries to the hash-prefix table [PR #8162](https://github.com/XRPLF/rippled/pull/8162)
- SponsorSignature's `TxnSignature` and `Signers` field descriptions updated to note that signatures must be generated using the new role-specific hash prefixes, and that standard transaction signatures are rejected [PR #8162](https://github.com/XRPLF/rippled/pull/8162)


amarantha-k (approved):
lgtm

PR #3941, opened 2026-09-15
Additional 3.4 updates

rachelflynn:
- Duplicate {account, oracle_document_id} entries in the oracles request array are now deduplicated instead of double-counting that oracle's price in the aggregation [PR #6586](https://github.com/XRPLF/rippled/pull/6586)
- delivered_amount is now documented as available for CheckCash and AccountDelete, and the ledger method's expanded output now includes it for all three, consistent with tx, account_tx, and subscribe [PR #5706](https://github.com/XRPLF/rippled/pull/5706)
- PaymentBurn-only delegated payments can no longer cross a zero balance into an implicit mint, unless the delegate also holds PaymentMint
- Reproduces PR #3928's known-amendments.md changes (the fixCleanup3_4_0 section, fixAMMOverflowOffer's retired-status flip, and LendingProtocolV1_1's graduation out of "Amendments in Development"), since that PR's merge was reset off release-3.4.0, plus folds in the new PaymentBurn bullet


amarantha-k (approved):
lgtm

PR #3943, opened 2026-09-15
XRPL AI starter kit updates for milestone 2

amarantha-k:
Summary of changes: 
- Update wallet skill file for human users
- Add links to MPP SDK and xrpl-up, both in Beta
- Update Getting Started to reflect updates to wallet skill
- Update AI Tools page to include skills

PR #3945, opened 2026-09-16
Tweak dependencies in attempt to fix WS tool production-only breakage

mDuo13:
The recent Redocly 0.136.2 upgrade broke the WebSocket Tool in production even though we added fixes that stopped it from breaking in local dev. This push attempts to fix the issue and some others:

- Run `npm dedupe` to synchronize different versions of the same dependency, since a CodeMirror version mismatch is the proximate cause for the tool breaking.
- Remove the `htmltojsx` devDependency which is responsible for several security reports on the repo. If this dependency is currently used, I don't see where/how, so it seems prudent to remove it until it's needed again.
- Re-remove the `use-query-params` dependency which I think was unintentionally re-added when merging #3943
- Remove `bun.lock` file which @Calvinjwala confirmed we don't need.


mDuo13:
Looks like the preview build still has the same breakage as production (can't reproduce in local dev): https://xrpl-dev-portal--885d39.preview.redocly.app/resources/dev-tools/websocket-api-tool

amarantha-k (approved):
Verified preview URL - the websocket tool page loads successfully. Happy to re-review if there are further updates on this PR.

PR #3952, opened 2026-09-17
Upgrade Redocly to Realm v0.137.0

mDuo13:
The Realm package needs to be bumped in sync with the Redocly Theme package or the build fails dramatically. But as long as they're updated together, everything looks groovy. (Tested [the whole checklist](https://xrpl.org/resources/contribute-documentation/realm-upgrade-checklist) without issues.)

Supersedes #3942.


amarantha-k (approved):
Looks good.

PR #3953, opened 2026-09-17
Rewrite & consolidate Deep Freeze docs

mDuo13:
- Remove separate Deep Freeze page and replace it with a "deep freeze" section of the existing freezes page.
    - Add redirects from the old page.
    - Update links that went to the old page.
- Rewrite existing content and remove extraneous details (e.g. implementation details of what the amendment changed in the payment processing engine).
- Update a few other bits that were out of date re: MPTs, Clawback, etc.

I have attempted to port over the parts of the existing Deep Freeze translation that were applicable and adapt some of them to better closely match the reworded English, but I didn't add a translation for the new section re: freezes' relation to clawback.


amarantha-k (dismissed):
Looks good. Only suggestion is to add a section similar to https://xrpl.org/docs/concepts/tokens/fungible-tokens/deep-freeze#how-does-mpt-freezelock-behavior-differ-from-iou that highlights the differences for MPT and IOU.

